PR #100, opened 2026-07-20
fix: merged-PR-reported-as-failure + factory agent-type alias fallthrough (found by the first real factory run)

0xfauzi:
Two defects surfaced live on 2026-07-20 during the first real factory runs (the knowledge/evolution A+ gate evidence capture). Both measured, both with regression tests.

## 1. Merged PRs reported as failures (`pr.py`) - R0.2

`gh pr merge --delete-branch` fails on its **local**-branch deletion whenever the branch is checked out in the component worktree - which it always is at merge time - and gh exits nonzero even though the PR merged. Measured: factory PR #99 sat MERGED on GitHub while the component was FAILED with `merge failed for PR #99: failed to delete local branch...`.

Fixes:
- Drop `--delete-branch` from both merge invocations.
- `_merge_and_wait` consults the actual PR state when the merge *command* fails and continues on a confirmed `MERGED` outcome - R0.2 gates completion on the merge **outcome**, not gh's exit code. A genuinely unmerged PR still fails (tested).
- Remote-branch cleanup becomes an explicit best-effort step after merge confirmation (a stale remote branch would break the next same-name push - the reason --delete-branch existed).

## 2. Factory path uses raw toml agent-type aliases (`cli.py`) - R2.4 mirror

`ralph run` canonicalizes `[agent] type = "claude"` via the R2.4 preflight; `ralph factory` passed it **raw** into `get_agent` in every engineer worker, whose unrecognized-type fallthrough is codex. Measured: toml `"claude"` + `MODEL=haiku` produced a `codex (haiku)` engineer (HTTP 400 on a ChatGPT account) - and `_cli_family` misread the engineer family, which would invert the R7.1 reviewer rotation.

Fixes: the factory command runs `_check_agent_preflight` on `base_config`; the decompose-agent path in both `decompose` and `factory` canonicalizes `effective_type` through `_agent_preflight` (which also subsumes the manual no-agent check).

## Tests

- gh stub in `test_pr_outcomes` is now **merge-aware** (view reports MERGED only after a successful stub merge; explicit override wins) so a failed merge command on an unmerged PR can never read as merged in tests.
- New: MERGED-rescue completes; unmerged failure still fails; `--delete-branch` never passed; remote branch cleaned post-merge; factory canonicalizes the alias before `run_factory`; unknown type blocks loudly before any engineer.
- Adjusted: spine merge-failure test pins the unmerged state; spine happy-path asserts the post-merge remote cleanup.

Gates: 1539 passed / 28 skipped, mypy --strict clean, ruff clean.

Found-by context: these runs are the roadmap's "two real factory runs" measurement; the fix is cherry-picked onto the throwaway `factory/e2e-base` runtime branch so run 2 can proceed - this PR is the mainline landing.

Per H1, not self-reviewed; gating reviewer is the user or `/code-review ultra`.

### DIFF
--- a/ralph_py/cli.py
+++ b/ralph_py/cli.py
@@ -1357,10 +1357,18 @@ def decompose(
         else os.environ.get("RALPH_AGENT_TYPE", "auto")
     )
 
-    if not effective_cmd and not CodexAgent.is_available() and not ClaudeCodeAgent.is_available():
-        ui_impl.err("No agent available (codex and claude not found in PATH)")
-        ui_impl.info("Install an agent or use --agent-cmd to specify a custom one")
+    # R2.4 mirror (measured 2026-07-20): canonicalize aliases like
+    # "claude" before get_agent, whose unrecognized-type fallthrough is
+    # codex; the preflight also covers the no-agent-available check.
+    canonical_type, type_error, type_hint = _agent_preflight(
+        effective_cmd, effective_type,
+    )
+    if type_error:
+        ui_impl.err(type_error)
+        if type_hint:
+            ui_impl.info(type_hint)
         sys.exit(1)
+    effective_type = canonical_type or effective_type
 
     agent = get_agent(effective_cmd, effective_model, effective_reasoning, effective_type)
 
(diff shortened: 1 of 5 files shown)


0xfauzi:
Second commit added: real-run finding #4 - fact utilization measured 0/7 on two clean runs because the knowledge prefix never asked the engineer to record consumed facts, while the metric (measure_fact_utilization) only counts verbatim opening-snippet echoes in diff/progress.txt. The prefix now instructs a 'Facts used:' bullet in progress.txt. Instruments the existing lower-bound metric; measurement logic unchanged. Gates: 1540 passed, mypy strict + ruff clean.

PR #123, opened 2026-07-20
rename: PR 2/4 - KSTRL_* env namespace + kstrl.toml, legacy dual-read

0xfauzi:
Second of the rename series, stacked on #122 (retarget to main after #122 merges - GitHub does this automatically when the base branch is deleted).

## Changes

- **`kstrl/envcompat.py`**: `get`/`contains`/`require` reading `KSTRL_X` first, then legacy `RALPH_X` with a once-per-variable DeprecationWarning. Typed with `os.environ.get`-style overloads for `--strict`.
- All 14 direct env-reader modules threaded through it (~145 sites), including the dynamic env-map sites (`timeout`, `feedforward`, `verify`) and calibration's injected-mapping reviewer override (inline dual-read since it takes a mapping, not os.environ).
- Notify hooks export **both** `KSTRL_NOTIFY_*` and `RALPH_NOTIFY_*` so existing hook scripts keep reading their values for one release.
- **Config file**: `resolve_config_file()` prefers `kstrl.toml`, honors `ralph.toml` with a warning instructing `mv ralph.toml kstrl.toml`. All 15 `load_toml_section` call sites + `KstrlConfig.load` + `KnowledgeConfig.load` routed through it. `ralph.toml.example -> kstrl.toml.example`; `ks init` scaffolds `kstrl.toml`.
- `kstrl_branch` field renames, `KSTRL_LINEAR_TOKEN` default (legacy env still works - the token tests that set `RALPH_LINEAR_TOKEN` were deliberately left as-is and pass through the fallback).
- conftest scrubs both namespaces; opt-in gates accept both `KSTRL_RUN_CALIBRATION`/`RALPH_RUN_CALIBRATION` spellings, so your pending calibration commands work unchanged.

## Tested

- Fast tier 1741 passed / 28 skipped; spine 25 passed; mypy --strict clean (73 files); ruff clean; gen_docs --check current.
- New `tests/test_envcompat.py` (16 tests): new-name precedence, legacy fallback + warning, warn-once semantics, KeyError on absence, config-level precedence (`KSTRL_BRANCH` beats `RALPH_BRANCH`), toml filename preference/fallback/warning with a real load.
- The 138 DeprecationWarnings in the suite run are the legacy fallback being exercised by the ~90 untouched RALPH_-setting tests - deliberate coverage of the compat path.

## Assumed

- Bare `FACTORY_*` names keep their existing explicit dual-accept (unchanged behavior, not routed through envcompat).
- Operator-visible paths (`scripts/ralph/`, `.ralph/`) and prompt bodies are untouched here - PR 3.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -60,7 +60,8 @@ scripts/ralph/understand_prompt.md
 # Ralph - runtime state (journals, worktrees, knowledge, locks; R3.4)
 .ralph/
 
-# Live per-checkout config; ralph.toml.example is the tracked template (R3.4)
+# Live per-checkout config; kstrl.toml.example is the tracked template (R3.4)
+/kstrl.toml
 /ralph.toml
 
 # Claude Code session dir: personal settings + live session worktrees (R3.4)
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -9,7 +9,7 @@
 ## Verification commands
 
 - **Test**: `uv run pytest tests/ -v`
-- **Calibration (opt-in, real LLMs)**: `RALPH_RUN_CALIBRATION=1 uv run pytest tests/test_calibration.py -v`
+- **Calibration (opt-in, real LLMs)**: `KSTRL_RUN_CALIBRATION=1 uv run pytest tests/test_calibration.py -v`
 - **Typecheck**: `uv run mypy kstrl/ --strict`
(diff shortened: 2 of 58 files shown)


0xfauzi:
Auto-closed by GitHub when the stacked base branch was deleted; continued as the replacement PR above (same rebased commit).

PR #173, opened 2026-07-22
R8.1: Policy envelope - declarative Phase 1 merge guardrails (closes #148)

### LINKED ISSUE #148: R8.1: Policy envelope - declarative [policy] block enforced in the mechanical verifier
## Why

Machine-made merge decisions are only defensible inside an explicit, written envelope. Today the rules are implicit and scattered (diff-scope check, allowed paths, bad patterns). The policy envelope makes them declarative, auditable, and enforceable at every autonomy level.

## Approach: build (no policy engine), integrate license metadata

OPA/conftest need a Go binary sidecar and Cedar's Python bindings are immature - all foreign runtimes for rules that fit in ten TOML lines. Build a `[policy]` section in `kstrl.toml` enforced by in-process mechanical checks in the Phase 1 verifier. Integrate `pip-licenses` or `licensecheck` for dependency license gating (explicit allowlist + partial-match deny for copyleft, since compound SPDX expressions defeat exact matching).

## Key design points

- Keys: `paths_deny` (workflows, `kstrl.toml`, `.kstrl/**`, key/env files), `max_files_changed`, `max_lines_changed`, `deps_allow_new`, `license_allow`/`license_deny_partial`, `secret_patterns`, `deploy`.
- Enforcement is harness-side on artifacts (git diff, `uv.lock`, dist metadata) - never on agent self-report. Matches the existing no-LLM verifier doctrine.
- Policy file hash recorded in the run manifest; a diff touching enforcement machinery (policy file, CI workflows, verifier code) is an instant halt at every autonomy level.
- Violations emit typed `Finding`s and route to the exception inbox (R8.3) with approve-once / approve-and-amend-policy / reject actions.

## Done when

Policy checks run in Phase 1 with typed findings; a planted violation in each category (deny path, blast radius, new dep, denied license, secret pattern) is caught in tests; the policy hash lands in the manifest; `kstrl.toml.example` and docs updated.

## Dependencies

None - this is the foundation item. R8.2 (autonomy ladder) consumes it.

Full plan: `docs/dark-factory-roadmap.md` (R8.1).

0xfauzi:
Implements **R8.1 Policy envelope** (closes [#148](https://github.com/0xfauzi/kstrl/issues/148)) in full - the keystone of R8 wave 1 (governance core): a declarative `[policy]` block enforced harness-side in the Phase 1 mechanical verifier on **artifacts** (git diff, `uv.lock`, dependency license metadata), never agent self-report. The autonomy ladder (R8.2) gates on this envelope.

Opt-in via `[policy] enabled = false` (like `[sandbox]`): **existing runs are unchanged**. When a repo opts in, a violation fails Phase 1 and blocks the merge.

## Checks

| Check | Reads | Notes |
|---|---|---|
| `paths_deny` | changed paths | gitignore-style `**` globs |
| `max_files_changed` / `max_lines_changed` | `git diff --numstat` | machine-generated lockfiles excluded; negative disables |
| `deps_allow_new` | `uv.lock` diff | new top-level `[[package]]` entries |
| `secret_patterns` | added diff lines | configurable regexes, all file types |
| **license gate** | dep license metadata | new-dep SPDX license: denied (copyleft) or not-allowlisted blocks; unresolvable is advisory |
| enforcement-machinery halt | changed paths | **hardcoded, non-overridable**: CI workflows + config file |
| `deploy` | - | stored + hashed only; enforced by R8.7 |

The check **fails closed** on an unreadable diff or malformed policy. The resolved envelope's SHA-256 is stamped into the run manifest (`policy_hash`) as the audit record of what guardrails were in force.

## License resolution — a measured correction (H4)

The roadmap assumed `pip-licenses` / installed dist metadata. **Measured against this repo's uv toolchain, that is false**: uv's installed venv materializes no `METADATA` file (empty `licenses/` dir; `importlib.metadata` and `uv pip show` both return nothing), so pip-licenses would resolve nothing. `kstrl/licensing.py` instead:

1. reads the license from **uv's cache** (`<cache>/**/<name>-<version>.dist-info/METADATA`) - offline, doctrine-aligned;
2. falls back to the **PyPI JSON API** on a cache miss (`KSTRL_POLICY_LICENSE_NET=0` forces offline-only).

It prefers `License-Expression` (PEP 639 SPDX), then maps `Classifier` lines to SPDX, then the `License` field. Classification (deny-wins substring for copyleft, compound-SPDX atom tokenizing for the allowlist) is pure in `kstrl.policy`. An unresolvable license is **advisory** (non-blocking) so a merge is not held hostage to a cache/network miss.

## Design notes

- Emits a `CheckResult` (the Phase 1 verifier's native currency - retry-context, events, evolution signatures all key off it), not a `Finding`. Routing to the exception inbox (R8.3) is a clean future seam.
- `kstrl/policy.py` (checks) and `kstrl/licensing.py` (resolution) keep pure logic and I/O separate; resolvers are injectable (`uv_cache`, `http_get`) so tests need neither a real cache nor network.

## Tests / gates

- 84 tests in `tests/test_policy_envelope.py`: a planted violation per category (incl. denied / not-allowlisted / unresolved licenses), license resolution against a fixture uv cache + injected PyPI, config load/env/hash, manifest round-trip, patched-git + **real-git end-to-end**.
- Real-resolver E2E (manual): `anyio` (MIT) allowed from the actual uv cache offline; an unknown package resolves to nothing and becomes a non-blocking advisory.
- `mypy kstrl/ --strict` clean (103 files), `ruff check kstrl/ tests/ scripts/` clean, full suite **2003 passed / 28 skipped / 0 failed**.
- No prompt-body changes, so no calibration (H2) or prompt-version (H3) obligations.

Docs: `kstrl.toml.example`, `ks init`, `docs/env-vars.md`, `scripts/gen_docs.py` (+ regenerated README), roadmap R8.1 marked `[x]` done with the measured-correction note.


kstrl/policy.py
```diff
@@ -0,0 +1,543 @@
...
+# the envelope itself - the CI that runs the checks and the config file
+# that defines them. Modifying any of these is a non-overridable hard
+# fail whenever the envelope is enabled, independent of ``paths_deny``
+# and of the autonomy level (R8.2). Hardcoded on purpose: a repo cannot
+# opt out of protecting its own guardrails.
+ENFORCEMENT_MACHINERY_PATHS: tuple[str, ...] = (
+    ".github/workflows/**",
+    "kstrl.toml",
+    "ralph.toml",
+)
```
0xfauzi, line 60: The hardcoded halt must protect the verifier implementation too. Issue #148 and the roadmap explicitly include verifier code, but this tuple only covers workflows and config files. As written, `evaluate_policy(["kstrl/verify.py"], ..., PolicyConfig(paths_deny=[]))` returns `ok=True`. Please include the complete enforcement surface and add a planted test for each protected code path.

kstrl/verify.py
```diff
@@ -829,6 +835,124 @@ def check_bad_patterns(cwd: Path, base_branch: str) -> CheckResult:
...
+            details=[
+                f"Error: {exc}",
+                "The change cannot be proven within policy; do not treat "
+                "this as permission to merge.",
+            ],
+            duration_seconds=time.monotonic() - start,
+        )
+
+    changed = git.get_diff_names(base_branch, cwd)
+    numstat = git.get_diff_numstat(base_branch, cwd)
```
0xfauzi: These metadata reads are fail-open. `get_diff_names()` and `get_diff_numstat()` return empty collections on timeout or nonzero exit, so a successful earlier content diff does not prove these later subprocesses succeeded. I reproduced a `kstrl.toml` diff passing as `0 files, 0 lines` when both return empty. Please propagate `GitDiffError` or derive the metadata from one validated diff, with timeout and nonzero-exit tests.

0xfauzi (commented):
Changes requested before merge.

The current implementation has two policy-enforcement bypasses:

1. The non-overridable enforcement-machinery set omits verifier code, although issue #148 and the roadmap explicitly require verifier-code changes to halt.
2. The changed-file and numstat reads fail open. Their helpers return empty collections on timeout or nonzero exit, and `check_policy_envelope` then evaluates the change as zero files and zero lines.

I reproduced both cases locally. A `kstrl/verify.py` change passes with an empty `paths_deny`, and a diff touching `kstrl.toml` passes when the two metadata helpers return empty results.

Please also reconcile these acceptance-criteria differences before marking R8.1 complete:

- An unresolved dependency license is allowed as advisory, which weakens the stated explicit-allowlist posture. `KSTRL_POLICY_LICENSE_NET` can also change the effective verdict without changing `policy_hash`.
- Issue #148 calls for typed `Finding`s, while this implementation emits only `CheckResult` and defers inbox integration.

Validation performed: the focused policy suite passed 84/84, the full local suite passed 2,003 tests with 28 skipped, and GitHub CI run 295 is green. These are logic and scope gaps rather than test regressions.

Note: GitHub does not permit a pull request author to set “Changes requested” on their own PR, so this is submitted as a comment review.

kstrl/policy.py
0xfauzi: Fixed in 1672829. Confirmed the repro: `evaluate_policy(["kstrl/verify.py"], ..., PolicyConfig(paths_deny=[]))` returned `ok=True`.

`ENFORCEMENT_MACHINERY_PATHS` now covers all three surfaces #148 names - policy file, CI workflows, and verifier code:

```
".github/workflows/**", "kstrl.toml", "ralph.toml",
"**/kstrl/verify.py", "**/kstrl/policy.py", "**/kstrl/licensing.py",
"**/kstrl/guards.py", "**/kstrl/fixtures.py",
```

The leading `**/` matches zero or more directories, so a nested/vendored checkout is covered too. Added `enforcement_paths_extra` for a repo to protect its own gate code - deliberately ADDITIVE (joined to the hardcoded tuple), so no config edit can shrink the protected set.

Tests: a planted violation per protected path with `paths_deny=[]` (`TestEnforcementMachineryCoversVerifierCode`), the same paths nested under `vendor/pkg/`, an extras-are-additive case, an extras-cannot-shrink case, and an over-blocking guard asserting `kstrl/pipeline.py`, `src/app/verify.py` and a bare `verify.py` still pass. Verified end-to-end on a real repo: committing a `kstrl/verify.py` edit now blocks with a `critical` `policy_enforcement_machinery` finding.

kstrl/verify.py
0xfauzi: Fixed in 1672829. You are right, and the old docstring on `get_diff_numstat` asserted exactly the flawed reasoning you named ("a swallowed error here cannot mask one") - a successful `get_diff_content` says nothing about two later, separate subprocesses. That comment is gone.

Both helpers now take `strict: bool = False`; with `strict=True` a timeout or nonzero exit raises `GitDiffError` instead of returning `[]`. `check_policy_envelope` makes all three reads inside one try block and fails closed on any of them:

```python
diff_text = git.get_diff_content(base_branch, cwd)
changed = git.get_diff_names(base_branch, cwd, strict=True)
numstat = git.get_diff_numstat(base_branch, cwd, strict=True)
```

I kept the lenient default rather than changing the global contract, because `check_diff_scope`, `check_bad_patterns`, `check_dead_code`, and `check_mutation_score` were all written against the `[]` behavior - flipping it for them is a separate change with its own blast radius, not something to smuggle into this PR. A test pins that lenient contract so the distinction stays deliberate.

Tests (`TestGitMetadataFailsClosed`): nonzero exit for each helper, timeout for each helper (patched `subprocess.run`), the lenient-default contract, and the policy check failing closed when either the names read or the numstat read raises - the failure now surfaces as an `infrastructure_error` finding rather than a vacuous pass.

0xfauzi:
All four points addressed in 1672829. Both bypasses reproduced first, then fixed and pinned with regression tests.

**1. Enforcement machinery omitted verifier code** - real bypass, fixed. The halt set now covers all three surfaces #148 names (policy file, CI workflows, verifier code) at any depth, plus an additive `enforcement_paths_extra` that config can never shrink. Planted test per protected path, plus over-blocking guards so ordinary source is unaffected. Details in the inline thread.

**2. Fail-open metadata reads** - real bypass, fixed. `get_diff_names`/`get_diff_numstat` gained `strict=True` (raises `GitDiffError`); the policy check uses it for all three reads and fails closed. The old docstring asserted the exact flawed reasoning you identified; it is gone. Timeout and nonzero-exit tests for both helpers. Details in the inline thread.

**3. Unresolved license / hash coverage** - agreed on both, and the inconsistency was mine: every other failure path in this check fails closed, while the license path failed open. Unresolved now **blocks by default** via `license_unresolved = "block" | "advisory"`, so the explicit-allowlist posture holds and accepting unprovable licenses is an explicit opt-in. The network toggle is now a real config field (`license_use_network`) instead of a bare env read, so it is inside `envelope_hash` - two runs with the same `policy_hash` enforced the same rules. `KSTRL_POLICY_LICENSE_NET` / `KSTRL_POLICY_LICENSE_UNRESOLVED` resolve into those fields *before* the hash is computed. Tested, including that the offline verdict names its reason and that both toggles move the hash.

**4. Typed `Finding`s** - implemented rather than deferred; you were right that the criterion is not satisfied by `CheckResult` alone. Added `Finding.policy_violation`, a `CheckResult.findings` field, and a lift in `_phase_verify` into the component finding stream, so a machine-made gate decision reaches the PR body, journal, and evolution record. Severity is `critical` for the machinery halt, `high` for other blocking violations, `advisory` for non-blocking notices; findings attach on passing checks too, since an advisory is still evidence. Verified this cannot disturb the security gate, which counts severities on its own `SecurityResult.findings`, not `comp.findings`. Inbox **routing** remains R8.3 - the item does not exist yet, and `CheckResult.findings` is the seam it will consume.

Verification: 118 tests in the policy suite (+34), `mypy --strict` and `ruff` clean, full suite **2037 passed / 28 skipped**. Both of your repros re-run end-to-end on real git repos: a `kstrl/verify.py` edit with `paths_deny=[]` now blocks with a critical finding, a `kstrl.toml` diff blocks, and an ordinary change still passes with zero findings. Roadmap R8.1 records all four corrections.

PR #174, opened 2026-07-27
R8.2: Autonomy ladder - earned, bounded, revocable levels (closes #149)

### LINKED ISSUE #149: R8.2: Autonomy ladder - earned, bounded, revocable autonomy levels (L1-L4)
## Why

Autonomy today is a scatter of independent flags (`pause_before_pr_merge`, hard-mode thresholds, budgets). The cATO model from DoD software-factory practice is the frame: autonomy is earned after demonstrated baseline compliance, bounded by an explicit envelope, and revocable with automatic demotion back to human-gated mode.

## Approach: build (~small state machine over existing flags)

Nothing integrable exists for a local CLI. Prior art converges: Claude Code permission modes (levels = bundles of flags), OpenHands confirmation policy crossed with a risk analyzer, and arXiv:2506.12469 (levels defined by what the human still does).

## Key design points

- **L1 Supervised**: plan approval + merge gate, strictest thresholds, no new deps, no deploy. Default for new repos.
- **L2 Gated-merge**: plans auto-accepted, human gates PR merge only. Entry: 5 components merged at L1 with zero policy violations, calibration compare green.
- **L3 Enveloped auto-merge**: merge gate off when fully green AND inside the policy envelope; any breach routes to inbox instead of merging. Entry: 15 consecutive L2 merges approved without edits, health metrics inside limits, recorded `ks autonomy promote` ack.
- **L4 Deploy**: L3 + release stage enabled. Entry: L3 held for 30 merged components + explicit promote.
- Promotion needs evidence AND a recorded human ack - agents cannot promote themselves. Demotion is automatic, one level per trigger (policy violation, calibration regression, health-metric breach, human rejects an auto-merge). Fast down, slow up: cool-down of 10 decisive runs after demotion.
- State in `.kstrl/autonomy.json`; every transition appended to events.jsonl and the evolution journal; demotions emit an inbox item with the triggering evidence.
- Numeric thresholds are placeholders until measured: replay the triggers over historical `experiments.tsv` and count would-have-fired transitions before enabling auto-demotion (no-assumptions rule).

## Done when

Levels drive the flag bundles; promotion requires a recorded ack; demotion fires correctly on planted trigger fixtures; the replay tool exists and its output is captured in the roadmap doc.

## Dependencies

R8.1 (the envelope defines L3). R8.4 supplies EWMA triggers later; initial triggers (policy violation, calibration regression, human reject) work without it.

Full plan: `docs/dark-factory-roadmap.md` (R8.2).

0xfauzi:
Implements **R8.2 Autonomy ladder** (closes [#149](https://github.com/0xfauzi/kstrl/issues/149)) - the governance core R8.3, R8.5, R8.6, and R8.7 all gate on, now unblocked by the R8.1 envelope (#173).

Autonomy today is a scatter of independent flags (`pause_before_pr_merge`, `review_mode`, `deps_allow_new`, deploy). Any one can be flipped in isolation, so "how much is this factory allowed to do without me?" has no single answer and no audit record. This makes it one ordered level with a derived flag bundle.

## The three invariants

| Invariant | How it is enforced |
|---|---|
| **Agents cannot promote themselves** | `promote()` requires an `actor`, an `ack`, evidence, AND a controlling TTY - an out-of-band signal an unattended agent subprocess lacks (strings alone prove nothing). Real but not absolute: an agent with shell access could write `.kstrl/autonomy.json`, so that file and `kstrl/autonomy.py` are in the R8.1 enforcement-machinery halt set. |
| **Fast down, slow up** | Demotion is automatic, immediate, needs no ack, drops exactly one level, and starts a cool-down that blocks re-promotion for N decisive runs. |
| **The bundle is derived, never stored** | Computed from the level at run start. Config that contradicts it is recorded as a manual override and **the bundle wins** - a hand-edited flag cannot grant autonomy the ladder never awarded. |

Two further guards worth calling out: `review_mode` stays `hard` at **every** level (a Goodhart guard - autonomy must never buy weaker verification), and corrupt or out-of-range stored state falls back to L1, because the safe direction for unknown autonomy is the least autonomy.

## Levels

| Level | Human's remaining role | Bundle |
|---|---|---|
| L1 Supervised | approves plan and merge | merge gate ON, plans human-approved, no new deps, no deploy |
| L2 Gated-merge | gates merge only | plans auto-accepted; merge gate still ON |
| L3 Enveloped auto | exception handling | merge gate off when green AND inside the R8.1 envelope; new deps permitted |
| L4 Deploy | exception handling | L3 + release stage reachable (R8.7) |

## Threshold replay - the R8 "no assumed thresholds" rule

`ks autonomy replay` replays every threshold over `.kstrl/experiments.tsv`. Captured output:

```
Runs recorded:        5
  decisive:           2
  infra-aborted:      3 (excluded)
Components merged:    2

Would-have-promoted:  0
Would-have-demoted:   0
Final level after replay: L1

VERDICT: INSUFFICIENT DATA (2 decisive runs vs a MIN_DECISIVE_RUNS floor of 8)
```

Read honestly: **every threshold in this PR remains an unmeasured placeholder.** Three of five recorded runs died on PR/push plumbing, which the replay excludes as infrastructure casualties - they say nothing about the factory's judgement. The remaining sample is one toy project over a single day and predates the TUI stack, the rename, and every fix since. This calibrates nothing; it establishes only that the ladder cannot promote past L1 on existing evidence, which is the safe failure. **Generating real evidence is the blocking prerequisite for L2+, not more code.** Recorded in the roadmap.

The replay never mutates state (tested), and exits `2` on insufficient data so a script cannot mistake it for a green run.

## Surface

- `kstrl/autonomy.py` - levels, thresholds (one named block), flag bundles, state machine, atomic persistence to `.kstrl/autonomy.json`, `[autonomy]` config
- `kstrl/autonomy_replay.py` - the replay tool, with an explicit decisive-run definition (infra-aborted runs excluded, so a string of broken runs cannot unlock a promotion)
- `ks autonomy status | promote | demote | history | replay`
- events: `autonomy_transition`, `autonomy_level_applied`. Transitions also land in the evolution journal, which is the durable cross-run record; `events.jsonl` is per-run, so a CLI transition outside a run reaches the journal only. All three writes go through one `commit_transition()`.
- `run_factory` derives the bundle and logs contradicting config

**Opt-in** (`[autonomy] enabled = false`): L1 is *stricter* than today's defaults - it forces the merge gate on - so enabling the ladder must be deliberate, never a surprise upgrade for an existing repo.

## Tests / gates

- 57 ladder tests: planted fixtures for all five demotion triggers, promotion refusals (no actor / no ack / no evidence / cool-down / level-skip), counter resets, persistence incl. corrupt-file fallback, config + clamping, replay behavior, and factory-wiring tests proving the bundle overrides contradicting config.
- `mypy --strict` clean (105 files), `ruff` clean, full suite **2094 passed / 28 skipped / 0 failed**.
- CLI invariants verified end-to-end: self-promotion refused, human promotion recorded, policy-violation demotion with cool-down, re-promotion blocked, full history.

R8.4 will enrich demotion with health-metric breaches; the `HEALTH_BREACH` trigger already exists for it to fire. Inbox routing for demotion notices lands with R8.3.

---

**Update (3b605e4):** seven review findings addressed - see the [summary comment](https://github.com/0xfauzi/kstrl/pull/174#issuecomment-5093310188). The substantive ones: L3 could auto-merge with no envelope; the ladder was inert (no production call sites, and infra-aborted runs wrongly counted as decisive evidence); promotion authority was a caller-supplied string; and transitions never reached the audit streams. Suite now 2118 passed.


kstrl/factory.py
```diff
@@ -1719,6 +1727,33 @@ def _run_factory_locked(
...
+    if autonomy_config.enabled:
+        autonomy_state = AutonomyState.load(root_dir)
+        level = effective_level(autonomy_state, autonomy_config)
+        bundle = flag_bundle_for(level)
+        overrides = manual_override_notes(
+            bundle,
+            configured_pause_before_pr_merge=factory_config.pause_before_pr_merge,
+            configured_review_mode=factory_config.review_mode,
+        )
+        factory_config.pause_before_pr_merge = bundle.pause_before_pr_merge
```
0xfauzi, line 1855: [P1] Require the R8.1 envelope before removing the merge gate. `policy_config` can still be disabled here (that is its default), and the pipeline skips `check_policy_envelope` when it is disabled, yet an L3 bundle unconditionally sets `pause_before_pr_merge=False`. That permits an "enveloped" L3 run to auto-merge with no envelope at all. This also applies only two of the six bundle fields: for example, `deps_allow_new_permitted=False` is ignored, so L1/L2 still accept `[policy] deps_allow_new=true`. Please derive an effective policy config before hashing/constructing the pipeline, clamp dependency permission from the bundle, and refuse L3+ unless the envelope is enabled.

kstrl/factory.py
```diff
@@ -1719,6 +1727,33 @@ def _run_factory_locked(
...
     manifest.save(manifest_path)
 
+    # R8.2: derive this run's permissions from the autonomy level. The
+    # bundle is computed at run start and WINS over contradicting config,
+    # so a hand-edited flag cannot grant autonomy the ladder never
+    # awarded; contradictions are recorded rather than silently dropped.
+    # Opt-in: when [autonomy] is disabled the config's own flags stand.
+    autonomy_config = AutonomyConfig.load(root_dir)
+    if autonomy_config.enabled:
+        autonomy_state = AutonomyState.load(root_dir)
```
0xfauzi, line 1844: [P1] Wire run outcomes into the persisted state. There are no production call sites for `record_decisive_run`, `record_merged_component`, `record_policy_violation`, or automatic `demote`/`save`; they appear only in tests and replay code. Consequently, a successful factory run leaves every counter at zero, promotion is impossible without `--force`, and a real policy violation does not trigger the promised immediate demotion. Please update and save the state from terminal run/check outcomes under the run lock, with end-to-end tests that execute a run and reload `.kstrl/autonomy.json`.

kstrl/autonomy_replay.py
```diff
@@ -0,0 +1,238 @@
...
+                DemotionTrigger.CALIBRATION_REGRESSION,
+                f"replay: {run.failed} component(s) failed in {run.run_id}",
+            )
+            if record is not None:
+                report.would_demote.append(
+                    f"{run.timestamp} {run.run_id}: "
+                    f"L{record.from_level} -> L{record.to_level} "
+                    f"({record.trigger})"
+                )
+        if not state.promotion_blockers():
```
0xfauzi: [P1] Advance the hypothetical state after a would-promote decision. This branch records eligibility but never changes `state.level` or resets its counters. I reproduced this with 10 successful decisive runs: the report emits L2 eligibility for runs 8, 9, and 10 and still returns `final_level == 1`. As written, L3/L4 thresholds, post-promotion demotions, and cool-down behavior are never replayed. Please perform a dedicated simulated promotion here and add a dataset large enough to traverse multiple levels and then demote.

kstrl/cli.py
```diff
@@ -3071,6 +3071,217 @@ def _evolve_apply(
...
+    else:
+        ui_impl.ok("  Criteria met; `ks autonomy promote --actor <you> --ack <why>`")
+    ui_impl.info(
+        "  Thresholds are UNMEASURED placeholders; run `ks autonomy replay`."
+    )
+    sys.exit(0)
+
+
+@autonomy_group.command(name="promote")
+@click.option("--actor", required=True, help="Who is acknowledging (a human)")
```
0xfauzi, line 3174: [P1] A caller-supplied string does not establish a human acknowledgement. With sandboxing off by default, an agent can run `ks autonomy promote --actor human --ack x --force` (and `--root` can target the main repository), satisfying every check and bypassing all evidence. This contradicts the central "agents cannot promote themselves" invariant. The promotion authority needs an out-of-band signal unavailable to the agent, such as a controlling-TTY-only confirmation or a separately authenticated approval; the evidence-bypass path must not be callable from the same unattended execution context.

kstrl/cli.py
```diff
@@ -3071,6 +3071,217 @@ def _evolve_apply(
...
+
+    root_dir = (root or Path.cwd()).resolve()
+    ui_impl = _autonomy_ui(ui, no_color)
+    state = AutonomyState.load(root_dir)
+    try:
+        record = state.promote(actor=actor, ack=ack, force=force)
+    except AutonomyError as exc:
+        ui_impl.err(f"Promotion refused: {exc}")
+        sys.exit(1)
+    state.save(root_dir)
```
0xfauzi: [P1] Append the transition to the promised audit streams when committing it. This path, and the demotion path below, only save `autonomy.json`; `AutonomyTransition` has no production emitter, and neither command writes the evolution journal. Therefore the documented claim that every transition is reconstructable from `events.jsonl` and the journal is false. Please centralize transition persistence so the state save and both audit appends happen together, and cover promotion and demotion through CLI-level tests.

kstrl/autonomy.py
```diff
@@ -0,0 +1,606 @@
...
+                trigger=str(h.get("trigger", "")),
+                evidence=h.get("evidence", {}) or {},
+            )
+            for h in data.get("history", [])
+            if isinstance(h, dict)
+        ]
+        return cls(
+            level=level,
+            since=str(data.get("since", "")) or _utc_now_iso(),
+            components_merged_at_level=int(data.get("components_merged_at_level", 0)),
```
0xfauzi: [P2] Validate the entire persisted payload before constructing state. Only JSON parsing and `level` conversion are guarded; a syntactically valid file such as `{"level":99,"components_merged_at_level":"not-an-int"}` raises `ValueError` here, and malformed numeric fields inside `history` fail similarly. That crashes `status` and an enabled factory run instead of resolving unknown autonomy safely. Please reject/quarantine any invalid schema or field type and fail closed at L1 (or refuse the run explicitly), with malformed-counter and malformed-history tests.

kstrl/cli.py
```diff
@@ -3071,6 +3071,217 @@ def _evolve_apply(
...
+        ui_impl.kv("in force", f"L{int(level)} (clamped by max_level)")
+    ui_impl.kv(
+        "enabled", "yes" if config.enabled else "no ([autonomy] enabled=false)",
+    )
+    ui_impl.kv("since", state.since or "-")
+    if state.last_promoted_by:
+        ui_impl.kv("promoted by", state.last_promoted_by)
+
+    ui_impl.subsection("Flag bundle")
+    for line in state.flag_bundle().describe():
```
0xfauzi: [P2] Render the effective bundle, not the stored bundle. When stored state is L4 and `max_level=2`, the command correctly prints `in force: L2` and then uses `state.flag_bundle()` to report the L4 permissions, including deploy permitted and merge gate off. Use `flag_bundle_for(level)` here so the displayed permissions match what `run_factory` will enforce.

0xfauzi (commented):
I found several blockers in the runtime integration and audit model. The implementation's unit tests, Ruff, mypy, and the full 2,094-test suite pass, but the tests currently exercise the state machine mostly in isolation and miss these end-to-end failures. In particular, L3 can disable the merge gate without an enabled policy envelope, normal runs never accumulate ladder evidence or trigger demotion, and the replay does not actually advance its simulated level.

kstrl/factory.py
0xfauzi: Fixed in 3b605e4. Confirmed: L3 dropped the merge gate regardless of the envelope, which is auto-merge inside nothing - and "Enveloped" is the whole justification for the level.

`resolve_runtime_level(state, config, policy_enabled=...)` now applies two independent ceilings and takes the **lowest**: `max_level` and an envelope ceiling of L2 when `[policy] enabled=false`. It returns `(level, notes)` rather than raising, so the run proceeds safely and says loudly what it withheld.

On the second half - only 2 of 6 fields applied - `deps_allow_new` is now clamped from the bundle before the policy hash is taken, so the manifest records the envelope actually enforced:

```python
if not bundle.deps_allow_new_permitted and policy_config.deps_allow_new:
    policy_config = replace(policy_config, deps_allow_new=False)
```

The ladder can only ever WITHHOLD a permission the envelope grants, never add one. The remaining bundle fields (`auto_accept_plan`, `auto_merge_when_green`, `deploy_permitted`) have no consumer yet - plan auto-accept is R8.3, deploy is R8.7 - so they stay declarative rather than being wired to nothing; `deploy_permitted` is the gate R8.7 will read.

Tests: `TestEnvelopeCeiling` (L3 and L4 clamp without the envelope, L3 allowed with it, lowest-ceiling-wins) plus an end-to-end case asserting an L3 repo with `[policy]` off keeps `pause_before_pr_merge=True`, and `TestBundleClampsPolicy` for the deps clamp.

kstrl/factory.py
0xfauzi: Fixed in 3b605e4. You are right, and this was the most serious one: the state machine was real but inert. Nothing in production touched the recorders, so counters stayed at zero, promotion was unreachable except via `--force`, and the "automatic demotion" in the PR description never fired.

`_record_autonomy_outcome()` now runs at run end, before the factory lock is released, and folds the terminal outcome into state: decisive-run count, merged components, policy violations, and an immediate demotion when a violation is present. Demotion happens at run end rather than mid-run deliberately - demoting while components are still executing would change the bundle underneath them, and those PRs were already gated by the level in force at start.

**A second bug surfaced while fixing this**, and it is worth flagging because it would have quietly inverted the ladder: I had `decisive = completed or failed`, so a run that died on infrastructure counted as decisive. A string of broken runs would have burned down a demotion cool-down and accrued promotion evidence. It also contradicted my own replay module, which excludes infra casualties. Decisive now means "at least one component reached a verdict that was actually about the factory judgement", detected via typed `infrastructure_error` findings - the two definitions have to agree or the replay predicts nothing.

Tests: `TestRunOutcomesReachState` runs the factory and reloads `.kstrl/autonomy.json` (evidence recorded, accumulates, nothing recorded when the ladder is disabled, policy violation demotes AND lands in the journal, advisory findings do not demote). The violation path drives findings through the real R8.1 route - a policy `CheckResult` carrying findings that the pipeline lifts - because pre-seeding the manifest does not survive `begin_attempt` clearing the stream.

kstrl/autonomy_replay.py
0xfauzi: Fixed in 3b605e4. Reproduced your exact case first - 10 clean decisive runs produced three separate "eligible for L2" lines and `final_level == 1`.

The branch now performs a real simulated promotion (which advances the level and resets the per-level counters) and records the actual transition:

```
2026-07-08 r8:  L1 -> L2 (eligible; human ack still required in reality)
2026-07-23 r23: L2 -> L3
2026-07-25 r25: L3 -> L4
```

The ack is a real-world requirement, not a replay one: the replay answers "what would the criteria have allowed?", so it grants the ack and says so in the line it emits.

Tests: `TestReplayAdvancesLevels` - level advances past L1, no duplicate eligibility for the same target, a 60-run dataset traverses L1→L2→L3→L4, a judgement failure appended to that dataset demotes from L4 to L3, and 200 runs never exceed L4.

Worth noting for the record: the captured `INSUFFICIENT DATA` verdict in the roadmap is unaffected by this bug - that sample never approaches the threshold - and I re-ran it after the fix to confirm 2 decisive runs and the same verdict.

kstrl/cli.py
0xfauzi: Fixed in 3b605e4, and this was the sharpest finding - the invariant I put at the top of the module docstring was enforced by nothing more than a non-empty string check.

Promotion now requires a **controlling TTY** on both stdin and stdout, checked before the state is even loaded, and `--force` is covered by the same gate. An unattended agent subprocess does not have one:

```
$ ks autonomy promote --actor human --ack x --force < /dev/null
ERROR: Promotion refused: promotion requires an interactive terminal
(forced promotions bypass evidence and a caller-supplied --actor string
does not establish human acknowledgement).
```

The obvious way around a TTY gate is to write the level straight to disk, so `.kstrl/autonomy.json` and `kstrl/autonomy.py` are now in the R8.1 `ENFORCEMENT_MACHINERY_PATHS` set - an agent-authored diff touching either is a non-overridable halt.

I have deliberately **not** claimed this is airtight. The module docstring now reads: the boundary is real but not absolute, since an agent with shell access could still write the file directly, which is why it is in the halt set. A cryptographically-authenticated approval is the honest fix and belongs with R8.3 rather than being half-built here.

Tests: `TestPromotionAuthority` - non-TTY cannot promote, non-TTY cannot force (and the message names the evidence bypass), TTY may promote, plus a check that both ladder-state paths are enforcement machinery.

kstrl/cli.py
0xfauzi: Fixed in 3b605e4. You are right that the documented claim was false - I wrote "every transition emits `autonomy_transition`" while nothing emitted it outside the sample fixture.

`commit_transition(state, record, root_dir, bus=..., run_id=...)` now performs all three writes in one place: state save, evolution-journal append, and event emit. Both CLI paths and the automatic demotion in `run_factory` go through it, so the three cannot drift apart again. State save happens first because it is the load-bearing write; an audit-append failure warns but never strands the ladder in an unsaved state (tested).

One correction to the claim itself rather than the code: `events.jsonl` is **per-run** (`.kstrl/runs/<run_id>/events.jsonl`), so a CLI transition outside a run has no event stream to append to. The evolution journal is the durable cross-run record and always receives the transition; the event is emitted when the transition happens inside a run. I have amended the roadmap and PR description to say exactly that instead of overclaiming.

Tests: `TestTransitionAudit` - journal receives promotions with actor and direction while state is saved, a bus inside a run receives `AutonomyTransition`, and a journal write failure warns without losing the state save. `TestRunOutcomesReachState::test_policy_violation_demotes_and_journals` covers the automatic path end-to-end.

kstrl/autonomy.py
0xfauzi: Fixed in 3b605e4. Confirmed: `{"level":99,"components_merged_at_level":"not-an-int"}` raised out of `load()` and took `status` and any enabled run down with it.

Every field is now validated before the state is constructed, and **anything** unrecognizable discards the whole file and returns a fresh L1 - never a partially-trusted level assembled from a damaged record. `_require_int` treats a wrong type as a rejection rather than coercing, and excludes `bool` explicitly (it is an `int` subclass, so `true` would otherwise launder into a plausible-looking counter of 1). `_parse_history` applies the same strictness per entry.

The rejection warns rather than passing silently, mirroring `knowledge.read_facts`: losing an earned level deserves a note.

Tests: malformed counter, malformed history entry, out-of-range level, non-object top level, and unreadable JSON all fall back to L1.

kstrl/cli.py
0xfauzi: Fixed in 3b605e4. `status` now resolves the level through the same `resolve_runtime_level()` the factory uses and renders `flag_bundle_for(level)`, so the displayed permissions are the ones a run would actually enforce.

This matters more after the other fix in this round: the effective level is now clamped by the policy envelope as well as `max_level`, so an L3 repo with `[policy] enabled=false` would otherwise have advertised auto-merge that the run refuses to grant. The clamp reasons are printed under `in force` rather than left implicit.

0xfauzi:
All seven addressed in 3b605e4. Your framing was correct and worth restating: the unit tests exercised the state machine in isolation while the *integration* went untested, so I shipped a ladder that was real but inert and claimed end-to-end behaviour it did not have. That is the H4 failure this repo has a rule about, and I hit it.

Four were substantive holes rather than polish:

- **L3 auto-merged with no envelope** - `resolve_runtime_level()` now clamps L3+ to L2 when `[policy] enabled=false` (lowest ceiling wins against `max_level`), and `deps_allow_new` is clamped from the bundle before the policy hash is taken. The ladder can withhold an envelope permission, never grant one.
- **The ladder was inert** - run outcomes now fold into state at run end inside the factory lock. **A second bug fell out of fixing this:** I had counted infra-aborted runs as decisive, so a string of broken runs would have burned down a cool-down and accrued promotion evidence - the exact opposite of what the mechanism is for, and inconsistent with my own replay module. Now excluded via typed `infrastructure_error` findings.
- **Promotion authority was a string** - now requires a controlling TTY, with `.kstrl/autonomy.json` and `kstrl/autonomy.py` added to the R8.1 enforcement-machinery halt set to close the write-the-file path. I have not claimed this is airtight; the docstring states the boundary is real but not absolute, and a properly authenticated approval belongs with R8.3.
- **Transitions never reached the audit streams** - `commit_transition()` now does state save + journal append + event emit together. One correction to the *claim* rather than the code: `events.jsonl` is per-run, so a CLI transition outside a run reaches the journal only. The roadmap and PR body now say that instead of overclaiming.

Plus the replay level-advance fix (reproduced your 10-run case first), fail-closed state validation, and effective-bundle rendering.

Verification: 81 ladder tests (+24), including end-to-end runs that reload `.kstrl/autonomy.json` and drive policy findings through the real R8.1 path. `mypy --strict` and `ruff` clean; full suite **2118 passed / 28 skipped**. The captured `INSUFFICIENT DATA` replay verdict was re-run after the replay fix and is unchanged (2 decisive runs, floor of 8) - that sample never approaches the threshold.

Issue #153, opened 2026-07-21
R8.6: Continuous intake - .kstrl/queue substrate, ks serve, GitHub Issues front-end


0xfauzi:
## Why

Intake today is one-shot: a human fires `ks factory --spec` and the run ends. The first US software factory (SDC, 1972-78) died because work was not required to flow through it - intake is a survival capability. The factory needs specs, features, and bugs flowing in without a human firing each run.

## Approach: build a thin local substrate, integrate the edges

- **Build**: maildir-style directory queue under `.kstrl/queue/` (`queued/ leased/ running/ done/ failed/ poison/`), item = spec file + `meta.json` sidecar, transitions via `os.replace` (the existing atomic-write pattern), flock singleton, pid/ttl leases with a reaper for sleep/crash recovery, every transition journaled. ~200-300 lines. Queue libraries were evaluated and rejected (litequeue near-dormant + Python floor conflict; persist-queue buys nothing at this concurrency; huey inverts control).
- **Integrate front-ends**: GitHub Issues as primary inbox - poll a `kstrl:queued` label via `gh` (ETag-cheap, ~30 req/hr vs 5,000/hr budget), write back state labels + result comments. Linear as an optional second polling adapter reusing the existing client. Both pull-based: no webhooks, no public endpoint. Linear Agents API deferred until it exits preview.
- **Integrate scheduling**: `ks serve` as a launchd KeepAlive LaunchAgent (StartInterval misses intervals elapsed during sleep); runs execute under `caffeinate -i` held only during active work.

## Key design points

- Only `infrastructure_error` failures auto-retry (backoff, `max_attempts`); spec-level failures go straight to `poison/` with a comment back to the source issue - no token-burning crash loops.
- Remote-sourced items default to `stop_at_pr` (merge stays human) - continuous intake must not silently delete the merge gate; `auto-merge` is per-item opt-in and ladder-gated.
- Queue-level `daily_budget_usd` hard stop, independent of per-run caps.
- Processed-ids ledger makes re-seen issues idempotent; front-end outages never block the local queue (adapters strictly additive).

## Done when

`ks queue add/ls/show/retry/rm/pause/resume` and `ks serve [--once]` work; the poison path and budget pause are tested; the GitHub adapter round-trips labels and comments against a test repo; the launchd plist and caffeinate behavior are documented.

## Dependencies

R8.2 (merge disposition per autonomy level), R8.3 (poison/budget notifications land in the inbox).

[...]

PR #175, opened 2026-07-27
R8.3: Exception inbox - one surface for human decisions (closes #150)

### LINKED ISSUE #150: R8.3: Exception inbox - ks inbox surface for over-the-loop operation
## Why

Over-the-loop operation means the human is pulled in only on boundary conditions - but today those conditions are scattered across SpecBlocker exit codes, FAILED components, MERGE_PENDING, evolve proposals, and calibration captures. Walk-away operation needs one surface listing everything awaiting a human decision.

## Approach: build a thin JSONL inbox, integrate ntfy.sh for push

The inbox itself is small (`.kstrl/inbox.jsonl` + CLI + a Textual screen in the home shell). Notification rides the existing notify hook: ntfy.sh is one HTTP POST, self-hostable, with priority levels - no new daemon, one-way only (actions happen in `ks inbox`, not from the phone).

## Key design points

- Item types: policy exception, merge gate (L1-L2), halted run, budget overrun, demotion notice, calibration drift.
- Linear-Triage-style one-key actions per type: approve / reject-with-comment / retry / edit-spec / snooze-with-TTL. Every decision is journaled (audit-trail doctrine).
- `approve-and-amend-policy` converts repeated approvals into envelope widening - the learning loop that shrinks the inbox over time.
- Open-item cap: beyond N open items the queue pauses intake (prevents the guilt-pile failure mode). Notify only action-required items and demotions; successes are silent; daily digest for informational items.

## Done when

All existing halt paths emit inbox items; actions round-trip (approving a merge-gate item resumes the component); the Textual screen renders and acts on items; an ntfy hook example is documented.

## Dependencies

R8.1 and R8.2 define the main item types; can land with the subset that exists today (halted run, merge gate, budget).

Full plan: `docs/dark-factory-roadmap.md` (R8.3).

0xfauzi:
Implements **R8.3 Exception inbox** (closes [#150](https://github.com/0xfauzi/kstrl/issues/150)), completing R8 Wave 1 and unblocking Wave 3 (continuous intake, #153).

Over-the-loop operation needs a single place to look. Before this, the things wanting a decision were scattered across surfaces sharing nothing: a FAILED component in the manifest, a MERGE_PENDING park in a status string, an R8.1 policy violation in a finding, an R8.2 demotion in the journal, a budget overrun in a log line. Each was discoverable only if you already knew to look for it — the opposite of what "humans handle exceptions" requires.

## Emitters are wired, not declared

This is the lesson from #174, so I led with it. **Every existing halt path feeds the inbox:**

| Halt path | Item kind |
|---|---|
| component FAILED, PR-flow failure | `halted_run` |
| MERGE_PENDING park | `merge_gate` |
| token-budget halt | `budget_overrun` |
| R8.1 policy findings (advisories excluded) | `policy_exception` |
| R8.2 demotion, with triggering evidence | `demotion_notice` |

Verified end-to-end rather than asserted: a run with a planted policy violation produces a `policy_exception`, a `halted_run`, **and** a `demotion_notice`, while the ladder drops L3 → L2. Inbox writes can never fail a run — but they warn, because a silently empty inbox reads exactly like a clean run.

This closes three IOUs left by earlier items: R8.1's "violations route to the inbox", R8.2's "every demotion emits an inbox item carrying the triggering evidence", and the surface R8.6 needs.

## Substrate

`.kstrl/inbox.jsonl`, append-only, folded on read. Every emission and decision is one appended line; a crash mid-write loses at most the last line and never corrupts a prior decision.

Design choices that carry weight:
- **Dedupe by key** — a component failing the same way across three retries is one item with a bumped count, not three. But a repeat *after* a decision opens a fresh item: you approved that failure once, and its recurrence is new information.
- **Snooze has a TTL, not a hide button.** A lapsed snooze reads as open again by the clock, so nothing has to remember to un-hide it. That is the difference between deferring a decision and losing one.
- **The open-item cap is a real backstop** — R8.6 consults it before admitting queue work. An operator already behind should not be handed more.
- **Reject demands a comment.** A bare "no" is not a decision anyone can act on later, least of all the person who made it.

## Surfaces

- `ks inbox ls | show | approve | reject | snooze | retry` — `retry` is a genuine round-trip, resetting the component to PENDING via the manifest so the next `ks factory` picks it up.
- `kstrl/tui/screens/inbox.py` — master-detail triage screen in the home shell, using the *same* `Inbox` methods the CLI calls, so a decision made in either place is indistinguishable and lands in the same log. Requeue is deliberately CLI-only: it mutates the manifest, and that should cost one more deliberate step than a keystroke.
- Notification stays **one-way**: `notifiable()` selects action-required kinds plus demotions, so successes are silent. kstrl still runs no inbound HTTP surface; an ntfy.sh example reuses the existing `[notify]` hook.

## Deliberately not built

`approve-and-amend-policy` (widening the R8.1 envelope from a repeated approval) and the daily digest. Both are learning-loop refinements that want real inbox traffic to design against, and neither is load-bearing for R8.6. Recorded in the roadmap rather than left implied.

## Tests / gates

- 44 inbox tests: append-only log and decision folding, dedupe semantics both ways, snooze TTL expiry, cap behaviour, torn-line tolerance, ambiguous-prefix refusal, the notification taxonomy, config surface, **end-to-end emitter runs**, and a live Textual screen test that approves an item and asserts the store.
- `mypy --strict` clean (107 files), `ruff` clean, full suite **2162 passed / 28 skipped / 0 failed**.

### DIFF
--- a/docs/dark-factory-roadmap.md
+++ b/docs/dark-factory-roadmap.md
@@ -377,7 +377,36 @@ exists and its output over historical `experiments.tsv` is captured here.
 
 ## R8.3 Exception inbox (L) - [#150](https://github.com/0xfauzi/kstrl/issues/150)
 
-Status: `[ ]` - Depends on: R8.1/R8.2 item types (can land with today's subset)
+Status: `[x]` - Shipped in `kstrl/inbox.py` + `ks inbox` +
+`kstrl/tui/screens/inbox.py`. Append-only `.kstrl/inbox.jsonl` folded on
+read; item kinds policy_exception / merge_gate / halted_run /
+budget_overrun / demotion_notice / calibration_drift; dedupe by key,
+snooze with a TTL that RETURNS the item, and an open-item cap that R8.6
+will consult before admitting queue work.
+
+**Emitters are wired, not declared.** Every existing halt path feeds it:
+component FAILED and PR-flow failure (halted_run), MERGE_PENDING
+(merge_gate), token-budget halt (budget_overrun), R8.1 policy findings
+(policy_exception, advisories excluded), and R8.2 demotions
+(demotion_notice, carrying the triggering evidence - the item R8.2
+promised). Verified end-to-end: a run with a planted policy violation
+produces a policy_exception, a halted_run, AND a demotion_notice while
+the ladder drops L3 -> L2.
+
+Closes three IOUs left by earlier items: R8.1's "violations route to the
+inbox", R8.2's "every demotion emits an inbox item", and the surface R8.6
+needs for `stop_at_pr`.
+
+Notification stays one-way, as specified: `notifiable()` selects
+action-required kinds plus demotions, so successes are silent; kstrl runs
+no inbound HTTP surface, and items are actioned in `ks inbox` or the TUI
+screen. An ntfy.sh example is documented in `docs/env-vars.md` under the
+existing `[notify]` hook.
+
+Not built: `approve-and-amend-policy` (widening the R8.1 envelope from a
+repeated approval) and the daily digest. Both are learning-loop
+refinements that want real inbox traffic to design against, and neither
+is load-bearing for R8.6.
 
 **Why.** Over-the-loop operation needs one surface for everything awaiting a
 human decision. Today: SpecBlocker exit codes, FAILED components,
--- a/docs/dark-factory-roadmap.md
+++ b/docs/dark-factory-roadmap.md
@@ -384,14 +384,21 @@ budget_overrun / demotion_notice / calibration_drift; dedupe by key,
 snooze with a TTL that RETURNS the item, and an open-item cap that R8.6
 will consult before admitting queue work.
 
-**Emitters are wired, not declared.** Every existing halt path feeds it:
-component FAILED and PR-flow failure (halted_run), MERGE_PENDING
-(merge_gate), token-budget halt (budget_overrun), R8.1 policy findings
+**Emitters are wired, not declared.** The halt paths that feed it:
+component FAILED and PR-flow failure (halted_run), MERGE_PENDING and the
+pre-merge checkpoint that no interactive UI can answer (merge_gate),
+token-budget halt (budget_overrun), R8.1 policy findings
 (policy_exception, advisories excluded), and R8.2 demotions
 (demotion_notice, carrying the triggering evidence - the item R8.2
 promised). Verified end-to-end: a run with a planted policy violation
 produces a policy_exception, a halted_run, AND a demotion_notice while
 the ladder drops L3 -> L2.
+
+`pause_before_pr_merge` on an unattended run no longer proceeds. It
+returns `CheckpointDecision.PARKED`: the merge is withheld, the
+component fails at `phase=pr / check=merge_gate`, and the decision goes
+to the inbox. Proceeding defeated the gate in exactly the unattended
+case R8.2's L1/L2 forces it on for.
 
 Closes three IOUs left by earlier items: R8.1's "violations route to the
 inbox", R8.2's "every demotion emits an inbox item", and the surface R8.6
@@ -400,13 +407,21 @@ needs for `stop_at_pr`.
 Notification stays one-way, as specified: `notifiable()` selects
 action-required kinds plus demotions, so successes are silent; kstrl runs
 no inbound HTTP surface, and items are actioned in `ks inbox` or the TUI
-screen. An ntfy.sh example is documented in `docs/env-vars.md` under the
-existing `[notify]` hook.
+screen. The push itself is a dedicated `[notify].on_inbox_item` command,
+empty by default: a failing component already fires `on_first_failure`
+and raises an item for the same event, so sharing one command would page
+twice for one thing. An ntfy.sh example is in `docs/env-vars.md`.
 
-Not built: `approve-and-amend-policy` (widening the R8.1 envelope from a
-repeated approval) and the daily digest. Both are learning-loop
-refinements that want real inbox traffic to design against, and neither
-is load-bearing for R8.6.
+Not built, and NOT claimed as done:
+
+- `approve-and-amend-policy` (widening the R8.1 envelope from a repeated
+  approval) and the daily digest. Both are learning-loop refinements that
+  want real inbox traffic to design against, and neither is load-bearing
+  for R8.6.
+- Inbox emission from `SpecBlockerError` exits and from terminal contract
+  failures. Review of PR #175 established that the original "every halt
+  path feeds it" claim was too broad: those two paths still bypass the
+  inbox.
 
 **Why.** Over-the-loop operation needs one surface for everything awaiting a
 human decision. Today: SpecBlocker exit codes, FAILED components,
--- a/kstrl/inbox.py
+++ b/kstrl/inbox.py
@@ -1,0 +1,568 @@
+"""R8.3 exception inbox: one surface for everything awaiting a human.
+
+Over-the-loop operation needs a single place to look. Today the things
+that want a decision are scattered across surfaces that share nothing: a
+FAILED component lives in the manifest, a MERGE_PENDING park lives in a
+status string, an R8.1 policy violation lives in a finding, an R8.2
+demotion lives in the evolution journal, a budget overrun lives in a log
+line. Each is discoverable only if you already know to look for it, which
+is the opposite of what "humans handle exceptions" requires.
+
+This module is the thin substrate: an append-only ``.kstrl/inbox.jsonl``
+plus the fold that turns it into a current view. Deliberately small - the
+value is in the emitters that feed it and the actions that resolve it,
+not in the store.
+
+**Append-only, never rewritten.** Every emission and every decision is one
+appended line; the current state of an item is the fold of its lines in
+order. A crash mid-write can lose at most the last line, never corrupt a
+prior decision, and "what did I approve and when?" is answerable by
+reading the file rather than trusting a mutated record.
+
+Design notes that carry weight:
+
+- **The open-item cap is a real backstop, not decoration.** An inbox that
+  grows without bound is a second job, and the roadmap's own failure mode.
+  ``InboxConfig.open_item_cap`` is what R8.6 will consult before admitting
+  more queue work.
+- **Snooze has a TTL, not a hide button.** A snoozed item returns; that is
+  the difference between deferring a decision and losing one.
+- **Notifications are one-way.** Items are actioned in ``ks inbox``, never
+  by a callback into the harness, so kstrl runs no inbound HTTP surface.
+"""
+
+from __future__ import annotations
+
+import json
+import os
+import tempfile
+import uuid
+from collections.abc import Iterable, Sequence
+from dataclasses import dataclass, field
+from datetime import UTC, datetime, timedelta
+from enum import StrEnum
+from pathlib import Path
+from typing import Any
+
+INBOX_FILENAME = ".kstrl/inbox.jsonl"
+INBOX_SCHEMA_VERSION = 1
+
+
+class ItemKind(StrEnum):
+    """What kind of decision an item is asking for.
+
+    The kinds map one-to-one onto the halt paths that existed before this
+    module; adding a kind means adding an emitter, never just a label.
+    """
+
+    POLICY_EXCEPTION = "policy_exception"      # R8.1 envelope violation
+    MERGE_GATE = "merge_gate"                  # a merge awaiting approval
+    HALTED_RUN = "halted_run"                  # a component/run stopped
+    BUDGET_OVERRUN = "budget_overrun"          # a cap was hit
+    DEMOTION_NOTICE = "demotion_notice"        # R8.2 autonomy revoked
+    CALIBRATION_DRIFT = "calibration_drift"    # detection rate moved
+
+    @property
+    def action_required(self) -> bool:
+        """Whether this kind blocks on a human, as opposed to informing.
+
+        Drives notification: action-required kinds and demotions notify;
+        informational kinds stay silent and batch into a digest. Alert
+        fatigue is the failure mode being defended against - a surface
+        that pages on success trains you to ignore it.
+        """
+        return self in {
+            ItemKind.POLICY_EXCEPTION,
+            ItemKind.MERGE_GATE,
+            ItemKind.HALTED_RUN,
+            ItemKind.BUDGET_OVERRUN,
+        }
+
+
+class ItemStatus(StrEnum):
+    OPEN = "open"
+    APPROVED = "approved"
+    REJECTED = "rejected"
+    SNOOZED = "snoozed"
+    RESOLVED = "resolved"     # actioned elsewhere / no longer relevant
+
+
+class Priority(StrEnum):
+    """Notification tier. Not a sort key alone - age matters too."""
+
+    HIGH = "high"
+    NORMAL = "normal"
+    LOW = "low"
+
+
+#: Default priority per kind. A demotion is high because autonomy was
+#: revoked and the evidence is perishable; drift is low because it is a
+#: trend, not an event.
+DEFAULT_PRIORITY: dict[ItemKind, Priority] = {
+    ItemKind.POLICY_EXCEPTION: Priority.HIGH,
+    ItemKind.MERGE_GATE: Priority.NORMAL,
+    ItemKind.HALTED_RUN: Priority.NORMAL,
+    ItemKind.BUDGET_OVERRUN: Priority.HIGH,
+    ItemKind.DEMOTION_NOTICE: Priority.HIGH,
+    ItemKind.CALIBRATION_DRIFT: Priority.LOW,
+}
+
+_PRIORITY_ORDER = {Priority.HIGH: 0, Priority.NORMAL: 1, Priority.LOW: 2}
+
+
+class InboxError(RuntimeError):
+    """An inbox action was refused (unknown id, already decided, cap hit)."""
+
+
+def _utc_now() -> datetime:
+    return datetime.now(UTC).replace(microsecond=0)
+
+
+def _iso(moment: datetime) -> str:
+    return moment.isoformat()
+
+
+def _parse_iso(value: str) -> datetime | None:
+    try:
+        parsed = datetime.fromisoformat(value)
+    except (TypeError, ValueError):
+        return None
+    return parsed if parsed.tzinfo else parsed.replace(tzinfo=UTC)
+
+
+@dataclass
+class InboxItem:
+    """One decision awaiting (or having received) a human.
+
+    ``dedupe_key`` collapses repeats: the same component failing the same
+    way across three retries is one item with a bumped count, not three
+    items. Without it the inbox becomes noise on exactly the runs where it
+    matters most.
+    """
+
+    id: str
+    kind: ItemKind
+    title: str
+    created_at: str
+    detail: str = ""
+    priority: Priority = Priority.NORMAL
+    component: str = ""
+    run_id: str = ""
+    dedupe_key: str = ""
+    evidence: dict[str, Any] = field(default_factory=dict)
+    status: ItemStatus = ItemStatus.OPEN
+    occurrences: int = 1
+    last_seen_at: str = ""
+    decided_at: str = ""
+    decided_by: str = ""
+    decision_comment: str = ""
+    snooze_until: str = ""
+
+    @property
+    def is_open(self) -> bool:
+        """Open now: never decided, or snoozed past its TTL.
+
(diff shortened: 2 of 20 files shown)


0xfauzi (commented):
Thorough review completed against issue #150 and the R8.3 roadmap. The CI gates are green, but I found multiple untested behavioral defects, including the missing human merge-gate round trip and incomplete halt-path coverage. These are blocking for the stated acceptance criteria. Focused reproductions also confirmed broken dedupe after a decision, a stale merge gate after successful re-poll, an empty inbox after scheduler timeout, and duplicate budget items. Suggested fixes and regression-test scopes are inline.

kstrl/pipeline.py
```diff
@@ -737,6 +804,14 @@ def _park_merge_pending(
             component=comp.id, pr_url=comp.pr_url, error=comp.error,
         ))
         self.notify.fire_merge_pending(comp.id, comp.error)
+        self._inbox_add(
+            ItemKind.MERGE_GATE,
```
0xfauzi, line 851: [P1] This is not the human merge gate described by #150. `_park_merge_pending` runs only after `push_create_and_merge_pr` has already attempted the merge and confirmation timed out. The actual L1/L2 checkpoint is `_phase_checkpoint`; when it cannot prompt, it currently proceeds automatically and emits no inbox item. Consequently `ks inbox approve` only changes JSONL status and cannot resume or authorize the component, so the explicit acceptance criterion "approving a merge-gate item resumes the component" is unmet. Please park the component before merge when approval is needed, emit the item there, persist enough state to resume safely, and make approve record the decision plus transition/resume that parked component. Add a non-interactive end-to-end test covering park -> approve -> next factory run -> PR flow.

kstrl/pipeline.py
```diff
@@ -737,6 +804,14 @@ def _park_merge_pending(
...
+            f"{comp.id} merge unconfirmed",
+            detail=comp.error,
+            component=comp.id,
+            dedupe_key=f"merge:{comp.id}",
```
0xfauzi, line 855: [P1] This gate is never closed when `repoll_merge_pending()` later observes `merged` or `closed`. I reproduced a `MERGE_PENDING` component with this open key, mocked the re-poll to `merged`, and got component status `completed` while the gate remained open. That stale item continues counting toward `open_item_cap`. Resolve this item when the PR merges; when it closes unmerged, resolve/replace it with a `halted_run`. Add inbox assertions to both existing re-poll tests.

kstrl/inbox.py
```diff
@@ -0,0 +1,568 @@
...
+            raise InboxError(
+                f"ambiguous item id {item_id!r}: matches "
+                + ", ".join(m.id[:8] for m in matches[:5])
+            )
+        return None
+
+    def find_by_dedupe_key(self, key: str) -> InboxItem | None:
+        if not key:
+            return None
+        for item in self.items():
```
0xfauzi: [P1] This returns the oldest matching generation because `items()` sorts matching items by ascending `created_at`. After an item is approved and the same failure recurs, the new open item is created correctly, but the following recurrence finds the old approved item again and creates a second open item instead of incrementing the existing one. I reproduced `approve(first); second=add(key); third=add(key)` yielding three distinct IDs and two open items with occurrence count 1. Select an existing open generation before any decided generation (and otherwise the newest decided generation), then add a regression test with at least two repeats after the first decision.

kstrl/pipeline.py
```diff
@@ -691,6 +708,21 @@ def fail_for_budget(self, comp: Component, phase: str) -> Transition:
...
+            detail=error,
+            component=comp.id,
+            dedupe_key=f"budget:{comp.id}",
+            evidence={
+                "phase": phase,
+                "total_tokens": self.run_usage.total_tokens,
+                "max_total_tokens": self.factory_config.max_total_tokens,
+            },
+        )
         return self.fail(
```
0xfauzi, line 735: [P2] `fail()` unconditionally emits a generic `halted_run`, so this path creates two open items for one budget event: `budget_overrun` and `halted_run`. I reproduced exactly that pair with a direct `fail_for_budget()` call. This consumes two cap slots and defeats the type-specific routing described above. Let `fail()` accept the item kind/evidence, or suppress its generic emission here, so a token-budget halt produces only `budget_overrun`; assert the exact item set in a budget-emitter test.

kstrl/inbox.py
```diff
@@ -0,0 +1,568 @@
...
+    merges differently, a file gains a line. The thing an operator can
+    lose by having it off is the record of a decision they already had to
+    make, so the safe default is on. ``open_item_cap`` is the backstop
+    R8.6 consults before admitting more queue work.
+    """
+
+    enabled: bool = True
+    open_item_cap: int = 50
+    snooze_hours: float = 24.0
+    notify_action_required: bool = True
```
0xfauzi, line 262: [P2] This public config knob has no runtime consumer. `notifiable()` is only called by unit tests, and no emitter invokes the notify hook based on either this flag or the selected items. Therefore `KSTRL_INBOX_NOTIFY=0` has no effect, while demotion notices never trigger the advertised inbox notification path. Either wire item creation to the existing one-way notify hook while honoring this flag and `notifiable()`, or remove the knob and documentation until that integration exists. Include enabled/disabled tests against the actual hook.

tests/test_inbox.py
```diff
@@ -0,0 +1,506 @@
...
+        kstrl_dir = tmp_path / "scripts" / "kstrl"
+        kstrl_dir.mkdir(parents=True, exist_ok=True)
+        (kstrl_dir / "prompt.md").write_text("p")
+        with patch.object(Path, "write_text", Path.write_text):
+            _run_factory(tmp_path, verification=VerificationResult(
+                passed=False, checks=[CheckResult("test_suite", False, "boom")],
+            ))
+        # _run_factory rewrites kstrl.toml with inbox enabled, so assert the
+        # config path directly instead:
+        assert InboxConfig.load(tmp_path).enabled is True
```
0xfauzi: [P2] This test named `test_disabled_inbox_writes_nothing` rewrites the config to `enabled=true` inside `_run_factory` and then asserts that it is true. It never exercises disabled behavior or checks that the inbox is empty, so it can pass if disabled mode is broken. Make `_run_factory` accept the inbox setting (or stop it from rewriting the caller's TOML), run a failing component with `enabled=false`, and assert `Inbox(...).items() == []`.

docs/dark-factory-roadmap.md
```diff
@@ -377,7 +377,36 @@ exists and its output over historical `experiments.tsv` is captured here.
...
 
-Status: `[ ]` - Depends on: R8.1/R8.2 item types (can land with today's subset)
+Status: `[x]` - Shipped in `kstrl/inbox.py` + `ks inbox` +
+`kstrl/tui/screens/inbox.py`. Append-only `.kstrl/inbox.jsonl` folded on
+read; item kinds policy_exception / merge_gate / halted_run /
+budget_overrun / demotion_notice / calibration_drift; dedupe by key,
+snooze with a TTL that RETURNS the item, and an open-item cap that R8.6
+will consult before admitting queue work.
+
+**Emitters are wired, not declared.** Every existing halt path feeds it:
```
0xfauzi: [P1] This completion claim is not supported by the implementation. A scheduler backstop produces `status=failed` with an empty inbox in a focused reproduction. Terminal contract failures, PRs found closed during merge re-poll, and the `SpecBlockerError` exits called out in #150 also bypass inbox emission. Please inventory every terminal exit/FAILED assignment, route each applicable boundary condition into the inbox, add one test per path, and keep R8.3 unchecked until those tests pass.

kstrl/pipeline.py
0xfauzi: Fixed in bf0da17. Agreed the merge gate is the human gate, not the post-merge timeout.

`_phase_checkpoint` no longer proceeds when the UI cannot prompt. It emits a `merge_gate` inbox item and returns the new `CheckpointDecision.PARKED`; the caller withholds the merge and fails the component with `phase="pr"`, `check="merge_gate"`. `ks inbox retry <id>` resets it to PENDING so the next `ks factory` picks it up.

Worth recording that the method contradicted itself before this: its own docstring said automation should "fail loudly rather than block indefinitely", while the code warned and merged. Proceeding defeated the gate in exactly the unattended case R8.2's L1/L2 forces `pause_before_pr_merge` on for.

Two tests encoded the old behavior and were updated (not deleted) to assert the new one:
- `tests/test_phase_e.py::TestE6HitlCheckpoint::test_non_interactive_ui_parks_for_approval` (renamed) now asserts the component is in `result.failed`, `failed_phase == "pr"`, `failed_check == "merge_gate"`, and exactly one open `merge_gate` item.
- `tests/test_event_stream.py::...::test_checkpoint_events_auto_resolution` now asserts `CheckpointResolved(decision="parked", decided_by="inbox")`.

kstrl/pipeline.py
0xfauzi: Fixed in bf0da17. A `merge_gate` item outliving its question was a real leak: it held an open-item cap slot forever and asked a human to decide something reality had already decided.

`repoll_merge_pending` now closes the loop in both directions:
- PR turns out MERGED -> `_inbox_resolve("merge:<comp>", "PR #N merged")` resolves the open item.
- PR closed without merging -> resolve the merge_gate item (it is no longer a merge decision) and raise a `halted_run` with `dedupe_key=halted:<comp>:pr-closed` instead, alongside the existing FAILED transition.

`_inbox_resolve` is non-fatal like `_inbox_add`: it catches `OSError | ValueError | InboxError` and warns, so a broken inbox cannot fail a run.

What I verified vs assumed: the resolve/raise code paths are exercised by the existing merge-repoll tests in the suite (all green), but I did not add a dedicated test that drives a closed-unmerged PR end-to-end through `repoll_merge_pending` with a stub `gh`. Calling that out rather than claiming coverage I did not add.

kstrl/inbox.py
0xfauzi: Fixed in bf0da17. This was the worst of the seven: dedupe fanning out is the exact noise dedupe exists to prevent.

`items()` is sorted ascending by age, so the old loop returned the OLDEST match. After a decision, the next occurrence re-found the decided original, `add()` saw a non-OPEN status and opened a fresh item; the occurrence after that re-found the same decided original again and opened another. One item per occurrence.

`find_by_dedupe_key` now prefers the OPEN generation and otherwise the newest decided one. One extra detail the fix needed: "newest" is by LOG POSITION, not `created_at`. `_utc_now()` truncates to the second, so two generations of a key raised in the same second tie, and `max(..., key=created_at)` returns the first max encountered - the oldest - putting the tie-break right back where the bug was. Added `Inbox._folded()`, which returns items in first-appearance log order (a total order); `items()` is now that, sorted.

Verified repro, `tests/test_inbox.py::TestDedupeAcrossGenerations`:
- `test_two_repeats_after_a_decision_share_one_item`: add, approve, add, add -> the two repeats share ONE id distinct from the approved original; `open_items()` has length 1 with `occurrences == 2`.
- `test_many_repeats_stay_one_open_item`: 5 repeats after a decision -> 1 open item, `occurrences == 5`.
- `test_open_generation_wins_over_older_decided` and `test_all_decided_falls_back_to_the_newest` pin the two branches of the selection rule directly.

`add()` still treats a snoozed item as not-OPEN, so I left `find_by_dedupe_key` matching on `status is OPEN` rather than the `is_open` property. Widening it to `is_open` would return a lapsed-snooze item that `add()` then declines to bump, which reintroduces a fan-out. Behaviour under an active snooze is unchanged by this PR.

kstrl/pipeline.py
0xfauzi: Fixed in bf0da17. One event was consuming two cap slots and burying its own reason: `fail_for_budget` raised `budget_overrun` and then delegated to `fail()`, which adds a generic `halted_run` for every terminal failure.

`fail_for_budget` now calls `_inbox_suppress_generic(comp.id)` before delegating; `fail()` checks that marker, consumes it, and skips the generic item. The same marker is used on the new PARKED merge-gate path (finding #1) for the same reason - the typed `merge_gate` item already says what happened.

Test rewritten to drive the real halt rather than assert a shape: `tests/test_inbox.py::TestBudgetEmitsOneItem` now runs the factory with `max_total_tokens=100` and an engineer result carrying 500 recorded tokens, so `fail_for_budget` fires for real, and asserts `[str(i.kind) for i in open_items()] == ["budget_overrun"]`. The previous version of that test manually `add()`-ed a budget item after a clean run, which proved nothing about the suppression - your finding applied to my test as much as to the code.

kstrl/inbox.py
0xfauzi: Fixed in bf0da17, and the first attempt at this fix was wrong in an instructive way - recording it honestly.

You were right that `[inbox] notify_action_required` was a documented no-op. My first wiring gave it a consumer by having `NotifyHooks.fire_inbox_item` REUSE the `on_first_failure` command, with `ComponentPipeline._inbox_add` calling it for any notifiable item. That double-notified: a failing component fires `fire_first_failure` AND raises a `halted_run` item describing the same event, so one failure sent two pushes. Two existing "fired exactly once" tests in `tests/test_notify.py` caught it. Those assertions were correct, so the wiring changed, not the tests.

What shipped instead: a dedicated `[notify].on_inbox_item` key, default empty, env `KSTRL_NOTIFY_ON_INBOX_ITEM`. `fire_inbox_item` fires that command only, so an operator who wants per-item pushes opts in, and the default stays silent. `notify_action_required` gates whether an item is offered to the notifier at all; `on_inbox_item` decides whether anything is actually sent. The condition key is `inbox_<kind>`, so `KSTRL_NOTIFY_EVENT` arrives as e.g. `inbox_merge_gate` and one command can route by kind - and once-per-condition still holds, so two components raising `halted_run` push once.

Config surface moved together in the same commit: `kstrl.toml.example`, `docs/env-vars.md` (both the `[notify]` table and the `[inbox]` ntfy.sh example, which now uses `on_inbox_item`), `scripts/gen_docs.py` KEY_DESCRIPTIONS, the regenerated README block, and the notify key list in `kstrl/config_report.py` so `ks config show` resolves it. No change needed in `kstrl/init_cmd.py` - `DEFAULT_KSTRL_TOML` has no `[notify]` block, and `EXPECTED_SCAFFOLD_KEYS` therefore has no `notify` section.

Tests: `tests/test_notify.py::TestFactoryFiresHooks::test_inbox_hook_is_silent_until_opted_in` (a real failing run raises 2 inbox items and pushes nothing) and `test_opted_in_inbox_hook_fires_without_doubling_the_failure_hook` (with both set, `on_first_failure` still fires exactly once and `on_inbox_item` fires once), plus unit coverage in `tests/test_inbox.py::TestNotifyWiring` using counting shell commands instead of a patched `_fire`.

tests/test_inbox.py
0xfauzi: Fixed in bf0da17. The test asserted the opposite of its own name: the `_run_factory` helper unconditionally rewrote `kstrl.toml` with the inbox ENABLED, so the test had to settle for `assert InboxConfig.load(tmp_path).enabled is True` with a comment explaining why. It proved the helper overwrote the config, nothing more.

`_run_factory` now takes an `inbox_enabled` parameter and writes it into the `[inbox]` section, so the test configures a genuinely disabled inbox, runs a failing component through the factory, and asserts `Inbox(tmp_path, InboxConfig()).items() == []` - no log written, nothing to fold. The `patch.object(Path, "write_text", Path.write_text)` no-op that was working around the same problem is gone.

Two other tests I had added in this PR were vacuous by the same standard and were rewritten to drive real paths rather than shapes: the budget one-item test (see the reply on finding #4) and `TestMergeGateIsTheHumanGate::test_non_interactive_gate_parks_and_emits`, which previously ended in `assert Inbox(...).items() is not None` because the helper had `create_prs=False` and never reached the gate. The helper now takes `create_prs` and `pause_before_pr_merge`, so that test actually reaches the checkpoint and asserts exactly one open `merge_gate` item for `comp-a`.

docs/dark-factory-roadmap.md
0xfauzi: Fixed in bf0da17. "Every existing halt path feeds it" was not true, and a roadmap that overclaims is worse than one that admits a gap.

`docs/dark-factory-roadmap.md` R8.3 now:
- opens the emitter paragraph with "The halt paths that feed it:" instead of "Every existing halt path", and enumerates them;
- adds an explicit "Not built, and NOT claimed as done" bullet for inbox emission from `SpecBlockerError` exits and from terminal contract failures, naming this review as what established the original claim was too broad. Those two paths still bypass the inbox.

Two related corrections in the same section while I was there: `merge_gate` now covers both MERGE_PENDING and the pre-merge checkpoint that no interactive UI can answer (finding #1), and the notification paragraph names the dedicated `[notify].on_inbox_item` key and why it is not a reuse of `on_first_failure` (finding #5).

I did not add emitters for those two paths in this PR. Closing the gap in the doc rather than silently widening the diff.

kstrl/pipeline.py
0xfauzi: Correction to my own reply above: I did add the coverage, in bc978f8. `tests/test_pipeline.py::TestMergePendingRepoll` seeds the `merge:comp-a` item the park would have raised, then asserts `open_items() == []` after a MERGED re-poll, and `[halted_run]` after a CLOSED one (proving the merge_gate item was resolved rather than left alongside the new item). Both branches were already driven by those tests; only the inbox assertions were missing.

0xfauzi:
All seven findings addressed in `bf0da17` (+ `bc978f8` for follow-up test coverage). Every one was valid; two of them applied to tests I had written, not just to the code. Per-finding detail is in the inline replies; summary and what was verified vs assumed below.

## What changed

| # | Finding | Fix |
|---|---|---|
| 1 | merge gate is not the human gate | `_phase_checkpoint` returns the new `CheckpointDecision.PARKED` when the UI cannot prompt: merge withheld, `merge_gate` item filed, component fails at `phase=pr / check=merge_gate` |
| 2 | stale merge_gate after re-poll | `_inbox_resolve` on MERGED; resolve + `halted_run` on closed-without-merge |
| 3 | dedupe returned the oldest generation | prefers the OPEN generation, else the newest decided, ranked by log position (not second-resolution `created_at`) |
| 4 | budget halt emitted 2 items | typed-item marker suppresses the generic `halted_run` from `fail()` |
| 5 | `notify_action_required` had no consumer | new dedicated `[notify].on_inbox_item` key (default empty), NOT a reuse of `on_first_failure` |
| 6 | vacuous disabled-inbox test | helper takes `inbox_enabled`; test asserts `items() == []` |
| 7 | roadmap claim too broad | roadmap now names the emitters and explicitly does NOT claim `SpecBlockerError` exits or terminal contract failures |

## Two things worth flagging

**Finding #1 exposed a contradiction, not just a bug.** `_phase_checkpoint`'s own docstring said automation should "fail loudly rather than block indefinitely", while the code warned and merged. Proceeding defeated the gate in exactly the unattended case R8.2's L1/L2 forces `pause_before_pr_merge` on for. Two tests encoded the old behaviour and were updated to assert the new one (`test_non_interactive_ui_parks_for_approval`, and the `CheckpointResolved` event now carrying `decision="parked" / decided_by="inbox"`). Neither was deleted or weakened.

**My first fix for #5 was wrong.** I initially had `fire_inbox_item` reuse the `on_first_failure` command, which double-notified: a failing component fires the failure hook AND raises an item for the same event. Two existing "fired exactly once" tests caught it. Those assertions were right, so the wiring changed rather than the tests. The shipped version is a separate opt-in key, so the default is silent.

## Behaviour changes a user can see

- `pause_before_pr_merge` on an unattended run now FAILS the component instead of merging it. This is the intended change but it is a behaviour change, not a refactor: unattended runs that previously merged will now park. `ks inbox retry <id>` resets the component to PENDING for the next `ks factory`.
- New config key `[notify].on_inbox_item` / `KSTRL_NOTIFY_ON_INBOX_ITEM`, default empty. The `[inbox]` ntfy.sh example in `docs/env-vars.md` was switched to it. Existing `on_first_failure` hooks are unaffected.

## Verified vs assumed

Verified: `uv run pytest tests/ -q` -> **2174 passed, 28 skipped, 0 failed** (was 2166 passed / 4 failed); `uv run mypy kstrl/ --strict` -> clean, 107 files; `uv run ruff check kstrl/ tests/ scripts/` -> clean; `uv run python scripts/gen_docs.py --check` -> README generated sections current. The dedupe repro is a real assertion, not a claim: add / approve / add / add yields ONE open item with `occurrences == 2`.

Assumed, and stated as such: I have not exercised a real ntfy.sh endpoint - the notify tests use counting shell commands. The inbox emitters for `SpecBlockerError` exits and terminal contract failures are still missing; that is now documented as not-done rather than fixed.

Per repo rule H1 I did not run `/code-review` on my own changes.

PR #176, opened 2026-07-27
fix: enforce max_total_tokens during the engineer loop, not just after it

0xfauzi:
## The defect

`[factory] max_total_tokens` was a **post-hoc detector, not a limiter**. It was consulted only at parent-process phase boundaries:

- `kstrl/pipeline.py` - after the engineer returns, before review, before security, before distill
- `kstrl/factory.py` - the scheduling gate and the end-of-run summary

`grep max_total_tokens kstrl/loop.py kstrl/agents/*.py` returned empty: the engineer loop and every adapter were structurally unaware of the run-level cap. With `[run] max_iterations = 10` and a 1800s per-iteration timeout, ONE component could spend for hours before the cap was consulted even once. It could stop the next phase; it could never stop the current one.

The only genuine in-loop control, `max_budget_usd`, is claude-sdk only. On the claude-code adapter there was no in-loop limit at all.

## The fix

**Loop side** (`kstrl/loop.py`): a new `LoopBudget` carries the cap plus the run's spend as of the worker's launch. It is evaluated BETWEEN iterations - after the completion return, before the no-progress breaker's stall probe - so a blown budget pays for neither another agent call nor another breaker test command.

**Unknown usage is not silently zero.** Calls that report nothing count as zero while *other* calls do report (the total stays a documented lower bound that still grows toward the cap). But if nothing in the whole run has reported after two agent calls, the cap provably cannot trip - a custom `agent_cmd` never reports usage - and the loop halts on *that* instead of pretending to enforce. One silent call stays an incident: the claude adapter records no counts for a timed-out or unparseable result, and a single bad iteration must not end a capped run. The threshold of 2 is a judgment call, documented as such next to the constant.

**Wiring**: `_submit_args` snapshots the budget per launch, `_run_component` forwards it and reports `ComponentResult.budget_exceeded`, and `process_result` routes that flag into the **existing** `fail_for_budget`. The audit trail is therefore identical wherever the breach is caught: `BudgetExceeded` event, typed `infrastructure_error` finding, exactly one `budget_overrun` inbox item, no duplicate `halted_run`. `fail_for_budget` gained a `reason` override so the unenforceable case does not record a false "N >= cap" sentence.

## What the cap actually guarantees now

**It is not a hard cap.** The engineer loop starts no new iteration once the run's reported spend reaches the cap. What stays unbounded, stated in the code, `kstrl.toml.example`, `docs/env-vars.md`, and `docs/sdk-spike.md`:

| Gap | Why |
|---|---|
| The iteration already in flight | Nothing interrupts a single agent call mid-run. Overshoot is up to one iteration per running worker; `agent_iteration` bounds that in wall clock, never in tokens |
| Concurrent workers | Each sees the run total as of its own launch. With `max_parallel = N`, up to N iterations can be in flight past the cap |
| Unreported spend | Every figure is a CLI self-report; totals are lower bounds whenever `unreported_calls > 0` |

`[agent] budget_usd` remains the only genuine per-turn ceiling, and only the claude-sdk adapter enforces it.

## Secondary: abort-path usage loss (fixed)

`_abort_inflight` killed the worker and cancelled its future without reading a result, so `usage_records` died with the process while organic failures carried usage correctly. Workers now publish an atomic per-iteration usage snapshot (mkstemp + `os.replace`). It is deliberately a **file, not an event**: the reducer sums `component_usage`, so streaming cumulative snapshots would double count in every rollup. On abort, a future that already delivered is authoritative; otherwise the snapshot is read back. The two routes are exclusive by construction. Spend inside the interrupted iteration remains unrecoverable - it was never reported to anyone before the kill.

## Behavior change to be aware of

A run with `max_total_tokens` set **and** a non-reporting agent (`agent_cmd`) now halts after two iterations with `token budget unenforceable`, where previously it ran to completion under a decorative cap. Runs without a cap (`= 0`, the default) are unaffected.

## Tests

26 new behavioral tests in `tests/test_usage_meter.py`: the loop stops early (iteration count asserted, not just the flag), prior-component spend counts, the unenforceable / mixed-reporting / one-silent-call policies, the scheduler hands down the right budget with the right prior spend, the in-loop halt produces the same audit state as the phase-boundary halt, and the abort path keeps the spend.

Verified by mutation, not by presence: disabling the in-loop check fails 5 tests, reverting the pipeline routing fails 2, removing the abort salvage fails 1. Existing budget tests are unmodified.

## Gates

- `uv run pytest tests/ -q` - **2200 passed, 28 skipped, 0 failed** (baseline 2174 passed)
- `uv run mypy kstrl/ --strict` - clean, 107 source files
- `uv run ruff check kstrl/ tests/ scripts/` - clean

No adversarial prompt was touched, so no calibration run is implicated (H2/H3).

### DIFF
--- a/kstrl/factory.py
+++ b/kstrl/factory.py
@@ -978,8 +978,14 @@ def _read_partial_usage(path: Path) -> UsageTotals | None:
 
     Returns None when the file is absent, unreadable, or not a usage
     dict - the abort path then records nothing rather than guessing.
-    ``unreported_calls`` is a derived property and is deliberately not
-    read back.
+    ``unreported_calls`` and ``tokenless_calls`` are derived properties
+    and are deliberately not read back.
+
+    Backward compatible on read: ``token_calls`` (added in R8 after the
+    review) is optional and defaults to 0 when the payload predates it.
+    That under-reports token coverage rather than inventing it, and the
+    file is rewritten at the next iteration boundary anyway - it is a
+    per-run, per-attempt scratch file, never a long-lived record.
     """
     try:
         data = json.loads(path.read_text(encoding="utf-8"))
@@ -989,8 +995,9 @@ def _read_partial_usage(path: Path) -> UsageTotals | None:
         return None
     totals = UsageTotals()
     int_fields = (
-        "calls", "known_calls", "input_tokens", "output_tokens",
-        "cache_read_tokens", "cache_creation_tokens", "total_tokens",
+        "calls", "known_calls", "token_calls", "input_tokens",
+        "output_tokens", "cache_read_tokens", "cache_creation_tokens",
+        "total_tokens",
     )
     for name in int_fields:
         value = data.get(name)
@@ -1007,6 +1014,27 @@ def _read_partial_usage(path: Path) -> UsageTotals | None:
     if totals.calls == 0:
         return None
     return totals
+
+
+def _clear_partial_usage(path: Path) -> None:
+    """Retire the previous attempt's usage snapshot (R8 review, P2-c).
+
+    The snapshot is keyed by run + component only, and a normal result
+    leaves the file on disk, so without this the NEXT attempt inherits
+    it: a retry cancelled before its own first iteration boundary got
+    salvaged with the previous attempt's tokens, which
+    ``process_result`` had already recorded - a clean double count.
+
+    Called by the scheduler in the PARENT immediately before every
+    submission, not by the worker, because the window that matters is
+    exactly the one where the worker never starts (cancelled or killed
+    during pool startup). Best effort: a snapshot we failed to delete is
+    an accounting risk, never a reason to fail a component.
+    """
+    try:
+        path.unlink()
+    except OSError:
+        return
 
 
 def _run_component(
@@ -1039,6 +1067,7 @@ def _run_component(
     sandbox_allow_network: bool = False,
     agent_budget_usd: float | None = None,
     events_dir_str: str | None = None,
+    usage_dir_str: str | None = None,
     run_id: str = "",
     token_budget: LoopBudget | None = None,
     redirect_output: bool = True,
@@ -1059,6 +1088,14 @@ def _run_component(
     the parent's UI so sequential runs keep live engineer output.
     Without ``events_dir_str`` the legacy PlainUI-on-stderr behavior is
     preserved for direct callers.
+
+    R8: ``usage_dir_str`` is where the per-iteration usage snapshot is
+    published so a killed worker's spend is recoverable. It is a
+    SEPARATE argument from ``events_dir_str`` on purpose - the factory
+    always sets it, including when progress logging is off, because
+    accounting must not be switchable off by an observability setting
+    (review finding P2-d). Direct callers that leave it None simply do
+    not publish snapshots.
 
     R8: ``token_budget`` carries the run-level ``max_total_tokens`` plus
     the spend recorded before this worker launched, so the engineer loop
@@ -1251,9 +1288,13 @@ def _run_component(
     # R8: durable per-iteration usage snapshot. The parent reads it only
     # when a worker was killed before it could return a ComponentResult
     # (PR B abort path), so it can never double count with result.usage.
+    # Keyed off usage_dir_str, NOT events_dir_str: the accounting file
+    # is written even when progress logging is disabled (P2-d).
     on_iteration_usage: Callable[[UsageTotals], None] | None = None
-    if run_paths is not None:
-        usage_path = run_paths.engineer_usage(component_id)
+    if usage_dir_str is not None:
+        usage_path = RunPaths(root=Path(usage_dir_str)).engineer_usage(
+            component_id,
+        )
         on_iteration_usage = functools.partial(
             _write_partial_usage, usage_path,
         )
@@ -1519,6 +1560,17 @@ def _salvage_aborted_usage(
       killed yields a complete (if slightly stale) snapshot. Spend
       inside the interrupted iteration is NOT recoverable - it was never
       reported to anyone before the kill.
+
+    The snapshot is attempt-scoped by the scheduler, which deletes it
+    immediately before every submission (``_clear_partial_usage``). That
+    is what makes this route safe: without it a completed attempt left
+    its file behind and a later attempt aborted before its own first
+    iteration boundary salvaged the PREVIOUS attempt's tokens on top of
+    the ones ``process_result`` had already recorded (review finding
+    P2-c, reproduced as a clean 2x double count).
+
+    Reads ``pipeline.usage_paths``, not ``run_paths``: accounting
+    storage exists even when progress logging is off (P2-d).
     """
     if future.done() and not future.cancelled():
         try:
@@ -1528,10 +1580,10 @@ def _salvage_aborted_usage(
         if result is not None and result.usage is not None:
             pipeline.record_engineer_usage(comp_id, result.usage)
         return
-    run_paths = pipeline.run_paths
-    if run_paths is None:
+    usage_paths = pipeline.usage_paths
+    if usage_paths is None:
         return
-    totals = _read_partial_usage(run_paths.engineer_usage(comp_id))
+    totals = _read_partial_usage(usage_paths.engineer_usage(comp_id))
     if totals is not None:
         pipeline.record_engineer_usage(comp_id, totals)
 
@@ -1848,6 +1900,17 @@ def _run_factory_locked(
     journal_path: Path | None = None
     run_paths: RunPaths | None = None
     run_file_sinks: list[EventSink] = []
+    # R8 (review finding P2-d): accounting storage is allocated for
+    # EVERY run, including progress_log_enabled = false. It used to be
+    # `run_paths`, which is gated on progress logging, so turning the
+    # observability opt-out on also deleted the only place a worker
+    # could publish its spend - a killed worker then recorded nothing
+    # and the run silently under-reported. Money is not observability:
+    # an opt-out may drop the narration, never the meter. This is the
+    # same directory as `run_paths` when progress logging is on; when it
+    # is off the run writes exactly one small JSON file per component
+    # attempt under .kstrl/runs/<run_id>/ and no events at all.
+    usage_paths = RunPaths.for_run(root_dir, run_id)
     if not factory_config.progress_log_enabled:
         progress_log = NullProgressLog()
     else:
@@ -2079,6 +2142,7 @@ def _run_factory_locked(
         bus=bus,
         journal_path=journal_path,
         run_paths=run_paths,
+        usage_paths=usage_paths,
         interaction=interaction,
         notify=notify,
         review_selection=review_selection,
@@ -2299,16 +2363,24 @@ def _run_factory_locked(
             # Chunk 6: worker event channel (None when progress logging
             # is disabled - transcripts and events off together).
             str(run_paths.root) if run_paths is not None else None,
+            # R8: accounting channel. Always set, precisely because the
+            # line above may be None (review finding P2-d).
+            str(usage_paths.root),
             run_id,
             # R8: run-level token ceiling + the spend recorded so far,
             # snapshotted AT LAUNCH so the engineer loop can halt itself
             # between iterations. With max_parallel > 1 a worker cannot
             # see a concurrent sibling's spend, only what the parent had
             # recorded when this worker started.
+            # prior_calls/prior_token_calls carry the run-wide tokenless
+            # call count down so the unenforceable rule does not reset on
+            # every attempt and component (review finding P1-a).
             LoopBudget(
                 max_total_tokens=factory_config.max_total_tokens,
                 prior_total_tokens=pipeline.run_usage.total_tokens,
                 prior_known_calls=pipeline.run_usage.known_calls,
+                prior_calls=pipeline.run_usage.calls,
+                prior_token_calls=pipeline.run_usage.token_calls,
             ),
         )
 
@@ -2384,6 +2456,12 @@ def _run_factory_locked(
                         attempt=comp.retries + 1,
                     ))
                     args = _submit_args(comp, wt_path)
+                    # R8 (P2-c): the usage snapshot is attempt-scoped by
+                    # deletion, and the deletion happens HERE - before
+                    # the worker exists - so an attempt cancelled during
+                    # pool startup cannot salvage its predecessor's
+                    # tokens on top of the ones already on the meter.
+                    _clear_partial_usage(usage_paths.engineer_usage(comp.id))
                     if isinstance(executor, _InlineExecutor):
                         # In-process worker: no fd redirection (it would
                         # hijack the parent terminal), and each transcript
--- a/kstrl/factory.py
+++ b/kstrl/factory.py
@@ -1016,7 +1016,7 @@ def _read_partial_usage(path: Path) -> UsageTotals | None:
     return totals
 
 
-def _clear_partial_usage(path: Path) -> None:
+def _clear_partial_usage(path: Path) -> bool:
     """Retire the previous attempt's usage snapshot (R8 review, P2-c).
 
     The snapshot is keyed by run + component only, and a normal result
@@ -1028,13 +1028,27 @@ def _clear_partial_usage(path: Path) -> None:
     Called by the scheduler in the PARENT immediately before every
     submission, not by the worker, because the window that matters is
     exactly the one where the worker never starts (cancelled or killed
-    during pool startup). Best effort: a snapshot we failed to delete is
-    an accounting risk, never a reason to fail a component.
+    during pool startup).
+
+    Returns True when the slot is provably clean - deleted, or already
+    absent. Returns False when a snapshot may still be on disk, and the
+    caller must then refuse disk salvage for that attempt: deletion IS
+    the attempt-scoping invariant, so swallowing the error left the
+    previous attempt's totals addressable as the current one and
+    re-counted them (review finding on 22e99b4). A worker that cannot
+    delete the file generally cannot overwrite it either, so "we failed
+    to clear it" and "what is there is stale" travel together.
+
+    Never fails the component: an accounting risk is not a reason to
+    stop work, but it IS a reason to distrust the number.
     """
     try:
         path.unlink()
+    except FileNotFoundError:
+        return True          # nothing to retire; the slot is clean
     except OSError:
-        return
+        return False
+    return True
 
 
 def _run_component(
@@ -1582,6 +1596,8 @@ def _salvage_aborted_usage(
         return
     usage_paths = pipeline.usage_paths
     if usage_paths is None:
+        return
+    if not pipeline.usage_salvage_is_safe(comp_id):
         return
     totals = _read_partial_usage(usage_paths.engineer_usage(comp_id))
     if totals is not None:
@@ -2443,6 +2459,19 @@ def _run_factory_locked(
                         pipeline.fail_for_budget(comp, "scheduling")
                         transitioned_without_launch += 1
                         continue
+                    # R8 review: a loop that emits COMPLETE returns before
+                    # its own budget check, so an adapter that finishes on
+                    # its first tokenless call never halts itself - the
+                    # ordinary success path for a custom agent_cmd. This
+                    # gate is what stops the run handing out NEW work
+                    # under a cap that can no longer fire.
+                    unenforceable = pipeline.token_budget_unenforceable()
+                    if unenforceable is not None:
+                        pipeline.fail_for_budget(
+                            comp, "scheduling", reason=unenforceable,
+                        )
+                        transitioned_without_launch += 1
+                        continue
                     pipeline.begin_attempt(comp)
                     manifest.save(manifest_path)
                     bus.emit(ComponentStarted(component=comp.id))
@@ -2466,7 +2495,24 @@ def _run_factory_locked(
                     # the worker exists - so an attempt cancelled during
                     # pool startup cannot salvage its predecessor's
                     # tokens on top of the ones already on the meter.
-                    _clear_partial_usage(usage_paths.engineer_usage(comp.id))
+                    if _clear_partial_usage(
+                        usage_paths.engineer_usage(comp.id)
+                    ):
+                        pipeline.mark_usage_salvage_safe(comp.id)
+                    else:
+                        # A snapshot we could not delete may still hold
+                        # the PREVIOUS attempt's totals, which
+                        # process_result already counted. Refuse disk
+                        # salvage for this attempt rather than risk
+                        # counting them twice - losing an aborted
+                        # attempt's spend understates a total; salvaging
+                        # a stale one corrupts it.
+                        pipeline.mark_usage_salvage_unsafe(comp.id)
+                        ui.warn(
+                            f"  Could not retire the previous usage "
+                            f"snapshot for {comp.id}; disk salvage is "
+                            f"disabled for this attempt"
+                        )
                     if isinstance(executor, _InlineExecutor):
                         # In-process worker: no fd redirection (it would
                         # hijack the parent terminal), and each transcript
--- a/kstrl/factory.py
+++ b/kstrl/factory.py
@@ -2324,6 +2324,7 @@ def _run_factory_locked(
 
     def _submit_args(comp: Component, wt_path: Path) -> tuple[Any, ...]:
         ctx_json = component_contexts.get(comp.id)
+        engineer_usage = pipeline.engineer_usage_totals()
         knowledge_prefix = ""
         if knowledge_config.enabled:
             try:
@@ -2372,15 +2373,19 @@ def _run_factory_locked(
             # between iterations. With max_parallel > 1 a worker cannot
             # see a concurrent sibling's spend, only what the parent had
             # recorded when this worker started.
-            # prior_calls/prior_token_calls carry the run-wide tokenless
+            # prior_calls/prior_token_calls carry the ENGINEER's tokenless
             # call count down so the unenforceable rule does not reset on
-            # every attempt and component (review finding P1-a).
+            # every attempt and component (review finding P1-a), while
+            # staying blind to other roles' timeouts - those say nothing
+            # about whether the engineer's adapter reports tokens.
+            # prior_total_tokens stays run-wide: the overrun check asks
+            # what the RUN has spent against the cap.
             LoopBudget(
                 max_total_tokens=factory_config.max_total_tokens,
                 prior_total_tokens=pipeline.run_usage.total_tokens,
                 prior_known_calls=pipeline.run_usage.known_calls,
-                prior_calls=pipeline.run_usage.calls,
-                prior_token_calls=pipeline.run_usage.token_calls,
+                prior_calls=engineer_usage.calls,
+                prior_token_calls=engineer_usage.token_calls,
             ),
         )
 
--- a/kstrl/factory.py
+++ b/kstrl/factory.py
@@ -9,6 +9,7 @@ import shutil
 import signal
 import subprocess
 import sys
+import tempfile
 import threading
 import time
 from collections.abc import Callable, Mapping
@@ -72,6 +73,7 @@ from kstrl.knowledge import (
     measure_fact_utilization,
 )
 from kstrl.linear import LinearConfig, build_linear_sink
+from kstrl.loop import LoopBudget
 from kstrl.manifest import Component, ComponentStatus, Manifest
 from kstrl.observability import (
     NotifyConfig,
@@ -504,6 +506,11 @@ class ComponentResult:
     # prompt against the same state is the exact spend the breaker
     # exists to stop) with a distinct journal event.
     no_progress: bool = False
+    # R8: the engineer loop halted ITSELF on the run-level token budget
+    # (between iterations). Routed to pipeline.fail_for_budget so the
+    # audit trail is identical to a breach caught at a phase boundary;
+    # ``error`` carries which budget condition fired.
+    budget_exceeded: bool = False
 
 
 @dataclass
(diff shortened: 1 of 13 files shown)


0xfauzi:
Pushed `fc44f69` after reviewing the diff. One real hole in the unenforceable guard, plus two doc/test corrections around it.

**The hole.** The halt summed prior phases into its decision:

```python
known_calls = self.prior_known_calls + loop_usage.known_calls
if loop_usage.calls >= _UNENFORCEABLE_CALLS and known_calls == 0:
```

So one reporting call anywhere earlier in the run disarmed it. Reproduced directly: with `prior_known_calls=3` and a silent loop, `halt_reason` returned `None` for 11 consecutive calls, while the pure-silent case correctly halted at 2. A reporting architect plus a silent engineer meant unbounded engineer spend under a nominal cap - the original defect in a narrower shape.

The mixed case is not weaker than the all-silent one, it is identical: `prior_total_tokens` is frozen at worker launch, so once this loop reports nothing the total can never grow toward the ceiling. Equally dead either way. It is judged on the loop alone now; `prior_known_calls` stays for provenance and is documented as not part of the decision.

Easy configuration to hit: `[agent] command` sets a custom engineer command while the adversarial roles keep a reporting adapter.

**Test that encoded it.** `test_prior_reported_calls_disarm_the_unenforceable_halt` asserted `iterations == 3` for a silent loop - it pinned the defect in place. Rewritten to assert the halt, with the reasoning in the docstring so it is not "fixed" back later. Added `test_one_silent_call_alongside_prior_reporting_keeps_going` to hold the other side: the threshold still separates "this adapter never reports" from "one unreadable result", so a lone silent iteration cannot kill a capped run.

**Docs.** `docs/env-vars.md` described the halt as covering "no call in the whole run" - the wording that described the hole. Corrected, and the gap table now points at the outright-halt case.

Verified unchanged: overrun still fires, pure-silent still halts at 2, a disabled cap stays inert. Full suite 2201 passed, `mypy --strict` and `ruff` clean.

Two things from the original report worth carrying into review, unchanged by this commit: the `ProcessPoolExecutor` path was never exercised across a real process boundary (pinned by a positional-args test plus a pickle round-trip), and the killed-worker snapshot recovery is tested with a never-completing `Future` rather than an actually SIGKILLed worker. Both are honest disclosures rather than defects, but they are the untested seams if this misbehaves later.

kstrl/loop.py
```diff
@@ -27,8 +28,106 @@
...
+            return None
+        total = self.prior_total_tokens + loop_usage.total_tokens
+        if total >= self.max_total_tokens:
+            return (
+                f"token budget exceeded: {total} total tokens recorded >= "
+                f"max_total_tokens ({self.max_total_tokens}); halting the "
+                "engineer loop instead of starting another iteration (R8)"
+            )
+        known_calls = self.prior_known_calls + loop_usage.known_calls
+        if loop_usage.calls >= _UNENFORCEABLE_CALLS and known_calls == 0:
```
0xfauzi: [P1] Include prior run calls in the unenforceable threshold

This condition only counts calls made by the current loop. LoopBudget carries prior_known_calls, but not prior calls, so the documented "after two agent calls in the whole run" rule resets at every attempt/component. I reproduced three sequential one-iteration loops with an unreported record: the accumulated run totals ended at calls=3, known_calls=0, total_tokens=0, while every budget_halt_reason remained empty.

This leaves max_total_tokens decorative whenever work is split into short loops, including max_iterations=1 and retries that fail after one call. Carry prior_calls into LoopBudget, or evaluate the unenforceable rule from pipeline.run_usage at engineer and scheduling boundaries. Please add a regression spanning attempts/components rather than only multiple iterations in one loop.

kstrl/loop.py
```diff
@@ -27,8 +28,106 @@
...
+        if not self.enabled:
+            return None
+        total = self.prior_total_tokens + loop_usage.total_tokens
+        if total >= self.max_total_tokens:
+            return (
+                f"token budget exceeded: {total} total tokens recorded >= "
+                f"max_total_tokens ({self.max_total_tokens}); halting the "
+                "engineer loop instead of starting another iteration (R8)"
+            )
+        known_calls = self.prior_known_calls + loop_usage.known_calls
```
0xfauzi: [P1] Do not use known_calls as evidence that token accounting works

UsageTotals.known_calls increments when a call reports either tokens or cost. Both Claude parsers can therefore produce a cost-only known call when total_cost_usd is present but token fields are absent. Two such records produce calls=2, known_calls=2, total_tokens=0, and this method returns None even though max_total_tokens can never advance.

Track token-reporting coverage separately, for example token_known_calls/reported_token_calls, and use that for this token-budget decision. Keep known_calls for the existing general usage-rollup semantics. Add a cost-only regression using the real UsageTotals.add_record behavior.

kstrl/factory.py
```diff
@@ -1166,18 +1248,36 @@ def _transcript(line: str) -> None:
         )
         stop_heartbeat = _start_heartbeat(worker_bus)
 
+    # R8: durable per-iteration usage snapshot. The parent reads it only
+    # when a worker was killed before it could return a ComponentResult
+    # (PR B abort path), so it can never double count with result.usage.
+    on_iteration_usage: Callable[[UsageTotals], None] | None = None
+    if run_paths is not None:
+        usage_path = run_paths.engineer_usage(component_id)
```
0xfauzi: [P2] Make the snapshot attempt-scoped before a retry can launch

This path is keyed only by run and component, and a normal result leaves the file behind. If the same component retries and is canceled or killed before its first new iteration snapshot, _salvage_aborted_usage reads the previous attempt's totals even though process_result already recorded them. I reproduced a 700-token completed attempt followed by a pending retry Future; the retry salvaged the same 700 tokens again.

Clear or version the snapshot before every submission, preferably in the parent so cancellation before worker startup is covered. An attempt-specific filename/token would also work. Please add a regression where attempt 1 records usage and attempt 2 is aborted before its first boundary.

kstrl/factory.py
```diff
@@ -1385,12 +1486,56 @@ def _abort_inflight(
...
+    if future.done() and not future.cancelled():
+        try:
+            result = future.result(timeout=0)
+        except Exception:  # noqa: BLE001 - a crashed worker reported nothing
+            return
+        if result is not None and result.usage is not None:
+            pipeline.record_engineer_usage(comp_id, result.usage)
+        return
+    run_paths = pipeline.run_paths
+    if run_paths is None:
```
0xfauzi: [P2] Preserve abort accounting when progress logging is disabled

With factory.progress_log_enabled=false, run_paths is None, _submit_args passes events_dir_str=None, and the worker never installs on_iteration_usage. A killed pending Future therefore reaches this return and records nothing, leaving the original abort-path usage loss unfixed for a supported configuration.

The accounting snapshot should not depend on the observability/event-log opt-out. Allocate private accounting storage independently, or document and deliberately enforce a different contract. Add a progress_log_enabled=false abort regression.

tests/test_usage_meter.py
```diff
@@ -880,6 +889,581 @@ def test_unknown_usage_cannot_trip_the_cap(self, tmp_path: Path) -> None:
...
+            return ComponentResult(
+                comp_id, success=True, iterations=1,
+                usage=_engineer_usage(4242),
+            )
+
+        with patch(
+            "kstrl.factory._run_component", side_effect=slow_component,
+        ), patch("kstrl.git.get_diff_content", return_value=""):
+            result = run_factory(
+                manifest, _factory_config(root), _make_base_config(root),
```
0xfauzi, line 1829: [P2] This test does not reach the killed-worker snapshot branch

_factory_config defaults max_parallel=1, so _InlineExecutor runs slow_component synchronously inside submit(). By the time the scheduler observes stop, the Future is already done and _salvage_aborted_usage uses result.usage. The test still passes if snapshot writing/reading is broken.

Please add coverage with an actually pending worker/Future that takes the disk snapshot route. That coverage should include a retry with a stale prior snapshot and progress_log_enabled=false, which expose the production gaps noted above.

0xfauzi (commented):
Do not merge in the current state. The between-iteration token arithmetic is well factored and the new tests cover the normal single-loop path, but I found two ways the documented unenforceable-budget policy remains bypassable and two correctness gaps in abort accounting.

The most important issue is that "none of the calls in the whole run reported tokens" is not represented by the state passed into LoopBudget. Short attempts or components can accumulate unlimited unreported calls without any individual loop reaching the threshold. Separately, known_calls is not a token-reporting signal because cost-only records increment it.

The abort snapshot also needs attempt scoping, and it is unavailable whenever progress logging is disabled. The end-to-end abort test currently exercises the already-delivered Future route, not the killed-worker snapshot route.

Verification performed at head 4d40342:
- uv run pytest tests/ -q: 2200 passed, 28 skipped
- uv run mypy kstrl/ --strict: clean
- uv run ruff check kstrl/ tests/ scripts/: clean
- focused reproductions for each inline finding

Please address the inline findings and add regressions for the cross-attempt/cross-component and killed-worker configurations before merging.

0xfauzi:
Follow-up after re-reviewing the new head `fc44f69`:

The new commit correctly fixes the separate case where prior reporting disarmed an entirely silent current engineer loop. It does not resolve the five inline findings in review 4791920374.

- The unenforceable threshold is still local to one loop. Three sequential one-iteration silent loops still accumulate `calls=3, known_calls=0, total_tokens=0` with an empty `budget_halt_reason` every time. Since `max_total_tokens` is run-level, short attempts/components can still spend indefinitely under an unenforceable cap. The updated docs now describe "this engineer loop", but the PR body still claims "the whole run", and the run-level safety gap remains.
- `loop_usage.known_calls` still includes cost-only records. Two cost-only records still produce `known_calls=2, total_tokens=0` and no unenforceable halt.
- The snapshot lifecycle, progress-log dependency, and inline-executor test are unchanged.

Suggested direction remains: model total prior calls and token-reporting calls explicitly, enforce the unknown-usage policy at parent boundaries as well as within a loop, and decouple/version abort snapshots per attempt.

Current-head verification:
- `uv run pytest tests/ -q`: 2201 passed, 28 skipped

Recommendation remains: do not merge until the open findings are addressed.

kstrl/loop.py
0xfauzi: Fixed in 3fa8ee2. Reproduced first: three sequential `max_iterations=1` loops with a silent adapter and a 500 cap gave `iterations=1 halt="" ` three times, run totals `calls=3 known_calls=0 total_tokens=0`. Exactly as you described.

**What I changed.** Split the rule into the two things it had conflated, as you suggested:

- CALL THRESHOLD, run-wide. New `LoopBudget.prior_calls` / `prior_token_calls` (`kstrl/loop.py:96-101`), populated in `_submit_args` from `pipeline.run_usage` (`kstrl/factory.py:2381-2382`).
- TOKEN EVIDENCE, this loop only (`loop_usage.token_calls == 0`), per fc44f69. `test_prior_reported_calls_do_not_disarm_the_unenforceable_halt` still passes unchanged in behavior; I only added the realistic priors and extended its docstring to say why it survives the new rule.

New condition at `kstrl/loop.py:189-196`.

**The trap you flagged, and what I decided.** The run-wide counter counts TOKENLESS calls (`prior_calls - prior_token_calls` plus this loop's), not all calls. Counting all calls would have been backwards: a call that DID report tokens is evidence the adapter works, so it must not push us toward "this adapter never reports". With tokenless-scoping, a lone unparseable call in an otherwise-reporting run leaves the run-wide count at 1 and the loop keeps going - `test_one_silent_call_alongside_prior_reporting_keeps_going` pins that, and it would have failed under the literal `prior_calls + loop_calls` form.

The residual case, accepted deliberately and documented in the docstring (`kstrl/loop.py:170-180`) and `docs/env-vars.md`: once a run has accumulated one tokenless call ANYWHERE - including a different role's adapter - the engineer's first tokenless iteration reaches the threshold and halts, where an isolated loop would have given it a second chance. Two independent tokenless calls in one run is adapter behavior rather than an incident, and the failure is loud, recorded and recoverable, whereas the alternative is a counter that never fires. `test_a_tokenless_call_after_an_earlier_one_halts_immediately` pins it so the trade is a test, not a comment.

Your edge case: `loop_usage.calls > 0` is now an explicit conjunct, so a worker launched into a run with high `prior_calls` cannot halt before its engineer has run. `test_a_high_prior_call_count_cannot_halt_a_loop_that_has_not_run`.

**On your alternative** (evaluate from `pipeline.run_usage` at the engineer and scheduling boundaries): more information, worse timing. The parent sees the true run-wide totals rather than a launch snapshot, but it only sees them after the engineer loop has already returned - which is the exact window this PR exists to close. It could supplement the in-loop check, never replace it. Feeding the priors down gives the in-loop rule the same information the parent has at launch AND keeps the halt at the earliest point it can fire. Adding a duplicate parent-side rule would mint a second halt reason for one condition and risk two audit records for one breach; the overrun half is already re-checked at every phase boundary via `token_budget_exceeded`. So: in-loop with run-wide priors, no duplicate.

After: loop 1 continues (one tokenless call is an incident), loop 2 halts with `token budget unenforceable ... 2 call(s) in this run have now reported none`. `test_unenforceable_threshold_does_not_reset_per_loop` threads the priors exactly as `_submit_args` does, so it pins the whole channel and not just the arithmetic.

kstrl/loop.py
0xfauzi: Fixed in 3fa8ee2. Confirmed your reading of `add_record` first: two `UsageRecord(cost_usd=...)` records give `calls=2 known_calls=2 total_tokens=0`, `halt_reason -> None`, and `run_loop` then ran all 10 iterations under a 500 cap that could never advance.

**What I changed.** Added `UsageTotals.token_calls` (`kstrl/agents/base.py:62`), incremented only when a token field or `total_tokens` was present (`add_record`, the new `token_known` flag), plus a `tokenless_calls` property. `known_calls` semantics are untouched - it still counts token OR cost - so every existing rollup keeps its meaning. The budget decision now reads `token_calls`.

**Serialization blast radius**, since you flagged it. `to_dict` gains `token_calls`, so:
- `ev.ComponentUsage` gains the field (it is constructed as `ComponentUsage(**totals.to_dict())`, so it had to).
- The v1 compat mapping in `kstrl/events.py` carries it, keeping `progress.jsonl` and `events.jsonl` in agreement about the same event. `ProgressLog.component_usage` is documented as taking `UsageTotals.to_dict()` verbatim, so this is the field following its own contract rather than a new one.
- `test_events.py::test_named_methods_parity` caught the drift immediately; I updated its fixture and commented why, rather than dropping the field from v1 and letting the two files disagree.

**Backward compatible on read.** `event_from_dict` only reads keys it knows, so new files decode in old readers and old files decode with `token_calls=0`. Every v1 consumer (`observability._as_int_field`, `evolution`, the TUI evolve screen) looks keys up defensively, so an added key is inert. `merge` sums it. `_read_partial_usage` treats it as optional and defaults to 0, documented at `kstrl/factory.py:984-989` - conservative rather than invented, and that file is a per-run, per-attempt scratch file rewritten each iteration anyway.

**Regression uses the real `add_record`**, not a hand-built total, since the defect lives in that method: `test_cost_only_record_is_known_but_tokenless` (asserts `known_calls == 2` AND `token_calls == 0` on the same object) and `test_cost_only_reporting_is_not_token_evidence`, which drives a full `run_loop` with a cost-only agent and now halts at 2 iterations instead of 10.

kstrl/factory.py
0xfauzi: Fixed in 3fa8ee2. Reproduced end to end: attempt 1 completes with 700 tokens (recorded by `process_result`) and leaves its snapshot behind, attempt 2's future is still pending when the stop lands, salvage reads the stale file - two `component_usage` engineer events, `[700, 700]`, for 700 tokens of work.

**What I changed.** Clear, in the parent, immediately before every submission - your preferred option, because it is the only placement that covers the window where the worker never starts. New `_clear_partial_usage` (`kstrl/factory.py:1019-1038`), called at `kstrl/factory.py:2459` right after `_submit_args` and before either executor branch. Best effort by design: a snapshot we failed to delete is an accounting risk, never a reason to fail a component.

I chose clearing over an attempt-scoped filename because the filename variant needs the attempt number threaded through the submit tuple (which a test pins positionally) and still leaves stale files from earlier attempts on disk to be reasoned about. Deletion makes the invariant "this file describes the attempt currently in flight, or nothing".

**Regression is exactly the scenario you described**: `test_a_stale_snapshot_from_a_finished_attempt_is_not_recounted` - attempt 1 records usage and writes a boundary snapshot, attempt 2 is aborted before its own first boundary. Asserts the event LIST (`== [700]`), not the sum, because a double count and a correct total are the same number of tokens but a different number of events.

Verified the test depends on the fix: with `_clear_partial_usage` patched to a no-op - i.e. the pre-fix path - it fails with `assert [700, 700] == [700]`.

kstrl/factory.py
0xfauzi: Fixed in 3fa8ee2. Reproduced: with `progress_log_enabled=false`, `.kstrl/runs/` was never created at all, and a killed pending future recorded nothing - `record_engineer_usage` was called zero times. Original abort-path loss, unfixed, in a supported configuration, exactly as you said.

**What I changed** - allocated the accounting storage independently, your first option:

- `usage_paths = RunPaths.for_run(root_dir, run_id)` is computed for EVERY run (`kstrl/factory.py:1903-1913`), outside the `progress_log_enabled` branch. `run_paths` stays gated on progress logging because it drives events and transcripts.
- The worker gets its own `usage_dir_str` argument, separate from `events_dir_str` (`kstrl/factory.py:1070`, passed at `kstrl/factory.py:2366-2368`), and installs `on_iteration_usage` from it (`kstrl/factory.py:1291-1299`). One switch no longer controls two unrelated things.
- `_salvage_aborted_usage` reads `pipeline.usage_paths`, not `run_paths` (`kstrl/factory.py:1583-1589`); `ComponentPipeline` gained the field.

Cost of the opt-out being honored but not total: a run with progress logging off now writes one small JSON per component attempt under `.kstrl/runs/<run_id>/` and no events. I think that is the right side of the trade - an observability opt-out may drop the narration, never the meter - and it is documented in `docs/env-vars.md` ("Usage accounting vs progress logging"), `kstrl.toml.example`, and the generated README line for `progress_log_enabled`.

**Regressions**, both with `progress_log_enabled=false`:
- worker half: `test_worker_persists_usage_without_an_events_dir` - `events_dir_str=None` still publishes the snapshot, and asserts no `events.jsonl` appears (the opt-out is still honored).
- end to end: `test_pending_worker_spend_survives_progress_log_disabled` - genuinely pending future, spend salvaged from disk, asserted through the end-of-run usage rollup (there is no progress log to read, and the rollup is what a human actually sees), plus assertions that neither `progress.jsonl` nor any `events.jsonl` was written.

Verified it depends on the fix: with `_read_partial_usage` stubbed to None it fails with "the aborted worker's spend never reached the meter".

tests/test_usage_meter.py
0xfauzi: Confirmed and fixed in 3fa8ee2. I demonstrated your claim before touching anything: patching `kstrl.factory._read_partial_usage` to return `None` - the disk route completely dead - and re-running `test_stop_mid_run_keeps_the_aborted_component_spend` unchanged. It passes. The test proved nothing about the snapshot.

Root cause is one step earlier than the executor: `_factory_config` sets `use_worktrees=False`, and `_run_factory_locked` forces `max_parallel=1` whenever worktrees are off - so the config knob is not even reachable from these fixtures, and `_InlineExecutor` runs the worker synchronously inside `submit()`.

**What I added.** A `_pending_executor(sync_calls, on_pending)` seam patched in for `_InlineExecutor`: the first N submissions run synchronously, later ones return a `Future` that never resolves - a pool worker SIGKILLed mid-loop. `on_pending` receives the submit args, which is how a test learns the run-scoped usage directory (the run id is minted inside `run_factory`); `_usage_dir_from_args` also pins the positional contract from the consumer end. Three tests take the disk route for real:

- `test_pending_worker_spend_is_salvaged_from_disk_end_to_end` - the worker publishes a boundary snapshot then is killed; the only surviving record of 700 tokens is the file.
- `test_a_stale_snapshot_from_a_finished_attempt_is_not_recounted` - the P2-c case.
- `test_pending_worker_spend_survives_progress_log_disabled` - the P2-d case.

Verified each depends on the branch it claims: with `_read_partial_usage` stubbed to None the first fails `assert [] == [700]` and the third fails on the rollup assertion; with `_clear_partial_usage` stubbed to a no-op the second fails `assert [700, 700] == [700]`.

I kept `test_stop_mid_run_keeps_the_aborted_component_spend` rather than deleting it - the already-delivered route is a real branch and needs its own test - but rewrote its docstring to state its scope and to record that it passed with snapshot write/read broken. The two routes are mutually exclusive branches of `_salvage_aborted_usage`, so each needs its own coverage.

0xfauzi:
All five findings addressed in 3fa8ee2. Every one was valid; each was reproduced against the previous head before being fixed, and each new regression was re-run with its own fix disabled to prove it fails.

| Finding | Reproduced before | Closed after |
|---|---|---|
| P1-a threshold reset per loop | 3x `max_iterations=1` loops: `calls=3 known_calls=0 total_tokens=0`, all `budget_halt_reason` empty | loop 2 halts `token budget unenforceable ... 2 call(s) in this run` |
| P1-b `known_calls` is not token evidence | 2 cost-only records: `known_calls=2 total_tokens=0`, `halt_reason=None`, loop ran all 10 iterations | halts at 2 iterations on `token_calls == 0` |
| P2-c snapshot not attempt-scoped | engineer usage events `[700, 700]` for 700 tokens of work | `[700]` |
| P2-d snapshot dies with progress logging off | `.kstrl/runs/` never created; `record_engineer_usage` called 0 times | spend salvaged and shown in the rollup, with no `progress.jsonl` and no `events.jsonl` |
| P2-e abort test misses its branch | shipped test passes with `_read_partial_usage` returning None | 3 new tests take the disk route; each fails when its route is disabled |

**Design decisions worth your attention**

1. The unenforceable rule is now two facts over two scopes: token evidence on THIS loop (fc44f69 was right, and its regression test still pins it), call threshold RUN-WIDE via new `LoopBudget.prior_calls` / `prior_token_calls`. The run-wide counter counts TOKENLESS calls only, not all calls - counting all calls would let a healthy reporting call push us toward "this adapter never reports" and would make one unparseable result fatal. Documented consequence, pinned by a test: once a run has one tokenless call anywhere, the engineer's first tokenless iteration halts.
2. I did not add the parent-side evaluation as an alternative. It has better information but only after the engineer loop returns - the window this PR exists to close - so it can supplement, never replace. Feeding the priors down gives the in-loop rule the parent's information at launch and keeps the halt at the earliest point it can fire, without minting a second halt reason for one condition.
3. `token_calls` is serialized (`to_dict` -> `ComponentUsage` -> both `events.jsonl` and the v1 `progress.jsonl`). The parity test caught the drift; I updated its fixture rather than let the two files disagree about the same event. Old readers ignore the key, old payloads decode it as 0.
4. `progress_log_enabled=false` now writes one small accounting JSON per component attempt. That is a deliberate, documented change to the "symmetric opt-out": an observability switch may drop the narration, never the meter.

**What the cap guarantees after this**

- No NEW engineer iteration and no new phase starts once the reported spend reaches the cap.
- A cap that cannot be enforced fails loudly instead of pretending: the run halts when this loop reports no token count and the run has seen two tokenless calls, and that determination no longer resets per attempt or component.
- Every route lands in one audit state: `budget_exceeded` event, one typed `infrastructure_error` finding, one `budget_overrun` inbox item, no burned retry.

**Still unbounded, unchanged by this PR**

- The iteration already in flight. Nothing interrupts a single agent call; `[timeout] agent_iteration` bounds wall clock, not tokens.
- Concurrent workers. Each sees the run total as of its own launch, so with `max_parallel = N` both the token overshoot and the tokenless-call count lag by up to N in-flight iterations.
- Unreported spend. Every figure is a CLI self-report, so totals are lower bounds and the token halt can arrive late.
- The first tokenless loop in a run: with `max_iterations = 1` it completes before the run-wide threshold is reached, and the halt lands on the next loop.

Gates: `uv run pytest tests/ -q` 2214 passed / 28 skipped / 0 failed (baseline 2201, +13 new tests), `uv run mypy kstrl/ --strict` clean, `uv run ruff check kstrl/ tests/ scripts/` clean. Docs regenerated via `scripts/gen_docs.py`. No prompt bodies touched, so no calibration delta applies (H2/H3).

0xfauzi:
Pushed `22e99b4` after reviewing `3fa8ee2`. All five findings are fixed; one of the fixes needed narrowing.

**P1-a was fixed with too broad a scope.** Sourcing the tokenless-call threshold from `pipeline.run_usage` aggregates every role, so calls that say nothing about the engineer's adapter counted as evidence about it. I reproduced a healthy claude-code run at 250k of a 500k cap, four reporting engineer calls behind it, where an architect timeout plus the engineer's first timeout halted the run with:

```
max_total_tokens (500000) can never trip on this adapter
```

False twice over: the adapter had reported four times, and the cap was halfway to tripping.

The threshold asks one question - does the ENGINEER's adapter report tokens? - so only engineer calls can answer it. `prior_calls`/`prior_token_calls` now come from a new `ComponentPipeline.engineer_usage_totals()` folding the `engineer` phase across every component and attempt. P1-a stays fixed (still survives `max_iterations=1` and retries, where a per-loop counter never reached the threshold) without an unrelated timeout condemning a working adapter. `prior_total_tokens` stays run-wide - the overrun check is about what the run has spent.

The message no longer makes a claim about the adapter. It states what is checkable: this loop reported no tokens, prior spend is frozen at N, so the cap cannot advance from here.

Verified after the change: the false-positive case no longer halts; silent-engineer-across-attempts still halts at 2; cost-only records still halt (P1-b); overrun and disabled-cap paths unchanged.

`test_a_tokenless_call_after_an_earlier_one_halts_immediately` documented the run-wide-across-roles trade in its docstring, so I rewrote it for the engineer-scoped rule rather than leaving a stale rationale. Added `TestUnenforceableHaltIsEngineerScoped` covering all four properties, and corrected `docs/env-vars.md`, which said "counted run-wide".

Credit where due on the other four: each reviewer scenario was reproduced before the fix and each new test was proven load-bearing by disabling its own fix - that is why P2-e's dead test showed up as dead.

Full suite 2218 passed, `mypy --strict` and `ruff` clean, `gen_docs --check` current.

**Guarantee now:** no new engineer iteration or phase starts once reported spend reaches the cap, and a cap that cannot be enforced fails loudly without resetting per attempt. **Still unbounded:** the iteration already in flight, concurrent workers (lag scales with `max_parallel`), unreported spend, and the engineer's first tokenless loop in a run when `max_iterations = 1`.

kstrl/loop.py
```diff
@@ -324,6 +536,28 @@ def run_loop(
...
             )
 
+        # R8 run-level token budget: the ONLY in-loop enforcement point
+        # for [factory] max_total_tokens. Checked here - after the
+        # completion return, before the breaker's stall probe - so a
+        # blown budget never pays for another agent call or another
+        # breaker test command. This bounds overshoot to the work
+        # already in flight (about one iteration per running worker); it
+        # cannot interrupt a call mid-flight. See LoopBudget.
+        if budget is not None:
```
0xfauzi, line 554: [P1] Completed tokenless loops never reach this check

The completion return immediately above skips this entire block. I reproduced three sequential one-iteration loops, threading prior_calls/prior_token_calls exactly like the scheduler, where each tokenless agent emitted COMPLETE: every result had completed=True and an empty budget_halt_reason, ending at engineer calls=3, token_calls=0, total_tokens=0. An end-to-end factory reproduction likewise completed two dependent components with priors (0, 0) then (1, 0) and no budget failure.

This is the normal success path for a custom agent_cmd, not only an artificial max-iterations case. The statement that "the halt lands on the next loop" is false when that next loop also completes on its first call.

Evaluate unenforceability before honoring COMPLETE, or perform the equivalent check in process_result immediately after recording engineer usage and before review/scheduling. Preserve the human-readable reason through ComponentResult and add a regression where multiple tokenless components each emit COMPLETE in one iteration.

0xfauzi (commented):
Re-review at 22e99b4: the new token_calls signal, engineer-scoped priors, independent accounting directory, stale-file clearing, and pending-Future tests address the five original findings on their covered paths. The fixes are well documented and the new regressions are materially stronger.

I found two remaining issues, including one P1 that still leaves max_total_tokens unenforceable for a realistic custom-agent success path. A tokenless loop that emits COMPLETE returns before LoopBudget.halt_reason is evaluated. Repeating that across components accumulates tokenless engineer calls without ever halting. The second issue is that best-effort snapshot deletion silently preserves and later salvages a stale attempt when deletion fails.

Verification at this head:
- uv run pytest tests/ -q: 2218 passed, 28 skipped
- uv run pytest tests/test_usage_meter.py tests/test_events.py -q: 107 passed
- uv run mypy kstrl/ --strict: clean
- uv run ruff check kstrl/ tests/ scripts/: clean
- uv run python scripts/gen_docs.py --check: current
- GitHub Actions CI run 308: success
- focused reproductions for both inline findings

Recommendation remains do not merge until the P1 completion-boundary bypass is fixed. The P2 should be fixed in the same pass because failed deletion invalidates the attempt-scoping invariant.

kstrl/factory.py
```diff
@@ -933,6 +940,103 @@ def _preflight_component_branches(
...
+
+    Called by the scheduler in the PARENT immediately before every
+    submission, not by the worker, because the window that matters is
+    exactly the one where the worker never starts (cancelled or killed
+    during pool startup). Best effort: a snapshot we failed to delete is
+    an accounting risk, never a reason to fail a component.
+    """
+    try:
+        path.unlink()
+    except OSError:
```
0xfauzi, line 1049: [P2] A failed delete must invalidate the snapshot

Deletion is the entire attempt-scoping invariant, but every OSError is swallowed. If unlink fails, the previous attempt's file remains; the next worker generally cannot replace it for the same filesystem reason, and a pending abort then salvages the stale totals that process_result already counted. With a 700-token snapshot in a non-writable component directory, _clear_partial_usage leaves the file present and _salvage_aborted_usage records the same 700 again.

Treat FileNotFoundError as the harmless case. For other failures, return an unsafe/invalid status and prevent that future from using disk salvage, or use an attempt-specific filename so stale data is never addressable as the current attempt. A warning alone would expose the problem but would not prevent incorrect accounting. Add a deletion-failure regression that asserts the stale snapshot is not salvaged.

kstrl/loop.py
0xfauzi: Fixed in `27b800c`. Confirmed the mechanism: `run_loop` returns on the completion marker before `LoopBudget.halt_reason` is ever evaluated, so an adapter finishing on its first tokenless call cannot halt itself. And you are right that this is the ordinary success path for a custom `agent_cmd`, not an artificial case - my own PR comment claimed "the halt lands on the next loop", which is false when the next loop also completes.

I closed it in the parent rather than by refusing to honor COMPLETE. Failing a component that genuinely finished would destroy valid work, and the cap exists to stop spending, not to discard what was already bought. New `ComponentPipeline.token_budget_unenforceable()` runs at the scheduling gate and refuses to start further components once the engineer has made `UNENFORCEABLE_CALLS` tokenless calls with no token report. Spend is bounded by the component in flight when the determination lands.

Worth naming: this is your original "evaluate at the engineer/scheduling boundary" suggestion, which the previous round declined as a *replacement* for the in-loop check. That reasoning was right about replacement and wrong to stop there - the two cover different paths. In-loop halts a loop that keeps iterating; the gate stops a run whose loops keep completing. Neither alone is sufficient, which is exactly what your reproduction showed.

Regression `TestCompletionBoundaryBypass` covers the gate firing after two tokenless completions, the single-call case that must NOT fire, a reporting engineer that never trips it, and inertness with no cap. I also added the limitation to `LoopBudget`'s WHAT STAYS UNBOUNDED and to `docs/env-vars.md`, since a bypass that is only fixed elsewhere should be findable from the place that looks like it should have caught it.

kstrl/factory.py
0xfauzi: Fixed in `27b800c`. You are right that a warning would expose the problem without preventing the bad accounting - deletion IS the attempt-scoping invariant, so swallowing the error left the previous attempt addressable as the current one.

`_clear_partial_usage` now returns a status: `FileNotFoundError` is the harmless case (nothing to retire), any other `OSError` reports the slot as unclean. The scheduler then marks that component salvage-unsafe and `_salvage_aborted_usage` refuses the disk route for it. I chose the status-plus-flag over an attempt-specific filename because the filename variant needs the attempt threaded through a submit tuple that a test pins positionally, and it leaves stale files on disk to reason about later; refusing to read an untrustworthy file makes the invariant local to the decision that depends on it.

The trade, stated so it is a choice rather than an accident: losing an aborted attempt's spend understates a total, while salvaging a stale one corrupts it. The guard prefers the understatement.

`TestFailedSnapshotDeletionInvalidates` covers missing file, successful delete, failed delete (patched `PermissionError`, file still present), and the behavioral half - an unsafe attempt does not salvage a planted 700-token snapshot, while a safe one still does, so the guard is not a blanket off switch. Verified load-bearing: removing the `usage_salvage_is_safe` check makes that test fail.

Issue #152, opened 2026-07-21
R8.5: Test-suite adequacy gate - oracle linting, patch coverage, diff-scoped mutation, fixtures


0xfauzi:
> **Stabilisation queue: P3, rank 15 of 21** (set 2026-09-09; the full ordered list and the coordinator's rules are in #369). Adequacy layers 1 and 2, advisory first. On hold until the owner lifts the P3 hold in #369.

## Why

The lights-out precondition per every source tradition is an evaluator-grade test suite. The evidence says this cannot be assumed for agent-written tests: 80.2% of 86k agent-authored test-file patches carry weak or no oracle signals (arXiv:2606.18168), LLM assertions encode actual rather than expected behavior, and test-gaming is measured, not hypothetical (ImpossibleBench: frontier models exploit tests at high rates). Green tests alone cannot be a merge gate at high autonomy.

## Approach: four layers, integrate the tooling

- **Layer 0 - test-diff discipline (mechanical, free)**: fail diffs that delete tests, add skip/xfail, or loosen assertions without spec-linked justification; oracle-signal linter on new/changed test files (W1-W5/S1-S3 taxonomy), requiring at least one strong-oracle assertion per new test file.
- **Layer 1 - patch coverage floor**: diff-cover `--fail-under` on changed lines (~85%), reusing the existing coverage run.
- **Layer 2 - diff-scoped mutation score**: Google-style - mutants only on changed+covered lines, max 1 per line, hard wall-clock cap with recorded sampling. Tools: mutmut (Python, coverage-limited), StrykerJS `--incremental` (JS/TS), cargo-mutants `--in-diff` (Rust). Gate >= 70% killed, advisory first, ratchet up only. Surviving mutants feed back to the engineer as concrete test targets (Meta ACH pattern) for one remediation iteration.
- **Layer 3 - fixtures oracle**: promote the existing approved input/output fixtures from opt-in to required at high autonomy - the only layer whose ground truth the engineer cannot rewrite. Mutation testing does not catch wrong-expectation tests; only spec-derived oracles do. The layers are complementary.

Verification independence: cross-family review becomes the default at L3+ (arXiv:2506.07962: same-family builder/reviewer pairs have measurably correlated blind spots). The calibration suite gains a same-family vs cross-family detection delta.

## Behavior by autonomy level

[...]

PR #178, opened 2026-07-28
R8.5 Layer 0: test-diff discipline and oracle-signal linting (part of #152)

0xfauzi:
Implements **Layer 0** of R8.5 ([#152](https://github.com/0xfauzi/kstrl/issues/152)). This is **not** the whole item - scope is stated explicitly below and the issue stays open.

Green tests cannot be a merge gate on their own. The evidence is specific: 80.2% of 86k agent-authored test patches carry weak or no oracle signals ([arXiv:2606.18168](https://arxiv.org/abs/2606.18168)), LLM assertions tend to encode the behavior code *has* rather than what it *should* have, and test-gaming is measured rather than hypothetical (ImpossibleBench). A suite an agent can edit is a suite an agent can weaken, and "the tests pass" then means only "the tests that survived pass".

## What Layer 0 checks

| Signal | Example |
|---|---|
| Deleted test | a `test_*` def removed and not re-added |
| Added skip/xfail | `@pytest.mark.xfail(reason='flaky')` |
| Net assertion loss | more `assert` lines removed than added, per file |
| Weak oracle | a new test file whose only assertions are `is not None`, `assert True`, `isinstance(...)` |
| No oracle | a test with no assertion at all |

"Falsifiable" is a deliberately low bar: a comparison against an expected value, or an asserted exception (`pytest.raises` counts). It does **not** judge whether an expected value is *correct* - nothing static can, and that is the fixtures oracle's job. It only separates "asserts something that could fail" from "asserts nothing", which is a far more reliable signal.

## Dogfooded, not just unit-tested

Pointed at kstrl's own last three merges, Layer 0 found:

- the one test I genuinely deleted during the budget-cap work (`test_non_interactive_ui_warns_and_proceeds`)
- **assertionless tests in kstrl's own suite**, including `test_inbox_write_failure_does_not_fail_the_run` - which I wrote, and whose comment reads "The run completed; that is the assertion". (Originally five; after the review fixes it is four - `test_crashed_worker_records_nothing` was never assertionless, its oracle is `assert_not_called()`, and the old classifier could not see it.)
- **zero** weak/no-oracle *files*, so the file-level gate stays quiet on real code

That is the shape you want: true positives on real history, no flood.

## Behavior

Opt-in (`[adequacy] enabled = false`) and **advisory first** - findings are recorded without failing, so turning it up later starts from evidence rather than a guess. With the R8.2 ladder on, Layer 0 blocks from L1 up per the roadmap's level table: autonomy may tighten this gate, never loosen it. Findings are typed (`adequacy_*`). BLOCKING findings reach the R8.3 inbox through a dedicated `ItemKind.TEST_ADEQUACY` emitter; advisories stay in the finding stream, since the inbox is for decisions. (The original body claimed the `CheckResult.findings` lift alone delivered this - it did not, and the review caught it.) Fails **closed** on an unreadable diff.

One precision detail worth noting: a *modified* test is not reported as deleted (removed def plus re-added def for the same name). Crying wolf on every edited test would train people to ignore the signal.

## Scope - what is NOT in this PR

- **Layer 1** patch coverage floor (`diff-cover --fail-under`) - adds a dependency and wants a measured floor, not the roadmap's ~85% placeholder.
- **Layer 2** diff-scoped mutation - `check_mutation_score` already exists and is file-scoped, but the R8.5 requirements on top (1 mutant per line, wall-clock cap, sampling in the audit trail, surviving mutants fed back as remediation targets) are unbuilt. Its >= 70% gate is explicitly "thresholds set from the empirical distribution".
- **Layer 3** fixtures required at L3+ - a small level-gate that belongs with the Layer 1/2 pass.
- **Cross-family review at L3+** and the calibration family-delta - user-run measurements (overlaps remediation R7.1).

The sequencing reason, recorded in the roadmap: **Layer 0 needs no data, so it degrades to nothing without any.** Layers 1-2 need an empirical distribution to set a threshold anyone should trust, and this repo currently has ~2 decisive runs. Shipping them against invented numbers is the failure this cycle keeps trying to avoid.

## Tests / gates

47 adequacy tests: oracle classification (9 parametrized assertion forms), `pytest.raises`, BoolOp strongest-operand, syntax-error fail-open, diff discipline including modified-vs-deleted, config and level gating, and four end-to-end cases against real git repos (advisory records, L1 blocks the same diff, a clean change passes with zero findings, unreadable diff fails closed).

`mypy --strict` clean (108 files), `ruff` clean, full suite **2273 passed / 28 skipped / 0 failed**.

---

**Update:** six review findings addressed - whole-file deletions evaded the gate entirely, truthiness (`bool(x)`, `or`-aggregation, bare calls) counted as strong, unittest/mock assertions read as no-oracle, the strong-oracle floor applied to modified rather than only new files, non-decorator skips were invisible, and the inbox integration was claimed but absent. Suite now 2332 passed. The measured false-positive profile is documented in `docs/env-vars.md`.


kstrl/adequacy.py
```diff
@@ -0,0 +1,491 @@
...
+    current = ""
+    prev = ""
+    for line in diff_text.splitlines():
+        if line.startswith("diff --git"):
+            current = ""
+        elif line.startswith("+++ ") and prev.startswith("--- "):
+            target = line[4:].strip()
+            if target.startswith("b/"):
+                target = target[2:]
+            current = "" if target == "/dev/null" else target
```
0xfauzi: [P1] Preserve the old path for deleted files

For a whole-file deletion, Git emits `--- a/tests/test_old.py` followed by `+++ /dev/null`. This assignment clears `current`, so every removed `def test_*` and `assert` line is skipped. A PR can therefore delete an entire test file without producing either `test_deleted` or `assertion_removed`, even though those are core Layer 0 signals. Please retain the `---` path when the target is `/dev/null` and add a regression using an actual deleted-file diff.

kstrl/adequacy.py
```diff
@@ -0,0 +1,491 @@
...
+        strengths = [
+            _classify_assert(ast.Assert(test=value, msg=None))
+            for value in test.values
+        ]
+        return (
+            OracleStrength.STRONG
+            if OracleStrength.STRONG in strengths
+            else OracleStrength.WEAK
+        )
+    return OracleStrength.STRONG
```
0xfauzi: [P1] Do not let weak truthiness expressions count as strong

The strongest-operand rule and the default `STRONG` return allow trivial weak-oracle bypasses. For example, `assert bool(result)` is classified strong even though it is equivalent to the weak `assert result`; and `assert result is not None or result == 3` is classified strong even though the first branch short-circuits for any non-`None` wrong value. Because one strong assertion carries the whole file, either form suppresses the finding. `And` and `Or` need different aggregation semantics, and unknown call/container expressions should not default to strong without evidence of an expected-value check.

kstrl/adequacy.py
```diff
@@ -0,0 +1,491 @@
...
+        if marker:
+            report.skipped.append((node.name, marker))
+
+        asserts = [n for n in ast.walk(node) if isinstance(n, ast.Assert)]
+        strengths = [_classify_assert(a) for a in asserts]
+        if _has_raises_context(node):
+            strengths.append(OracleStrength.STRONG)
+
+        if not strengths:
+            report.without_assertions.append(node.name)
```
0xfauzi, line 494: [P2] Recognize assertion method calls before declaring a test assertionless

This only collects Python `assert` statements plus raises contexts. Common valid oracles such as `self.assertEqual(result, 3)` and `mock.assert_called_once_with(3)` therefore produce `strength == NONE` and land in `without_assertions`. With Layer 0 blocking at L1, touching or adding a unittest/mock-based test file can fail verification despite having an exact-value or interaction oracle. Please classify standard unittest/mock assertion methods, or otherwise avoid labeling calls that are established assertion APIs as having no oracle.

kstrl/verify.py
```diff
@@ -1020,6 +1026,91 @@ def _check_licenses(
...
+            continue
+        full = cwd / rel
+        if not full.exists():
+            continue          # deleted; the diff analysis covers it
+        try:
+            sources[rel] = full.read_text(encoding="utf-8", errors="replace")
+        except OSError:
+            continue
+
+    adequacy_findings = evaluate_layer0(diff_text, sources, config)
```
0xfauzi: [P2] Apply the strong-oracle requirement only to newly added files

`changed` includes modified files as well as added files, and every surviving Python test path is passed to `evaluate_layer0`. That means merely editing a legacy test file whose pre-existing tests use only weak oracles now emits `weak_oracle` and can block at L1. The config and issue contract say the strong-oracle floor is per NEW test file; modified files should still receive diff-discipline checks, but need status-aware filtering before whole-file oracle linting.

kstrl/adequacy.py
```diff
@@ -0,0 +1,491 @@
...
+                if _ASSERT_RE.match(body):
+                    result.removed_assertions[current] = (
+                        result.removed_assertions.get(current, 0) + 1
+                    )
+            elif line.startswith("+") and not line.startswith("+++"):
+                body = line[1:]
+                added_def = _DEF_RE.match(body)
+                if added_def:
+                    result.added_tests.add((current, added_def.group(1)))
+                if _SKIP_RE.search(body):
```
0xfauzi: [P2] Detect non-decorator skip and xfail additions

The regex only recognizes decorator syntax beginning with `@`. Adding `pytest.skip("disabled")` inside a test, module-level `pytestmark = pytest.mark.skip(...)`, or `pytest.param(..., marks=pytest.mark.xfail(...))` disables execution without producing `test_skipped`. Since the AST-collected `report.skipped` is not emitted either, these common forms evade the advertised added-skip/xfail discipline. Please cover the supported pytest skip mechanisms with syntax-aware detection and regressions.

docs/dark-factory-roadmap.md
```diff
@@ -491,7 +491,41 @@ duckdb extra is wired for the query subcommand only.
...
 
-Status: `[ ]` - Depends on: R8.2 for level-gated behavior (lands advisory-first)
+Status: `[~]` - **Layer 0 only.** Shipped in `kstrl/adequacy.py` +
+`check_test_adequacy`: test-diff discipline (deleted tests, added
+skip/xfail, net assertion loss) and oracle-signal linting (each new test
+file needs one falsifiable assertion; tests that assert nothing are
+reported). Opt-in via `[adequacy] enabled`, ADVISORY first; with the R8.2
+ladder on, Layer 0 blocks from L1 up per the level table. Findings are
+typed (`adequacy_*`) and reach the R8.3 inbox through the existing
+`CheckResult.findings` lift.
```
0xfauzi: [P2] Wire adequacy findings into the inbox before claiming this integration

The `CheckResult.findings` lift only appends findings to the component stream. In `ComponentPipeline`, `_inbox_add` is guarded by `category.startswith(POLICY_CATEGORY_PREFIX)`, and `ItemKind` has no adequacy kind, so neither advisory nor blocking `adequacy_*` findings reach the R8.3 inbox. Please add the intended inbox emitter/kind (with deduping and tests), or remove this claim if the audit trail is the only intended destination.

0xfauzi (commented):
I found six correctness and integration gaps in Layer 0. The most important ones let whole-file deletions and weak truthiness assertions evade the gate; the others can either miss disabled tests or block valid/common test styles. Inline comments include concrete reproductions.

kstrl/adequacy.py
0xfauzi: Fixed in c962e68 (`kstrl/adequacy.py:576-596`).

Reproduced first against a real git deletion: a repo with `tests/test_old.py` holding `test_a`/`test_b`, deleted on a branch, `git diff main...HEAD` produces `--- a/tests/test_old.py` / `+++ /dev/null`, and `analyze_test_diff` returned `deleted_tests: []`, `net_assertion_loss: {}`. Exactly as you described - the loudest possible weakening reported nothing.

`analyze_test_diff` now reads BOTH header lines through a shared `_diff_path` helper and keeps the `---` side when the `+++` target is `/dev/null` (and vice versa for an added file, which was already correct because the target is the non-empty side). Same fixture after: `[(tests/test_old.py, test_a), (tests/test_old.py, test_b)]` and `{tests/test_old.py: 2}`.

Two regressions, both on git-produced diffs rather than hand-written strings, since the bug was in reading git's header:
- `TestDiffDiscipline::test_whole_file_deletion_reports_every_test` - builds the repo, deletes the file, asserts `+++ /dev/null` is actually in the diff (so the fixture cannot silently stop testing the case), then asserts both tests and the assertion loss.
- `TestCheckEndToEnd::test_deleting_a_whole_test_file_is_caught` - end to end through `check_test_adequacy` at L1: fails, with both `adequacy_test_deleted` and `adequacy_assertion_removed`.

No false-positive risk traded: a path that was previously dropped is now read.

kstrl/adequacy.py
0xfauzi: Fixed in c962e68 (`kstrl/adequacy.py:137-238`).

All three reproduced before the change - `_classify_assert` returned STRONG for `assert bool(result)`, `assert result is not None or result == 3` and `assert compute()`.

The rule is now stated in the `_is_truthiness_call` docstring and referenced from `OracleStrength` and `_classify_expr`:

> A Call is STRONG only when its arguments contain a comparison, directly or inside a comprehension. Otherwise it is asserted for truthiness and is WEAK.

- `bool(...)` needs no special case: no comparison in the args, so it is weak for the same reason `assert result` is. `bool(result == 3)` stays strong.
- BoolOp now aggregates per operator (`_classify_expr:212-224`): `and` takes the STRONGEST operand (every operand must hold), `or` takes the WEAKEST (the first that holds satisfies it, and short-circuits before any comparison).
- Bare calls default WEAK.
- Bonus from the same rewrite: `assert not ready()` is weak and `assert not (a == b)` is strong. `not` inverts the predicate, not the strength.

Pinned in BOTH directions, as you asked - `TestOracleClassification::test_truthiness_is_weak` (bool(), or-with-shape, or-with-name, bare call, method call, negated call) and `::test_stated_expectations_stay_strong` (`all(x > 0 for x in items)`, `any(v == 3 for v in vs)`, `all(len(r) == 2 for r in rows)`, `'x' in render()`, `sorted(xs) == [1, 2]`, `compute() != 0`, the `and` case, `bool(result == 3)`, `not (a == b)`).

False positive I knowingly traded, stated plainly: a predicate carrying its expectation as a plain argument - `assert s.startswith("prefix")`, `assert re.match(r"^a", s)` - is now WEAK where it used to be STRONG. I decided against a "call has a literal argument" escape hatch because it re-admits `assert flag_set(0)`. The mitigations are that one strong assertion anywhere in the file clears the finding, and after P2-d the floor only applies to files the diff ADDS. Documented at `adequacy.py:150-155` rather than left for someone to discover.

kstrl/adequacy.py
0xfauzi: Fixed in c962e68 (`kstrl/adequacy.py:255-315`, wired at `:451-454`).

Reproduced: `lint_test_source` on a two-line `unittest.TestCase` whose test body is `self.assertEqual(compute(), 3)` returned `strength=none, without_assertions=["test_a"]`, and the same for `m.assert_called_once_with(3)`. At L1 that fails verification on a file with an exact-value oracle - agreed, the worst false positive in the PR.

Two frozensets, split on one question: does the assertion constrain the VALUES, or only that something happened?

- `_STRONG_ASSERT_METHODS` - assertEqual / assertNotEqual / assertAlmostEqual / assertIn / assertIs / assert*Equal / assertCountEqual / assertRegex / assertGreater / assertLess / assertRaises(Regex) / assertWarns / assertLogs, and mock's assert_called_with / assert_called_once_with / assert_any_call / assert_has_calls / assert_not_called plus the await twins.
- `_WEAK_ASSERT_METHODS` - assertTrue / assertFalse / assertIsNone / assertIsNotNone / assertIsInstance / assertNotIsInstance, and assert_called / assert_called_once.

`assert_not_called` sits in STRONG deliberately: it pins zero interactions, so nothing can slip through wrong-valued. `assert_called_once` sits in WEAK because a call with ANY arguments satisfies it, which is the same weakness as `is not None`.

Weak-family methods upgrade to STRONG when their arguments contain a comparison (`self.assertTrue(compute() == 3)`), by the same rule as P1-b - one classifier, not two.

Both families covered: `TestOracleClassification::test_value_pinning_methods_are_strong` (9 cases incl. the upgrade), `::test_presence_only_methods_are_weak` (5), and `::test_unittest_class_file_is_not_assertionless`, which asserts `without_assertions == []` on a real TestCase file.

Confirmed against kstrl's own suite, not just fixtures: re-running the dogfooding over `6597814~3..6597814`, `test_crashed_worker_records_nothing` drops off the assertionless list (5 -> 4). Its only oracle is `pipeline.record_engineer_usage.assert_not_called()`. That was a live false positive in this repo.

kstrl/verify.py
0xfauzi: Fixed in c962e68 (`kstrl/git.py:334-402`, `kstrl/verify.py:1051-1093`, `kstrl/adequacy.py:729-806`).

Reproduced end to end: a repo with a legacy `tests/test_legacy.py` containing `assert build() is not None`, then a branch whose entire diff adds a `# tidy up` comment. `check_test_adequacy(..., autonomy_level=1)` returned `passed=False`, `1 test-adequacy finding(s) [blocking]`, `weak_oracle`. A comment blocked a run.

Status-aware helper, no extra shell-out as you asked: `git.get_diff_name_status` now runs the `--name-status -z -M -C` command, and `get_diff_names` is a thin dedupe over it, so there is one git invocation and one parser. `_parse_name_status_records` returns `(status, path)`; `_parse_name_status_z` is kept as a wrapper so its existing contract and callers are untouched. `strict` carries the identical lenient/strict contract, documented as such - lenient returns `[]` (fail-OPEN), strict raises `GitDiffError`.

`check_test_adequacy` derives `new_paths = {path for status, path if status.startswith("A")}` and passes it to `evaluate_layer0`, which now takes `new_paths` as a REQUIRED keyword. I made it required rather than defaulted in either direction: a default of "all new" reproduces this bug for the next caller, and a default of "none new" silently disables the floor. mypy catches an omission at the call site instead.

The split:
- ADDED file: whole-file oracle lint (strong-oracle floor + every assertionless test + every skipped test).
- MODIFIED file: all diff discipline, plus the assertionless and skip checks restricted to test defs THIS diff added. A module-level `pytestmark` skip counts on a modified file only when the diff added tests under it - those arrive dead.
- A rename/copy destination is `R`/`C`, never `A`, so moved tests are not treated as new.

Tests: `test_modified_file_is_not_held_to_the_oracle_floor`, `test_modified_file_still_gets_diff_discipline`, `test_assertionless_test_added_to_a_modified_file_is_flagged`, `test_pre_existing_skip_in_a_modified_file_is_quiet`, plus both directions end to end - `test_editing_a_weak_legacy_file_does_not_block` and `test_a_new_weak_file_still_blocks`.

Dogfooding delta from this one: over `6597814~3..6597814` the finding count drops 5 -> 3, both losses being assertionless tests in files that merge merely edited (`test_config_control_plane.py`, `test_notify.py`). Those are real weak spots in the suite but they are not that change's doing, which is exactly your point.

kstrl/adequacy.py
0xfauzi: Fixed in c962e68 (`kstrl/adequacy.py:349-441` for detection, `:527-548` for the diff path, `:781-806` for emission).

Reproduced, all three invisible to the old `_SKIP_RE`: `pytest.skip("disabled")` in a body, `pytestmark = pytest.mark.skip(...)`, and `pytest.param(..., marks=pytest.mark.xfail(...))` each produced `added_skips=[]`. And separately, `TestFileReport.skipped` was populated and then never read by `evaluate_layer0` - collected and dropped, as you said.

**Syntax-aware, per your steer.** `_added_skip` dedents the added line and `ast.parse`s it, then walks the tree with `_skip_reference`. That handles all three forms uniformly: `pytest.param(1, marks=...),` parses as a tuple, trailing comma and all, and `pytestmark = ...` is an ordinary assignment. The regex survives only as a fallback for lines that cannot stand alone - a bare decorator, a multi-line fragment - and it is now anchored on a call or an `@`, so `if row.skip:` no longer matches.

**Name matching is per dotted SEGMENT, not substring.** `pytest.mark.skip` and `unittest.skip` match; `runner.skip_count` and `test_skip_behaviour` do not. The old `_skip_marker` used `token in joined`, so it would have matched anything with "skip" in the name.

**Module-level markers.** `_module_skip_marker` reads `pytestmark = ...` (including a list form) and records it under the sentinel `<module>`, because one line there disables every test in the file while each test still looks healthy to the AST.

**Emission.** `evaluate_layer0` now emits `TEST_SKIPPED` from the AST report, with the detail distinguishing "N test(s) disabled in this file" from "module-level marker disables every test in this file". Identical findings from the two detectors (the diff line and the file AST) collapse.

**Deliberate quiet case.** A `pytest.skip()` nested in an `if` is NOT reported by the file lint: a platform guard is not a disabled test, and flagging every one buries the case that matters. Only an unconditional top-level statement in the test body counts. Pinned by `test_guarded_body_skip_is_not_a_finding`.

Regressions per form: `test_added_skip_forms_are_detected` (7 spellings incl. `self.skipTest`, `pytestmark = [pytest.mark.xfail(...)]`, `@unittest.skip`), `test_ordinary_lines_are_not_skips` (4 near-misses), `test_unconditional_body_skip_is_collected`, `test_module_level_pytestmark_is_collected`, `test_a_test_named_after_skipping_is_not_a_skip`, and the emission tests `test_skipped_test_in_a_new_file_is_emitted` / `test_module_level_pytestmark_is_emitted`.

docs/dark-factory-roadmap.md
0xfauzi: Fixed in c962e68 (`kstrl/inbox.py:64`, `:80-84`, `:112`; `kstrl/pipeline.py:1636-1662`; docs at `docs/dark-factory-roadmap.md:494-527` and `docs/env-vars.md:243-253`).

Confirmed false before touching anything: `grep ItemKind\. kstrl/` showed the only `_inbox_add` call in the verification path was gated on `POLICY_CATEGORY_PREFIX`, and `ItemKind` had six members, none for adequacy. The claim appeared in the PR body, the roadmap and env-vars.md; none of it was wired.

Wired rather than deleted, as you asked:

- `ItemKind.TEST_ADEQUACY = "test_adequacy"`, `action_required=True` (a blocked change is waiting on a decision), `DEFAULT_PRIORITY` NORMAL - lower than a policy exception, which is a breach rather than a quality bar.
- Emitted in `ComponentPipeline._run_verify_phase`, in the same loop as the policy emitter and under the same severity condition. Dedupe key `adequacy:{comp}:{category}:{location}`, so the same file failing the same way across retries collapses onto one item instead of fanning out, while two different files stay distinct.
- Advisory findings deliberately do NOT create items; the comment at `pipeline.py:1636-1643` says why in as many words: "an ADVISORY one is recorded in the finding stream and stops there, because the inbox is a queue of decisions, not of notes."

Tests, exactly the pair you specified: `TestEmittersFireDuringRuns::test_blocking_adequacy_finding_emits_one_item` asserts `len(items) == 1` for a `severity="high"` adequacy finding and checks the component and the evidence `location` survive; `::test_advisory_adequacy_finding_emits_nothing` asserts no `test_adequacy` item for the advisory. The `test_action_required_taxonomy` parametrization gained its row so the kind cannot be added without a stance on notification.

Docs now state what is true and nothing more: findings reach the finding stream either way, blocking ones additionally raise an inbox item with the dedupe key named, advisories do not and why. The R8.3 section's emitter list and item-kind list both include it, since that is where someone would look to check.

0xfauzi:
All six findings addressed in c962e68, on this branch. Each was reproduced before the fix and re-checked after; details are in the inline replies.

| # | What was wrong | Fix | Reproduced -> after |
|---|---|---|---|
| P1-a | `+++ /dev/null` cleared the current path, so a deleted test FILE reported nothing | keep the `---` side (`adequacy.py:576-604`) | `deleted_tests: []` -> both tests + `{tests/test_old.py: 2}` |
| P1-b | `bool(x)`, `or`-aggregation, bare calls all read STRONG | comparison-in-arguments rule + per-operator BoolOp (`:137-238`) | 3 x STRONG -> 3 x WEAK, with the property oracles still STRONG |
| P2-c | `unittest`/`mock` files read as "asserts nothing" | STRONG/WEAK method sets (`:240-315`) | `strength=none` -> `strong`, `without_assertions=[]` |
| P2-d | oracle floor hit MODIFIED files | `git.get_diff_name_status` + required `new_paths` (`git.py:313`, `verify.py:1051`) | a comment-only diff blocked at L1 -> passes |
| P2-e | only decorator skips seen; collected skips never emitted | AST-first detection, segment-exact names, emission (`:333-441`, `:781-806`) | 3 forms invisible -> all detected, near-misses still quiet |
| P2-f | the inbox claim was false | `ItemKind.TEST_ADEQUACY` + emitter + docs (`inbox.py:64`, `pipeline.py:1634`) | no kind, no emitter -> 1 item for blocking, 0 for advisory |

**Dogfooding re-run, since the PR body cites it.** Same range (`git diff 6597814~3 6597814`), and the numbers MOVED - both moves are the fixes working:

- PR body claimed: 1 deleted test found, **5** assertionless tests, 0 weak/no-oracle files.
- Now: 1 deleted test (unchanged), **3** findings from `evaluate_layer0` instead of 5, and the file-level lint finds **4** assertionless tests instead of 5. Still 0 weak/no-oracle files.
- 5 -> 4 assertionless is P2-c: `test_crashed_worker_records_nothing` was never assertionless. Its oracle is `pipeline.record_engineer_usage.assert_not_called()`. That was a live false positive in this repo, not a fixture.
- 5 -> 3 findings is P2-d: two of the five sat in files that change merely EDITED (`test_config_control_plane.py`, `test_notify.py`). Genuinely weak tests, but not that change's doing. The two that remain - `test_inbox_write_failure_does_not_fail_the_run` (added file) and `test_unwritable_snapshot_path_is_swallowed` (def added by the diff) - are true positives, and I checked both sources by hand.

The PR body's "five assertionless tests" line is now wrong by one and should read four; the "reach the R8.3 inbox through the existing CheckResult.findings lift" line was wrong and is now true by construction rather than by assertion.

**False positives knowingly traded**, so they are on the record rather than discovered later:

1. `assert s.startswith("prefix")` and `assert re.match(r"^a", s)` are now WEAK. They state an expectation in a plain argument, but admitting that re-admits `assert flag_set(0)`. One comparison anywhere in the file clears the finding and the floor only applies to ADDED files, so the exposure is a new test file that never compares anything.
2. `any(is_ok(x) for x in xs)` is WEAK for the same reason; `any(x == 3 for x in xs)` is STRONG.
3. A `pytest.skip()` guarded by an `if` is NOT reported by the file lint (platform guard). An added guard still shows in the diff-level check, consistent with how `skipif` decorators were already treated.

**Gates**: `uv run pytest tests/ -q` 2332 passed / 28 skipped / 0 failed (baseline 2273 - the delta is new tests, no test was weakened or removed; the five `evaluate_layer0` call sites changed only to pass the new required `new_paths`, noted in the class docstring). `mypy kstrl/ --strict`, `ruff check kstrl/ tests/ scripts/`, `gen_docs.py --check` all clean. Per H1 I did not run `/code-review` on this.

PR #179, opened 2026-07-28
fix: one agent-type vocabulary; reject unknown types instead of guessing

0xfauzi:
Two layers disagreed about what `[agent] type` means, and the disagreement silently changed which model wrote your code.

`kstrl/cli.py` kept a private `_AGENT_TYPE_ALIASES` table canonicalizing `"claude"` to `"claude-code"`. `get_agent` had no such table, and its final unguarded `return CodexAgent(...)` swallowed anything unrecognized:

```
get_agent(agent_type="claude")     -> CodexAgent
get_agent(agent_type="typo-here")  -> CodexAgent
get_agent(agent_type="custom")     -> CodexAgent   (no command set)
```

Any caller not routed through the CLI preflight - a test, an embedder, a future entry point - got a different model than the documented spelling names, with nothing said. Reproduced before fixing.

**A correction I owe the record:** earlier in this cycle I described `type = "claude"` as "not a valid type" and treated it as a broken config. It is a **supported alias**. The defect was never the spelling; it was that only one of two layers honored it. I repeated that mischaracterization several times without checking it against the alias table.

## The fix

- `AGENT_TYPE_ALIASES` and `canonical_agent_type` now live beside the adapters they select, and the CLI imports the same object. A test pins `_AGENT_TYPE_ALIASES is AGENT_TYPE_ALIASES`, so the two tables cannot drift apart again.
- `get_agent` canonicalizes, then **raises** `UnknownAgentTypeError` on a genuine typo rather than falling back. A fallback that silently swaps the model is worse than a loud failure - you cannot debug a wrong answer you were never told about.
- `type = "custom"` with no command raises instead of resolving to codex. The CLI already said this; the library did not.
- Case and surrounding whitespace are tolerated.
- `VALID_AGENT_TYPES` excludes the internal `""` alias, so messages do not read "expected one of , auto, ...".

An explicit `agent_cmd` still short-circuits before type validation - pre-existing contract, still covered.

## Why now

This came out of preparing a real factory run. The previous attempt died on environment issues before it could exercise anything, and the surviving question was what else in the startup path fails quietly. This is one answer.

## Tests

27 tests: every documented spelling resolves to its adapter, the `"claude"` regression in one line, case and whitespace handling, unknown types raising with the valid list in the message, custom with and without a command, and a parametrized sweep asserting no alias can reach the fallthrough by accident.

Full suite **2359 passed / 28 skipped / 0 failed**, `mypy --strict` and `ruff` clean.

### DIFF
--- a/kstrl/agents/__init__.py
+++ b/kstrl/agents/__init__.py
@@ -8,6 +8,10 @@ from kstrl.agents.custom import CustomAgent
 from kstrl.sandbox import SandboxConfig
 
 __all__ = [
+    "AGENT_TYPE_ALIASES",
+    "UnknownAgentTypeError",
+    "VALID_AGENT_TYPES",
+    "canonical_agent_type",
     "Agent",
     "ClaudeCodeAgent",
     "ClaudeSdkAgent",
@@ -15,6 +19,46 @@ __all__ = [
     "CustomAgent",
     "get_agent",
 ]
+
+
+class UnknownAgentTypeError(ValueError):
+    """``[agent] type`` names an adapter that does not exist.
+
+    Raised rather than tolerated because the old behavior was the worst
+    possible one: an unrecognized type fell through every branch to the
+    codex fallback, so `type = "claude"` silently ran Codex while the
+    operator believed they were running Claude. A typo changed which
+    model wrote the code and nothing said so.
+    """
+
+
+#: The one agent-type vocabulary, shared with the CLI preflight. Two
+#: layers previously disagreed: the CLI accepted "claude" as an alias for
+#: "claude-code", while get_agent did not and fell through to the codex
+#: fallback, so a programmatic caller passing the DOCUMENTED spelling got
+#: a different model than the CLI would have chosen. One table, one
+#: vocabulary.
+AGENT_TYPE_ALIASES: dict[str, str] = {
+    "": "auto",
+    "auto": "auto",
+    "claude": "claude-code",
+    "claude-code": "claude-code",
+    "claude-sdk": "claude-sdk",
+    "codex": "codex",
+    "custom": "custom",
+}
+
+#: Accepted spellings for ``[agent] type``.
+VALID_AGENT_TYPES: tuple[str, ...] = tuple(
+    sorted(name for name in AGENT_TYPE_ALIASES if name)
+)
+
+
+def canonical_agent_type(agent_type: str | None) -> str | None:
+    """Canonical spelling for a configured type, or None if unknown."""
+    if agent_type is None:
+        return None
+    return AGENT_TYPE_ALIASES.get(agent_type.strip().lower())
 
 
 def get_agent(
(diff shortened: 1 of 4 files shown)


kstrl/agents/__init__.py
```diff
@@ -50,6 +94,25 @@ def get_agent(
...
+                "your code and nothing said so."
+            )
+        if canonical == "custom":
+            raise UnknownAgentTypeError(
+                'agent type "custom" is configured but no agent command is '
+                "set; set [agent] command in kstrl.toml, AGENT_CMD, or "
+                "--agent-cmd. Without one there is nothing to run, and this "
+                "used to fall through to the codex adapter instead of saying so."
+            )
+        agent_type = canonical
```
0xfauzi, line 115: [P1] Keep the factory family resolver in sync with this canonicalization

`kstrl.factory._cli_family()` still compares the raw type against only `("claude-code", "claude-sdk")` and otherwise treats it as Codex. After this line, `get_agent(agent_type="claude")` constructs `ClaudeCodeAgent`, but `_cli_family(None, "claude", ...)` returns `"codex"`. Consequently, a programmatic `run_factory` call (or any path that has not gone through CLI preflight) records a Claude engineer as Codex and, when both CLIs are installed, selects `claude-code` as the supposed cross-family reviewer. That is actually same-family review and defeats the R7.1 safety invariant. The same mismatch exists for the newly supported case/whitespace variants. Please have `_cli_family` use `canonical_agent_type` (and reject unknown values consistently), with a regression test covering `engineer_type="claude"`.

0xfauzi (commented):
The adapter factory change is well tested in isolation, but one dispatch mirror still implements the old semantics. That mismatch breaks the factory's cross-family review invariant for the newly accepted aliases, so this should be fixed before merge. I verified the focused tests, reviewer-rotation tests, ruff, and strict mypy locally; the PR's CI run is also green.

kstrl/agents/__init__.py
0xfauzi: Fixed. You found a third implementation of the same vocabulary, and it was the one that mattered most.

Reproduced with both CLIs installed:

```
get_agent(agent_type="claude")   -> ClaudeCodeAgent
_cli_family(None, "claude", ...) -> "codex"
```

So a Claude engineer was recorded as codex, and the R7.1 rotation - whose whole job is to pick the OTHER family - selected claude-code. Same-family review, with the audit trail asserting the opposite. Correlated blind spots are the thing R7.1 exists to prevent, so a silent mismatch here is worse than a crash: the safety property fails while the record says it held.

**I introduced this.** Before #179 the two functions agreed by both being wrong about "claude", so the rotation still landed on a genuinely different family. Fixing one side and not its mirror converted a mislabeling into a broken invariant. That is a sharper lesson than the original bug: consolidating a vocabulary means finding every implementation of it, and I found two of three.

`_cli_family` now canonicalizes through the shared table and raises on an unknown type, matching `get_agent` - a config that cannot construct an agent should not first be assigned a family.

**A second bug surfaced while fixing it**, worth flagging since it was mine and not in your report: `canonical_agent_type` returns `None` for both "unset" and "unrecognized". My first version routed `agent_type=None` into the unknown branch, so the ordinary no-type-configured case raised. Unset is now split out before that check and pinned by a test covering both `claude_available` values.

Regressions in `TestFamilyResolverMatchesTheAdapter`: family agrees with the constructed adapter across every spelling including case and whitespace variants, the "claude" case in one line, unset auto-detects both ways, unknown raises, a custom command stays an unknown family, and the end of the chain - a "claude" engineer no longer resolves to the family it actually used. `tests/test_reviewer_rotation.py` passes unchanged.

Full suite 2372 passed, `mypy --strict` and `ruff` clean.

PR #180, opened 2026-07-28
feat: add [factory] max_cost_usd, a cost-denominated run ceiling

0xfauzi:
## Why - measured, not hypothetical

A real factory run halted on `[factory] max_total_tokens = 500000` at
**1,864,081 tokens**. The breakdown from the run's own journal:

| field | value |
|---|---|
| `input_tokens` | 52 |
| `output_tokens` | 20,855 |
| `cache_read_tokens` | **1,781,669 (95.6%)** |
| `cache_creation_tokens` | 61,505 |
| `total_tokens` | 1,864,081 |
| `cost_usd` | **1.216512** |

`UsageTotals.total_tokens` counts cache reads at par with input tokens, and
cache reads cost roughly an order of magnitude less. So an operator who set a
500k "budget" expecting a spend ceiling got halted at **$1.22**. The cap
measures something real but nearly uncorrelated with cost. That is the defect
this PR fixes.

Second measured fact, stated because it bounds what this PR can promise: that
was **ONE engineer call of 376s**. The in-loop check runs BETWEEN iterations,
so a single iteration overshot the entire cap by **3.7x**. A cost ceiling has
the identical structural limit and this PR does not claim otherwise anywhere.

## What this adds

`[factory] max_cost_usd` (float, default `0.0` = unbounded, matching the
`max_total_tokens` convention) as a first-class ceiling **alongside** the token
cap. Both may be set; whichever trips first halts. Reachable through all three
surfaces: `kstrl.toml`, `KSTRL_FACTORY_MAX_COST_USD`, `--max-cost-usd`.

The existing machinery was extended, not duplicated:

- **`LoopBudget`** (`kstrl/loop.py`) gains `max_cost_usd` / `prior_cost_usd` /
  `prior_cost_calls`, and `halt_reason` evaluates both ceilings.
- **`UsageTotals`** (`kstrl/agents/base.py`) gains `cost_calls`, the mirror of
  `token_calls`, plus a `costless_calls` property. It follows the pattern
  `token_calls` established: serialized in `to_dict`, carried by
  `ComponentUsage` and the v1 compat mapping, decoded backward-compatibly
  (absent key -> 0, so old and new readers interoperate both ways).
- **`ComponentPipeline`** (`kstrl/pipeline.py`) gains `cost_budget_exceeded`,
  `breached_ceiling`, `budget_exceeded`, `cost_budget_unenforceable` and
  `budget_unenforceable`. `fail_for_budget` names WHICH ceiling tripped in its
  message, its `infrastructure_error` finding, the `budget_exceeded` event
  (new `ceiling` field) and the `budget_overrun` inbox item.
- **`_submit_args`** (`kstrl/factory.py`) snapshots the cost priors per launch,
  the same channel the token priors already use.

## The semantics that needed care: per-ceiling unenforceability

The UNENFORCEABLE rule was per-token (`token_calls == 0` after N tokenless
engineer calls). With two ceilings it is now **per ceiling**, and this resolves
a real inconsistency:

| adapter behavior | token ceiling | cost ceiling |
|---|---|---|
| codex: token total, no cost | enforceable | unenforceable |
| claude with a missing `usage` dict: `total_cost_usd`, no tokens | unenforceable | enforceable |
| custom `agent_cmd`: reports nothing | unenforceable | unenforceable |

**Only the last row halts.** The cost-only case is not hypothetical - review
finding P1-b on PR #176 established that `known_calls` increments on cost
alone. Under the old rule that adapter was condemned as unenforceable even
though it could have enforced a spend ceiling perfectly well. If any configured
ceiling can still fire, the run keeps going.

The engineer-scoping fix is preserved: the tokenless/costless counts come from
`ComponentPipeline.engineer_usage_totals()`, not run-wide totals, because
another role's timeout is not evidence about the engineer's adapter.

## What each ceiling actually guarantees

**Guaranteed:** the engineer loop starts no NEW iteration, and the parent
starts no new phase and no new component, once the run's *reported* spend has
reached a configured ceiling.

**Unbounded, identically for both ceilings** - `max_cost_usd` is **not** a hard
cap and is not documented as one:

| gap | why |
|---|---|
| the iteration already in flight | nothing interrupts a single agent call mid-flight; `[timeout] agent_iteration` bounds wall clock, never tokens and never dollars. Measured above: 3.7x the cap inside one call |
| concurrent workers | each sees the run total as of ITS launch, so overshoot scales with `max_parallel` |
| unreported spend | every figure is a CLI self-report; totals are lower bounds whenever `unreported_calls > 0` |

`[agent] budget_usd` remains claude-sdk-only and adapter-internal (it bounds a
single turn). The docs now say explicitly how it differs from this run-level
ceiling, and `max_total_tokens` is documented as a poor proxy for cost with the
measured example above.

## Tests

36 new tests. Both directions of the per-ceiling rule, both overrun paths,
whichever-trips-first, `max_cost_usd = 0` inert, the halt message and emitted
finding naming the ceiling, exactly one `budget_overrun` with no duplicate
`halted_run` (that suppression is exercised for the cost path too so it cannot
regress), and the priors threaded exactly as `_submit_args` threads them.

**Mutation-checked, not just green.** Two deliberate mutations were run against
the new suite to confirm it has teeth:

1. halting as soon as ANY ceiling is dead (the wrong rule) -> 4 failures,
   including both "live ceiling keeps the run going" tests.
2. cost overrun never firing (pre-change behavior) -> 4 failures.

Existing budget tests pass unmodified. Two test fixtures changed for cause, both
documented in their docstrings/comments: the v1 golden-parity payload in
`test_events.py` gains `cost_calls` (exactly as it gained `token_calls` in R8 -
both sides of the parity must carry it), and `_budget_for` in
`test_usage_meter.py` gains an optional `cost_cap`. The TUI snapshot was
regenerated for one label change: `5% of cap` -> `5% of token cap`, now that
there are two caps to distinguish.

## Gates

| gate | result |
|---|---|
| `uv run pytest tests/ -q` | **2408 passed, 28 skipped, 0 failed** (baseline 2372) |
| `uv run mypy kstrl/ --strict` | clean, 108 source files |
| `uv run ruff check kstrl/ tests/ scripts/` | clean |
| `uv run python scripts/gen_docs.py --check` | current |


kstrl/factory.py
```diff
@@ -162,10 +162,30 @@ class FactoryConfig:
...
+    # cost_usd, which - exactly like the token total - is a CLI
+    # self-report and a LOWER BOUND whenever some calls report no cost.
+    #
+    # NOT a hard cap, and it must never be described as one: it has the
+    # same between-iterations enforcement granularity as the token
+    # ceiling. Measured: one engineer call of 376s overshot the entire
+    # 500k token cap by 3.7x, and nothing in this design would have
+    # interrupted it. Distinct from [agent] budget_usd, which is
+    # adapter-internal (claude-sdk only) and bounds a single turn.
+    max_cost_usd: float = 0.0
```
0xfauzi, line 244: [P1] Reject non-finite and negative cost ceilings at the configuration boundary

Every input path currently converts this value with plain `float`, so TOML `max_cost_usd = nan`, `KSTRL_FACTORY_MAX_COST_USD=nan`, and `--max-cost-usd nan` are accepted. I reproduced `FactoryConfig.load(...).max_cost_usd == nan`; `LoopBudget.cost_enabled` is then false and `halt_reason()` returns `None` even with $1,000,000 of recorded spend. The same value also crashes `render_cost_meter()` with `ValueError: cannot convert float NaN to integer`. `inf` produces an enabled ceiling that can never be reached, and negative values silently disable it. Because this is a safety limit, please validate the resolved value as finite and non-negative with a clear configuration error, including at the `run_factory` boundary so programmatically constructed `FactoryConfig` objects cannot bypass the check.

kstrl/pipeline.py
```diff
@@ -798,49 +889,80 @@ def fail(
...
         self._add_findings(comp, [Finding.infrastructure_error(
             phase=phase, explanation=error,
         )])
         self.bus.emit(ev.BudgetExceeded(
             component=comp.id,
             total_tokens=self.run_usage.total_tokens,
             max_total_tokens=self.factory_config.max_total_tokens,
+            cost_usd=round(self.run_usage.cost_usd, 6),
+            max_cost_usd=self.factory_config.max_cost_usd,
+            ceiling=ceiling or "",
```
0xfauzi: [P2] Preserve the ceiling identity for unenforceable cost budgets

For a cost-only run on a costless adapter (the documented Codex case), `budget_unenforceable()` supplies a reason but `breached_ceiling()` returns `None` because no numeric threshold was crossed. This writes `ceiling=""`. I reproduced the resulting audit trail: the inbox title is `comp-a exceeded the budget budget`, `LinearSink` reports `run token budget exceeded (200/0)`, and the reducer reconstructs `token budget exceeded: 200 >= 0`, even though `max_total_tokens` is disabled and the actual reason is `max_cost_usd` being unenforceable. That directly violates this PR's promise that every audit surface names the correct knob. Please carry typed ceiling/condition metadata from the loop or scheduling decision into `fail_for_budget` rather than deriving it only from numeric breach state; the both-dead case may need to represent multiple ceilings explicitly.

0xfauzi (commented):
The end-to-end budget plumbing and per-ceiling coverage model are strong, and all 2,408 tests pass locally. I found two correctness gaps that should be fixed before merge: accepted non-finite configuration can disable the new safety ceiling, and cost-only unenforceable halts lose their ceiling identity and are reported as token overruns downstream.

kstrl/factory.py
0xfauzi: Reproduced, then fixed in 5a66ad7.

Confirmed all three shapes were accepted and each failed differently:

- `max_cost_usd=nan` -> `enabled=False`, no halt at $1M spend (`nan > 0` is False, so it disabled itself while still reading as configured)
- `max_cost_usd=inf` -> `enabled=True`, no halt (enabled and unreachable)
- `max_cost_usd=-5.0` -> `enabled=False`, no halt

Added `BudgetConfigError` + `_validate_cost_ceiling()` at four boundaries: `from_env`, `load` (toml), the env-override path, and `run_factory`. The last one matters because `FactoryConfig` is constructible directly by tests, embedders and the SDK path, so a limit enforced only on the config-file path is not a limit. `0` still means unbounded.

kstrl/pipeline.py
0xfauzi: Confirmed and fixed in 5a66ad7.

Your scenario reproduced exactly: cost-only ceiling + a costless adapter gave `breached_ceiling() -> None`, so `ceiling=""` reached the audit surfaces and they fell back to naming the token ceiling - one that was not even configured on that run.

Root cause is that "breached" and "unenforceable" are different facts. Nothing crossed a threshold, but the ceiling that failed still has a name. Added `ComponentPipeline.unenforceable_ceilings()` (returns only ceilings that are both *configured* and *dead*) and a `ceiling` parameter on `fail_for_budget`, threaded from the in-loop halt and the scheduling gate.

After:
```
breached_ceiling()       -> None
unenforceable_ceilings() -> ["max_cost_usd"]
identity carried to audit surfaces: "max_cost_usd"
```
Both-dead names both; a live ceiling is never named; a disabled ceiling is never named.

0xfauzi:
Both findings fixed in 5a66ad7.

**P1 - unbounding `max_cost_usd` accepted.** `nan`, `inf` and negatives all passed a bare `float()`, then failed in three different silent directions - two disabled the ceiling while still reading as configured, one enabled a ceiling nothing could reach. Validation now runs at `from_env`, toml `load`, the env-override path, and the `run_factory` boundary. That last one is the point: `FactoryConfig` is constructible directly, so validating only the config-file path leaves the limit bypassable.

**P2 - unenforceable halts lost ceiling identity.** "Breached" and "unenforceable" are different facts. `breached_ceiling()` correctly returns `None` when nothing crossed a threshold, but the ceiling that *failed* still has a name, and `ceiling=""` made every audit surface blame the token ceiling - including on runs where it was disabled. Added `unenforceable_ceilings()` (configured AND dead) plus a `ceiling` parameter on `fail_for_budget`.

**Tested:** 2423 passed, 28 skipped. `ruff` clean, `mypy --strict` clean across 108 files.

New tests are mutation-checked rather than assumed load-bearing:

| mutation | result |
|---|---|
| disable `_validate_cost_ceiling` | 8 fail |
| `unenforceable_ceilings()` returns a hardcoded list | 4 fail |

One honest note: my first mutation attempt - removing the `max_total_tokens > 0` guard inside `unenforceable_ceilings()` - failed to break anything. `token_budget_unenforceable()` already guards `cap <= 0` internally, so that outer check is redundant belt-and-braces, not load-bearing logic. Left in place for locality, but it is not what the tests are pinning.

kstrl/pipeline.py
0xfauzi: [P1] Follow-up: the new identity selection still produces false audit data when a numeric breach and a different dead ceiling coexist. Reproduction: with `max_total_tokens=500`, `max_cost_usd=100`, and two token-only calls totaling 600 tokens, `LoopBudget.halt_reason()` correctly returns a token breach; in the parent, `breached_ceiling()` is `max_total_tokens` while `unenforceable_ceilings()` is `["max_cost_usd"]`. The current `", ".join(unenforceable_ceilings()) or breached_ceiling()` chooses `max_cost_usd`, so `fail_for_budget` emits `cost budget exceeded: $0 >= $100` and records the cost ceiling even though the token ceiling actually tripped. Numeric breach identity must take precedence; only consult dead ceilings when no ceiling was breached.

There is a second representation issue in the both-dead case: `ceiling="max_total_tokens, max_cost_usd"` is not understood by `LinearSink` or the reducer, which only special-case exact `"max_cost_usd"`; both currently render the combined event as `token budget exceeded: 0 >= 500`. Please carry the halt condition and one-or-many ceiling identities structurally, or update every consumer to handle the multi-ceiling value explicitly, with end-to-end tests covering these two cases.

kstrl/factory.py
0xfauzi: Follow-up: the value is now rejected correctly at the library boundary, but the CLI does not surface the promised clear configuration error. I reproduced `ks factory` through `CliRunner` with a valid manifest and TOML `max_cost_usd = nan`: it exits 1 with `output == ""` and an uncaught `BudgetConfigError` exception. The real entry point will therefore emit a traceback rather than a normal `ui.err(...)` message. `FactoryConfig.load()` is called outside a catch in `factory`, `run`, retry/config-report paths, and `run_factory()` can also raise after a `--max-cost-usd` override. Please catch `BudgetConfigError` at the user-facing command boundary and render the message with a deliberate exit code. The safety bypass itself is fixed; this remaining issue is the CLI error contract.

kstrl/pipeline.py
0xfauzi: Both parts reproduced, then fixed in 1c7f2c7.

**Precedence.** Confirmed with your exact numbers - `max_total_tokens=500`, `max_cost_usd=100`, two token-only calls totalling 600:

```
breached_ceiling()       -> "max_total_tokens"
unenforceable_ceilings() -> ["max_cost_usd"]
current selection        -> "max_cost_usd"     <-- wrong
reducer                  -> "cost budget exceeded: $0.000000 >= $100.0"
```

That sentence is not just mislabelled, it is arithmetically impossible - $0 is not >= $100 - for a run that blew its token cap at 600/500.

The rule now lives once, in `ComponentPipeline.budget_halt_identity()`: a breach is a threshold actually crossed with numbers behind it, so dead ceilings are consulted only when nothing was breached. Both call sites stopped spelling it out, which is what let them diverge.

**Representation.** Fixed structurally rather than by teaching consumers to parse a joined string. `condition` (`"breached"` / `"unenforceable"`) and `ceilings` (one or many) are carried end to end: `LoopBudget.halt_verdict()` returns a `BudgetHalt`, `ComponentResult` carries the pair across the pool boundary (pickle-checked), and `BudgetExceeded` gained both fields.

Only `"breached"` licenses a comparison sentence, so the unenforceable halt no longer claims one. Both surfaces classify through a single shared `events.budget_halt_kind()` rather than each re-testing `== "max_cost_usd"` locally - a third private copy of that rule is exactly how a mislabel became a divergence in #179.

After:
```
reducer -> budget ceiling unenforceable (max_total_tokens, max_cost_usd): no configured ceiling can still fire
linear  -> ks run `r`: run budget ceiling unenforceable (max_total_tokens, max_cost_usd) - no configured ceiling can still fire; ...
```

`ceiling` stays as the joined legacy string, since events.jsonl is append-only and payloads written before this change must keep their old reading - covered by a test.

End-to-end tests added for both cases, plus the disabled-ceiling case and legacy decode. Mutation-checked: restoring the old precedence fails 1, dropping the reducer's unenforceable branch fails 2.

kstrl/factory.py
0xfauzi: Confirmed and fixed in 1c7f2c7.

Reproduced your case exactly - valid manifest, TOML `max_cost_usd = nan`:

```
exit=1 exc=BudgetConfigError output=""
```

You are right about the cause: `BudgetConfigError` subclasses `ValueError`, and `config show` happens to sit behind a `except ValueError` that renders it - which is why that one path looked fine while `factory` did not. Per-command catching means every call site has to remember, and the set is larger than it looks (`factory`, `run`, `retry`, config report, `launch`, plus `run_factory` itself after a `--max-cost-usd` override).

So it is caught centrally in the click group instead - no entry point can leak one, including entry points added later. Exit 1 with an `error:` line, matching what `config show` already did for the same defect rather than inventing a second contract.

Verified across `factory --manifest` (nan and negative), `config show`, `status`, and the flag-override path:
```
ks factory --manifest m.json --max-cost-usd inf
exit: 1 | exc: SystemExit
error: max_cost_usd must be a finite number, got inf; use 0 to disable the ceiling...
no work started: True
```
The last assertion matters - the override is rejected before anything launches. Mutation-checked: removing the central catch fails 3 tests.

0xfauzi:
Both follow-ups fixed in 1c7f2c7. Reproduced each first.

**Precedence (P1).** Your repro confirmed, and the consequence is worse than a mislabel: the reducer emitted `cost budget exceeded: $0.000000 >= $100.0` for a run that blew its *token* cap at 600/500 - a sentence that is arithmetically impossible. The rule now lives once in `budget_halt_identity()`; both call sites stopped spelling it out, which is what let them diverge in the first place.

**Representation (P2).** Fixed structurally, not by teaching consumers to parse a joined string. `condition` + `ceilings` travel end to end - `LoopBudget.halt_verdict()` -> `BudgetHalt`, `ComponentResult` across the pool boundary, `BudgetExceeded` on the wire. Only `"breached"` licenses a `N >= cap` sentence, so the unenforceable halt stopped claiming one. Both surfaces classify through one shared `events.budget_halt_kind()`; a third private copy of that rule is precisely how #179 turned a mislabel into a divergence. Legacy `ceiling` retained - events.jsonl is append-only - with a test pinning the old reading.

**CLI contract (P3).** Caught centrally in the click group rather than per command. The call set is larger than it looks (`factory`, `run`, `retry`, config report, `launch`, and `run_factory` itself after a `--max-cost-usd` override), and per-site catching means each one has to remember. Central means future entry points cannot leak one either. Exit 1 + `error:`, matching what `config show` already did rather than inventing a second contract.

**Tested:** 2446 passed, 28 skipped. `ruff` and `mypy --strict` clean across 108 files.

Mutation-checked rather than assumed load-bearing:

| mutation | result |
|---|---|
| restore dead-ceiling-first precedence | 1 fails |
| drop the reducer unenforceable branch | 2 fail |
| remove the central CLI catch | 3 fail |

One thing worth flagging that I hit while reproducing: `ks factory --spec` validates the ceiling only *after* decomposition, so an invalid value there is caught having already spent an LLM call. The `--manifest` path and the `--max-cost-usd` override both now fail before any work starts, and that is what the new test asserts. Moving the spec path's check earlier is a separate change - happy to do it here or file it, your call.

0xfauzi:
CI caught one thing my local run could not, worth stating plainly: the three new CLI tests were **environment-dependent**. `ks factory` aborts on agent detection before it reads the ceiling, so they passed on my machine (`claude` on PATH) and failed on the runner (no agent installed) - measuring the environment, not the fix. Same class of mistake as the one flagged in #179.

Fixed in d562a6b: both tests now pass `--agent-cmd` and force `PATH=""`, which is the harsher of the two environments and a fixed one regardless of host. Re-mutation-checked under the new setup - removing the central catch still fails 3.

The fix itself was never wrong; the test was. All 5 checks green now: 2446 passed, 28 skipped locally, CI test/lint/spine/coverage/GitGuardian all pass.

0xfauzi:
Folded in as `a758d51`.

**The `--spec` early-validation fix.** The check now sits next to the existing agent preflight, before anything spends. Both sources are covered because they fail independently: the file/env value (the `FactoryConfig.load` is hoisted above the decompose branch) and the two flags, which reach `run_factory` without passing any config loader. The flags are validated but deliberately *not applied* there - applying them early would change what `_collect_toml_notes` reports as overridden further down.

Verified against a recording fake agent rather than by reading the control flow:

```
toml nan                   exit=1 agent_invoked=False
toml -3.0                  exit=1 agent_invoked=False
flag --max-cost-usd inf    exit=1 agent_invoked=False
flag --max-total-tokens -5 exit=1 agent_invoked=False
```

`agent_invoked` is a marker file the fake agent touches when it runs, so these tests fail if any call happens before validation. A valid-ceiling case is included too, asserting the preflight does *not* block ordinary runs - otherwise it would read as fixed while breaking everything.

**One thing I fixed beyond what you asked**, flag it if you disagree. While adding the preflight I measured the sibling knob:

```
[factory] max_total_tokens = -5  ->  loads as -5
LoopBudget.token_enabled: False
```

Same defect as the one you found for cost, in the knob that predates it: the negative makes `max_total_tokens > 0` false, so the ceiling disables itself while still reading as configured. Fixing one and leaving the other seemed worse than the small scope increase, so it is now validated at the same four boundaries via a public `validate_token_ceiling` alongside `validate_cost_ceiling`. Happy to split it out if you would rather keep this PR to the cost ceiling.

Also: preflight errors name the **flag** (`--max-cost-usd`) rather than the generic knob, since with three sources the operator needs to know which one to fix. That changed one earlier assertion, updated accordingly.

**Tested:** 2456 passed, 28 skipped. `ruff` and `mypy --strict` clean.

| mutation | result |
|---|---|
| un-hoist the load (validate after decompose again) | 2 fail |
| drop flag validation from the preflight | 2 fail |
| drop the token ceiling validator | 3 fail |

PR #182, opened 2026-07-28
fix: write the engineer's progress log inside the component's scope

0xfauzi:
## The defect

The harness told the engineer to write a file that the harness's own diff-scope check then rejected.

`DECOMPOSE_PROMPT` rule 12 (`kstrl/decompose.py:200-203`) tells the architect that a component's `allowedPaths` include exactly `scripts/kstrl/feature/<component-id>/` -- "(the agent updates progress.txt and PRD passes there)". The architect duly emits e.g. `["hmac_signer/", "tests/", "scripts/kstrl/feature/hmac-sign-verify/"]`.

The code disagreed:

- `config.KstrlConfig.progress_file` defaulted to `scripts/kstrl/progress.txt`, one level **above** the feature subtree.
- `factory._run_factory_locked` relativized that single path once and handed the same run-wide value to every component worker (`_submit_args` -> `_run_component(progress_file_str=...)`).

So the engineer wrote `scripts/kstrl/progress.txt`, Phase 1 `diff_scope` reported "files outside allowed scope", and the component was FAILED and retried from base.

**Measured cost of one such retry on a real paid run: $12.93.** In that run the engineer eventually diagnosed the harness's own bug itself and committed `chore: restore PRD diff scope - untrack uv.lock and progress.txt`, relocating progress into `scripts/kstrl/feature/hmac-sign-verify/progress.md` -- but only after the failure had already been paid for.

## The fix

The prompt was already right, so the **code moves to match the prompt** -- no `*_PROMPT` body changed, so no `*_PROMPT_VERSION` bump or snapshot update is required (H3).

The progress path is now derived from the same source of truth as `prdPath`: it is the PRD's **sibling**. That makes it inside `allowedPaths` by construction rather than by coincidence.

- `config.component_progress_path(prd_path, configured)` -- the single derivation.
- `KstrlConfig.component_progress_file(prd_path, root_dir)` -- the single place the factory resolves it per component; called from `_submit_args`, so the path is per-component instead of run-wide.
- `factory._run_component(progress_file_str=None)` -- None now means "derive it next to this component's PRD", so a direct caller gets the in-scope path too. The safe path is the default, not an opt-in.

### Explicit configuration still wins

`[paths] progress` / `PROGRESS_FILE` is forced on every component exactly as before. Explicitness is **tracked** (`KstrlConfig.progress_file_explicit`, mirroring the existing `kstrl_branch_explicit`) rather than inferred by comparing the value against the default -- the compare-against-default heuristic is the one R2.1 already removed from `VerifyConfig.load`, and it silently drops a setting that happens to equal the default.

### The standalone loop is untouched

`ks run` has no feature subtree, and the derivation reproduces its historical path for free: a PRD at `scripts/kstrl/prd.json` yields `scripts/kstrl/progress.txt`. Covered by a test.

### Two secondary readers of the same path

Both would have gone stale the moment the engineer moved, so they now use the same resolver:

- `verify.run_mechanical_verification` -- the opt-in `require_self_critique` check read `<worktree>/scripts/kstrl/progress.txt`, i.e. a file the engineer was never told to write. It would have failed components for the harness's own path confusion. `VerifyConfig.progress_file_path` is now `None`-defaulted meaning "derive it", matching how `test_command` / `dead_code_command` already express "unset". (A separate explicit-flag field was tried first and rejected: `scripts/gen_docs.py` treats **every** scalar field of `VerifyConfig` as a documented kstrl.toml key, so an internal flag leaks into the generated config reference.)
- `pipeline._phase_distill` -- the knowledge fact-utilization metric had a **second hardcoded copy** of `scripts/kstrl/progress.txt`. It sat inside a bare `try/except`, so it silently read nothing for every decomposed component and under-reported utilization.

### Config surface

`kstrl.toml.example` no longer ships a live `[paths] progress` -- copied verbatim it would have re-pinned the out-of-scope path for every component, reintroducing the bug through configuration. The scaffolded `kstrl.toml` (`ks init`) already had every key commented out; only the comments changed. README's config reference is regenerated from `scripts/gen_docs.py`.

## Verification

| | Before | After |
|---|---|---|
| `uv run pytest tests/` | 2459 passed, 28 skipped | **2472 passed, 28 skipped** |
| collected | 2487 | 2500 (+13 new) |

### Mutation check (the fix was reverted and the tests watched to fail)

The four call sites were reverted to their pre-fix behavior with the new API left in place, so each failure is attributable:

```
6 failed, 7 passed
FAILED TestProgressPathInsideAllowedPaths::test_default_progress_path_is_inside_component_allowed_paths
FAILED TestProgressPathInsideAllowedPaths::test_progress_path_is_a_sibling_of_the_component_prd
FAILED TestFactoryWiring::test_worker_gets_the_feature_subtree_path
FAILED TestFactoryWiring::test_run_component_default_derives_from_the_prd
FAILED TestSelfCritiqueReadsTheComponentLog::test_reads_the_log_next_to_the_component_prd
FAILED TestFactUtilizationReadsTheComponentLog::test_metric_reads_the_component_progress_log
```

The invariant test reproduces the production failure verbatim:

```
E  AssertionError: scripts/kstrl/progress.txt is outside the component's allowedPaths
   ['hmac_signer/', 'tests/', 'scripts/kstrl/feature/hmac-sign-verify/'];
   Phase 1 diff_scope would fail the component
E  +  where False = path_is_allowed('scripts/kstrl/progress.txt', [...])
```

Restored -> 13 passed. The 7 that pass under mutation are the guard tests (explicit-config precedence, single-component layout unchanged), which are supposed to hold either way.

### The invariant test

`tests/test_progress_scope.py::test_default_progress_path_is_inside_component_allowed_paths` asserts the property that was actually violated, with **both sides produced by production code**: the PRD is written by `decompose._generate_component_prd` and re-read by `PRD.load` (so `allowedPaths` is what the factory would really enforce), the progress path comes from `KstrlConfig.component_progress_file`, and containment is judged by `guards.path_is_allowed` -- the same function `check_diff_scope` calls.

## Not fixed (deliberately)

- **`uv.lock` untracked.** The same run also tripped on `uv.lock`. Out of scope per the brief, and I agree it was the experiment's repo setup rather than a harness defect. There is a defensible generalisation -- a lockfile the agent is *required* to regenerate but is not allowed to commit is a scope trap of the same shape -- but the honest fix is for the architect to include lockfiles in `allowedPaths` when the stack has them, not for `check_diff_scope` to grow a hardcoded artifact denylist. A denylist would silently widen every component's write scope, which is exactly what R1.5 hardened against.
- **The in-loop guard never sees the component's `allowedPaths`.** `_run_component` builds `KstrlConfig(allowed_paths=base_config.allowed_paths)` -- the run-wide flag, not `prd.allowed_paths`. So `guards.enforce_allowed_paths` is a no-op in normal factory runs and a scope violation is only discovered after the whole engineer loop has been paid for, at Phase 1. Forwarding the component's own `allowedPaths` (plus `guard_ignored_paths` for harness-owned files) would turn a whole wasted component into a single wasted iteration. Real money, but a behavior change to the engineer loop that deserves its own PR and its own calibration thought.
- **Divergence between `[paths] progress` and `[verify] progress_file_path`.** Setting one but not the other still points writer and reader at different files. Pre-existing, unchanged by this PR, and arguably correct as an explicit-means-explicit contract.

## Assumed vs verified

- **Verified:** the invariant, all four call sites, the mutation results, the explicit-config precedence for toml and env (including an explicit value that equals the default), the single-component layout, the full suite / mypy / ruff, and that no `*_PROMPT` body changed.
- **Assumed, not verified:** that this alone makes the real hmac-sign-verify run pass. That needs a live factory run to confirm; this PR fixes the specific harness-caused `diff_scope` failure the run exhibited, and I did not re-run the $13 experiment.
- **Note:** the progress log is now committed alongside the component PRD (the repo's `.gitignore` only ignores the exact path `scripts/kstrl/progress.txt`). That is the same treatment the per-component `prd.json` already gets and is what the architect's `allowedPaths` anticipates, so it is intended, not incidental.

### DIFF
--- a/README.md
+++ b/README.md
@@ -265,7 +265,7 @@ interactive = false  # human-in-the-loop mode for the legacy loop
 [paths]
 prompt = "scripts/kstrl/prompt.md"              # engineer prompt file
 prd = "scripts/kstrl/prd.json"                  # PRD file
-progress = "scripts/kstrl/progress.txt"         # progress log the agent appends to
+progress = "scripts/kstrl/progress.txt"         # progress log the agent appends to; set = forced on every factory component, unset = each writes beside its own PRD
 codebase_map = "scripts/kstrl/codebase_map.md"  # brownfield codebase notes
 allowed = []                                    # diff-scope allowlist, e.g. ["src/", "tests/"]; empty = unrestricted
 
@@ -319,20 +319,20 @@ allow_network = false  # re-open outbound network inside the sandbox (off = deny
 
 # Phase 1 mechanical verification
 [verify]
-test_command = ""                                  # empty = smart default (uv run pytest)
-typecheck_command = ""                             # empty = smart default (uv run mypy)
-lint_command = ""                                  # empty = smart default (uv run ruff check)
-check_diff_scope = true                            # fail on changes outside allowed paths
-check_bad_patterns = true                          # scan the diff for secret-like patterns
-dead_code_cleanup = false                          # optional dead-code check
-dead_code_command = ""                             # empty = smart default when dead_code_cleanup is on
-mutation_testing = false                           # optional mutation testing
-mutation_threshold = 50.0                          # minimum mutation kill rate (percent)
-mutation_timeout = 600.0                           # seconds for the mutation run
-subprocess_timeout = 300.0                         # seconds per verification subprocess
-require_self_critique = false                      # fail Phase 1 if the ## Self-Critique block is missing/sparse
-self_critique_min_bullets = 3                      # minimum substantive bullets in the block
-progress_file_path = "scripts/kstrl/progress.txt"  # progress file the self-critique check reads
+test_command = ""              # empty = smart default (uv run pytest)
+typecheck_command = ""         # empty = smart default (uv run mypy)
+lint_command = ""              # empty = smart default (uv run ruff check)
+check_diff_scope = true        # fail on changes outside allowed paths
+check_bad_patterns = true      # scan the diff for secret-like patterns
+dead_code_cleanup = false      # optional dead-code check
+dead_code_command = ""         # empty = smart default when dead_code_cleanup is on
+mutation_testing = false       # optional mutation testing
+mutation_threshold = 50.0      # minimum mutation kill rate (percent)
+mutation_timeout = 600.0       # seconds for the mutation run
+subprocess_timeout = 300.0     # seconds per verification subprocess
+require_self_critique = false  # fail Phase 1 if the ## Self-Critique block is missing/sparse
+self_critique_min_bullets = 3  # minimum substantive bullets in the block
+progress_file_path = ""        # progress file the self-critique check reads; empty = the log beside the component's PRD
 
 # Phase 1 approved-fixtures oracle (R7.2; default off)
 [fixtures]
--- a/README.md
+++ b/README.md
@@ -265,7 +265,7 @@ interactive = false  # human-in-the-loop mode for the legacy loop
 [paths]
 prompt = "scripts/kstrl/prompt.md"              # engineer prompt file
 prd = "scripts/kstrl/prd.json"                  # PRD file
-progress = "scripts/kstrl/progress.txt"         # progress log the agent appends to; set = forced on every factory component, unset = each writes beside its own PRD
+progress = ""                                   # progress log the agent appends to; empty = each factory component writes beside its own PRD (inside its allowedPaths), set = that one path is forced on every component
 codebase_map = "scripts/kstrl/codebase_map.md"  # brownfield codebase notes
 allowed = []                                    # diff-scope allowlist, e.g. ["src/", "tests/"]; empty = unrestricted
 
--- a/kstrl/cli.py
+++ b/kstrl/cli.py
@@ -31,7 +31,12 @@ from kstrl.agents.base import Agent
 from kstrl.agents.logging import LoggingAgent
 from kstrl.breaker import BreakerConfig
 from kstrl.commandrun import CommandRun, open_command_run
-from kstrl.config import KstrlConfig, _parse_paths, resolve_config_file
+from kstrl.config import (
+    KstrlConfig,
+    _parse_paths,
+    reconcile_progress_paths,
+    resolve_config_file,
+)
 from kstrl.config_report import build_config_report
 from kstrl.config_report import normalize_ui_mode as _normalize_ui_mode
 from kstrl.decompose import SpecBlockerError, decompose_spec
@@ -1271,6 +1276,8 @@ def feature(
     # R2.4 preflight: accept whichever agent the resolved config selects.
     _check_agent_preflight(base_config, ui_impl)
 
+
+
     sandbox_cfg = SandboxConfig.load(root_dir)
     if sandbox_cfg.enabled and base_config.agent_cmd:
         ui_impl.warn(
@@ -2225,6 +2232,36 @@ def factory(
     # through to the codex default - and _cli_family misreads the
     # engineer family, inverting the R7.1 reviewer rotation.
     _check_agent_preflight(base_config, ui_impl)
+
+    # --no-verify leaves no reader, so there is nothing to reconcile.
+    if v_config is not None:
+        # R8 review: the progress log's writer ([paths] progress) and its
+        # reader ([verify] progress_file_path) default to the same
+        # derivation, so they agree until exactly ONE is set - and then the
+        # self-critique check inspects a file the engineer never wrote and
+        # fails for a reason the operator cannot see from either setting.
+        writer_path, reader_path = reconcile_progress_paths(
+            base_config.progress_file if base_config.progress_file_explicit else None,
+            v_config.progress_file_path,
+        )
+        if (
+            writer_path is not None
+            and reader_path is not None
+            and writer_path != reader_path
+        ):
+            # Both named, and named differently. Not overridden - an operator
+            # who set two paths meant two paths - but said out loud, because
+            # the failure it produces is otherwise unattributable.
+            ui_impl.warn(
+                f"progress log writer ([paths] progress = {writer_path}) and "
+                f"reader ([verify] progress_file_path = {reader_path}) point "
+                "at different files; the self-critique check will inspect a "
+                "file the engineer does not write"
+            )
+        v_config.progress_file_path = reader_path
+        if writer_path is not None:
+            base_config.progress_file = Path(writer_path)
+            base_config.progress_file_explicit = True
 
     # Ensure prompt file exists
     if not base_config.prompt_file.exists():
--- a/kstrl/cli.py
+++ b/kstrl/cli.py
@@ -34,7 +34,7 @@ from kstrl.commandrun import CommandRun, open_command_run
 from kstrl.config import (
     KstrlConfig,
     _parse_paths,
-    reconcile_progress_paths,
+    reconcile_progress_config,
     resolve_config_file,
 )
 from kstrl.config_report import build_config_report
@@ -2240,28 +2240,12 @@ def factory(
         # derivation, so they agree until exactly ONE is set - and then the
         # self-critique check inspects a file the engineer never wrote and
         # fails for a reason the operator cannot see from either setting.
-        writer_path, reader_path = reconcile_progress_paths(
-            base_config.progress_file if base_config.progress_file_explicit else None,
-            v_config.progress_file_path,
-        )
-        if (
-            writer_path is not None
-            and reader_path is not None
-            and writer_path != reader_path
-        ):
-            # Both named, and named differently. Not overridden - an operator
-            # who set two paths meant two paths - but said out loud, because
-            # the failure it produces is otherwise unattributable.
-            ui_impl.warn(
-                f"progress log writer ([paths] progress = {writer_path}) and "
-                f"reader ([verify] progress_file_path = {reader_path}) point "
-                "at different files; the self-critique check will inspect a "
-                "file the engineer does not write"
-            )
-        v_config.progress_file_path = reader_path
-        if writer_path is not None:
-            base_config.progress_file = Path(writer_path)
-            base_config.progress_file_explicit = True
+        # The reconciliation lives in config.py (and is applied in ONE path
+        # domain, review finding 3) so a test can exercise the same wiring
+        # this command runs.
+        mismatch = reconcile_progress_config(base_config, v_config, root_dir)
+        if mismatch is not None:
+            ui_impl.warn(mismatch)
 
     # Ensure prompt file exists
     if not base_config.prompt_file.exists():
--- a/kstrl/config.py
+++ b/kstrl/config.py
@@ -61,6 +61,34 @@ def relative_to_root(path: Path, root_dir: Path) -> str:
 COMPONENT_PROGRESS_FILENAME = "progress.txt"
 
 
+def reconcile_progress_paths(
+    writer_explicit: str | Path | None,
+    reader_explicit: str | Path | None,
+) -> tuple[str | None, str | None]:
+    """Keep the progress log's WRITER and READER pointing at one file.
+
+    ``[paths] progress`` (or ``PROGRESS_FILE``) configures where the
+    engineer WRITES its progress log; ``[verify] progress_file_path``
+    (or ``KSTRL_VERIFY_PROGRESS_FILE``) configures where the
+    self-critique check READS it. Both default to the same derivation,
+    so they agree until exactly one of them is set - and then the
+    reader silently inspects a file the engineer never wrote, which
+    fails the self-critique gate for a reason the operator cannot see.
+
+    Setting one propagates it to the other. Setting BOTH is left alone,
+    including when they differ: an operator who named two paths meant
+    two paths, and the caller warns rather than overriding. Returns
+    ``(writer, reader)`` as strings, or None where nothing is set.
+    """
+    writer = str(writer_explicit) if writer_explicit is not None else None
+    reader = str(reader_explicit) if reader_explicit is not None else None
+    if writer is not None and reader is None:
+        return writer, writer
+    if reader is not None and writer is None:
+        return reader, reader
+    return writer, reader
+
+
 def component_progress_path(
     prd_path: str | Path,
     configured: str | Path | None = None,
--- a/kstrl/config.py
+++ b/kstrl/config.py
@@ -7,7 +7,7 @@ import re
 import tomllib
 from dataclasses import dataclass, field
 from pathlib import Path
-from typing import Any
+from typing import Any, Protocol
 
 
 def _parse_bool(value: str | None) -> bool:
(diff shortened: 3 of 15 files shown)


README.md
```diff
@@ -265,7 +265,7 @@ interactive = false  # human-in-the-loop mode for the legacy loop
 [paths]
 prompt = "scripts/kstrl/prompt.md"              # engineer prompt file
 prd = "scripts/kstrl/prd.json"                  # PRD file
-progress = "scripts/kstrl/progress.txt"         # progress log the agent appends to
+progress = "scripts/kstrl/progress.txt"         # progress log the agent appends to; set = forced on every factory component, unset = each writes beside its own PRD
```
0xfauzi: [P1] This generated reference still emits `progress` as an active TOML key, so copying it recreates the exact defect this PR fixes. I loaded a `kstrl.toml` containing this documented line and `component_progress_file('scripts/kstrl/feature/demo/prd.json', root)` resolved to `scripts/kstrl/progress.txt`; the production `path_is_allowed` check returned false for the component's feature allowlist. Commenting the key only in `kstrl.toml.example` is insufficient because `scripts/gen_docs.py` regenerates this active block, and `ks config show` likewise reports the obsolete root path as the effective default. Please teach the generator/reporting surface to represent this setting as unset or per-component-derived, emit the README line commented out (or omit it), and add a regression that loads the generated example and asserts the progress path remains inside the component scope.

kstrl/config.py
```diff
@@ -128,6 +183,23 @@ def load(
...
+
+        The single place the factory decides where a component's
+        engineer writes its progress log: an explicit configuration wins
+        for every component, otherwise the path is derived from the
+        component's own PRD so it sits inside the component's
+        allowedPaths (see ``component_progress_path``).
+        """
+        configured = (
+            relative_to_root(self.progress_file, root_dir)
+            if self.progress_file_explicit else None
```
0xfauzi: [P2] A caller that supplies the public dataclass field directly now has its value silently ignored unless it also knows to set this new bookkeeping flag. For example, `KstrlConfig(progress_file=Path('docs/custom-progress.md')).component_progress_file('scripts/kstrl/feature/demo/prd.json', root)` returns the derived feature path, not the supplied custom path. Before this PR, `run_factory(..., base_config)` honored `base_config.progress_file`, so this breaks programmatic users even though TOML/env loaders set the flag correctly. Please represent the unset state with a sentinel or otherwise make constructor/assignment overrides unambiguously explicit without requiring callers to coordinate two fields, and add a `run_factory` regression using a directly constructed config.

0xfauzi (commented):
The per-component derivation itself is well-factored and the CLI-loaded TOML/env paths are covered convincingly. All repository gates pass. I found two configuration-surface regressions that the new tests do not cover, one of which recreates the original out-of-scope path when users copy the generated README reference. Suggested fixes and regression tests are inline.

0xfauzi:
Pushed `e82ab20` - the two deferred items are now fixed, plus a third problem found while fixing them.

**1. The in-loop guard was inert.** `_run_component` received `base_config.allowed_paths`, the run-wide `--allowed-paths` flag, which is empty in every ordinary factory run. So `if config.allowed_paths` in `run_loop` was False and `guards.enforce_allowed_paths` never executed. `_component_scope` now resolves the component's architect-emitted `allowedPaths`.

Two design points worth stating rather than burying:

- **Read from `root_dir`, not the worktree.** Phase 1 deliberately re-reads the PRD from the worktree, but an in-loop guard sourced from a file the agent can edit would let the agent widen its own scope mid-run. The pre-run PRD is the honest input for enforcement; the worktree copy stays the input for verification.
- **It fails OPEN where Phase 1 fails CLOSED.** That asymmetry is deliberate: this is a cheap tripwire, not the gate. An unreadable PRD failing closed here would kill a component before the engineer had done anything, and Phase 1 still refuses the same PRD (R1.5), so nothing merges unverified. There is a test asserting both halves.

**2. Found while fixing 1: the guard threw away its own evidence.** `ok, _ = guards.enforce_allowed_paths(...)` discarded the violations, and the factory then rendered the halt as `"Did not complete"`. An agent told only that something went wrong will repeat the edit - the diagnosis it needs is exactly the list it was denied. `LoopResult` now carries `guard_violations` and the component error names the files, worded to mirror `check_diff_scope` so the same failure reads identically wherever it is caught.

**3. Writer/reader divergence.** `reconcile_progress_paths` propagates whichever of `[paths] progress` / `[verify] progress_file_path` is set. Two explicitly different paths are left alone and warned about - an operator who named two paths meant two paths - but the silent single-setting case, where the self-critique check inspects a file the engineer never writes, is gone.

**Still not fixed, and I want to be explicit about why:** the lockfile scope trap. The right fix is the architect listing lockfiles in `allowedPaths` when the stack has them, which is a `DECOMPOSE_PROMPT` change - and under H2 a prompt edit ships only with a calibration re-run against the saved baseline. That is real LLM spend and a separate decision, so it stays out of this PR rather than going in untested.

**Tested:** 2482 passed, 28 skipped. `ruff` and `mypy --strict` clean.

| mutation | result |
|---|---|
| revert guard to the run-wide flag (the inert bug) | 2 fail |
| revert reconciliation to a no-op | 2 fail |

One caveat I cannot close from here: making the guard live changes engineer-loop behavior, and I have not proven the false-positive rate on a real run. In the run that motivated this, the engineer created `uv.lock` and `progress.txt` early and cleaned them up four iterations later - with the guard live it would now fail at iteration 1 instead. That is the intended saving, but it does remove the self-correction path, and only a live run will show whether that trade holds.

0xfauzi (commented):
Re-reviewed `e82ab20`. The full suite, strict mypy, ruff, generated-doc check, and current GitHub checks are green, but both earlier findings still reproduce. The new commit also adds runtime behavior that its helper-level tests do not cover: path reconciliation resolves the writer and reader in different filesystem contexts, and the newly activated guard has no per-worker baseline, so it misses committed agent edits while attributing pre-existing operator dirt to the agent. I left inline fixes for those issues and for the incomplete unreadable-PRD handling.

kstrl/cli.py
```diff
@@ -2226,6 +2233,36 @@ def factory(
...
     _check_agent_preflight(base_config, ui_impl)
 
+    # --no-verify leaves no reader, so there is nothing to reconcile.
+    if v_config is not None:
+        # R8 review: the progress log's writer ([paths] progress) and its
+        # reader ([verify] progress_file_path) default to the same
+        # derivation, so they agree until exactly ONE is set - and then the
+        # self-critique check inspects a file the engineer never wrote and
+        # fails for a reason the operator cannot see from either setting.
+        writer_path, reader_path = reconcile_progress_paths(
```
0xfauzi: [P1] Normalize this path before copying it to the verification config. `KstrlConfig.load(root)` resolves `[paths] progress = "docs/custom-progress.md"` to an absolute path under the main checkout. The writer later calls `component_progress_file`, which relativizes that value and therefore writes `<worktree>/docs/custom-progress.md`, but this assignment leaves the reader absolute, so `run_mechanical_verification` reads `<root>/docs/custom-progress.md`. I reproduced the reconciled strings comparing equal while the runtime writer and reader paths compared unequal. This still fails self-critique, or can pass against a stale root file. Reconcile in one path domain, preferably root-relative for paths inside `root_dir`, and add an integration test that loads TOML, runs the factory wiring, and compares the actual worker write path with the actual verification read path.

kstrl/factory.py
```diff
@@ -2511,7 +2558,23 @@ def _submit_args(comp: Component, wt_path: Path) -> tuple[Any, ...]:
...
+            # component, versus ~$2.50 for the single iteration it takes
+            # to catch it here.
+            #
+            # Read from root_dir, NOT the worktree: Phase 1 deliberately
+            # re-reads the PRD from the worktree (fail-closed on load
+            # error), but an in-loop guard sourced from a file the agent
+            # may edit would let the agent widen its own scope mid-run.
+            # The pre-run PRD is the honest input for enforcement; the
+            # worktree copy remains the input for verification.
+            _component_scope(comp, root_dir, base_config),
```
0xfauzi, line 2606: [P1] Activating the component scope here needs a per-worker change baseline. The downstream guard calls `git.get_changed_files()`, which only reports staged, unstaged, and untracked files. In a real repo I committed an out-of-scope agent edit and `enforce_allowed_paths` returned `(True, [])`, so the claimed early tripwire is still bypassed whenever the engineer commits. Conversely, an uncommitted `operator-notes.txt` that existed before a `--no-worktrees` run was reported as the agent's violation; in interactive mode, `Revert and continue` can delete or restore that operator-owned change. Capture the worker's initial HEAD and dirty state, then compare the complete agent delta since that baseline, including commits, or refuse a dirty no-worktree run. Add regressions for both a committed violation and pre-existing dirt.

kstrl/factory.py
```diff
@@ -1006,6 +1007,31 @@ def _kept(path: Path) -> bool:
...
+    gate. A PRD that cannot be read here returns None, the loop runs
+    unguarded, and Phase 1 still fails CLOSED on the same unreadable PRD
+    (R1.5) - so nothing merges unverified. Failing closed here instead
+    would convert an unreadable PRD into an immediate component failure
+    before the engineer has done anything, which is a worse trade for a
+    tripwire whose only job is to save iterations.
+    """
+    try:
+        prd = PRD.load(root_dir / comp.prd_path)
+    except (FileNotFoundError, ValueError):
```
0xfauzi: [P2] The documented fail-open behavior only covers a missing or parse-invalid PRD. Other ordinary read failures escape this helper and abort scheduling before Phase 1: making `prd.json` a directory reproduces an uncaught `IsADirectoryError`, and permission/I/O failures are also `OSError` subclasses not covered here. Catch the intended read-error family (for example `OSError` plus `ValueError`) and make Phase 1's fail-closed loader handle the same family so an unreadable PRD becomes a verification result rather than a factory exception. Add a directory or permission-error regression, not only the missing-file case.

README.md
0xfauzi: Still reproducible on `e82ab20`. The new commit does not change `README.md`, `scripts/gen_docs.py`, or the config-report surface, and the generated README still contains the active `progress = "scripts/kstrl/progress.txt"` key. Loading that documented config still pins a decomposed component to the out-of-scope root log. Leaving this thread unresolved.

kstrl/config.py
0xfauzi: Still reproducible on `e82ab20`: `KstrlConfig(progress_file=Path("docs/custom-progress.md"))` leaves `progress_file_explicit=False`, and `component_progress_file(...)` still returns the derived feature path. The new CLI-only reconciliation does not repair direct construction or programmatic `run_factory` callers. Leaving this thread unresolved.

0xfauzi:
All five findings addressed in `16c4f87`. Each was reproduced first; I re-verified every fix independently rather than relying on the implementation report.

**[P1] Generated docs shipped the defective key.** `gen_docs.py` and the config-report surface now represent the setting as unset / per-component-derived. Verified: no active `progress =` line pinning the root path survives in the generated README.

**[P2] Two-field coordination.** Replaced with a `None` sentinel; `progress_file_explicit` is gone. Verified:
```
KstrlConfig(progress_file="docs/custom-progress.md") -> docs/custom-progress.md   (was: silently derived)
KstrlConfig()                                        -> scripts/kstrl/feature/demo/progress.txt
```
`resolved_progress_file` keeps the standalone `ks run` path concrete.

**[P1] Path domains.** `reconcile_progress_config` normalises to root-relative before comparing. Verified on the actual runtime paths, not the reconciled strings:
```
worker WRITES -> /tmp/WORKTREE/docs/custom-progress.md
verify  READS -> /tmp/WORKTREE/docs/custom-progress.md
```
Two explicitly different paths are still left alone and warned about.

**[P1] Guard baseline.** You were right, and it was worse than a gap - the guard I activated was close to useless, since the engineer commits after every story. Reproduced:
```
agent writes OUTSIDE.md, git add -A, git commit
get_changed_files()    -> set()
enforce_allowed_paths  -> ok=True, violations=[]
```

One correction to your suggested fix, which I want to be explicit about. I first anchored the baseline on the worker's initial HEAD as you described, and an existing test caught that this **rejects the retry's own fix**: a retry resumes the branch carrying the previous attempt's out-of-scope commit, so *deleting* that file registers as a fresh out-of-scope change. The baseline is now the **base branch** - the same point `check_diff_scope` measures from, where add-then-delete nets to nothing. Pre-existing dirt is still subtracted, so operator files in a `--no-worktrees` run are never attributed to the agent. All four cases verified: committed violation caught, operator dirt excluded, self-revert not over-reported, `baseline=None` preserves `ks run`.

**[P2] Exception family.** Now `OSError` + `ValueError`, matched in Phase 1's fail-closed loader. Verified: a `prd.json` that is a directory returns `None` instead of aborting scheduling.

---

**Two regressions the fixes exposed**, both caught by existing tests rather than by me:

- Moving the catch earlier meant my in-loop message - not `check_diff_scope`'s - reaches attempt 2, dropping the R0.4 guidance. `base_branch` is now threaded to the worker and the message carries the base-branch and complete-allowed-paths lines verbatim.
- A scope violation caught in-loop emitted **no `verification_result` at all**, silently changing the journal's shape. The guard now emits the same `diff_scope` failure record Phase 1 would - reporting only that one check, since claiming tests and typecheck ran when the loop halted before them would be false.

**Tested:** 2507 passed, 28 skipped. `mypy --strict` and `ruff` clean.

kstrl/factory.py
0xfauzi: Post-merge audit of `16c4f87`: the baseline fixes the original committed-file blind spot, but three integration defects remain.

1. **[P1] The guard and Phase 1 select different base refs.** `capture_workspace_baseline(base_ref="main")` calls the new `resolve_ref`, which prefers local `main`; `check_diff_scope` uses `resolve_base_ref`, which prefers `origin/main`. With stale local `main` and a worker cut from current `origin/main`, I measured the guard baseline at the stale SHA and `get_changed_files_since` reported an upstream `docs/upstream.md` before the agent changed anything, while Phase 1 selected `origin/main`. Reuse `resolve_base_ref` here so both layers resolve the same commit, and add a stale-local-main regression.

2. **[P1] Pre-existing dirt is subtracted by filename, not by captured content.** If `operator-notes.txt` is dirty at baseline and the agent overwrites it, `get_changed_files_since` still returns an empty set. This can silently destroy operator work in `--no-worktrees` mode. Capture content/index state for dirty paths and compare against that snapshot, or refuse a dirty no-worktree run when enforcement is active; do not exempt the pathname for the whole run.

3. **[P2] Interactive revert clears the guard view but not committed history.** After the agent commits `OUTSIDE.md`, choosing `Revert and continue` deletes/stages the path and `get_changed_files_since` becomes empty, but `git diff main...HEAD` still contains `OUTSIDE.md` and Phase 1 fails it; `git status --short` is `D  OUTSIDE.md`. Either create a path-limited reversal commit before reporting success or keep the guard unresolved until the reversal is committed. Add an assertion against the production `get_diff_names(base_branch)`, not only `get_changed_files_since`.

Focused tests still pass (`48 passed`), so these cases are not covered by the new suite.

PR #183, opened 2026-07-29
fix: split an oversized single-file diff on hunk boundaries

0xfauzi:
## The defect

Hard-mode review chunks the component diff so each reviewer pass fits the prompt cap. `git.split_diff_for_prompt` split **only** on `diff --git` file boundaries, so a single file whose own diff exceeded the per-chunk budget had no boundary to split on and raised `DiffUnsplittableError`.

From a paid end-to-end factory run:

```
ERROR: Diff unsplittable for hmac-sign-verify: single-file diff segment is 55710 chars,
       over the 49880-char per-chunk budget; cannot split on file boundaries
       (diff --git a/tests/test_purity.py b/tests/test_purity.py)
```

## Severity: cost, not correctness

This is **not** an unwinnable retry, and I am not claiming it was. The existing handler (`pipeline.py`, `except git.DiffUnsplittableError`) fails closed into the retry path with an explicit "the diff is too large to review" signal, and in the observed run the engineer recovered on the next iteration - it committed `feat: US-005 ... (split purity tests to fit review cap)` and took the file from 55,710 to 14,912 chars. The mechanism works.

The problem is what recovery costs. That retry was a full engineer-loop invocation: **$3.99 measured for that single call**, in a run where engineer calls cost $1.70-$7.42 each. The money bought a repackaging of a diff that the harness could not chunk - not a fix to a defect in the code under review. A test file legitimately larger than the chunk budget is a normal outcome, not agent misbehaviour, so charging the engineer for it is the wrong trade.

## The fix

`split_diff_for_prompt` now splits an over-budget file further on `@@` hunk boundaries (`_split_file_segment` in `kstrl/git.py`). Design points:

- **Every part is a self-describing reviewable unit.** Each part repeats the file header verbatim (`diff --git` / `index` / `---` / `+++`) so a reviewer holding part 2 knows what it is looking at.
- **Every part is labelled.** A `# [kstrl R1.4] file part i of n: <path>` marker names the file and the part number, so a part cannot be read as the file's whole change. Continuation parts additionally say the header below them is *a repeat for context, not a second change*, so the repeated header is not reviewed as duplicated work.
- **The chunk-level provenance header stays honest.** Chunks carrying only whole files keep the pre-R8 wording byte-for-byte; chunks carrying a within-file part say `split on file/hunk boundaries` instead. A reviewer is never told "file boundaries" while holding a fragment of a file.
- **No prompt bodies were touched.** These are provenance lines in `git.py`, not `*_PROMPT` constants - no H3 version bump or H2 calibration is implicated, and `tests/test_prompt_versions.py` (which AST-walks `kstrl/` for module-level `*_PROMPT` constants) still passes unchanged.
- **The budget is unchanged.** `DEFAULT_PROMPT_DIFF_CHAR_LIMIT` is still 50,000 and `_CHUNK_HEADER_RESERVE` is still 120; a test asserts the cap explicitly so the fix cannot later be "solved" by raising it.
- **Existing helpers reused,** not duplicated: the same `_CHUNK_HEADER_RESERVE`, the same greedy packer, the same `DiffUnsplittableError`. `truncate_diff_for_prompt` is deliberately untouched - it is the advisory-mode path, which is explicitly allowed to be `partial`.

## The residual case I did not handle, and why

**A single hunk larger than the budget still raises `DiffUnsplittableError` and still fails closed through the retry path.** That is deliberate, on two grounds:

1. R1.4 forbids truncating a diff that is being reviewed for correctness. Below hunk granularity there is no boundary that preserves a reviewable unit, and silently truncating would let an unreviewed tail merge - the exact failure H-16 was written to prevent.
2. A >49KB single hunk is hundreds of contiguous changed lines with no unchanged context line for git to break on. No reviewer, human or LLM, can hold that at this cap either. Demanding the engineer restructure it is a legitimate ask about the change, not a harness workaround - which is precisely the distinction that made the original behaviour wrong.

**Is the retry path still correct for the residual case?** Yes, and I left its shape alone. Re-running the engineer loop is the right response when the demand is genuinely on the code. I made one narrow change inside it: the retry context said *"Reduce the change so each file's diff fits the 50KB review cap"*, which is now stale guidance - files may exceed the cap, individual hunks may not. It now names hunk granularity. That is a one-string edit in the failure path, mutation-tested; I did not expand scope further.

## Verification

Reproduction first: the new tests were written against the unfixed code and **watched fail**.

| | before | after |
|---|---|---|
| `uv run pytest tests/ -q` | 2507 passed, 28 skipped | **2518 passed, 28 skipped** (+11) |

**Reproduction.** With the fix reverted, the new end-to-end test fails with the production error string, reproducing the run exactly:

```
Review diff unsplittable at the prompt cap: single-file diff segment is 55715 chars,
over the 49880-char per-chunk budget; cannot split on file boundaries
```

(55,715 synthetic vs 55,710 observed; same 49,880 budget.)

**Mutation check.** Fix reverted, tests run, failures watched, fix restored, tests re-run green:

- `kstrl/git.py` reverted to `origin/main`: **11 of 12** new/changed tests fail; 41 -> 30 passed in `tests/test_truncation_policy.py`. Restored: 41/41.
- The 12th is `test_multi_file_packing_is_unchanged_by_within_file_splitting`, the no-drift guard, which passes both before and after **by design**. To prove it is not inert I mutated the chunk header string by three characters on the fixed code and watched it fail, then restored it.
- The pipeline retry-guidance assertion was mutated separately (guidance string reverted to its old wording) and watched fail, then restored.

**Round-trip property.** Asserted two ways, since one alone is not enough:

- *Byte-exact reassembly* - stripping every injected provenance line (and the repeated header on continuation parts) and concatenating reproduces the input byte for byte. Nothing dropped, nothing altered.
- *Each hunk exactly once* - the list of `@@` hunk headers across the raw concatenation of the chunks equals the original's, in order. Nothing duplicated across chunks either.

**Not just synthetic fixtures.** I built a throwaway git repo with a 1,300-function test file, scattered edits through it, and ran the real `git diff` output through the splitter: **134,383 chars, 434 hunks, one file -> 3 chunks**, max chunk 49,893 (cap 50,000), byte-exact reassembly `True`, all 434 hunks present exactly once, every chunk carrying the real file header including the `index` line.

**Verified vs assumed.** Verified: chunk sizes, hunk coverage, reassembly, the residual raise, no-drift on multi-file packing, the end-to-end pipeline path (component completes and every hunk reaches a reviewer prompt), and behaviour on real git output. Assumed, not measured: that the *reviewer's* detection quality is unchanged when a file arrives in parts - that is a calibration question, out of scope here since no prompt body changed.

### DIFF
--- a/kstrl/git.py
+++ b/kstrl/git.py
@@ -826,6 +826,16 @@ _FILE_PART_CONTINUED_MARKER = (
 )
 
 
+# Cap on the part-count fixed-point rounds in _split_file_segment.
+# The iteration provably converges (see the proof there) in at most one
+# round per distinct DIGIT WIDTH of the part count, and a part count
+# cannot exceed the hunk count of a string held in memory, so 24 rounds
+# is far above the reachable ceiling (len(str(2**63)) == 19). It exists
+# only so an unforeseen non-monotonicity fails loudly instead of
+# spinning; hitting it is a bug, not a diff shape.
+_PART_COUNT_FIXED_POINT_ROUNDS = 24
+
+
 @dataclass(frozen=True)
 class _DiffUnit:
     """One indivisible piece of a diff for packing purposes: either a
@@ -849,6 +859,43 @@ def _file_segment_label(header: str) -> str:
     return first_line[:200] or "(unnamed diff segment)"
 
 
+def _part_marker_reserve(path: str, parts: int) -> int:
+    """Width of the widest part-marker line for a file rendered as
+    ``parts`` parts.
+
+    Part indices run ``1..parts`` and ``len(str(i)) <= len(str(parts))``,
+    so formatting both fields with ``parts`` bounds every part's marker
+    exactly - no guessed digit padding. Depends on ``parts`` ONLY through
+    its digit count, which is what bounds the fixed-point iteration in
+    :func:`_split_file_segment`.
+    """
+    return max(
+        len(_FILE_PART_MARKER.format(i=parts, n=parts, path=path)),
+        len(_FILE_PART_CONTINUED_MARKER.format(i=parts, n=parts, path=path)),
+    )
+
+
+def _pack_hunks(hunks: list[str], content_budget: int) -> list[list[str]]:
+    """Group ``hunks`` into the fewest ordered, contiguous runs that each
+    fit ``content_budget``.
+
+    Order-preserving next-fit: since a part must be a contiguous slice of
+    the file (that is what keeps reassembly byte-exact), closing a group
+    only when the next hunk would overflow is optimal - any partition
+    into contiguous runs is dominated by this one. Callers must have
+    checked that every hunk fits ``content_budget`` on its own.
+    """
+    groups: list[list[str]] = [[]]
+    size = 0
+    for hunk in hunks:
+        if groups[-1] and size + len(hunk) > content_budget:
+            groups.append([])
+            size = 0
+        groups[-1].append(hunk)
+        size += len(hunk)
+    return groups
+
+
 def _split_file_segment(segment: str, budget: int) -> list[_DiffUnit]:
     """R8: split ONE file's diff segment into ``<=budget`` parts on
     ``@@`` hunk boundaries, repeating the file header on every part.
@@ -858,6 +905,11 @@ def _split_file_segment(segment: str, budget: int) -> list[_DiffUnit]:
     test file, and recovery cost a full engineer-loop pass ($3.99) to
     repackage a diff the harness simply could not chunk. A file bigger
     than the per-chunk budget is a normal outcome, not misbehaviour.
+
+    Produces the FEWEST parts the budget allows: the width reserved for
+    the ``file part i of n`` marker is derived from the part count the
+    packer actually settles on (see the fixed point below), so no hunk is
+    rejected over headroom the rendering never uses.
 
     Raises :class:`DiffUnsplittableError` when even hunk granularity is
     not enough (one hunk over the budget, or no hunks at all). That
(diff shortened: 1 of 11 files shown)


0xfauzi (commented):
Reviewed the complete PR diff and surrounding chunked review/security paths at `200e30c6`. The core hunk-boundary split works for the production-shaped case and preserves the cap and byte coverage, but I found two actionable edge defects: the calibrated reviewer prompts still describe file-only chunking, and the conservative marker-width calculation can still reject a diff that has a valid under-cap hunk partition.

Verification: `uv run pytest -q` (2,518 passed, 28 skipped), focused `tests/test_truncation_policy.py tests/test_pipeline.py` (73 passed), Ruff, mypy on the changed source files, `git diff --check`, and a targeted default-50KB reproduction for the false rejection. GitHub CI and GitGuardian checks are green.

kstrl/git.py
```diff
@@ -782,10 +782,157 @@ class DiffUnsplittableError(ValueError):
...
+# before; chunks carrying a within-file part get the accurate wording
+# instead (a reviewer must not be told "file boundaries" while holding
+# a fragment of a file).
+_CHUNK_HEADER = (
+    "# [kstrl R1.4] diff chunk {i} of {n}: oversized diff split on file "
+    "boundaries; other files are in other chunks\n"
+)
+_CHUNK_HEADER_PARTIAL_FILE = (
+    "# [kstrl R1.4] diff chunk {i} of {n}: oversized diff split on "
+    "file/hunk boundaries; the rest is in other chunks\n"
```
0xfauzi, line 810: [P2] Keep the reviewer prompts consistent with within-file chunks

This new header says the chunk may be split on hunk boundaries, but both calibrated prompt bodies still explicitly say that `# [kstrl R1.4] diff chunk ...` means the diff was split on **file boundaries** and that only cross-file interactions are invisible (`REVIEWER_PROMPT` around lines 408-413 and `SECURITY_PROMPT` around lines 447-452). For a continuation part, another relevant hunk from this same file is now absent, so those instructions give the reviewer an incorrect model of what it cannot see and can make a same-file, cross-hunk defect look complete. The part marker partially contradicts that guidance, which is not a reliable hard-mode contract. Please update both prompt bodies to describe file/hunk splitting and omitted-part interactions, bump their prompt versions/snapshots as required, and add a regression that a within-file chunk is accompanied by truthful review instructions.

kstrl/git.py
```diff
@@ -782,10 +782,157 @@ class DiffUnsplittableError(ValueError):
...
+
+    # The marker cannot be rendered until the part count is known, so
+    # reserve its worst-case width now. Part indices and the part count
+    # are both bounded by the hunk count, so formatting with that count
+    # bounds the marker's width exactly - no guessed digit padding.
+    widest = len(hunks)
+    reserve = max(
+        len(_FILE_PART_MARKER.format(i=widest, n=widest, path=path)),
+        len(_FILE_PART_CONTINUED_MARKER.format(i=widest, n=widest, path=path)),
+    )
```
0xfauzi: [P3] Do not size the marker from the hunk count

`len(hunks)` is only an upper bound on the eventual part count, but its digit width is used to reduce `content_budget` before grouping. That can reject a hunk which fits once the actual part count is known. At the default 50,000-character limit I reproduced this with one 49,706-character hunk followed by 999 tiny hunks: the current code raises because the 1,000-hunk reserve leaves 49,705 characters, while the exact two-part rendering is valid at 49,983 and 15,258 characters including both outer and file-part headers. This therefore still buys the engineer a harness-only retry even though a compliant hunk-boundary partition exists. Please derive marker widths from the actual part count, for example by packing to a fixed point or repacking after rendering exact markers, and cover this boundary with a regression test.

0xfauzi: Fixed in 6358b60. Confirmed real - I reproduced your case before changing anything.

**Approach: pack to a fixed point.** The part count and the marker width are mutually dependent, so I solve them together rather than bounding one with the other. `_split_file_segment` now starts from `parts_assumed = 1`, packs, and re-packs with the reserve implied by the count it just got, until the count reproduces itself. I picked this over "repack after rendering exact markers" because it terminates by construction rather than by inspection, and because the fixed point it lands on is the *least* one - the fewest parts, so the fewest reviewer passes.

**Termination** (also written out at `kstrl/git.py`, in the comment above the loop): the reserve is non-decreasing in the assumed part count, since it grows only with that count's digit width; therefore `content_budget` is non-increasing and the packed group count `f(n)` is non-decreasing. Starting at `n = 1`, the smallest possible count, gives `f(n_0) >= n_0`, and monotonicity carries that forward, so the sequence is non-decreasing and bounded above by the hunk count. It settles; it cannot oscillate. Because `n` enters only through its digit width, a round that does not change the width *is* the fixed point, so it runs at most one round per digit width (<= ~19 for any in-memory string; 2 rounds for both cases below). The loop is bounded at 24 rounds regardless and raises `DiffUnsplittableError` if that bound is ever reached, rather than looping - `test_non_convergence_fails_closed_instead_of_looping` patches the bound down to 1 and asserts the raise, so the non-converging branch is covered, not just asserted.

**Your case, before and after.** Fixture: `xx.py`, 51-char file header, one hunk of exactly 49,706 chars, then 999 tiny hunks, at the default 50,000 cap.

| | result |
|---|---|
| before | `DiffUnsplittableError: single hunk in xx.py is 49706 chars, over the 49700 chars left for hunk content after the 51-char file header` |
| after | 2 chunks: **49,988** and **30,248** chars, both <= 50,000 |

Chunk 1 is `108` (outer chunk header) + `122` (`file part 1 of 2` marker) + `51` (file header) + `49,706` (the hunk) - the exact two-part rendering, zero wasted reserve. My byte counts differ slightly from your 49,983 / 15,258 because the fixture's path length and tiny-hunk bodies differ; the mechanism and the knife edge are identical - at this path/header, 49,706 is exactly the 1-digit content budget and 6 chars over the 1,000-hunk one, so the fixture sits on the boundary by construction rather than by luck.

A second, non-knife-edge case is covered too (`test_marker_width_comes_from_parts_not_hunks`): 100 hunks of 2,352 chars at a 5,000 cap. Sizing from the 3-digit hunk count costs 2 chars per part - just enough to stop two hunks sharing a part, so the old code emitted **100** chunks where **50** suffice. Same defect, visible as doubled review cost rather than a hard failure.

**What did not change:**

- The per-chunk budget. Not raised.
- The residual floor. A single hunk over the budget still raises `DiffUnsplittableError` and still fails closed. The check now runs inside the loop, and it is sound at every round, not just the first: the iterate is a lower bound on any self-consistent part count, so a hunk that overflows there overflows every valid rendering too. `test_single_oversized_hunk_among_many_still_raises` plants a 60,000-char hunk among 999 small ones so the fix cannot paper the floor over by shrinking the marker away.
- Multi-file packing (still guarded by the golden-value test), the file header repeated verbatim on every part, byte-exact reassembly, every hunk exactly once in order.

**Evidence:**

- 6 new tests, all in `TestPartMarkerWidthFromPartCount`. Every one mutation-checked - I reverted and watched each fail, then restored: 4 fail on mutation 1 (seed the fixed point from `len(hunks)` again, i.e. the exact defect), 1 on mutation 2 (delete the residual-hunk raise; it then surfaces as the defensive part-size error instead), 1 on mutation 3 (delete the non-convergence raise).
- Differential harness, pre-fix vs post-fix implementation on 4,000 randomized multi-file diffs across five caps: 3,733 splittable, **0** cases where the new code raises where the old succeeded, **0** where it produces more chunks, 89 where it produces fewer. The cap, byte-exact-reassembly, hunk-exactly-once and marker-consistency (`i` runs `1..n`, `n` equals the parts actually rendered) properties were re-asserted on both implementations on every trial - i.e. the round-trip property is checked after the change, not only before.
- Real diffs: 72,174 chars / 62 hunks and 232,032 chars / 151 hunks from this repo chunk byte-identically to the pre-fix code (5, 2 and 6 chunks at 20K/50K caps).
- `uv run pytest tests/ -q` -> 2,524 passed, 28 skipped (baseline on this branch was 2,518 + the 6 new). `uv run mypy kstrl/ --strict` and `uv run ruff check kstrl/ tests/` clean.

**Assumed, not verified:** that next-fit is optimal for *ordered contiguous* grouping (used to argue the fixed point is the minimum part count) is an analytic claim, not a measured one - though the differential's "never more chunks than the old code" result is consistent with it. No factory run was executed against this change; the evidence is unit-level, differential and property-based.

The [P2] prompt-body finding is deliberately untouched here: it triggers H3 (version bump + snapshot) and H2 (calibration re-run), so it is being handled as its own decision rather than folded into this commit.

0xfauzi:
Both findings addressed; H3 and H2 both discharged.

**[P3] marker sizing** - fixed in `6358b60` by packing to a fixed point (part count and marker width are mutually dependent, so they are solved together), with a written termination proof and a bounded loop that fails closed rather than spinning. I verified your knife edge independently: one 49,706-char hunk + 999 tiny ones now yields **2 chunks at 49,988 / 25,932**, both under cap, all 1,000 hunks present exactly once. (My first repro attempt raised - that was my error, I built a 69-char file header instead of your 51, so the hunk genuinely exceeded budget and the raise was correct.)

**[P2] prompt contract** - fixed in `9e57021`. You were right that the part marker contradicting the standing instruction is not a reliable contract. Both bodies now distinguish the two granularities and state what a part actually hides: other hunks **of the same file**. The security wording names the concrete failure - a taint introduced in one part and sanitized in another is invisible from either alone.

`REVIEWER_PROMPT_VERSION` and `SECURITY_PROMPT_VERSION` 1.1.1 -> 1.2.0, both snapshot tuples moved in the same commit. New `TestChunkHeadersMatchTheReviewContract` asserts the harness's emitted marker text appears in both prompts, so code and contract cannot drift apart silently again.

**H2 calibration** (3 runs/fixture, haiku, vs `baseline-20260720`):

| role | before | after |
|---|---|---|
| security | 1.00 | **1.00** |
| security_hard | 1.00 | **1.00** |
| architect | 1.00 | **1.00** |
| reviewer | 1.00 | **0.92** |

The reviewer figure is one fixture going 3/3 -> 2/3 (`dead_code`). Above its 0.65 floor and inside the comparator tolerance, so nothing flags it - but I edited `REVIEWER_PROMPT`, so I am not going to call it clean. At n=3 I cannot distinguish a small real regression from noise. Recorded, not dismissed; happy to re-run the reviewer fixtures if you want a firmer number.

---

**The calibration suite was already broken, independently of this PR.** The run initially failed `architect_allowed_paths` 1.00 -> 0.00. That is not a detection result:

```
component url-slugifier allowedPaths contains forbidden harness internal:
  scripts/kstrl/feature/url-slugifier/  matches  kstrl/
```

The fixture's forbidden-path check used a substring match. The kstrl rename (#122/#124/#172) changed the forbidden entry from `ralph_py/` to `kstrl/` - and `kstrl/` **is** a substring of `scripts/kstrl/feature/<id>/`, the one subtree `DECOMPOSE_PROMPT` explicitly instructs the architect to include. The `20260720` baseline recorded that exact output as gate-clean, so the architect never changed; the checker did. It has been rejecting correct output ever since the rename, unnoticed because calibration is opt-in and takes ~77 minutes.

`_is_within` replaces the substring test with a path-prefix one (7 cases pinned, including the near-miss `src/kstrl_helper.py`). Re-running that fixture against the fix: **PASSED**. The stored baseline is annotated in `_results/README.md` so its void `architect_allowed_paths` figure cannot mislead a future comparison.

Worth stating plainly: H2 calls calibration the truth signal for prompt changes, and one of its gates was permanently red on correct behaviour. A gate that always fails trains everyone to wave it through.

**Tested:** 2531 passed, 28 skipped. `mypy --strict` and `ruff` clean.

PR #184, opened 2026-07-29
fix: make the cost ceiling's coverage honest and visible

0xfauzi:
## The defect, as measured

`[factory] max_cost_usd` (PR #180) is a dollar-denominated run ceiling. It counts only calls whose adapter reported a cost. On a paid factory run that meant it bounded **one role**.

Per-phase `component_usage` events from that run:

| phase | calls | tokens | cost | cost_calls | token_calls |
|---|---|---|---|---|---|
| engineer | 5 | 8,036,800 | $9.9929 | 5 | 5 |
| review | 2 | 78,157 | $0.0000 | 0 | 2 |
| engineer | 1 | 7,939,537 | $7.3448 | 1 | 1 |
| review | 3 | 115,476 | $0.0000 | 0 | 3 |
| engineer | 1 | 7,404,185 | $7.4238 | 1 | 1 |
| engineer | 1 | 2,947,879 | $3.9930 | 1 | 1 |

The run's **total cost equalled the engineer total exactly**: 193,633 reviewer tokens over 5 calls contributed **$0**. The operator set `--max-cost-usd 25.0` believing it bounded the run.

**Tokens are counted across roles** (26,522,034 run vs 26,328,401 engineer). Only the dollar figure under-counts, because the cross-family reviewer (codex) reports a token total and no cost. Verified in code before designing: `ComponentPipeline._record_usage` merges *every* phase into `run_usage`, and `cost_budget_exceeded` reads `run_usage.cost_usd` — the meter is wired correctly. This is an adapter capability gap.

### Why no surface caught it

Reproduced against the pre-fix code with the numbers above:

- Nothing was **breached** at the point the reviews were logged.
- Nothing was **unenforceable**: `cost_budget_unenforceable()` returned `None`, correctly — the engineer reports cost on every call.
- The usage rollup's lower-bound footer keys off `unreported_calls`, which was **0**, because every call reported *something*. The footer stayed silent on exactly the run it exists for.

The machinery modelled "the ceiling works" and "the ceiling is dead". A ceiling that counts some roles and not others is neither.

## The fix

Adds the missing middle term — **PARTIAL coverage** — and reports it. Extends the existing `cost_calls` / `token_calls` / `costless_calls` axis rather than adding a parallel notion.

- `RoleCoverage` / `CeilingCoverage` / `usage_coverage()` in `kstrl/agents/base.py`: folds the meter by role for either axis. `CEILING_AXES` states the ceiling→axis mapping once, so no surface re-derives "`max_cost_usd` means the cost axis" locally.
- `ComponentPipeline.ceiling_coverage()` / `coverage_notes()`, next to the existing `*_budget_exceeded` / `*_budget_unenforceable` predicates.
- **Up front**, at the plan stage: `Cost ceiling` / `Token ceiling` preflight lines stating *what each ceiling counts*.
- **At the first evidence**: a one-shot-per-ceiling `budget_coverage` event plus a `BUDGET COVERAGE:` warning, emitted from `_record_usage` the first time a configured ceiling stops covering every metered call — typically right after the first cross-family review, minutes into a run.
- **At the halt**: the coverage clause is appended to the component error, and `coverage` rides on the `budget_exceeded` event and the `budget_overrun` inbox item.
- **At the summary**: per-axis `note:` lines under the usage rollup, naming the uncovered roles. `-` in the cost column now means "no call in this row reported a cost" (keyed on `cost_calls`) rather than "the total was 0".

Sample output for the measured run:

```
note: cost coverage is PARTIAL: 8 of 13 metered call(s) reported a cost, so
max_cost_usd bounds only those; uncovered: review (5 of 5 call(s), 193,633
token(s) unpriced). No price is inferred for uncovered calls, so the cost
total is a lower bound
```

### On the "up front" question

The preflight states the ceiling's **semantics**, not a coverage prediction. At the plan stage zero calls have been made, so any per-role claim would have to come from a hard-coded adapter capability table — which this repo does not have, and which would go stale the day codex starts reporting cost. The measured per-role figure follows as soon as evidence exists, which is still early enough to act on.

## What I deliberately did NOT do

- **No invented prices.** The uncovered magnitude is reported in **tokens**. There is no token→dollar table anywhere in this change and the notes assert no `$` figure for uncovered calls. Repo policy: never guess numerical values. A fabricated cost in an audit trail is worse than a missing one.
- **No token attribution for partially covered roles.** `uncovered_tokens` sums only roles where *no* call reported a cost. Where a role reported cost on some calls, the aggregate does not say which tokens were unpriced, so no figure is claimed. Pinned by a test.
- **No change to ceiling semantics.** `max_cost_usd` still halts on reported dollars only.
- **No prompt edits** (H3/H2 untouched).

## Recommendation on the out-of-scope policy question

*Should `max_cost_usd` halt when coverage is partial?* **No — not as a default.** Recommendation: leave the ceiling as-is and add an opt-in strict knob later if the user wants one.

Reasoning, and the failure modes I weighed:

1. **It would fail closed on healthy runs.** Cross-family review is the *recommended* configuration (R7.1 deliberately picks the opposite family from the engineer). Partial cost coverage is therefore the *normal* state for a well-configured run, not an anomaly. A halt-on-partial default would break the recommended setup on its first review call.
2. **The existing halt rule is "the ceiling can never fire".** That is a provable statement about the future. "Partial coverage" is not — the covered subset still grows monotonically toward the ceiling, so the ceiling *does* fire, just later than an operator who assumed full coverage expects. Halting on it would conflate "cannot bound" with "bounds less than you thought".
3. **Second-order effect:** an operator who hits a partial-coverage halt has exactly two escapes — raise the ceiling or disable it. Both are worse than the status quo, and the second one is the outcome the ceiling exists to prevent.
4. **What the operator actually lacked was information**, which is what this PR supplies. If, with the disclosure in place, the user still wants enforcement, the natural shape is `[factory] cost_coverage = "report" | "require_full"` (default `report`), reusing `CeilingCoverage.partial` — the predicate already exists.

## Verification

| | before | after |
|---|---|---|
| `uv run pytest tests/ -q` | 2507 passed, 28 skipped | **2539 passed, 28 skipped** (+32) |

**Reproduction.** The defect was first reproduced with a standalone probe replaying the six `component_usage` rows above through the unmodified `_format_usage_rollup` and `ComponentPipeline`: run cost == engineer cost, `unreported_calls == 0` (footer silent), `cost_budget_unenforceable() is None`, `unenforceable_ceilings() == []`. Those facts are now pinned by `TestMeasuredCoverageGap`, so the premise cannot rot.

**Mutation check — every new behavior was reverted and watched to fail** (selector-scoped runs over `tests/test_usage_meter.py` + `tests/test_events.py`):

| reverted behavior | result |
|---|---|
| rollup per-axis footer | 2 failed, 4 passed |
| halt-message coverage clause | 1 failed |
| `coverage` on the `budget_exceeded` event | 1 failed |
| `_announce_coverage_gaps()` call | 2 failed, 3 passed |
| preflight ceiling lines | 2 failed, 1 passed |
| `-` keyed on `cost_calls` | 1 failed |
| silent-only token attribution | 1 failed |
| `budget_coverage` progress-log mirror | 3 failed |
| `coverage` in `ProgressLog.budget_exceeded` | 3 failed |
| coverage list on the halt | 3 failed, 2 passed |
| `note()` full-coverage suppression | 4 failed |

The first pass found a real gap: the fully-covered filter in `fail_for_budget` was unobserved *and* made "no gap" indistinguishable from a legacy payload. Fixed (coverage is now recorded for every configured named ceiling) and re-mutated.

**Both-sinks / legacy decode** (the #181 regression class): `events.jsonl` and `progress.jsonl` payloads are asserted byte-identical for `budget_coverage` and for a partially-covered `budget_exceeded`; the pre-existing generic guard (`test_every_event_field_survives_the_progress_log`) covers `coverage`, and a new twin covers every `BudgetCoverage` field. Legacy payloads lacking either field decode to empty defaults, asserted.

**Verified vs assumed.** Verified: the numbers above, that the meter is correctly wired, that the old footer was silent, all three commands, and every mutation result. Assumed (stated, not measured): that codex will continue to report no cost — which is precisely why nothing here hard-codes an adapter capability table.

**Not touched:** `kstrl/git.py` and the `DiffUnsplittableError` handler (concurrent work).


kstrl/pipeline.py
```diff
@@ -383,6 +396,42 @@ def _record_usage(
         self.bus.emit(ev.ComponentUsage(
             component=comp_id, phase=phase, **totals.to_dict(),
         ))
+        self._announce_coverage_gaps()
```
0xfauzi, line 399: [P1] Preserve partial coverage in the Textual dashboard

This emits the new warning/event, but the primary live dashboard still presents the affected totals as exact. `reducer.apply()` adds `ComponentUsage.cost_usd` and `unreported_calls` but drops `cost_calls`/`token_calls`; it also returns early for this run-scoped `BudgetCoverage` event. `render_cost_meter()` therefore continues to key its `+ lower bound` marker only on `unreported_calls`. A focused reproduction with `calls=2`, `known_calls=2`, `token_calls=2`, `cost_calls=1`, `$5` reported against a `$10` cap, followed by `BudgetCoverage`, renders `1.0k tok · $5.00 · 50% of cost cap` with no lower-bound marker or persistent warning. That is the original misleading state for the measured cross-family run, and `ActivityFeed.humanize()` also drops both `Log` warnings and `BudgetCoverage`. Suggested fix: fold per-axis call coverage into `RunState` (and the v1 upconversion path), handle `BudgetCoverage` as run-scoped state/feed data, and render token/cost lower-bound indicators independently in the cost meter.

kstrl/factory.py
```diff
@@ -1878,7 +1886,11 @@ def _row(label: str, phase: str, totals: UsageTotals) -> str:
             tokens_in = f"{totals.input_tokens:,}"
             tokens_out = f"{totals.output_tokens:,}"
             tokens_total = f"{totals.total_tokens:,}"
-            cost = f"{totals.cost_usd:.4f}" if totals.cost_usd > 0 else "-"
+            # "-" means "no call in this row reported a cost", never
+            # "it was free". Keyed on cost_calls rather than on the
+            # total so a genuinely reported $0.0000 is not rendered as
+            # silence - the same distinction the cost ceiling makes.
+            cost = f"{totals.cost_usd:.4f}" if totals.cost_calls > 0 else "-"
```
0xfauzi: [P2] Apply the same coverage distinction to the token columns

The cost cell now correctly distinguishes a reported zero from no report, but the three token cells are still gated by `known_calls`. A cost-only invocation has `known_calls=1` and `token_calls=0`, so the rollup prints `0  0  0` even though no token figure was reported; the new footer then says token coverage is EMPTY and the total is a lower bound. I reproduced this with one `UsageRecord(cost_usd=1.0)`: both the component and TOTAL rows showed numeric zero tokens. Suggested fix: render the token cells only when `totals.token_calls > 0`, otherwise use `-`, and add the cost-only-row counterpart to the new reported-zero-cost test.

kstrl/pipeline.py
```diff
@@ -996,6 +1104,26 @@ def fail_for_budget(
...
+        #
+        # EVERY configured named ceiling is recorded, including ones that
+        # covered every call. An empty ``coverage`` would otherwise mean
+        # two different things - "no gap" and "written before this
+        # landed" - and a reader could not tell verified from unknown
+        # (the distinction E9 added ``infrastructure_error`` for). The
+        # PROSE stays quiet on full coverage, because ``note()`` returns
+        # "" there: the operator-facing sentence is unchanged wherever
+        # there is nothing to disclose.
+        coverage = [
```
0xfauzi, line 1133: [P2] Record coverage for every configured ceiling as documented

This list iterates `ceilings`, which is the causal halt identity, not all configured ceilings. With both caps enabled, a token breach makes `budget_halt_identity()` return only `("max_total_tokens",)`, so a simultaneously partial `max_cost_usd` is omitted from the halt event and inbox evidence. I reproduced a two-call run with both caps configured: token coverage was recorded, while the cost coverage note was `PARTIAL` but absent from `coverage`. This contradicts the new docs and the comment below that promise every configured named ceiling, and makes a missing per-ceiling entry ambiguous. Suggested fix: keep `ceilings` for what caused the halt, but build `coverage` by iterating all enabled keys in `CEILING_AXES` (letting `ceiling_coverage()` filter disabled ceilings), with a dual-cap regression test.

0xfauzi (commented):
Reviewed the complete change at `4b1f7ce533048856164057cff3924041c8c25469` and found three actionable gaps in the new coverage reporting. GitHub CI is green. Local verification also passed: `uv run pytest tests/ -q` (2539 passed, 28 skipped), `uv run mypy kstrl/ --strict`, and `uv run ruff check kstrl/ tests/`. The inline comments include focused reproductions and suggested fixes.

0xfauzi:
All three findings confirmed and fixed in 4214b89. Each was reproduced with a failing test using your exact shapes before any code changed, and every new test was mutation-checked (revert the fix, watch it fail, restore).

---

## Finding 1 [P1] - the dashboard rendered partial coverage as exact

**Reproduced.** Your shape (`calls=2, known_calls=2, token_calls=2, cost_calls=1`, $5 against a $10 cap, followed by `BudgetCoverage`) rendered:

```
1.0k tok · $5.00 · 1% of token cap · 50% of cost cap
```

No marker, no warning. All four defects were real:

| defect | fix |
|---|---|
| `reducer.apply()` dropped `cost_calls`/`token_calls` | folded into `RunState` **and** `ComponentState`, with `tokens_are_lower_bound` / `cost_is_lower_bound` derived per axis |
| early return for run-scoped `BudgetCoverage` | handled **above** the `if not event.component: return` guard, into `RunState.coverage_gaps` (frozen `CoverageGap`, keyed by axis) |
| one marker keyed on `unreported_calls` | token and cost markers are now **independent**, on the figure and on the cap percentage |
| `humanize()` dropped `BudgetCoverage` | narrated with the uncovered roles and the unpriced TOKEN count |

Now renders (asserted on `render_cost_meter(...).plain`, per your note that state alone was not the defect):

```
1.0k tok · $5.00+ · 1% of token cap · 50%+ of cost cap  + lower bound (cost)
```

The token axis stays unmarked, since you are right that the two differ - a test pins `1.0k+` and `%+ of token cap` as absent in exactly that case. The percentage carries the marker too: `50% of cost cap` is what an operator reads as headroom, and it overstates it by the amount nobody measured.

The marker does **not** require the `BudgetCoverage` event (the usage stream alone carries the fact); the event corroborates and supplies the roles.

**One subtlety the fix had to handle.** Keying naively on `usage_calls > token_calls` would mark every PRE-R8 run dir on disk, where the axis counters do not exist and decode to 0 - claiming a gap that was never measured, which is the same class of false statement this PR exists to remove. `known_calls > 0` with both axis counters at 0 is impossible in a post-R8 payload (a known call reported tokens or cost), so it identifies a legacy one; those calls are counted as coverage-**unknown** and excluded from both sides of the arithmetic. Two tests pin both directions. The fake-run fixture now emits the axis counters, since it means "a modern run", not "an old log".

**Beyond the four bullets** (same defect class, same shared state, so leaving them would have re-created the finding one screen over): the dashboard's component table, `home_data.tokens_lower_bound`, and the `ks status` run/component usage lines - the last of which now also prints the gap's own sentence. Say the word if you would rather have those split out.

Raw `Log` warnings stay out of the feed deliberately - factory warnings are schemaless prose, and admitting them would put every homogeneity/config warning in the feed. The typed event carries the same fact, and the module docstring now says so.

**No price is invented on any new surface.** The uncovered magnitude is reported in tokens on the meter legend, the feed line and `ks status`. A test asserts the meter line contains exactly one `$` (the reported total); a mutation that adds `~$0.02 unpriced` to the legend fails it.

## Finding 2 [P2] - token columns gated on `known_calls`

**Reproduced** with a single `UsageRecord(cost_usd=1.0)`: `known_calls=1, token_calls=0`, and both the component and TOTAL rows printed

```
comp-a   engineer   1   0   0   0   1.0000   0
```

while the footer under them said token coverage was EMPTY and the total was a lower bound.

Each cell is now gated by its own axis (`token_calls > 0` for the three token cells, `cost_calls > 0` for cost); the `known_calls` branch is gone, since it is exactly the union of the two. `-` means "no call in this row reported this figure", never "it was zero" - so the counterpart you asked for is there: a call that genuinely reported **0 tokens** still renders `0`, not `-`, and a mutation gating on `total_tokens > 0` instead fails that test.

## Finding 3 [P2] - coverage recorded only for the causal ceiling

**Reproduced** with a two-call dual-cap run (both caps configured, token cap breached, cost axis PARTIAL): `coverage` held `{max_total_tokens}` only.

`coverage` is now built by iterating `CEILING_AXES`; `ceilings` is untouched and still names only what fired (`["max_total_tokens"]`, `condition: "breached"`). `ceiling_coverage()` returns `None` for a disabled ceiling, so an absent entry keeps exactly one meaning - a test with only the token cap configured pins that the cost entry stays absent.

The halt event and the inbox evidence now both carry `max_total_tokens` (`uncovered_calls: 0`) and `max_cost_usd` (`uncovered_roles: ["review"]`) - verified from `events.jsonl` and from `Inbox.items()`, plus the existing both-sinks-agree assertion.

---

## Verification

- **Suite**: 2539 -> 2561 passed, 28 skipped. `mypy kstrl/ --strict` and `ruff check kstrl/ tests/` clean.
- **New tests**: 22. `TestPerAxisCoverageFold` (5), `TestCostMeterPerAxisLowerBound` (7), `TestBudgetCoverageLine` (3), `TestEveryConfiguredCeilingRecordsItsCoverage` (4), two rollup cases, one `ks status` case.
- **Mutation checks**: 15, each watched failing and restored. Every new test is killed by at least one, including the guardrails - `M4` (count all calls as unknown) kills the legacy-suppression test, `M5` (add a dollar estimate to the legend) kills the no-price test, `M6` (gate the rollup on `total_tokens`) kills the reported-zero-tokens test, `M7` (drop `coverage` from `V1CompatSink`) kills the sink-parity test, `M8` (report a disabled ceiling) kills the absent-entry test.
- **Dual-sink parity**: no event payload changed, so nothing new to mirror. A test now also pins that `budget_coverage` reaches the same reducer state through the v1 `progress.jsonl` path, and `M7` proves the existing parity assertion bites.
- **Snapshot**: `test_overview_snapshot` updated - the only deltas are `5% of token cap` -> `5%+ of token cap` and `+ lower bound` -> `+ lower bound (tokens, cost)`.

**Assumed, not verified**: that no external consumer parses the meter legend or the rollup's `-` cells positionally. `budget_coverage` in the checkpoint screen's spend line (`kstrl/tui/screens/checkpoint.py:75`) is still keyed on `unreported_calls` - it reads a `UsageTotals` rather than `RunState`, so it is a different data path and I left it alone rather than widen the change further. Happy to take it in a follow-up if you want it consistent.

Not merged.

PR #185, opened 2026-07-29
R8.6 PR 1/4: maildir work-queue substrate and `ks queue`

0xfauzi:
First of four slices for [#153](https://github.com/0xfauzi/kstrl/issues/153) (R8.6 continuous intake). This is the **substrate only** - it adds a place for work to wait and the verbs to manage it. **Nothing drains the queue yet.**

Slicing agreed before writing code:

| PR | Contents | Status |
|---|---|---|
| **1** | **queue substrate + `ks queue` CLI** | **this PR** |
| 2 | `ks serve`, lease reaper, retry classifier, `daily_budget_usd`, caffeinate | next |
| 3 | GitHub Issues adapter, processed-ids ledger, `ks queue sync` | |
| 4 | launchd plist + docs | |

## What landed

`kstrl/workqueue.py` (~600 lines, no new dependency - queue libraries were evaluated and rejected in the R8.6 verdict):

- maildir layout `.kstrl/queue/{queued,leased,running,done,failed,poison}/`
- item = **directory** holding a spec file + `meta.json`, transitioned with one `os.replace`
- flock mutex, pid/ttl lease fields, append-only `journal.jsonl`, `pause.json` marker
- `QueueConfig` with the house `from_env()` + `load(root_dir)` pair, `[queue]` section

`ks queue add|ls|show|retry|rm|pause|resume`.

## The two invariants worth reviewing closely

Both are money-safety properties rather than style, because an unattended `ks serve` will drive this substrate in PR 2. Both are mutation-checked.

**1. The attempt is charged before the rename into `running/`.** Every transition writes `meta.json` first and renames second - the rename is the commit point. So a crash in that window over-counts an attempt (the item gets one fewer retry, which is safe) instead of under-counting one (a retry nobody recorded, which is an unbounded loop). At a measured \$1.70-2.60 per first attempt and \$3.99-7.42 per retry, an uncounted attempt is not a bookkeeping slip.

**2. The item directory is authoritative; the sidecar's `state` field is a mirror.** Readers derive state from the parent directory name and overwrite the mirror, so the same crash window cannot leave an item whose location and metadata disagree. The mirror self-heals on the next write.

An item is a directory rather than two sibling files specifically so one rename moves the spec and its sidecar together - two files would need two renames with an interruptible gap between them.

## Fail-closed choices

Corrupt or unreadable state resolves toward the *gated* value in every direction:

- an unreadable `merge_disposition` decodes to `stop_at_pr`, never `auto_merge` - a corrupt field must not be able to **grant** a permission
- an unreadable pause marker reads as **PAUSED** - never resume unattended spend on the strength of a file you cannot parse
- a missing or unparseable lease expiry counts as **expired** - a lease nobody can read must not wedge the queue forever
- `ks queue retry` refuses an item that has spent its attempts unless `--reset-attempts` is passed, and the automatic paths can never reset the counter: a retry policy that can zero its own bound is not a bound

The spec is **copied** into the item at enqueue, so editing or deleting the original afterwards cannot change what eventually runs.

## Tested vs assumed (H4)

**Tested.** 103 new tests (73 substrate, 30 CLI). Every invariant test was mutation-checked: the invariant was broken, the test was **watched failing**, then restored. 14/14 mutations caught -

attempt-charge ordering; in-memory rollback on an uncommitted move; `merge_disposition` fallback; pause fail-closed; lease-expiry default; attempt reset on requeue; directory-over-sidecar authority; removal of a running item; ambiguous-prefix guard; transition legality table; poison reason requirement; journal-failure isolation; staging-dir cleanup; pause admission gate.

One of these caught a defect in my own first draft: the initial crash-window test patched `os.replace` globally, which also broke `_atomic_write`'s own commit and so proved nothing about the window between the two renames. Fixed to fail only the directory rename.

**Assumed, not tested.** POSIX-only behavior of `flock` and directory `os.replace` (consistent with the existing A4 worktree lock and run-level factory lock; the codebase is POSIX-first). No live factory run was performed - this PR executes nothing.

## Verification

```
uv run pytest tests/ -v     2585 -> 2688 passed, 28 skipped   (baseline measured on 7e3269f before starting)
uv run mypy kstrl/ --strict Success: no issues found in 109 source files
uv run ruff check kstrl/ tests/ scripts/   All checks passed!
```

No prompt bodies changed, so H2/H3 do not apply. `scripts/gen_docs.py` gained the `[queue]` section and README was regenerated by the generator, not by hand. `KSTRL_QUEUE_` was added to the conftest env-scrub prefixes so ambient env cannot alter the `from_env`/`load` tests.

## Notes for review

- Module is `workqueue.py`, not `queue.py`: `kstrl/agents/proc.py` imports the stdlib `queue`. Absolute imports make `kstrl/queue.py` technically safe, but it is a trap not worth leaving.
- The queue mutex (`queue.lock`) is deliberately separate from `.kstrl/factory.lock` and short-lived, so `ks queue ls` stays responsive while a run is in flight. PR 2's serve singleton will be a third, long-held lock.
- `meta.json` carries `target_repo` from day one. It is unused here - it exists so that moving to a global `~/.kstrl/queue` (roadmap open question 3) is a config change rather than a migration of on-disk items.
- Per H1 I have not run `/code-review` on this. You are the gating reviewer.

### DIFF
--- a/kstrl/workqueue.py
+++ b/kstrl/workqueue.py
@@ -1,0 +1,1073 @@
+"""R8.6 continuous intake: the local work queue substrate.
+
+Intake before this module is one-shot - a human fires ``ks factory
+--spec`` and the run ends. The queue is what lets work arrive without a
+human firing each run, so it is the survival capability R8.6 exists to
+add, not plumbing around one.
+
+**Why a directory queue and not a library.** Queue libraries were
+evaluated and rejected in the R8.6 verdict (litequeue is near-dormant and
+conflicts with our Python floor; persist-queue buys nothing at a
+concurrency of one; huey inverts control - it would own the process the
+factory needs to own). A maildir-style directory tree gives durability,
+crash recovery, and human inspectability with no dependency and no
+schema migration story.
+
+**Layout.** Each state is a directory under ``.kstrl/queue/``; each item
+is a DIRECTORY holding its spec file plus ``meta.json``::
+
+    .kstrl/queue/
+      queued/<item_id>/{<spec>, meta.json}
+      leased/   claimed by a worker, nothing spent yet
+      running/  executing; money is being spent
+      done/     finished green
+      failed/   finished red and eligible to retry
+      poison/   finished red and NOT eligible to retry
+      journal.jsonl
+      pause.json    (absent = running)
+      queue.lock    (short-lived per-transition mutex)
+
+The item is a directory rather than two sibling files so that one
+``os.replace`` moves the spec and its sidecar together. Two sibling files
+would need two renames and could be interrupted between them, leaving an
+item whose spec and metadata disagree about what state it is in.
+
+**The directory is the source of truth.** ``meta.json`` also carries a
+``state`` field, but it is a convenience mirror: readers derive state
+from the parent directory name and overwrite the mirror. That makes the
+crash window between "write meta" and "rename dir" harmless in both
+directions - whichever step got interrupted, the directory still says
+what is true and the mirror self-heals on the next write.
+
+**Ordering is a money-safety property, not a style choice.** Every
+transition writes ``meta.json`` FIRST and renames SECOND, because the
+rename is the commit point. ``attempts`` is therefore incremented before
+the rename that starts execution: a crash in the window over-counts an
+attempt (the item gets one fewer retry - safe) instead of under-counting
+one (the item retries without the attempt being recorded - an unbounded
+retry loop). A measured factory engineer iteration costs ~$1.70-2.60 on
+a first attempt and $3.99-7.42 on a retry, so an uncounted attempt is
+not a bookkeeping slip.
+
+The retry POLICY - which failures may be retried at all - deliberately
+does not live here. This module records attempts and moves items; R8.6
+PR 2 (``ks serve``) decides, on positive evidence only, whether a
+failure was an infrastructure error. See ``docs/dark-factory-roadmap.md``
+R8.6.
+"""
+
+from __future__ import annotations
+
+import json
+import os
+import secrets
+import shutil
+import tempfile
+import warnings
+from collections.abc import Iterator
+from contextlib import contextmanager
+from dataclasses import dataclass, field
+from datetime import UTC, datetime, timedelta
+from enum import StrEnum
+from pathlib import Path
+from typing import IO, Any
+
+from kstrl.statedir import state_dir
+
+QUEUE_DIR_NAME = "queue"
+META_FILENAME = "meta.json"
+JOURNAL_FILENAME = "journal.jsonl"
+PAUSE_FILENAME = "pause.json"
+LOCK_FILENAME = "queue.lock"
+QUEUE_SCHEMA_VERSION = 1
+
+
+class QueueError(RuntimeError):
+    """A queue operation could not be completed."""
+
+
+class QueueLockedError(QueueError):
+    """Another process holds the queue mutex."""
+
+
+class ItemState(StrEnum):
+    """Where an item is. Also the on-disk directory name.
+
+    ``LEASED`` and ``RUNNING`` are deliberately distinct even though a
+    worker passes through both in quick succession: the reaper (PR 2)
+    treats them differently. A leased item whose owner died spent
+    nothing, so it returns to ``QUEUED`` for free. A running item whose
+    owner died spent real money and needs its failure classified before
+    anything decides to spend more.
+    """
+
+    QUEUED = "queued"
+    LEASED = "leased"
+    RUNNING = "running"
+    DONE = "done"
+    FAILED = "failed"
+    POISON = "poison"
+
+
+#: Every state directory, created eagerly so a scan never has to
+#: distinguish "no such directory" from "no items in it".
+ALL_STATES: tuple[ItemState, ...] = tuple(ItemState)
+
+#: Legal moves. Anything absent raises rather than being silently
+#: allowed: an illegal transition means a caller's state machine is
+#: wrong, and in this module a wrong state machine spends money.
+_LEGAL_TRANSITIONS: dict[ItemState, frozenset[ItemState]] = {
+    ItemState.QUEUED: frozenset({ItemState.LEASED, ItemState.POISON}),
+    # A lease can be dropped back to queued by the reaper (owner died
+    # before spending) or fail outright if the worker could not start.
+    ItemState.LEASED: frozenset({
+        ItemState.QUEUED, ItemState.RUNNING, ItemState.FAILED,
+        ItemState.POISON,
+    }),
+    ItemState.RUNNING: frozenset({
+        ItemState.DONE, ItemState.FAILED, ItemState.POISON,
+    }),
+    # Terminal-ish: a human (or the retry policy) can requeue, and a
+    # failed item that exhausts its attempts is poisoned.
+    ItemState.FAILED: frozenset({ItemState.QUEUED, ItemState.POISON}),
+    ItemState.POISON: frozenset({ItemState.QUEUED}),
+    ItemState.DONE: frozenset(),
+}
+
+
+class MergeDisposition(StrEnum):
+    """Whether a finished item's PR merges without a human.
+
+    ``STOP_AT_PR`` is the default for everything, and R8.6 requires it be
+    the default for remote-sourced items specifically: continuous intake
+    must not silently delete the human merge gate. ``AUTO_MERGE`` is
+    per-item opt-in AND ladder-gated - the R8.2 flag bundle
+    (``auto_merge_when_green``) can withhold it, and this field can only
+    ever request it, never grant it.
+    """
+
+    STOP_AT_PR = "stop_at_pr"
+    AUTO_MERGE = "auto_merge"
+
+
+class ItemSource(StrEnum):
+    """Where the item came from. ``source_ref`` names the specific one."""
+
+    LOCAL = "local"
+    GITHUB = "github"
+    LINEAR = "linear"
+
+
+def _utc_now() -> datetime:
+    return datetime.now(UTC)
+
+
+def _iso(moment: datetime) -> str:
+    return moment.isoformat()
+
+
+def _parse_iso(value: str) -> datetime | None:
+    if not value:
+        return None
+    try:
+        parsed = datetime.fromisoformat(value)
+    except ValueError:
+        return None
+    return parsed if parsed.tzinfo else parsed.replace(tzinfo=UTC)
+
+
+def mint_item_id() -> str:
+    """``q-YYYYMMDD-HHMMSS.ffffff-<nonce>``.
+
+    Shaped like :func:`kstrl.runid.mint_run_id` and for the same reason:
+    the microsecond stamp makes same-second ids order deterministically
+    by creation time, which is what gives the queue FIFO ordering within
+    a priority band, and the nonce guards same-microsecond collisions.
+    """
+    stamp = _utc_now().strftime("%Y%m%d-%H%M%S.%f")
+    return f"q-{stamp}-{secrets.token_hex(3)}"
+
+
+def _warn_rejected(path: Path, reason: str) -> None:
+    """Announce a skipped item rather than swallowing it.
+
+    A silently dropped queue item is work that vanished. Mirrors
+    ``autonomy._warn_rejected_state``.
+    """
+    warnings.warn(
+        f"queue: rejected item {path} ({reason}); skipping",
+        RuntimeWarning,
+        stacklevel=3,
+    )
+
+
+def _as_int(data: dict[str, Any], key: str, default: int) -> int:
+    value = data.get(key, default)
+    if isinstance(value, bool) or not isinstance(value, int):
+        return default
+    return value
+
+
+def _as_str(data: dict[str, Any], key: str, default: str = "") -> str:
+    value = data.get(key, default)
+    return value if isinstance(value, str) else default
+
+
+@dataclass
+class QueueItem:
+    """One unit of intake: a spec plus everything decided about it.
+
+    Not frozen: an item's ``state``, ``attempts``, and lease fields are
+    exactly what a transition mutates. The frozen container in this
+    module is :class:`QueueConfig`, which genuinely never changes after
+    load.
+    """
+
+    item_id: str
+    title: str
+    spec_filename: str
+    state: ItemState = ItemState.QUEUED
+    #: Higher runs first; ties break by ``item_id`` (creation order).
+    priority: int = 0
+    created_at: str = ""
+    updated_at: str = ""
+    #: Execution attempts CHARGED so far - incremented before the rename
+    #: into ``running/``, never after. See the module docstring.
+    attempts: int = 0
+    max_attempts: int = 3
+    merge_disposition: MergeDisposition = MergeDisposition.STOP_AT_PR
+    source: ItemSource = ItemSource.LOCAL
+    #: Identifies the specific origin, e.g. ``0xfauzi/kstrl#153``. The
+    #: processed-ids ledger (PR 3) dedupes on this.
+    source_ref: str = ""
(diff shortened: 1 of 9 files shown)


kstrl/workqueue.py
```diff
@@ -0,0 +1,1073 @@
...
+        if not isinstance(data, dict):
+            _warn_rejected(item_path, f"{META_FILENAME} is not an object")
+            return None
+        item = QueueItem.from_dict(data)
+        if item is None:
+            _warn_rejected(item_path, "missing item_id or spec_filename")
+            return None
+        # The DIRECTORY is authoritative; the sidecar's copy is a mirror
+        # that may be one crash behind. See the module docstring.
+        item.state = state
```
0xfauzi, line 737: [P1] Keep the directory name authoritative for item identity. `_load_item_dir()` trusts `meta.json`'s `item_id`, and later `item_dir()` and `remove()` use that value as a path component. I reproduced this by changing the sidecar inside `queued/q-safe` to `"item_id": "../../../outside-dir"`: `items()` loaded it, then `remove()` deleted `<root>/outside-dir` while leaving `queued/q-safe` intact. A corrupt or edited sidecar can therefore redirect destructive operations anywhere reachable via `..`. Validate the directory entry itself, reject symlinks, and either require the mirrored ID to equal `item_path.name` or overwrite it from that authoritative basename before returning the item.

kstrl/workqueue.py
```diff
@@ -0,0 +1,1073 @@
...
+        self.ensure_dirs()
+        directory = self.state_path(ItemState.QUEUED) / item.item_id
+        if directory.exists():
+            raise QueueError(f"queue item {item.item_id} already exists")
+        # Build in a hidden staging directory and rename into place, so a
+        # scan never sees an item whose spec has not landed yet.
+        staging = self.state_path(ItemState.QUEUED) / f".staging-{item.item_id}"
+        staging.mkdir(parents=True, exist_ok=False)
+        try:
+            _atomic_write(staging / spec_filename, content)
```
0xfauzi, line 950: [P1] Validate `spec_filename` as a single safe basename before joining it to a queue path. The string form is the API intended for later remote adapters, and `Queue.add("remote body", spec_filename="../../../../escaped.md")` currently writes `escaped.md` at the project root while still publishing a queue item. A similarly edited sidecar makes `read_spec()` escape the item directory. Reject absolute paths, separators, `.`/`..`, and empty names both on enqueue and decode, then ensure the resolved spec path remains inside the item directory.

kstrl/workqueue.py
```diff
@@ -0,0 +1,1073 @@
...
+
+    def items(self, states: tuple[ItemState, ...] | None = None) -> list[QueueItem]:
+        """Every item in the given states, in run order."""
+        found: list[QueueItem] = []
+        for state in states or ALL_STATES:
+            directory = self.state_path(state)
+            if not directory.is_dir():
+                continue
+            for entry in sorted(directory.iterdir()):
+                if not entry.is_dir():
```
0xfauzi, line 748: [P1] Exclude staging directories from state scans. `add()` creates `.staging-<id>` under `queued/`, but `items()` accepts every directory. Simulating a process death after `meta.json` is written but before the final rename makes `items()` return that staging entry as a real queued item, even though `item_dir()` for the returned ID does not exist; a death earlier produces a permanent rejected-item warning on every scan. This defeats the stated invariant that scans never see partially published work and will wedge the future claim path. Skip `.staging-*` entries unconditionally and add an explicit stale-staging recovery/cleanup policy.

kstrl/workqueue.py
```diff
@@ -0,0 +1,1073 @@
...
+        item = QueueItem(
+            item_id=mint_item_id(),
+            title=resolved_title,
+            spec_filename=spec_filename,
+            state=ItemState.QUEUED,
+            priority=priority,
+            created_at=now,
+            updated_at=now,
+            max_attempts=(
+                self.config.max_attempts if max_attempts is None else max_attempts
```
0xfauzi: [P2] Validate the effective per-item attempt budget here. `ks queue add spec.md --max-attempts 0` currently exits successfully and persists an item with `max_attempts == 0`, even though `QueueConfig` rejects the same value. This leaves a newly accepted item with no valid execution budget. Reject values below 1 in `Queue.add()` and use a positive `click.IntRange` for the CLI option so every construction path preserves the invariant.

kstrl/workqueue.py
```diff
@@ -0,0 +1,1073 @@
...
+        )
+
+    def start(self, item: QueueItem, *, run_id: str = "", actor: str = "") -> QueueItem:
+        """Begin execution. THIS is where the attempt is charged.
+
+        Charged here rather than on completion because completion is the
+        step a crash, a sleep, or a killed process can skip. An attempt
+        that ran but was never counted is an unbounded retry loop.
+        """
+        return self.transition(
```
0xfauzi, line 1067: [P1] Enforce `max_attempts` at the spending boundary, not only in the CLI retry policy. With `max_attempts=1`, the sequence start -> fail -> requeue -> lease -> start succeeds and produces a running item with `attempts == 2`. The existing bounded-retry test avoids the second start conditionally, so it does not prove that the substrate supplies the advertised bound. Any bug in the future reaper/classifier can therefore spend indefinitely. Under the queue lock, `start()` should refuse or poison an item whose `attempts_remaining` is zero before transitioning to running.

kstrl/workqueue.py
```diff
@@ -0,0 +1,1073 @@
...
+        Refuses while the item is running: removing the directory out
+        from under a live worker loses the audit trail for money already
+        spent.
+        """
+        if item.state is ItemState.RUNNING:
+            raise QueueError(
+                f"{item.item_id} is running; stop the run before removing it"
+            )
+        directory = self.item_dir(item)
+        shutil.rmtree(directory, ignore_errors=True)
```
0xfauzi: [P2] Do not report removal when deletion failed. `ignore_errors=True` suppresses permission, filesystem, and symlink errors, after which this method appends a `removed` journal record and the CLI prints success even though the item can still exist. That makes both the operator result and audit trail false. Let `rmtree` failures propagate as `QueueError`/`OSError` (and have the CLI render them), or verify nonexistence before journaling success.

kstrl/workqueue.py
```diff
@@ -0,0 +1,1073 @@
...
+    # ---------------------------------------------------------------- pause
+
+    @property
+    def pause_path(self) -> Path:
+        return self.path / PAUSE_FILENAME
+
+    def pause_state(self) -> PauseState:
+        try:
+            raw = self.pause_path.read_text(encoding="utf-8")
+        except OSError:
```
0xfauzi: [P1] Distinguish an absent pause marker from an unreadable one. This broad `except OSError` returns an unpaused state for `PermissionError`, I/O errors, and other read failures. Patching `Path.read_text` to raise `PermissionError("denied")` makes `pause_state().active()` false, directly contradicting the PR's fail-closed guarantee and allowing unattended spend to resume. Catch `FileNotFoundError` as the normal unpaused case, but return a paused state with an unreadable-marker reason for every other read error.

0xfauzi (commented):
I found seven reproducible queue-safety defects in the new substrate. The most serious allow corrupt sidecar data to redirect deletion outside the queue and allow execution to exceed the configured attempt bound. The full existing suite is green, but it does not cover these cases.

Local verification at this head:
- `uv run pytest -q tests/`: 2688 passed, 28 skipped
- focused queue tests: 103 passed

I would treat the P1 findings as merge blockers.

kstrl/workqueue.py
0xfauzi: Confirmed and fixed in 2dd6e50. Reproduced your exact scenario first: `remove()` deleted `<root>/outside-dir` while `queued/q-safe` stayed put.

Took the "overwrite from the authoritative basename" option, because it matches the doctrine already in the module - the directory is authoritative for `state`, and identity had no business being different. `_load_item_dir` now rejects symlinks, rejects an unsafe directory name, and overwrites `item_id` from `item_path.name`, warning when the two disagree so tampering is visible rather than silently normalized. `item_dir()` also refuses an unsafe id, so an item constructed in memory (never through `from_dict`) cannot reach a join either.

Test: `TestPathSafety::test_a_crafted_item_id_cannot_redirect_removal` asserts the unrelated directory and its contents survive. Mutation-checked in three places: sidecar trust, symlink rejection, `item_dir` guard.

kstrl/workqueue.py
0xfauzi: Confirmed and fixed in 2dd6e50. `Queue.add("remote body", spec_filename="../../../../escaped.md")` did write `escaped.md` at the project root and publish a normal-looking item.

Added `is_safe_component()` (rejects separators, `.`/`..`, empty, NUL, leading dots) and applied it at all three points you named: on enqueue (raises `QueueError`), on decode (`from_dict` returns None, so the item is skipped with a warning), and as a containment re-check in `spec_path()` for items built in memory that never passed through `from_dict`.

Your framing that this is the API remote adapters will call is the reason I treated the sidecar as untrusted input generally rather than patching the one call site - that assumption gets much more load-bearing in PR 3.

Tests: three in `TestPathSafety` plus parametrized coverage of the unsafe/safe name sets. Each mutation-checked separately.

kstrl/workqueue.py
0xfauzi: Confirmed and fixed in 2dd6e50. This one changed the layout rather than adding a filter: staging now lives at `.kstrl/queue/.staging/`, outside the state directories entirely, so a scan cannot reach it even if a name filter were removed. Still the same filesystem, so the publishing `os.replace` is unchanged.

Scans also skip dotted entries as defence in depth (and so a queue written by the previous build does not wedge), and `sweep_staging()` is the explicit recovery policy you asked for. No age heuristic: `add` runs under the queue mutex, so anything in staging while holding that mutex is abandoned by definition. `ks queue add` calls it under the lock.

Your point about the permanent rejected-item warning turned out to be the sharper half. My first regression test passed with the filter reverted, because the new name-safety check rejected `.staging-*` anyway - just noisily, which IS the defect you reported. The test now asserts SILENCE, and a second test models a SIGKILL (rename fails, cleanup never runs) rather than a clean exception, which is what actually distinguishes the new staging location.

kstrl/workqueue.py
0xfauzi: Confirmed and fixed in 2dd6e50. `ks queue add spec.md --max-attempts 0` did exit 0 and persist an unrunnable item.

Both paths, as you specified: `click.IntRange(min=1)` on the option and a check in `Queue.add` so programmatic construction cannot bypass it either.

Worth noting the test needed strengthening: asserting `exit_code != 0` passed with the `IntRange` reverted, because the library guard caught it and exited 1. It now asserts exit code 2 and clicks range message, so the option constraint is what is actually pinned.

kstrl/workqueue.py
0xfauzi: Confirmed and fixed in 2dd6e50, and I think this was the most important of the seven. Reproduced exactly as described: `attempts == 2` against `max_attempts == 1`.

`Queue.start()` now raises `QueueBudgetExhausted` when `attempts_remaining <= 0`, under the same call as the attempt charge. I chose refuse over poison so the substrate does not make a policy decision - PR 2s daemon catches it and poisons, and a caller that ignores the exception leaves the item leased for the reaper rather than running.

Your observation about the old test is the part I want to acknowledge directly: `test_retries_are_bounded_by_max_attempts` skipped the second `start()` when no attempts remained, so it asserted the callers good manners and would have passed no matter what the substrate did. Replaced with a test that drives start -> fail -> requeue -> lease -> start and asserts the refusal, plus one that confirms a reset item can start again so the bound is a bound and not a one-way door.

This is the bug in this PR that would actually have cost money, given PR 2 drives it unattended at $1.70-2.60 per first attempt.

kstrl/workqueue.py
0xfauzi: Confirmed and fixed in 2dd6e50. Took the propagate option: `shutil.rmtree(directory)` without `ignore_errors`, plus an explicit existence check afterwards, and the journal record is written only after the directory is confirmed gone. The CLI catches `(QueueError, OSError)` and prints "Could not remove ..." with exit 1.

Two tests, because the two failure shapes are different: a raising `rmtree` (PermissionError) and a silently no-op one. Both assert the item is still listed AND that no `removed` entry reached the journal - the false audit trail being the half that outlives the session.

kstrl/workqueue.py
0xfauzi: Confirmed and fixed in 2dd6e50. Your reproduction is exact, and this one was the most embarrassing of the seven: the PR body advertised the unreadable pause marker as a fail-closed guarantee, and it only held for `JSONDecodeError`. Any OS-level read failure fell through the broad `except OSError` to an unpaused state.

`FileNotFoundError` is now the only read failure that means unpaused; every other `OSError` returns paused with the error in the reason. Tests cover the `PermissionError` path directly and through `next_ready()` (the actual admission gate), plus the absent-marker case so the normal path stays pinned.

0xfauzi:
## All seven findings closed in 2dd6e50

Every one reproduced against the submitted code before being fixed — the [repro harness](https://github.com/0xfauzi/kstrl/pull/185) confirmed all 7, then confirmed 0 after. Five were enforcement bypasses, and **two broke guarantees this PR's own description claimed** (F3's "scans never see partially published work", F7's fail-closed pause marker).

| # | Finding | Fix |
|---|---|---|
| F1 | P1 sidecar `item_id` redirected `rmtree` outside the queue | directory is authoritative for identity; symlinks rejected; `is_safe_component` on every path component |
| F2 | P1 `spec_filename` traversal on enqueue and decode | validated at enqueue, at decode, and re-checked in `spec_path` |
| F3 | P1 staging dirs returned as phantom items | staging moved **outside** the state dirs to `.kstrl/queue/.staging/`; dot filter as defence in depth; `sweep_staging()` under the lock |
| F4 | P2 `--max-attempts 0` admitted an unrunnable item | `click.IntRange(min=1)` **and** a `Queue.add` check |
| F5 | P1 `max_attempts` advertised but not enforced | `Queue.start` raises `QueueBudgetExhausted` at the spending boundary |
| F6 | P2 failed deletion reported as success | errors propagate; journal written only after the directory is confirmed gone |
| F7 | P1 unreadable pause marker failed **open** | only `FileNotFoundError` means unpaused |

## Three of my regression tests were not discriminating

Worth calling out separately, because it changes what the "mutation-checked" claim is worth. When I mutation-checked the fixes, **3 of 14 mutations were MISSED** — the tests passed with the fix reverted, because a second redundant guard caught the same case. They were rewritten to pin the specific mechanism:

- **F3 scan filter** — now asserts **silence**. The name-safety check rejects `.staging-*` anyway, just noisily, and a permanent rejected-item warning on every scan *is* the defect you reported.
- **F3 staging location** — now models a **SIGKILL** (rename fails, cleanup never runs) instead of a clean exception. After a *successful* add, staging is empty either way, so the original test proved nothing.
- **F4 CLI option** — now asserts **exit 2** and click's range message, not merely a nonzero exit, which the library guard produced regardless.

Second run: **14/14 caught**.

## F5 is the one that would have cost money

`start -> fail -> requeue -> lease -> start` produced `attempts == 2` against `max_attempts == 1`. The bound existed only in the CLI's retry path, and PR 2's callers are an unattended daemon and a reaper. You were right that the old test asserted the caller's good manners rather than the substrate's guarantee — it would have passed no matter what `start()` did.

## Verification

```
uv run pytest tests/     2688 -> 2724 passed, 28 skipped
uv run mypy kstrl/ --strict    Success: no issues found in 109 source files
```

Review corrections are recorded in `docs/dark-factory-roadmap.md` under R8.6, following the R8.1/PR #173 precedent.

**Assumed, not tested:** POSIX-only `flock` and directory-`os.replace` semantics. No factory run — this PR still executes nothing.

Not merging without your approval. Holding PR 2 until you've looked at this.

PR #186, opened 2026-07-30
R8.6 PR 2/4: `ks serve`, the retry classifier, and four backstops

0xfauzi:
Second of four slices for [#153](https://github.com/0xfauzi/kstrl/issues/153). PR 1 (#185) built a place for work to wait; this drains it without a human firing each run — which makes it the module where a mistake spends money while nobody is watching.

| PR | Contents | Status |
|---|---|---|
| 1 | queue substrate + `ks queue` CLI | merged (#185) |
| **2** | **`ks serve`, lease reaper, retry classifier, `daily_budget_usd`, caffeinate** | **this PR** |
| 3 | GitHub Issues adapter, processed-ids ledger, `ks queue sync` | next |
| 4 | launchd plist + docs | |

## The retry rule, stated precisely

R8.6 says "only `infrastructure_error` failures auto-retry". That phrasing leaves the **unknown** case undefined, and the unknown case is where the money goes. `classify_run` implements it as **positive evidence only, failing closed**:

| Evidence | Verdict |
|---|---|
| exit 0 | success |
| launch failed before any spend | retry (free) |
| killed by signal / our timeout | retry (external cause, not a spec verdict) |
| exit 2 | spec failure — the architect halted on a blocker |
| every failed component carries `infrastructure_error` | retry |
| any failed component failed on its merits | spec failure |
| nonzero exit, nothing failed | **unclassifiable → poison** |
| manifest unreadable or absent | **unclassifiable → poison** |

This is deliberately the inverse of "retry unless proven to be a spec failure", which is the shape that produces an overnight crash loop. The same fail-open mistake is already on record in this repo: R8.1 review correction #2 found the changed-file reads returned `[]` on failure, so a `kstrl.toml` diff evaluated as "0 files, 0 lines" and passed every path and size rule.

Two details that make it work:

- **The run lock is probed before launching.** `ks factory` returns 2 both for a held lock and for an architect halt, and those need opposite treatment. Probing first keeps exit 2 unambiguous.
- **The classifier reuses `Finding.is_infrastructure_error`**, not a second copy of the predicate. Two copies of that rule drifting apart is exactly how a spec failure becomes retryable.

## Four backstops

A correct classifier is **not sufficient** — a *persistent* infrastructure fault is retryable by the rules and still burns money.

1. `max_attempts`, enforced inside `Queue.start` (PR 1, review finding F5)
2. Exponential backoff, 60s doubling, capped at 30 min
3. `daily_budget_usd`, checked **before** admitting each item, pausing until the next **local** midnight so a Friday budget hit is not a dead weekend
4. A **consecutive-poison breaker** that pauses the whole queue. If `main` is broken then every run fails verification, each failure is individually legitimate, and no per-item bound ever notices — only a cross-item signal does.

## The budget honesty problem, resolved as flagged

`daily_budget_usd` can only count cost an adapter **reported**, and the codex adapter reports tokens with no cost. With a cost-blind agent the budget is **unenforceable**, not approximate — the same condition PR #184 named for `max_cost_usd`. So:

- the ledger stores the day's spend **with its coverage**, and labels the total a `FLOOR` whenever any run under-reported
- unreported calls are **never** converted into an estimated dollar figure
- `ks serve` **refuses to run unattended** under an unenforceable budget unless `[serve] allow_uncovered_cost = true`

## Design notes

**Subprocess, not in-process `run_factory`.** A crash cannot take the daemon with it, `factory.lock` is released by process death whatever happens, and a timeout is enforceable by killing a process tree. The cost is that classification reads artifacts from disk — which is also what makes it work after a crash.

**The reaper keeps PR 1's LEASED/RUNNING split load-bearing.** A dead `leased` item spent nothing and requeues free; a dead `running` item already charged its attempt and is treated as the infrastructure casualty a suspend or OOM kill actually is.

**`caffeinate -i` wraps each run**, so it dies with the child and the machine sleeps between items rather than being pinned awake by the daemon.

## Tested vs assumed (H4)

**Tested.** 118 new tests (101 serve, 17 CLI). **36/36 mutations caught** — every classifier branch, every backstop, the reaper, the merge gate, both locks, the accounting, caffeinate, and the loop.

**Three harness defects were found and fixed before I trusted any number**, because the first run reported 30/34 with a *different* missed set each time:

1. **Stale `.pyc` made results non-deterministic.** Several mutations change the file size by the same few bytes inside one mtime tick, so CPython reused cached bytecode and the mutation was invisible. Verified by hand that two "MISSED" mutations do fail their tests.
2. **My first fix swept `.venv`**, recompiling all of site-packages — a 0.3s test file started taking minutes.
3. **One mutation hung the harness, and killing it left the mutation live in the source tree.** The `finally` restore never ran. Caught by checking every mutation's replacement text against the source; exactly one was stale.

The harness now purges only the project's caches, times out at 180s (a hang counts as caught), and restores on SIGTERM/SIGINT.

Three of the original tests were also **not discriminating** and were rewritten: the attempt-charge mutation edited an in-memory copy the test could not see; `test_an_elapsed_pause_window_resumes_itself` passed with `resume()` removed because an elapsed window already reads as inactive; and the `--once` mutation had a malformed snippet that never applied.

**Defect 3 also found a real weakness in the code.** `--once` depended on a `break` condition — a fault there turned it into an unbounded 60s-sleep loop, which under launchd is a job that never exits and blocks every later interval. `serve()` now returns `[_cycle()]` **before** entering the loop, so single-shot mode cannot spin regardless of any conditional.

**Assumed, not tested.** POSIX-only `flock` / `os.kill(pid, 0)` semantics. `caffeinate` behavior is tested only at the argv level (`-i`, availability, non-Darwin) — that it actually prevents sleep is asserted by macOS, not by this suite, and PR 4 documents it. **No factory run was performed**: the `FactoryRunner` Protocol exists so the whole loop is drivable with a stub, since a suite that spawned the real thing would cost dollars per assertion.

## Verification

```
uv run pytest tests/     2724 -> 2842 passed, 28 skipped
uv run mypy kstrl/ --strict    Success: no issues found in 110 source files
```

No prompt bodies changed, so H2/H3 do not apply. `[serve]` was registered in `scripts/gen_docs.py` and README regenerated by the generator.

## One thing I deliberately did NOT fix — needs your call

`run_factory` assigns `factory_config.pause_before_pr_merge = bundle.pause_before_pr_merge` **unconditionally**, so at L3+ the ladder overrides an explicit `stop_at_pr` request and logs it as "manual override ignored".

`manual_override_notes` documents itself as guarding against a hand-edited flag **granting** autonomy the ladder never awarded — but the implementation is symmetric and also rejects a **more restrictive** request. Changing that is an R8.2 decision, not an R8.6 one, so `resolve_merge_gate` **refuses** such an item (poison + inbox item) rather than letting the merge gate disappear silently.

Unreachable today (`[autonomy] enabled` defaults false, and L2+ entry is still blocked on the user-run measurements), but it needs an R8.2 decision before L3 is real.

### DIFF
--- a/kstrl/cli.py
+++ b/kstrl/cli.py
@@ -4046,8 +4046,21 @@ def serve(
             "paused",
             f"yes - {pause.reason}" if pause.active() else "no",
         )
-        spend = ledger.read()
-        floor = " (a FLOOR: some calls reported no cost)" if spend.lower_bound else ""
+        from kstrl.serve import ServeStateError
+
+        try:
+            spend = ledger.read()
+        except ServeStateError as exc:
+            ui_impl.err(str(exc))
+            sys.exit(2)
+        floor = (
+            f" (a FLOOR: {spend.uncovered_calls} call(s) reported no cost"
+            + (
+                "; unmetered: " + ", ".join(spend.unmetered_phases)
+                if spend.unmetered_phases else ""
+            )
+            + ")"
+        ) if spend.lower_bound else ""
         ui_impl.kv(
             "today", f"${spend.spent_usd:.2f} over {spend.runs} run(s){floor}",
         )
@@ -4056,13 +4069,13 @@ def serve(
             f"${config.daily_budget_usd:.2f}"
             if config.daily_budget_usd > 0 else "unset (no cap)",
         )
-        ui_impl.kv("consecutive poison", str(consecutive_poison_count(queue)))
+        ui_impl.kv("consecutive poison", str(consecutive_poison_count(ledger)))
         ui_impl.kv(
             "factory lock", "held by another run" if factory_lock_held(root_dir)
             else "free",
         )
         for label, admission in (
-            ("poison breaker", check_poison_breaker(queue, config)),
+            ("poison breaker", check_poison_breaker(ledger, config)),
             ("cost coverage", check_cost_coverage(ledger, config)),
             ("budget", check_budget(ledger, config)),
             ("inbox cap", check_inbox_cap(root_dir)),
@@ -4107,11 +4120,19 @@ def serve(
     ui_impl.info("")
     ui_impl.kv("cycles", str(len(results)))
     ui_impl.kv("items run", str(len(ran)))
-    # Nonzero only when an item ended up needing a human, so a launchd
-    # KeepAlive job's exit status means something.
-    poisoned = [r for r in ran if r.verdict is not None and not r.verdict.may_retry
-                and str(r.verdict) != "success"]
-    sys.exit(1 if poisoned else 0)
+    # Derived from the TERMINAL queue outcome, not from the classifier's
+    # retry permission. Review #186 F10: an infra verdict whose last
+    # attempt was spent is poisoned by serve_cycle, yet
+    # Verdict.RETRY_INFRA.may_retry stays true - and the reaper and
+    # merge-gate poison paths set no ran_item at all - so the old filter
+    # exited 0 on work that was waiting for a human.
+    needs_human = [r for r in results if r.needs_human]
+    if needs_human:
+        ui_impl.warn(
+            f"{len(needs_human)} cycle(s) produced work awaiting a human; "
+            "see `ks inbox ls` and `ks queue ls --state poison`"
+        )
+    sys.exit(1 if needs_human else 0)
 
 
 class _ServeUiObserver:
--- a/kstrl/cli.py
+++ b/kstrl/cli.py
@@ -3976,6 +3976,160 @@ def queue_resume(root: Path | None, ui: str, no_color: bool) -> None:
     sys.exit(0)
 
 
+@cli.command()
+@click.option(
+    "--once", is_flag=True,
+    help="Run a single poll cycle and exit (launchd/cron fallback mode)",
+)
+@click.option(
+    "--max-cycles", type=click.IntRange(min=0), default=0,
+    help="Stop after this many cycles; 0 runs until interrupted",
+)
+@click.option(
+    "--dry-run", is_flag=True,
+    help="Report what the next cycle would do without spending anything",
+)
+@_queue_root_option
+@_queue_ui_option
+@_queue_no_color_option
+def serve(
+    once: bool,
+    max_cycles: int,
+    dry_run: bool,
+    root: Path | None,
+    ui: str,
+    no_color: bool,
+) -> None:
+    """Drain the continuous-intake queue (R8.6).
+
+    Runs one factory invocation at a time, holding a daemon singleton
+    lock. Under launchd this is invoked with --once on an interval;
+    StartInterval fires immediately for intervals that elapsed while the
+    machine slept, so a laptop that was closed catches up on wake.
+
+    Only infrastructure failures are retried, and only with positive
+    evidence. Spec-level failures go to poison/ and wait for a human.
+    """
+    from kstrl.serve import (
+        ServeConfig,
+        ServeError,
+        ServeLockedError,
+        SpendLedger,
+        check_budget,
+        check_cost_coverage,
+        check_inbox_cap,
+        check_poison_breaker,
+        consecutive_poison_count,
+        factory_lock_held,
+    )
+    from kstrl.serve import serve as run_serve
+    from kstrl.workqueue import Queue, QueueConfig, summarize
+
+    root_dir = (root or Path.cwd()).resolve()
+    ui_impl = _autonomy_ui(ui, no_color)
+    try:
+        config = ServeConfig.load(root_dir)
+    except (ServeError, ValueError) as exc:
+        ui_impl.err(str(exc))
+        sys.exit(2)
+    queue = Queue(root_dir, QueueConfig.load(root_dir))
+    ledger = SpendLedger(root_dir)
+
+    if dry_run:
+        # Deliberately reports the same gates the loop evaluates, in the
+        # same order, so "what would it do" cannot drift from "what it
+        # does" without a test noticing.
+        ui_impl.section("Serve dry run")
+        ui_impl.kv("queue", summarize(queue.counts()))
+        pause = queue.pause_state()
+        ui_impl.kv(
+            "paused",
+            f"yes - {pause.reason}" if pause.active() else "no",
+        )
+        spend = ledger.read()
+        floor = " (a FLOOR: some calls reported no cost)" if spend.lower_bound else ""
+        ui_impl.kv(
+            "today", f"${spend.spent_usd:.2f} over {spend.runs} run(s){floor}",
+        )
+        ui_impl.kv(
+            "daily budget",
+            f"${config.daily_budget_usd:.2f}"
+            if config.daily_budget_usd > 0 else "unset (no cap)",
+        )
+        ui_impl.kv("consecutive poison", str(consecutive_poison_count(queue)))
+        ui_impl.kv(
+            "factory lock", "held by another run" if factory_lock_held(root_dir)
+            else "free",
+        )
+        for label, admission in (
+            ("poison breaker", check_poison_breaker(queue, config)),
+            ("cost coverage", check_cost_coverage(ledger, config)),
+            ("budget", check_budget(ledger, config)),
+            ("inbox cap", check_inbox_cap(root_dir)),
+        ):
+            if admission.allowed:
+                ui_impl.info(f"  gate {label}: ok")
+            else:
+                ui_impl.warn(f"  gate {label}: BLOCKS - {admission.reason}")
+        candidate = queue.next_ready()
+        ui_impl.kv(
+            "next item",
+            f"{candidate.item_id[:12]} - {candidate.title}"
+            if candidate else "nothing ready",
+        )
+        sys.exit(0)
+
+    ui_impl.info(
+        f"ks serve on {root_dir} "
+        f"(poll {config.poll_interval_seconds:.0f}s, "
+        f"budget "
+        + (f"${config.daily_budget_usd:.2f}/day" if config.daily_budget_usd > 0
+           else "unset")
+        + f", caffeinate {'on' if config.caffeinate else 'off'})"
+    )
+    observer = _ServeUiObserver(ui_impl)
+    try:
+        results = run_serve(
+            root_dir,
+            once=once,
+            config=config,
+            observer=observer,
+            max_cycles=max_cycles,
+        )
+    except ServeLockedError as exc:
+        ui_impl.err(str(exc))
+        sys.exit(2)
+    except KeyboardInterrupt:
+        ui_impl.info("Interrupted; the current item keeps its lease for the reaper.")
+        sys.exit(0)
+
+    ran = [r for r in results if r.ran_item]
+    ui_impl.info("")
+    ui_impl.kv("cycles", str(len(results)))
+    ui_impl.kv("items run", str(len(ran)))
+    # Nonzero only when an item ended up needing a human, so a launchd
+    # KeepAlive job's exit status means something.
+    poisoned = [r for r in ran if r.verdict is not None and not r.verdict.may_retry
+                and str(r.verdict) != "success"]
+    sys.exit(1 if poisoned else 0)
+
+
+class _ServeUiObserver:
+    """Adapts the daemon's narration onto the console UI."""
+
+    def __init__(self, ui_impl: UI) -> None:
+        self._ui = ui_impl
+
+    def info(self, message: str) -> None:
+        self._ui.info(message)
+
+    def warn(self, message: str) -> None:
+        self._ui.warn(message)
+
+    def err(self, message: str) -> None:
+        self._ui.err(message)
+
+
 def main() -> None:
     """Main entry point."""
     cli()
--- a/kstrl/serve.py
+++ b/kstrl/serve.py
@@ -1,0 +1,1524 @@
+"""R8.6 continuous intake: the daemon that drains the work queue.
+
+PR 1 built a place for work to wait. This is what runs it without a
+human firing each run - and therefore the module where a mistake spends
+money while nobody is watching. A measured factory engineer iteration
+costs ~$1.70-2.60 on a first attempt and $3.99-7.42 on a retry (retries
+carry accumulated context), and a 5-story component with adversarial
+review ran ~$29 and 96 minutes without finishing. Those numbers are why
+almost every decision below errs toward NOT launching a run.
+
+**The retry rule, stated precisely.** R8.6 says "only
+``infrastructure_error`` failures auto-retry". That phrasing leaves the
+UNKNOWN case undefined, and the unknown case is where the money goes, so
+this module implements it as *positive evidence only, failing closed*: an
+item is retried only when we can read affirmative evidence that the
+failure was infrastructural. An unreadable manifest, a missing artifact,
+a run that died before recording anything, an exit code we do not
+recognise - all poison, none retry. This is deliberately the inverse of
+"retry unless proven to be a spec failure", which is the shape that
+produces an overnight crash loop.
+
+The same class of mistake is already on record in this repo: R8.1 review
+correction #2 found the changed-file reads were fail-OPEN, so a
+``kstrl.toml`` diff evaluated as "0 files, 0 lines" and passed every
+path and size rule. Defaulting to permissive on missing evidence looks
+harmless until the evidence goes missing.
+
+**Four independent backstops**, because a correct classifier is not
+sufficient - a *persistent* infrastructure fault is retryable by the
+rules and still burns money:
+
+1. ``max_attempts`` per item, enforced by ``Queue.start`` itself.
+2. Exponential backoff between attempts, so a fast-failing item cannot
+   spin.
+3. ``daily_budget_usd``, checked BEFORE admitting each item.
+4. A consecutive-poison breaker that pauses the whole queue. If ``main``
+   is broken then every run fails verification, each failure is
+   individually legitimate, and per-item bounds never notice; only a
+   cross-item signal does.
+
+**Honesty about the budget (H4).** ``daily_budget_usd`` can only count
+cost that an adapter reported. The codex adapter reports tokens and no
+cost, so with a cost-blind agent the budget is not approximate - it is
+*unenforceable*, the same condition PR #184 named for ``max_cost_usd``.
+This module therefore records the day's spend together with its
+coverage, never converts unreported calls into an estimated dollar
+figure, and refuses to run unattended when a budget is configured but
+cost coverage is absent (``allow_uncovered_cost`` is the explicit
+override).
+"""
+
+from __future__ import annotations
+
+import json
+import os
+import shutil
+import socket
+import subprocess
+import sys
+import time
+from collections.abc import Iterator
+from contextlib import contextmanager
+from dataclasses import dataclass, field
+from datetime import UTC, datetime, timedelta
+from enum import StrEnum
+from pathlib import Path
+from typing import Any, Protocol
+
+from kstrl.statedir import state_dir
+from kstrl.workqueue import (
+    ItemState,
+    MergeDisposition,
+    Queue,
+    QueueBudgetExhausted,
+    QueueConfig,
+    QueueItem,
+    queue_lock,
+    queue_root,
+)
+
+SERVE_LOCK_FILENAME = "serve.lock"
+SPEND_FILENAME = "spend.json"
+
+#: Retry backoff: ``base * 2 ** (attempts - 1)``, capped. Not tunable by
+#: config on purpose - the cap is a safety floor on how fast an item can
+#: consume its attempts, not a preference.
+BACKOFF_BASE_SECONDS = 60.0
+BACKOFF_CAP_SECONDS = 1800.0
+
+
+class ServeError(RuntimeError):
+    """The daemon cannot run."""
+
+
+class ServeLockedError(ServeError):
+    """Another ``ks serve`` holds the singleton lock on this root."""
+
+
+class Verdict(StrEnum):
+    """What the evidence says about a finished run.
+
+    ``UNCLASSIFIABLE`` is a first-class outcome rather than an error
+    path: it is the common case for a crash, and treating it as "probably
+    fine, retry" is the bug this whole module is arranged to avoid.
+    """
+
+    SUCCESS = "success"
+    RETRY_INFRA = "retry_infra"
+    SPEC_FAILURE = "spec_failure"
+    UNCLASSIFIABLE = "unclassifiable"
+
+    @property
+    def may_retry(self) -> bool:
+        """Only ONE verdict authorizes spending again."""
+        return self is Verdict.RETRY_INFRA
+
+
+@dataclass(frozen=True)
+class Outcome:
+    """A verdict plus the evidence that produced it.
+
+    ``evidence`` is written to the inbox item, so a human deciding
+    whether to requeue sees what the classifier actually read rather
+    than having to trust its label.
+    """
+
+    verdict: Verdict
+    reason: str
+    evidence: dict[str, Any] = field(default_factory=dict)
+
+
+@dataclass(frozen=True)
+class RunOutcome:
+    """Raw result of one factory invocation, before interpretation."""
+
+    returncode: int
+    timed_out: bool = False
+    #: Non-empty when the launch itself failed (binary missing, etc).
+    launch_error: str = ""
+
+
+class FactoryRunner(Protocol):
+    """How the daemon executes one queue item.
+
+    A Protocol so tests drive the entire loop without spawning a factory.
+    That is not only a speed concern: a suite that ran the real thing
+    would cost dollars per assertion.
+    """
+
+    def __call__(
+        self,
+        *,
+        root_dir: Path,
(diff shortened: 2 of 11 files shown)


kstrl/serve.py
```diff
@@ -0,0 +1,1524 @@
...
+        "--no-color",
+    ]
+    command.append(
+        "--pause-before-pr-merge" if pause_before_pr_merge
+        else "--no-pause-before-pr-merge"
+    )
+    env = dict(os.environ)
+    env["KSTRL_NO_TUI"] = "1"
+    try:
+        completed = subprocess.run(
```
0xfauzi: [P1] Supervise and terminate the whole factory process tree. `subprocess.run(..., timeout=...)` kills only its direct child; there is no new session or `killpg`. I reproduced the same primitive with a shell that spawned `sleep`: after `TimeoutExpired`, the descendant was still alive. On macOS the direct child is also the `caffeinate` wrapper, so the actual factory can outlive the timeout while this code records an infrastructure failure and requeues it. The lease compounds this because `queue.lease()` records the daemon PID, not the factory child: if the daemon dies while the child survives, a successor reaps and requeues an active run. Use `Popen(start_new_session=True)`, persist the child PID/process group as the running lease owner, and terminate plus wait for the entire group before returning a timeout or allowing recovery.

kstrl/serve.py
```diff
@@ -0,0 +1,1524 @@
...
+def _run_id_from_manifest(manifest_path: Path) -> str:
+    """The run id the factory recorded, or "" if it never got that far."""
+    try:
+        raw = manifest_path.read_text(encoding="utf-8")
+        data = json.loads(raw)
+    except (OSError, json.JSONDecodeError):
+        return ""
+    if not isinstance(data, dict):
+        return ""
+    run_id = data.get("run_id")
```
0xfauzi: [P1] Bind accounting and classification to an invocation-specific run ID. `Manifest.save()` writes `runId`, but this reads `run_id`, so it returns `""` for every real manifest. `read_run_spend(root, "")` then deliberately loads the newest historical run, and `classify_run()` reads the shared manifest path. I reproduced an early failed invocation charging prior spend and receiving `retry_infra` from an old manifest. Fixing only the key still leaves the stale-artifact window when launch/decomposition fails before rewriting the manifest. Mint or capture the run ID before launch, return it in `RunOutcome`, require artifacts owned by that exact invocation, and never use the empty-ID/latest-run fallback for a money or retry decision.

kstrl/serve.py
```diff
@@ -0,0 +1,1524 @@
...
+        project_name=project_name,
+        pause_before_pr_merge=gate.pause_before_pr_merge,
+        timeout_seconds=cfg.factory_timeout_seconds,
+    )
+
+    # 6. Charge the spend before deciding anything, so a classification
+    #    bug cannot also lose the accounting.
+    manifest_path = root_dir / "scripts" / "kstrl" / "manifest.json"
+    run_id = _run_id_from_manifest(manifest_path)
+    spend = read_run_spend(root_dir, run_id)
```
0xfauzi: [P1] Include architect/decomposition usage in the daily spend. `ks factory --spec` calls `decompose_spec()` before `run_factory`; the run ID, run event bus, and usage paths are created only inside `run_factory`. The decomposition code calls the agent but emits no `ComponentUsage`, and the CLI passes no bus there. Therefore every queued item's mandatory architect calls, including a blocker exit 2, are absent from `read_run_spend()`. A later engineer/reviewer stream can even make coverage look complete while this paid phase is missing. Plumb one run ID and usage meter through decomposition and execution, or charge decomposition separately with its own coverage before applying the daily gate.

kstrl/serve.py
```diff
@@ -0,0 +1,1524 @@
...
+        here: failing closed would mean an unparseable ledger halts the
+        queue permanently with no way for the daemon itself to recover.
+        The exposure is bounded by ``max_attempts``, the backoff, and the
+        poison breaker, and by the ledger being rewritten atomically so
+        a torn write is not a normal event.
+        """
+        stamp = today or _local_today()
+        try:
+            raw = self.path.read_text(encoding="utf-8")
+        except OSError:
```
0xfauzi: [P1] Fail closed when the spend ledger is unreadable. I charged $9, corrupted the ledger, and `check_budget(... daily_budget_usd=5)` saw $0 and allowed another run. The cited backstops do not bound this exposure across distinct successful queue items, so a damaged or permission-denied ledger can disable the only queue-wide spend limit indefinitely. An atomic writer prevents ordinary torn writes but does not make read failures safe. Pause with an actionable ledger error, and use an atomic backup or append-only source from which an operator can recover explicitly rather than silently starting a fresh day.

kstrl/serve.py
```diff
@@ -0,0 +1,1524 @@
...
+    """How many items poisoned in a row, most recent first.
+
+    Derived from the journal rather than a counter file: the journal is
+    already the audit trail, and a separate counter is one more thing
+    that can disagree with reality. Only terminal transitions count -
+    a requeue or a lease in between is not a verdict.
+    """
+    terminal = {"done", "poison"}
+    streak = 0
+    for entry in reversed(queue.journal_entries()):
```
0xfauzi: [P1] Do not build the poison breaker on the best-effort journal. `Queue._journal()` intentionally swallows append failures and `journal_entries()` returns `[]` on any read error while skipping malformed lines. With three authoritative poison directories present, making `journal_entries()` unreadable makes this gate report a zero streak and allow more spending. Persist the streak in authoritative atomic state updated with terminal transitions, or fail closed when the journal cannot prove the streak. A money backstop cannot depend on narration that the queue explicitly permits itself to lose.

kstrl/serve.py
```diff
@@ -0,0 +1,1524 @@
...
+        # evidence of an EXTERNAL cause rather than a verdict on the
+        # spec, so it is the one crash shape that legitimately retries.
+        return Outcome(
+            Verdict.RETRY_INFRA,
+            f"killed by signal {-run.returncode} (external cause, not a "
+            "verdict on the spec)",
+            {"signal": -run.returncode},
+        )
+
+    if run.returncode == 2:
```
0xfauzi, line 727: [P1] A prior lock probe cannot make exit code 2 unambiguous. The probe acquires and immediately releases `factory.lock`; a manual factory can acquire it in the gap before this child does, causing the child to exit 2 for lock contention and this branch to poison the item as a spec failure. `factory_lock_held()` also returns false when opening the lock file fails. This is a TOCTOU race around the exact distinction the classifier relies on. Give lock contention and architect refusal distinct exit codes or structured outcomes from `ks factory`, or reserve the lock continuously across launch instead of inferring cause from a stale probe.

kstrl/serve.py
```diff
@@ -0,0 +1,1524 @@
...
+            dedupe_key=f"queue-poison:{item_id}",
+            evidence={"item_id": item_id, "cause": "interrupted run"},
+        ),)
+
+    # 2. An elapsed resume_after clears the pause on its own; that is what
+    #    makes the daily-budget stop self-healing rather than a weekend
+    #    of dead queue.
+    pause = queue.pause_state()
+    if pause.paused and not pause.active(moment):
+        queue.resume(actor="serve")
```
0xfauzi: [P1] Make elapsed-pause clearing atomic with operator pause writes. This read and `resume()` happen outside `queue_lock`. I reproduced the interleaving where the daemon reads an expired budget pause, an operator writes a fresh emergency pause, and then the daemon overwrites it with `resume()` and starts the queued item. Hold the queue mutex across reading the marker and conditionally clearing that same marker, and recheck under the lock so a newer manual pause always wins.

kstrl/serve.py
```diff
@@ -0,0 +1,1524 @@
...
+    override rather than quietly running with a cap that does nothing.
+
+    Only fires once a run has actually reported: with no data yet there
+    is no evidence of a gap, and refusing on absence of evidence would
+    make the daemon unusable on a fresh repo.
+    """
+    if config.daily_budget_usd <= 0 or config.allow_uncovered_cost:
+        return Admission(allowed=True)
+    spend = ledger.read(today)
+    if spend.runs == 0:
```
0xfauzi: [P1] Do not spend one full run before discovering that the configured budget is unenforceable. With a fresh ledger and `daily_budget_usd > 0`, this branch allows the first cost-blind run; only the next cycle sees its uncovered calls and pauses. I reproduced exactly `first_ran=True, second_paused=True`. That contradicts the documented promise that serve refuses unattended execution under an unenforceable budget, and the first run is not bounded by this budget at all. Resolve cost-reporting capability before claiming work, or require the explicit `allow_uncovered_cost` authorization until coverage has been positively established.

kstrl/serve.py
```diff
@@ -0,0 +1,1524 @@
...
+        return Admission(allowed=True)
+    if spend.spent_usd > 0 and not spend.lower_bound:
```
0xfauzi: [P2] Determine coverage from call counts, not whether dollars are positive. `DailySpend` discards `RunSpend.cost_calls` and `usage_calls`, so a fully covered run reporting a legitimate $0 cost is blocked with the false message that `0 ... calls reported no cost`. A launch/no-usage result also increments `runs` and hits the same branch on the next cycle. Persist covered and total call counts in the ledger, avoid counting a launch failure as a metered run, and gate on uncovered calls rather than `spent_usd > 0`.

kstrl/cli.py
```diff
@@ -3976,6 +3976,160 @@ def queue_resume(root: Path | None, ui: str, no_color: bool) -> None:
...
+        ui_impl.info("Interrupted; the current item keeps its lease for the reaper.")
+        sys.exit(0)
+
+    ran = [r for r in results if r.ran_item]
+    ui_impl.info("")
+    ui_impl.kv("cycles", str(len(results)))
+    ui_impl.kv("items run", str(len(ran)))
+    # Nonzero only when an item ended up needing a human, so a launchd
+    # KeepAlive job's exit status means something.
+    poisoned = [r for r in ran if r.verdict is not None and not r.verdict.may_retry
```
0xfauzi: [P2] Base the exit status on the terminal queue outcome, not the classifier's retry permission. An infrastructure verdict with its last attempt exhausted is poisoned by `serve_cycle`, but `Verdict.RETRY_INFRA.may_retry` remains true, so this filter excludes it and `ks serve --once` exits 0. Merge-gate and reaper poison paths also have no `ran_item` and are excluded. Add an explicit `needs_human` or terminal-state field to `CycleResult`, set it on every poison/refusal path, and derive the launchd status from that.

kstrl/serve.py
```diff
@@ -0,0 +1,1524 @@
...
+            observer=obs,
+        )
+
+    with serve_lock(root_dir):
+        if once:
+            return [_cycle()]
+
+        results: list[CycleResult] = []
+        while True:
+            results.append(_cycle())
```
0xfauzi: [P2] Do not retain every poll result for the lifetime of an unbounded daemon. The normal `max_cycles == 0` path appends one `CycleResult` forever and has no return path that uses the list; the CLI catches `KeyboardInterrupt` outside and discards it. This makes memory growth unbounded by construction. Keep results only for `--once` or bounded/test runs, use a bounded deque for diagnostics, or expose cycles as an iterator.

0xfauzi (commented):
I found eleven reproducible or directly traceable defects in the unattended execution path. The most serious gaps allow a timed-out factory process tree to keep spending, reuse previous-run artifacts for billing and retry classification, omit architect spend from the daily ledger, and erase safety backstops when their state files become unreadable.

Verification at this head:
- `uv run pytest -q tests/`: 2842 passed, 28 skipped
- focused serve/queue tests: 224 passed
- generated docs check: clean
- GitHub CI: green

I would treat all P1 findings as merge blockers. This is a comment review because GitHub does not permit the connected PR owner account to request changes on its own PR.

kstrl/serve.py
0xfauzi: Confirmed and fixed in c240422. Reproduced your primitive exactly: after `TimeoutExpired`, `sleep` descendants were still running.

`run_supervised` now uses `Popen(start_new_session=True)` and signals the whole group. Two details beyond what you specified, both because my own tests found them:

1. **The pgid is captured at SPAWN, not at kill time.** Looking it up in the killer fails once the direct child is reaped - and that is precisely the case where descendants survive, so a lazy lookup reports "reaped" for the one situation the function exists to detect.
2. **`verify.py` already had a guard I was missing.** A mocked `Popen` pid coerces to 1 via `MagicMock.__index__`, and `killpg(1, sig)` is `kill(-1, sig)` - every process the user owns. `_safe_pgid` now carries the same `pgid > 1 and pgid != getpgrp()` guards `verify._signal_process_group` documents, with a test that pins them.

On the lease: `Queue.adopt_lease` re-points the lease at the child pid, so a successor daemon judging liveness looks at the factory rather than at the daemon. And a timeout whose group cannot be *confirmed* dead now poisons rather than retrying - handing the item back while a factory may still be writing is the outcome worth avoiding.

I also extracted `run_supervised` from `subprocess_factory_runner`: the runner hardcodes `ks factory` as its argv, so the supervision was unreachable from tests. `serve.py` is enrolled in `POPEN_ALLOWLIST` with the same justification `verify.py` carries.

kstrl/serve.py
0xfauzi: Confirmed and fixed in c240422. The key mismatch is exact - I reproduced `_run_id_from_manifest` returning `""` for a manifest whose run id was `factory-REAL-123`, because `Manifest.to_dict` writes `runId`.

Fixing the key alone would have left the stale-artifact window you named, so ownership is now positive: a pre/post snapshot of `.kstrl/runs/` names, and

- only run dirs created BY this invocation are charged;
- the manifest is read only when its `runId` is among them, otherwise `classify_run` receives `None` and returns unclassifiable;
- `read_run_spend("")` returns zeros without consulting the reducer at all.

Worth noting the test lesson: fixing only the key is undetectable by a stale-manifest test, because a misread id yields `""` and lands on the same unclassifiable branch. It took a positive-path test (an owned manifest that must be read and classified) to pin it.

kstrl/serve.py
0xfauzi: Confirmed - and this one is deeper than wiring. `decompose_spec` emits no `ComponentUsage` at all, so plumbing a bus would not have fixed it; the architect phase is structurally unmetered, and metering it belongs with the architect's instrumentation rather than R8.6.

So I took your second option and refused to estimate the gap. The ledger records `unmetered_phases = ("architect",)` for every queued item, and that alone makes `DailySpend.lower_bound` true, so the day's total is labelled a FLOOR everywhere it is reported.

That reshaped the budget contract into three honest cases rather than two:

- **full coverage** - the cap is exact;
- **partial coverage** - a lower-bound cap: it fires at or after the threshold, never before. Still a real bound, and now always labelled;
- **zero coverage** - the total stays $0 forever and the cap can never fire. Refused (see F8).

The architect keeps it permanently in the middle case until its usage is emitted. I would rather ship a cap that is honestly late than one that claims to be exact.

kstrl/serve.py
0xfauzi: Confirmed and fixed in c240422. Reproduced: $9 charged, ledger corrupted, $5 budget, another run allowed - and your point that the per-item backstops do not bound this across distinct items is the part that made it a P1 rather than a nuisance.

`ServeStateError` now halts the cycle with the path and an actionable message, files an inbox item, and `FileNotFoundError` is the only read failure that still means "first run". Same correction PR #185 F7 applied to the pause marker; I should have applied it here at the same time, and the fact that I wrote a paragraph rationalising the fail-open should have been the tell.

kstrl/serve.py
0xfauzi: Confirmed and fixed in c240422. Reproduced: three poison directories on disk, streak 3, then blanking the journal reported 0 and the gate allowed spending.

The streak is now authoritative state in `ServeState` (`.kstrl/queue/spend.json`), written atomically on each terminal transition. Two properties I added tests for while moving it:

- a success resets it, a poison increments it;
- it does NOT reset with the daily spend rollover - a poison streak is not a daily fact, and resetting it at midnight would hand a broken base branch a fresh budget every night.

Your framing is the right one and I have quoted it in the code comment: a money backstop cannot depend on narration the queue explicitly permits itself to lose.

kstrl/serve.py
0xfauzi: Confirmed and fixed in c240422 - you are right that the probe cannot make exit 2 unambiguous, and I overclaimed in both the code comment and the PR body.

I did not change the factory's exit-code contract (that felt like R8.6 reaching into another subsystem mid-review), so the classifier now decides from the child's own OUTPUT, which is evidence rather than inference:

- output contains the lock-refusal marker -> `RETRY_INFRA` (contention is infrastructural);
- output contains the spec-blocker marker -> `SPEC_FAILURE`;
- neither -> `UNCLASSIFIABLE`, i.e. poison, rather than guessing which refusal it was.

`factory_lock_held` also fails closed when the lock file cannot be opened, and its comment now says plainly that it is a courtesy check that does not disambiguate anything.

If you would rather have distinct exit codes from `ks factory` - which would be strictly better than marker matching - I am happy to do it, but it changes a public contract `tests/test_instance_safety.py` currently pins at 2, so I would want that as its own change.

kstrl/serve.py
0xfauzi: Confirmed and fixed in c240422. Both the read and the conditional clear now happen inside one `queue_lock` block, with the marker re-read under the lock, and `_pause_queue` takes the lock for its write too - losing the operator's pause is the dangerous direction.

One honest note on testing this: my first attempt patched `pause_state` to write a competing pause inside the daemon's lock window, which is not a reachable interleaving once both sides take the mutex - the test was simulating something the fix makes impossible. It is now pinned three ways instead: an operator pause with no `resume_after` is never cleared, a live budget pause survives the cycle that observes it, and a structural check that the read's nearest enclosing `with` is the queue lock.

kstrl/serve.py
0xfauzi: Confirmed and fixed in c240422. I reproduced your exact result - `first_ran=True, second_paused=True` - and you are right that it contradicted the documented promise.

Coverage capability is now resolved BEFORE the first claim, from a persisted `cost_coverage_seen` flag in `ServeState`. With a budget set and no call having ever reported a cost figure, `ks serve` refuses immediately and names `allow_uncovered_cost` as the explicit authorization. The flag survives the daily rollover, because adapter capability is not a daily fact.

Combined with F9 this also removes the "spend a run to learn" shape entirely: the gate no longer needs a completed run to have an opinion.

kstrl/serve.py
0xfauzi: Confirmed and fixed in c240422. All three parts:

- `DailySpend` now persists `covered_calls` and `total_calls`, and coverage is judged from those rather than from `spent_usd > 0`. A fully-metered run that legitimately cost $0 is allowed - that false rejection is now a named regression test.
- `charge(metered_run=False)` records a launch failure's spend without incrementing `runs`, so it cannot trip the gate on the next cycle.
- The gate keys on whether ANY call reported cost, which is what actually determines whether the cap can ever fire; partial coverage is allowed as a lower-bound cap and labelled a floor.

`lower_bound` is also derived from the counts now rather than stored as a sticky flag, so the summary cannot disagree with the numbers it summarises.

kstrl/cli.py
0xfauzi: Confirmed and fixed in c240422. `CycleResult.needs_human` is now set on every poison and refusal path - the main classifier path, the reaper's poison, the merge-gate refusal, the exhausted-attempts poison, the orphaned-process poison, the budget pause, and the unreadable-ledger halt - and the CLI derives its exit status from that alone.

Your two cases are both covered by name: an exhausted `RETRY_INFRA` verdict (where `may_retry` stays true, which is exactly what fooled the old filter) and a reaper poison with no `ran_item` at all.

kstrl/serve.py
0xfauzi: Confirmed and fixed in c240422. The unbounded daemon path now keeps a `deque(maxlen=RECENT_CYCLE_WINDOW)` for diagnostics; `--once` and bounded/test runs still return every result, since those callers actually consume the list.

"Unbounded by construction" is the accurate description - there was no return path that could ever surface it.

0xfauzi:
## All eleven findings closed in `c240422`

Every one reproduced against the submitted code before being fixed. Eight were enforcement bypasses in the unattended spend path, and **two broke guarantees this PR's own description claimed** (F4's fail-closed ledger, F6's "the probe makes exit 2 unambiguous").

| # | Finding | Fix |
|---|---|---|
| F1 | P1 timeout killed only the direct child; factory outlived it | `start_new_session` + `killpg`, pgid captured **at spawn**, child adopts the lease, unconfirmed reap poisons |
| F2 | P1 read `run_id`; `Manifest` writes `runId` → charged/classified another run | pre/post run-dir snapshot; only owned artifacts count |
| F3 | P1 architect spend invisible (`decompose` emits no usage at all) | named `unmetered_phases`; day's total always a FLOOR, never estimated |
| F4 | P1 corrupt ledger disabled the only queue-wide cap | `ServeStateError` halts; only `FileNotFoundError` means "first run" |
| F5 | P1 poison breaker built on the best-effort journal | authoritative streak in `ServeState`; survives the daily rollover |
| F6 | P1 lock probe TOCTOU on exit 2 | classify from the child's **output**; unclassifiable when neither marker appears |
| F7 | P1 pause read/clear raced operator pauses | both under the queue mutex, re-read inside it |
| F8 | P1 one run spent before unenforceability was noticed | persisted `cost_coverage_seen`, checked before the first claim |
| F9 | P2 coverage inferred from dollars | persisted covered/total **call counts**; launch failures aren't metered runs |
| F10 | P2 exit status followed `may_retry` | `needs_human` on every poison/refusal path drives the exit code |
| F11 | P2 unbounded `CycleResult` retention | bounded deque on the daemon path |

## F3 was deeper than a wiring bug

`decompose_spec` emits **no usage events at all**, so plumbing a bus would not have fixed it — the architect is structurally unmetered, and metering it belongs with the architect's instrumentation, not R8.6. So I took your second option and refused to estimate the gap. That reshaped the budget contract into three honest cases instead of two:

- **full coverage** — the cap is exact
- **partial coverage** — a *lower-bound* cap: fires at or after the threshold, never before. Still a real bound, and now always labelled a FLOOR
- **zero coverage** — the total stays $0 forever and the cap can never fire. Refused (F8)

The architect keeps every queued item permanently in the middle case. I'd rather ship a cap that is honestly late than one that claims to be exact.

## Two fixes came from the existing code, not the review

- **`verify.py` already had a guard I was missing.** A mocked `Popen` pid coerces to `1`, and `killpg(1, sig)` is `kill(-1, sig)` — every process the user owns. `_safe_pgid` now carries the same guards `verify._signal_process_group` documents.
- **`serve.py` is enrolled in `POPEN_ALLOWLIST`** with the same `communicate(timeout)` + group-kill justification `verify.py` carries. The repo's own audit caught the bare `Popen`.

## Mutation check: 36/36 — after nine rewrites

The first run reported **27/36**. Nine tests were not discriminating, and diagnosing them was the most useful part of this pass:

- **The autouse `_no_spend` fixture patched the very function under test.** The test imported `read_run_spend` in its body, so it exercised the stub and passed regardless of what the real function did. Now bound at module import, before any fixture can patch it.
- **A spy counted the liveness probe as a kill.** `terminate_process_group` calls `killpg(pgid, 0)` to probe and `killpg(pgid, SIGTERM)` to kill; recording both let "never signal the group" pass on the probe alone.
- **Three were masked by redundant guards** — a wrong manifest key and an unowned manifest land on the *same* branch, so the stale-manifest test couldn't tell them apart. Fixed with positive-path tests.
- **One "missed" mutation was not a defect.** Dropping the spawn-time pgid is behaviourally *identical* on the timeout path (the child is still alive there); the captured value only matters for the already-reaped case, which that path cannot reach. That test now asserts the wiring and says so, rather than pretending the outcome differs.

`run_supervised` was extracted from `subprocess_factory_runner` because the runner hardcodes `ks factory` as its argv, leaving the money-critical supervision unreachable from tests.

## Verification

```
uv run pytest tests/     2842 -> 2895 passed, 28 skipped
uv run mypy kstrl/ --strict    Success: no issues found in 110 source files
```

Also verified no mutation text was left in the source after the run.

**Assumed, not tested.** POSIX-only `flock` / `killpg` / `os.kill(pid, 0)`. `caffeinate` is tested at the argv level only — that it actually prevents sleep is asserted by macOS, and PR 4 documents it. No factory run was performed.

## A scope note you may want to act on

Closing these reshaped `serve.py` more than a review pass usually implies: a new `ServeState`/`ServeStateError` layer replacing the naive ledger, `run_supervised` extracted, `Queue.adopt_lease` added, and the budget contract restated. The diff you review now is meaningfully different from the one you reviewed before.

I'd argue against splitting F4/F5/F8/F9 apart — they are all consequences of the ledger holding the wrong things, and separating them would mean landing an inconsistent ledger twice. But if you'd rather see the accounting rework as its own PR on top of a minimal fix set, say so and I'll split it.

Still open from the original PR body: the **R8.2 merge-gate tension** (at L3+ the ladder overrides an explicit `stop_at_pr`; `resolve_merge_gate` refuses rather than letting the gate vanish). Unreachable today, but it needs an R8.2 decision before L3 is real.

PR #187, opened 2026-07-30
R8.6 PR 3/4: GitHub Issues as the remote inbox

0xfauzi:
Third of four slices for [#153](https://github.com/0xfauzi/kstrl/issues/153).

> **Stacked on #186.** Base is `feat/r8-6-serve`, not `main`, because the terminal writeback hooks into `serve`'s poison/done paths. Rebases onto `main` cleanly once #186 merges.

| PR | Contents | Status |
|---|---|---|
| 1 | queue substrate + `ks queue` | merged (#185) |
| 2 | `ks serve`, classifier, backstops | open (#186) |
| **3** | **GitHub Issues adapter, processed-ids ledger, `ks queue sync`** | **this PR** |
| 4 | launchd plist + docs | next |

## What authorizes work

**The trigger is the LABEL, not the issue.** Applying a label requires write access to the repository, so on a public repo a stranger can open an issue but cannot queue a factory run. That property is the entire access-control story for this adapter, and it's why it watches a label rather than a title prefix or a mention.

**Remote items never auto-merge.** `STOP_AT_PR` is forced regardless of labels or config — an issue label is the last place a merge decision should be settable from.

## Strictly additive

Every `gh` call returns a `GhResult` rather than raising, so a missing binary, timeout, auth failure, and rate limit are all survivable: a GitHub outage produces an empty sync, not a stalled queue. `serve`'s writeback is wrapped so a broken front-end cannot alter or roll back a local transition that already committed.

## Idempotency has two halves

`Queue.find_by_source_ref` covers items still in the queue; the processed-ids ledger covers items that have **already left it**. Without the ledger, a completed issue would be re-enqueued on the very next poll, forever.

The ledger reads a corrupt file as **empty** — deliberately, and unlike PR 2's spend ledger where failing open disabled the only queue-wide cap. Here the blast radius genuinely is bounded by `find_by_source_ref`, the per-sync cap, and the daily budget. There is a test that pins exactly that claim by deleting the ledger and asserting no duplicate.

Other bounds: `max_items_per_sync` (a label applied to fifty issues must not queue fifty runs); a spec size cap with truncation **announced in the spec**; an empty issue body skipped without paying an architect call to learn it says nothing; and one state label at a time, so an issue can never carry `kstrl:running` and `kstrl:done` together.

## Live round-trip — the "done when" gate

Against `0xfauzi/claude-skills`, per your instruction:

| Step | Result |
|---|---|
| `kstrl:*` labels created | 5 labels |
| `ks queue sync --dry-run` | `polled: 1`, wrote nothing, label unchanged |
| `ks queue sync` | `polled: 1`, `enqueued: 1` |
| Label writeback | `kstrl:queued` → `kstrl:running` |
| Item provenance | `source: github (0xfauzi/claude-skills#1)`, `merge: stop_at_pr` |
| Re-sync | `enqueued: 0`, `skipped: already processed`, item count still 1 |
| Terminal writeback | exactly one label (`kstrl:poison`) + comment with recovery instructions |

Fixture issue closed afterwards; temp queue root removed. **No factory run was performed.** The `kstrl:*` labels were left in place (they're the adapter's operational contract) — delete them if you don't want intake wired to that repo.

## Tested vs assumed (H4)

**Tested.** 65 new tests, all with `gh` stubbed — a suite that hit the API would be slow, rate-limited, and would post public comments on every run. Mutation check **17/17**.

One mutation initially survived: removing the in-queue dedupe check. It's redundant with the ledger *unless the ledger is lost* — which is precisely the documented fail-open case. The test now deletes the ledger to pin it, so the justification in the docstring is actually enforced.

**Measured, and it contradicts the plan.** This does **not** use ETag-conditional requests: `gh issue list` exposes no ETag, so the saving R8.6 attributed to ETags is not realised. It isn't needed at this cadence — one call per poll interval is ~60/hour against 5,000/hour. Recorded in the roadmap rather than left as an implied capability.

**Also measured.** GitHub's issue-list endpoint can lag a label write. A sync issued immediately after labelling returned `polled: 0`; the same sync a minute later returned `polled: 1`. Confirmed transient by re-running rather than assumed. Invisible at any realistic poll interval, but a test that labels and syncs in one breath will look flaky — noted in the roadmap.

**Assumed, not tested.** `gh` version-specific output shapes beyond the fields parsed here. Rate-limit behaviour under sustained polling was not exercised.

## Prompt injection

Handled where it already was: `decompose` wraps the spec between per-run random delimiters as untrusted data (R5.3). This module deliberately does **not** pattern-match issue bodies — a spec legitimately discussing prompts would be rejected, and rejecting work is not the same as containing it. What it adds is a size cap.

## Verification

```
uv run pytest tests/     2895 -> 2960 passed, 28 skipped
```

`[intake_github]` registered in `scripts/gen_docs.py`; README regenerated by the generator. The `repo` field needed an `ENUM_SENTINELS` entry because it validates `owner/name` and the generic sentinel fails it — used the existing mechanism rather than weakening the validation.

### DIFF
--- a/kstrl/cli.py
+++ b/kstrl/cli.py
@@ -3976,6 +3976,94 @@ def queue_resume(root: Path | None, ui: str, no_color: bool) -> None:
     sys.exit(0)
 
 
+@queue_group.command(name="sync")
+@click.option(
+    "--dry-run", is_flag=True,
+    help="Poll and report what would be enqueued, writing nothing",
+)
+@_queue_root_option
+@_queue_ui_option
+@_queue_no_color_option
+def queue_sync(
+    dry_run: bool, root: Path | None, ui: str, no_color: bool,
+) -> None:
+    """Pull labelled GitHub issues into the queue (R8.6).
+
+    Polls open issues carrying the trigger label and enqueues the ones
+    not already seen. Remote items ALWAYS stop at the PR for a human.
+
+    The label is the authorization: applying it needs write access to the
+    repository, so an issue from a stranger cannot queue a factory run.
+    """
+    from kstrl.intake_github import (
+        GitHubIntakeConfig,
+        IntakeError,
+        ProcessedLedger,
+        poll_queued,
+        resolve_repo,
+    )
+    from kstrl.intake_github import sync as run_sync
+    from kstrl.workqueue import queue_lock
+
+    root_dir, queue = _queue_for(root)
+    ui_impl = _autonomy_ui(ui, no_color)
+    try:
+        config = GitHubIntakeConfig.load(root_dir)
+    except (IntakeError, ValueError) as exc:
+        ui_impl.err(str(exc))
+        sys.exit(2)
+    if not config.enabled:
+        ui_impl.err(
+            "GitHub intake is off. Set [intake_github] enabled = true in "
+            "kstrl.toml (or KSTRL_INTAKE_GITHUB_ENABLED=1)."
+        )
+        sys.exit(1)
+
+    if dry_run:
+        repo, error = resolve_repo(config, root_dir)
+        if error:
+            ui_impl.err(error)
+            sys.exit(1)
+        issues, poll_error = poll_queued(config, repo, root_dir)
+        if poll_error:
+            ui_impl.err(poll_error)
+            sys.exit(1)
+        ledger = ProcessedLedger(root_dir).load()
+        ui_impl.section(f"Would sync from {repo}")
+        ui_impl.kv("label", config.queued_label)
+        ui_impl.kv("polled", str(len(issues)))
+        if not issues:
+            ui_impl.info("  nothing labelled")
+        for issue in issues:
+            ref = issue.source_ref(repo)
+            if ledger.contains(ref):
+                verdict = "skip (already processed)"
+            elif queue.find_by_source_ref(ref) is not None:
+                verdict = "skip (already queued)"
+            elif not issue.body.strip():
+                verdict = "skip (empty body)"
+            else:
+                verdict = "ENQUEUE"
+            ui_impl.info(f"  {ref:<28} {verdict}  {issue.title}")
+        sys.exit(0)
+
+    with queue_lock(root_dir):
+        result = run_sync(queue, config, root_dir)
+
+    ui_impl.section(f"Sync from {result.repo or 'unknown repo'}")
+    ui_impl.kv("polled", str(result.polled))
+    ui_impl.kv("enqueued", str(len(result.enqueued)))
+    for ref in result.enqueued:
+        ui_impl.ok(f"  queued {ref}")
+    for ref, reason in sorted(result.skipped.items()):
+        ui_impl.info(f"  skipped {ref}: {reason}")
+    for error in result.errors:
+        ui_impl.err(f"  {error}")
+    # Nonzero on error so a cron/launchd wrapper notices, but note that
+    # partial success is real: items already enqueued stay enqueued.
+    sys.exit(1 if result.errors else 0)
+
+
 @cli.command()
 @click.option(
     "--once", is_flag=True,
--- a/kstrl/cli.py
+++ b/kstrl/cli.py
@@ -16,6 +16,8 @@ if TYPE_CHECKING:
     from kstrl.evolution import EvolutionConfig
     from kstrl.interaction import InteractionChannel
 
+from dataclasses import replace
+
 import click
 from click.core import ParameterSource
 
@@ -3995,15 +3997,9 @@ def queue_sync(
     The label is the authorization: applying it needs write access to the
     repository, so an issue from a stranger cannot queue a factory run.
     """
-    from kstrl.intake_github import (
-        GitHubIntakeConfig,
-        IntakeError,
-        ProcessedLedger,
-        poll_queued,
-        resolve_repo,
-    )
+    from kstrl.intake_github import GitHubIntakeConfig, IntakeError
     from kstrl.intake_github import sync as run_sync
-    from kstrl.workqueue import queue_lock
+    from kstrl.workqueue import QueueError, QueueLockedError, queue_lock
 
     root_dir, queue = _queue_for(root)
     ui_impl = _autonomy_ui(ui, no_color)
@@ -4019,48 +4015,51 @@ def queue_sync(
         )
         sys.exit(1)
 
+    # --dry-run runs the PRODUCTION planner with writes disabled, rather
+    # than a second decision tree. Review #187 F4/F11: the old dry-run
+    # had its own logic that ignored the admission cap, and a dry run
+    # that disagrees with the real thing is worse than none.
     if dry_run:
-        repo, error = resolve_repo(config, root_dir)
-        if error:
-            ui_impl.err(error)
-            sys.exit(1)
-        issues, poll_error = poll_queued(config, repo, root_dir)
-        if poll_error:
-            ui_impl.err(poll_error)
-            sys.exit(1)
-        ledger = ProcessedLedger(root_dir).load()
-        ui_impl.section(f"Would sync from {repo}")
-        ui_impl.kv("label", config.queued_label)
-        ui_impl.kv("polled", str(len(issues)))
-        if not issues:
-            ui_impl.info("  nothing labelled")
-        for issue in issues:
-            ref = issue.source_ref(repo)
-            if ledger.contains(ref):
-                verdict = "skip (already processed)"
-            elif queue.find_by_source_ref(ref) is not None:
-                verdict = "skip (already queued)"
-            elif not issue.body.strip():
-                verdict = "skip (empty body)"
-            else:
-                verdict = "ENQUEUE"
-            ui_impl.info(f"  {ref:<28} {verdict}  {issue.title}")
-        sys.exit(0)
+        config = replace(config, dry_run=True)
 
-    with queue_lock(root_dir):
-        result = run_sync(queue, config, root_dir)
+    try:
+        with queue_lock(root_dir):
+            result = run_sync(queue, config, root_dir)
+    except QueueLockedError as exc:
+        ui_impl.err(
+            f"{exc}. Another queue operation is in progress; retry shortly."
+        )
+        sys.exit(1)
+    except (QueueError, OSError) as exc:
+        ui_impl.err(f"Sync failed: {exc}")
+        sys.exit(1)
 
-    ui_impl.section(f"Sync from {result.repo or 'unknown repo'}")
+    heading = "Would sync from" if dry_run else "Sync from"
+    ui_impl.section(f"{heading} {result.repo or 'unknown repo'}")
+    ui_impl.kv("label", config.queued_label)
     ui_impl.kv("polled", str(result.polled))
-    ui_impl.kv("enqueued", str(len(result.enqueued)))
-    for ref in result.enqueued:
-        ui_impl.ok(f"  queued {ref}")
-    for ref, reason in sorted(result.skipped.items()):
-        ui_impl.info(f"  skipped {ref}: {reason}")
+    if dry_run:
+        ui_impl.kv("would enqueue", str(len(result.would_enqueue)))
+    else:
+        ui_impl.kv("enqueued", str(len(result.enqueued)))
+
+    for entry in result.planned:
+        ref = entry.issue.source_ref(result.repo)
+        verdict = "ENQUEUE" if entry.decision.admits else str(entry.decision)
+        line = f"  {ref:<28} {verdict:<22} {entry.issue.title}"
+        if entry.decision.admits:
+            ui_impl.ok(line)
+        elif entry.decision is entry.decision.__class__.REFUSE_UNAUTHORIZED:
+            ui_impl.err(f"{line}\n      {entry.reason}")
+        else:
+            ui_impl.info(f"{line}  ({entry.reason})")
+    if not result.planned:
+        ui_impl.info("  nothing labelled")
+
     for error in result.errors:
         ui_impl.err(f"  {error}")
-    # Nonzero on error so a cron/launchd wrapper notices, but note that
-    # partial success is real: items already enqueued stay enqueued.
+    # Nonzero on error so a cron/launchd wrapper notices. Partial success
+    # is real: items already enqueued stay enqueued.
     sys.exit(1 if result.errors else 0)
 
 
--- a/kstrl/intake_github.py
+++ b/kstrl/intake_github.py
@@ -4,12 +4,28 @@ An issue labelled ``kstrl:queued`` becomes a queue item; the queue's
 verdict comes back as a state label and a comment. Polling only - no
 webhooks, no public endpoint, nothing to keep reachable.
 
-**What authorizes work.** The trigger is the LABEL, not the issue. Adding
-a label to a repository requires write access, so on a public repo a
-stranger can open an issue but cannot queue a factory run against it.
-That property is the whole access-control story for this adapter, and it
-is the reason the adapter watches a label rather than, say, a title
-prefix or a mention.
+**What authorizes work, stated accurately.** The trigger is the LABEL,
+not the issue: a stranger can open an issue but cannot label it. What
+that boundary actually is, however, is narrower than it first appears,
+and the first version of this module overclaimed it:
+
+- Applying a label needs the **Triage** role or above, NOT write/push
+  access. On an organization repository a triager who cannot push a line
+  of code can still authorize factory spend (review #187 F2).
+- Any GitHub Action in the repo with ``issues: write`` can apply the
+  label, so a workflow can trigger spend with no human involved.
+
+So this is a permission designed for *managing issues*, borrowed to
+authorize *money*. Issue #188 replaces it with an explicit actor
+allowlist, which is what makes the authorization this project's own
+decision rather than an inherited one. Until then the residual risk is
+exactly the two bullets above, bounded by the adapter being off by
+default.
+
+What IS enforced here is that the authorized bytes are the bytes that
+run: an issue edited after it was labelled is refused, because GitHub
+lets an issue author rewrite the body after a maintainer labelled it
+(review #187 F1).
 
 **Strictly additive.** A front-end outage must never block the local
 queue (R8.6). Every ``gh`` call therefore returns a result object instead
@@ -41,8 +57,9 @@ from __future__ import annotations
 import json
 import os
 import subprocess
-from collections.abc import Callable
+from collections.abc import Callable, Sequence
 from dataclasses import dataclass, field
+from enum import StrEnum
 from pathlib import Path
 from typing import Any
 
@@ -61,6 +78,15 @@ GITHUB_LEDGER_FILENAME = "github_processed.json"
 #: bounded enough that a pathological body cannot become a pathological
 #: prompt. Truncation is announced in the spec itself, never silent.
 MAX_SPEC_CHARS = 60_000
+
+#: Poll paging. The window GROWS until the admission cap can be filled or
+#: the inbox is exhausted, because skipped issues do not consume the cap
+#: (review #187 F6).
+POLL_PAGE_SIZE = 30
+MAX_POLL_PAGES = 4
+MAX_POLL_LIMIT = 200
+#: Breadth to gather relative to the cap before stopping.
+POLL_OVERSCAN = 4
 
 
 class IntakeError(RuntimeError):
@@ -285,18 +311,24 @@ class RemoteIssue:
         return f"{repo}#{self.number}"
 
 
-def parse_issue_list(payload: str) -> list[RemoteIssue]:
-    """Decode ``gh issue list --json``, skipping anything malformed.
+def parse_issue_list(payload: str) -> tuple[list[RemoteIssue], str]:
+    """Decode ``gh issue list --json``; returns ``(issues, error)``.
 
-    Tolerant by design: one unparseable entry must not discard the whole
-    poll, because the queue would then stall on a single bad issue.
+    Tolerant PER ENTRY - one unparseable issue must not discard the whole
+    poll, because the queue would then stall on a single bad issue - but
+    STRICT about the top level. Review #187 F7: a malformed payload used
+    to collapse to the same empty list as a healthy ``[]``, so a `gh`
+    output-shape change or a truncated response looked like a successful
+    empty poll and no cron or launchd wrapper could alert on it.
     """
     try:
         data = json.loads(payload or "[]")
-    except json.JSONDecodeError:
-        return []
+    except json.JSONDecodeError as exc:
+        return [], f"could not parse `gh issue list` output: {exc}"
     if not isinstance(data, list):
-        return []
+        return [], (
+            f"`gh issue list` returned {type(data).__name__}, expected a list"
+        )
     issues: list[RemoteIssue] = []
     for entry in data:
         if not isinstance(entry, dict):
@@ -320,9 +352,11 @@ def parse_issue_list(payload: str) -> list[RemoteIssue]:
             labels=labels,
         ))
     # Oldest first: FIFO across the remote inbox, matching the queue's own
-    # ordering within a priority band.
+    # ordering within a priority band. The poll ALSO asks GitHub to sort
+    # ascending - sorting a truncated page cannot establish FIFO on its
+    # own (review #187 F6).
     issues.sort(key=lambda issue: issue.number)
-    return issues
+    return issues, ""
 
 
 def resolve_repo(config: GitHubIntakeConfig, root_dir: Path) -> tuple[str, str]:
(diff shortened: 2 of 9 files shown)


0xfauzi (commented):
Blocking review: 6 P1 findings and 6 P2 findings. The focused adapter tests, full suite, strict mypy, ruff, generated-doc check, and GitHub CI all pass, but adversarial checks exposed authorization, dry-run, atomicity, state-reconciliation, and availability defects. Because this review is being submitted from the PR owner's account, GitHub does not permit a formal request-changes review; the P1 comments below are merge blockers.

kstrl/intake_github.py
```diff
@@ -0,0 +1,705 @@
...
+    against a 5,000/hour budget.
+    """
+    result = run_gh(
+        [
+            "issue", "list",
+            "--repo", repo,
+            "--label", config.queued_label,
+            "--state", "open",
+            "--limit", str(max(config.max_items_per_sync * 2, 10)),
+            "--json", "number,title,body,url,labels",
```
0xfauzi, line 430: [P1] Bind authorization to an immutable issue revision

This fetch trusts the current title and body solely because the trigger label is present. A public issue author can submit a benign body, wait for a maintainer to label it, and then edit the body before this poll; GitHub explicitly permits issue authors to edit their descriptions. The label would then authorize different bytes from the ones the maintainer reviewed, and those bytes become factory input. Fetch and verify the label event plus a content revision, reject content changed after authorization, or have the trusted actor create an immutable/signed snapshot that intake consumes. See [GitHub's issue editing rules](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/editing-an-issue).

kstrl/cli.py
```diff
@@ -3976,6 +3976,94 @@ def queue_resume(root: Path | None, ui: str, no_color: bool) -> None:
...
+def queue_sync(
+    dry_run: bool, root: Path | None, ui: str, no_color: bool,
+) -> None:
+    """Pull labelled GitHub issues into the queue (R8.6).
+
+    Polls open issues carrying the trigger label and enqueues the ones
+    not already seen. Remote items ALWAYS stop at the PR for a human.
+
+    The label is the authorization: applying it needs write access to the
+    repository, so an issue from a stranger cannot queue a factory run.
```
0xfauzi, line 3998: [P1] Do not equate label permission with repository write permission

The security premise in this command is false for organization repositories: GitHub's built-in Triage role can apply and dismiss labels but cannot push to the repository. That gives a non-code contributor the ability to trigger factory spend and code generation under credentials that may be able to push. Resolve the actor who applied the trigger and verify an explicit acceptable permission or allowlist, or explicitly define Triage as authorized and revise the threat model. See the [GitHub repository role matrix](https://docs.github.com/en/organizations/managing-user-access-to-your-organizations-repositories/managing-repository-roles/repository-roles-for-an-organization).

kstrl/intake_github.py
```diff
@@ -0,0 +1,705 @@
...
+        try:
+            item = queue.add(
+                spec_from_issue(issue, repo),
+                title=issue.title,
+                priority=config.default_priority,
+                # FORCED, never configurable from the remote side.
+                merge_disposition=MergeDisposition.STOP_AT_PR,
+                source=ItemSource.GITHUB,
+                source_ref=ref,
+                target_repo=repo,
```
0xfauzi, line 993: [P1] Refuse source repositories that do not match the execution root

`target_repo` is only stored as metadata. `serve_cycle` always invokes the factory with its own `root_dir`, and no code routes execution using this field. Therefore an explicit `[intake_github].repo = "B"` in checkout A admits B's issue and executes it against A, potentially producing changes and a PR in the wrong repository. Validate that the resolved inbox repository matches the checkout remote before admission, or implement verified checkout routing for `target_repo` before accepting cross-repository intake.

kstrl/intake_github.py
```diff
@@ -0,0 +1,705 @@
...
+            result.skipped[ref] = "already in the queue"
+            continue
+        if not issue.body.strip():
+            # An empty body would produce a spec with nothing but a
+            # provenance header. Skip loudly rather than paying an
+            # architect call to discover it says nothing.
+            result.skipped[ref] = "issue body is empty; nothing to build"
+            continue
+
+        try:
```
0xfauzi, line 984: [P1] Make configured dry-run side-effect free

`GitHubIntakeConfig(dry_run=True)` only suppresses the later `gh` writes; execution still reaches `queue.add` and `ledger.record`. A focused reproduction returned `enqueued=('o/r#1',)`, left one queue item, and created `github_processed.json`. With `ks serve` active, a supposed dry run can therefore launch paid work. The generated config says dry-run should poll and log, and the PR reports that dry-run wrote nothing. Branch before all local mutations, ideally by sharing one side-effect-free planning path with the CLI's `--dry-run` implementation.

kstrl/intake_github.py
```diff
@@ -0,0 +1,705 @@
...
+                source_ref=ref,
+                target_repo=repo,
+                spec_filename="spec.md",
+                actor="intake-github",
+            )
+        except (QueueError, OSError) as exc:
+            errors.append(f"{ref}: could not enqueue ({exc})")
+            continue
+
+        ledger.record(ref, item_id=item.item_id, when=stamp)
```
0xfauzi: [P1] Do not publish an item before the fallible ledger commit

`queue.add` has already atomically published the item when `ledger.record` can raise `OSError`. Patching `record` to raise `OSError("disk full")` made `sync` escape with that exception while `o/r#1` remained in `queued/`. The batch aborts, the remote trigger remains stale, and there is no durable processed entry; after removal plus a failed terminal writeback the issue can be admitted again. Make admission and dedupe recording transactional, or roll back the newly created queued item when the ledger cannot be committed and return a structured error.

kstrl/intake_github.py
```diff
@@ -0,0 +1,705 @@
...
+    not needed at this cadence: one call per poll interval is ~60/hour
+    against a 5,000/hour budget.
+    """
+    result = run_gh(
+        [
+            "issue", "list",
+            "--repo", repo,
+            "--label", config.queued_label,
+            "--state", "open",
+            "--limit", str(max(config.max_items_per_sync * 2, 10)),
```
0xfauzi: [P2] Scan until the admission cap is filled or the inbox is exhausted

This fixed `2 * max_items_per_sync` window can be consumed entirely by already-processed, already-queued, or empty issues, none of which count toward the admission cap. Those skip paths also leave a stale trigger label in place, so an eligible issue outside the window can remain invisible indefinitely. In addition, `gh issue list` does not request oldest-first ordering; sorting only this truncated page does not establish FIFO. Request `sort:created-asc`, paginate until the new-item cap is filled or no results remain, and reconcile stale trigger labels for known items.

kstrl/intake_github.py
```diff
@@ -0,0 +1,705 @@
...
+
+def parse_issue_list(payload: str) -> list[RemoteIssue]:
+    """Decode ``gh issue list --json``, skipping anything malformed.
+
+    Tolerant by design: one unparseable entry must not discard the whole
+    poll, because the queue would then stall on a single bad issue.
+    """
+    try:
+        data = json.loads(payload or "[]")
+    except json.JSONDecodeError:
```
0xfauzi: [P2] Report malformed poll output as an error

A top-level JSON decode failure is collapsed to the same empty list as a valid `[]`. In a focused reproduction, `GhResult(ok=True, stdout="{bad")` made `sync` report `ok=True`, `polled=0`, and no errors. A `gh` output-shape change or truncated response will therefore look like a successful empty poll, so launchd or cron cannot alert. Preserve tolerant per-entry parsing after a valid list, but return a parse error from `poll_queued` for an invalid top-level payload.

kstrl/intake_github.py
```diff
@@ -0,0 +1,705 @@
...
+                actor="intake-github",
+            )
+        except (QueueError, OSError) as exc:
+            errors.append(f"{ref}: could not enqueue ({exc})")
+            continue
+
+        ledger.record(ref, item_id=item.item_id, when=stamp)
+        enqueued.append(ref)
+        label_error = apply_state_label(
+            config, repo, issue.number, "running", root_dir,
```
0xfauzi: [P2] Reflect the queue's actual state instead of marking admission as running

At this point `queue.add` has placed the item in `QUEUED`, yet the adapter removes the trigger and labels it `running`. A paused or backlogged item can remain visibly running for hours without a process, and the `failed` label is never used while a retry waits in the queue. Keep the queued label until `queue.start`, then drive remote labels from actual queue transitions, including retry and terminal transitions.

kstrl/serve.py
```diff
@@ -1572,6 +1573,38 @@ def _file_inbox_item(
...
+    observer: ServeObserver,
+) -> None:
+    """Tell the source front-end what happened, best effort.
+
+    R8.6 requires the adapters be strictly additive, so this swallows
+    every failure into a warning: the queue transition it describes has
+    already committed locally, and a GitHub outage must not be able to
+    roll it back or halt the daemon.
+    """
+    if item is None or item.source is not ItemSource.GITHUB:
```
0xfauzi, line 1591: [P1] Cover every terminal poison transition

The new helper is called only from the normal success and normal final-verdict paths. Existing terminal exits for reaper exhaustion, merge-gate refusal, `QueueBudgetExhausted`, and an unreaped timeout all call `queue.poison(...)` and return without invoking it. A focused merge-gate reproduction ended with the remote item in local `poison` and `_report_remote_outcome` called zero times, leaving the issue labelled `running` forever. Centralize a post-transition hook, or explicitly invoke writeback after every committed terminal transition.

kstrl/serve.py
```diff
@@ -1907,6 +1940,10 @@ def _adopt(child_pid: int) -> None:
             queue.finish_ok(current, actor="serve")
             ledger.record_terminal(poisoned=False)
             obs.info(f"  {running.item_id[:12]} done")
+            _report_remote_outcome(
```
0xfauzi: [P2] Release the queue mutex before remote network I/O

This success-path writeback is still inside `with queue_lock(...)`. `report_outcome` can perform two `gh` calls, each with the configured timeout, so an unavailable GitHub service blocks every local queue transition while the adapter waits. A focused callback attempting `queue_lock` here observed that the lock was held. Capture the finished item under the mutex, exit the critical section, and perform best-effort label/comment I/O afterward, as the poison path already does.

kstrl/cli.py
```diff
@@ -3976,6 +3976,94 @@ def queue_resume(root: Path | None, ui: str, no_color: bool) -> None:
...
+        for issue in issues:
+            ref = issue.source_ref(repo)
+            if ledger.contains(ref):
+                verdict = "skip (already processed)"
+            elif queue.find_by_source_ref(ref) is not None:
+                verdict = "skip (already queued)"
+            elif not issue.body.strip():
+                verdict = "skip (empty body)"
+            else:
+                verdict = "ENQUEUE"
```
0xfauzi: [P2] Apply the admission cap in dry-run output

The dry-run planner marks every otherwise eligible result as `ENQUEUE` and never counts `max_items_per_sync`. With a cap of 5 and ten returned issues, a focused CLI invocation printed ten `ENQUEUE` verdicts even though real `sync` would admit five. Since the option promises to report what would be enqueued, count planned admissions and mark the remainder as skipped by the cap, preferably by sharing the production planner rather than maintaining a second decision tree.

kstrl/cli.py
```diff
@@ -3976,6 +3976,94 @@ def queue_resume(root: Path | None, ui: str, no_color: bool) -> None:
...
+            ui_impl.info(f"  {ref:<28} {verdict}  {issue.title}")
+        sys.exit(0)
+
+    with queue_lock(root_dir):
```
0xfauzi: [P2] Convert queue-lock contention into a normal CLI error

This nonblocking mutex raises `QueueLockedError`, but unlike `queue_add` this command has no surrounding `except QueueError`. A concurrent transition therefore produces an uncaught exception and traceback instead of an actionable retry message. Wrap lock acquisition and `run_sync` in the same operator-facing queue error handling used by the other queue commands.

0xfauzi:
## Two design decisions recorded — **this PR's code is unchanged**

Taken after a first-principles pass on the label trigger. Both deferred, so the diff under review is the same one.

**1. Actor allowlist → #188, after this merges.** The adapter treats "the label is present" as authorization to spend money, and that authorization is *inherited* from GitHub write access — a permission that means "trusted with the codebase," not "may spend $30 unattended." The path worth closing: **any Action in the repo with `issues: write` can apply the label**, triggering spend with no human involved. Deferred safely because the adapter defaults to off.

**2. Linear is a priority/intent feed, not a second trigger** (recorded on #153). Linear is write-only in kstrl today and already derives status *from* GitHub for free, so intake-on-Linear would invert a deliberate design and pay for what it currently gets free.

## One thing I'd correct in my own PR description

I described the label-requires-write-access property as "the entire access-control story," and presented that as a strength. It is accurate but it is not *sufficient* — it's a borrowed boundary, and I should have said so rather than framing an inherited permission as a designed one. #188 is what makes it designed.

Nothing about that changes the code here; it changes how confidently the security property should be read.

kstrl/intake_github.py
0xfauzi: Confirmed and fixed in b3036b1. This was the sharpest finding on the PR and I had not considered that the label is a point-in-time act while the body is mutable.

`verify_authorization` now makes one GraphQL call per candidate that fetches the trigger label's `createdAt` (plus its actor) and the body's `lastEditedAt` together, and refuses when the edit is newer than the label - with "re-apply the label to authorize the current text" so the recovery is obvious.

It fails CLOSED on every uncertainty, which I think is the part that matters: an unreadable response, an unparseable timestamp, no issue node, no trigger-label event at all, or a label event that is only a STATE label. "We could not check" is not evidence that the bytes are the authorized ones.

Two notes:
- Only the *most recent* trigger-label event counts, so re-labelling after an edit re-authorizes deliberately. That is the intended recovery path rather than an escape hatch.
- The actor is captured on the `Authorization` and surfaced in the refusal reason. #188 turns it into an allowlist decision; I did not half-implement that here.

kstrl/cli.py
0xfauzi: Confirmed and fixed in b3036b1 - and you are right that the premise was simply false, not merely incomplete. Triage can label without push access, and I had also missed that any Action with `issues: write` can label, so a workflow can trigger spend with no human at all.

I did not silently redefine Triage as authorized. I corrected the claim everywhere it appeared - module docstring, `kstrl.toml.example`, the CLI help, and the PR body - and stated the residual risk as the two concrete paths rather than a hedge. #188 is open to replace the inherited permission with an explicit actor allowlist, which the user chose to land after this merges since the adapter defaults to off.

Worth saying plainly: I had described this property as "the entire access-control story" and presented it as a strength. It was a borrowed boundary that I framed as a designed one.

kstrl/intake_github.py
0xfauzi: Confirmed and fixed in b3036b1. You are right that `target_repo` is inert - nothing routes execution by it, and `serve_cycle` always invokes the factory with its own `root_dir`.

I took the "validate before admission" option rather than implementing checkout routing, because routing is a much larger feature and admitting work you cannot correctly execute is the actual defect. `sync` now resolves the CHECKOUT's repo separately from the configured inbox and refuses when they differ, naming both in the error and saying what to do. Case-insensitive, since GitHub owners/names are.

An unresolvable checkout also refuses rather than proceeding: not being able to confirm where a run would execute is not a reason to run it.

kstrl/intake_github.py
0xfauzi: Confirmed and fixed in b3036b1. Your reproduction is exact, and this one embarrassed me twice over: the generated config says dry-run polls and logs, and the PR body claimed it wrote nothing.

The root cause was two decision trees, which is also your F11. There is now one side-effect-free `plan_sync` that returns a `Decision` per issue, and:

- `sync` branches on `config.dry_run` BEFORE any local mutation, returning the plan plus `would_enqueue`;
- the CLI's `--dry-run` forces `config.dry_run=True` and renders `result.planned` directly.

So the dry run and the real thing cannot disagree, because there is only one implementation. Tests assert no queue item, no ledger file, no `issue edit`, and no `issue comment`.

kstrl/intake_github.py
0xfauzi: Confirmed and fixed in b3036b1. Reproduced exactly: `sync` escaped with `OSError("disk full")` while the item stayed in `queued/` with no processed entry.

Admission and dedupe now commit together: if `ledger.record` raises, the queued item is rolled back with `queue.remove` and the failure is returned as a structured per-item error rather than propagating. Two properties I added tests for:

- the batch is NOT aborted - other planned items still get their chance, and each failure is reported separately;
- if the rollback ITSELF fails, the error says so and names the item to remove by hand, rather than reporting a clean failure over a dirty queue.

I chose rollback over a write-ahead record because the ledger is advisory dedupe, not authorization: a spurious processed entry would silently drop real work, which is worse than a rolled-back admission.

kstrl/intake_github.py
0xfauzi: Confirmed and fixed in b3036b1 - and this one corrected my architecture, not just a constant.

My first attempt paginated inside `poll_queued` using a breadth heuristic (gather `4 x cap` then stop). That was structurally wrong for exactly the reason you gave: whether a wider window is needed depends on how many results are ELIGIBLE, and only the planner knows that. My own test caught it - a full page of already-processed issues satisfied the breadth check and admitted nothing.

The loop now lives in `sync`, where planning happens: poll at a widening limit, plan, and stop when the cap is filled or the page came back short (`exhausted`). Authorization is memoized per sync so re-planning costs no extra API calls. `sort:created-asc` is requested explicitly, since sorting one truncated page cannot establish FIFO.

On stale trigger labels for known items: I did not add reconciliation. Skipped issues keep their trigger label, which is visible in `ks queue sync --dry-run` output with the reason. Sweeping labels for items the queue already knows about is a separate behaviour I would rather propose than slip in - say the word and I will open an issue for it.

kstrl/intake_github.py
0xfauzi: Confirmed and fixed in b3036b1. Reproduced: `stdout="{bad"` gave `ok=True, polled=0, errors=()`.

`parse_issue_list` now returns `(issues, error)`: strict at the top level (a decode failure or a non-list payload is an error), tolerant per entry (one unparseable issue still does not discard the poll). `poll_queued` propagates it, so `sync` reports a real error and a cron or launchd wrapper gets a nonzero exit.

Both halves are pinned by tests, including that a genuine `[]` is still not an error and that per-entry tolerance survives.

kstrl/intake_github.py
0xfauzi: Confirmed and fixed in b3036b1. You are right that admission labelling the issue `running` was simply inaccurate - the item is QUEUED at that point.

Admission no longer relabels at all. `serve` drives the remote label from real transitions:

- `running` when `queue.start` actually charges the attempt;
- `failed` when a retry is requeued, with the backoff and attempt count in the comment - this is the case you noted was never used;
- `done` / `poison` at the terminal transitions.

The trigger label therefore survives until the run starts, which also means a paused or backlogged item does not read as running. Tests assert the exact state sequence for success, retry, and poison.

kstrl/serve.py
0xfauzi: Confirmed and fixed in b3036b1. Your merge-gate reproduction is exact: the item ended in local `poison` with zero writeback calls, leaving the issue labelled `running` forever.

All four terminal exits now report: reaper exhaustion, merge-gate refusal, `QueueBudgetExhausted`, and the unreaped-timeout poison, plus the normal verdict paths. Each has a named test.

I did not add a general post-transition hook. Centralising it properly means the writeback needs the committed item and the reason, and those differ per path - a hook that reconstructed them would be a second source of truth for what happened. Explicit calls after each committed transition seemed more honest, and the mutation check covers all four independently so a future path that forgets one will not pass silently. If you would rather have the hook, I am happy to do it.

kstrl/serve.py
0xfauzi: Confirmed and fixed in b3036b1. The success path was the last writeback still inside the mutex - the poison path already ran outside it, which made the inconsistency worse rather than better.

`serve_cycle` now captures the committed item inside the critical section, exits, and does the label/comment I/O afterwards. That applies to all five paths, not just this one.

The test asserts it behaviourally rather than structurally: the writeback callback tries to take `queue_lock` non-blocking and the test fails if it is refused. A structural check on indentation would have passed the moment the code moved.

kstrl/cli.py
0xfauzi: Confirmed and fixed in b3036b1. Reproduced: ten `ENQUEUE` verdicts against a cap of five.

Fixed by sharing the production planner, as you suggested, rather than teaching the second tree to count. `--dry-run` sets `config.dry_run=True` and renders `result.planned`, so the cap, the skip reasons, and the authorization refusals are all whatever `sync` actually decided. The test asserts the dry run reports exactly three ENQUEUE with a cap of three AND that the subsequent real sync admits exactly three.

kstrl/cli.py
0xfauzi: Confirmed and fixed in b3036b1. `queue sync` now wraps lock acquisition and the sync in the same handling the other queue commands use: `QueueLockedError` becomes "Another queue operation is in progress; retry shortly." with exit 1, and `QueueError`/`OSError` become a reported failure rather than a traceback.

The test asserts no exception escapes the CliRunner, not merely that the exit code is nonzero - a traceback would also have produced nonzero.

0xfauzi:
## All twelve findings closed in `b3036b1`

Every one reproduced against the submitted code first. **Two invalidated claims this PR itself made**, which is the part worth reading before the table.

| # | Finding | Fix |
|---|---|---|
| F1 | P1 label authorized bytes the author could then rewrite | GraphQL binds label time to `lastEditedAt`; edit-after-label refused |
| F2 | P1 **"requires write access" was false** — Triage suffices | claim corrected everywhere; residual risk named; #188 opened |
| F3 | P1 cross-repo intake would PR into the wrong repo | inbox must match the checkout; unresolvable checkout refuses |
| F4 | P1 `dry_run` still enqueued — a dry run could spend | single side-effect-free `plan_sync` shared with the CLI |
| F5 | P1 item published before the fallible ledger commit | rollback + structured per-item error, batch not aborted |
| F6 | P2 fixed window hid eligible work; no FIFO | `sort:created-asc` + widening driven by the planner |
| F7 | P2 malformed poll looked like a healthy empty one | strict top level, tolerant per entry |
| F8 | P2 labels didn't track the queue | `running` at `start`, `failed` on retry, terminal at end |
| F9 | P1 four terminal paths never reported | all four report; each independently tested |
| F10 | P2 remote I/O under the queue mutex | every writeback after the critical section |
| F11 | P2 dry-run ignored the admission cap | shares the production planner |
| F12 | P2 lock contention was a traceback | actionable retry message |

## The two claims that were wrong

**F1 — I hadn't considered that the label is a point-in-time act while the body is mutable.** Submit something benign, wait for a maintainer to label it, rewrite the body: those bytes become factory input under an authorization granted for different ones. Now bound, failing closed on every uncertainty including a label event that is only a *state* label.

**F2 — the security premise was simply false.** Triage can label without push access, and any Action with `issues: write` can label. I'd written "requires write access" in the module docstring, the config comment, the CLI help, and the PR body, and called it "the entire access-control story." It was a *borrowed* boundary presented as a designed one. Corrected in all four places rather than hedged.

## Two places the review corrected my architecture, not a constant

**F6.** My first fix paginated inside `poll_queued` with a breadth heuristic. Structurally wrong for the reason you gave — eligibility is only knowable *after* planning — and my own test caught it: a full page of already-processed issues satisfied the breadth check and admitted nothing. The loop moved into `sync`, with authorization memoized so re-planning is free.

**F4 + F11 had one root cause:** two decision trees. Rather than teach the second one to count, there is now one `plan_sync` and both callers render its output. A dry run *cannot* disagree with the real thing because there is only one implementation.

## Two things I did NOT do, deliberately

- **F6 stale-label reconciliation.** Skipped issues keep their trigger label; it's visible in `--dry-run` with the reason. Sweeping labels for items the queue already knows is a separate behaviour I'd rather propose than slip in. Say the word and I'll open an issue.
- **F9 centralized post-transition hook.** The writeback needs the committed item and the path-specific reason; a hook reconstructing those would be a second source of truth for what happened. Explicit calls after each committed transition, with the mutation check covering all four independently so a future forgotten path can't pass silently. Happy to do the hook if you prefer it.

## Verification

```
uv run pytest tests/     2960 -> 2997 passed, 28 skipped
```

**Mutation check: 28/28, caught on the first run** — no non-discriminating tests to fix this time. I also moved the harness's mutation snippets into JSON: nested triple-quoted Python had cost three failed attempts across this PR and #186, and a harness that won't parse teaches nothing.

Two tests are deliberately behavioural rather than structural: the mutex test has the writeback callback *try to take the lock* and fails if refused (an indentation check would pass the moment the code moved), and the dry-run test asserts the subsequent real sync admits exactly what the dry run promised.

Issue #188, opened 2026-07-30
R8.6 follow-up: gate the intake trigger on an actor allowlist, not inherited write access


0xfauzi:
> **Stabilisation queue: P2, rank 11 of 21** (set 2026-09-09; the full ordered list and the coordinator's rules are in #369). Protects authorization: intake treats repo write access as spend authority; not live while intake is off. Keep intake default-off until this lands.

Follow-up to #153 / #187. Deferred deliberately: the adapter defaults to off, so this gap is not live until GitHub intake is enabled.

## Problem

The GitHub intake adapter treats "the `kstrl:queued` label is present" as authorization to spend money on an unattended factory run ($2-7 per engineer call). Applying a label requires repo **write access** — so the authorization is *inherited* from a permission that means something else.

GitHub's write permission means "I trust you with the codebase." It does **not** mean "you may spend $30 of my money on an unattended agent run." GitHub has no vocabulary for the second, so the adapter borrowed a boundary designed for the first.

For a single-operator repo the two are identical. They diverge as soon as:

- a collaborator gets triage/write access, or
- **any GitHub Action in the repo with `issues: write` applies the label** — a workflow could trigger factory spend with no human involved at all.

That second path is the one worth closing.

## Fix

Make the authorization kstrl's own, explicit decision:

```toml
[intake_github]
allowed_actors = ["0xfauzi"]   # empty = fall back to today's behaviour
```

The trigger becomes "`kstrl:queued` applied **by an approved actor**", verified from the issue timeline. Confirmed feasible:

```
gh api repos/OWNER/REPO/issues/N/timeline \
  --jq '.[] | select(.event=="labeled") | {label: .label.name, actor: .actor.login}'
```

Costs one extra API call per candidate issue.

## Implementation notes

- Check the actor of the **trigger** label only, never the state labels the adapter writes back. Otherwise the adapter — running under the operator's own token, which *is* on the allowlist — could self-authorize.
- An unreadable timeline must fail **closed** (do not admit), consistent with the rest of R8.6's admission gates.
- Skipped issues need a visible reason, not silence: an operator whose label did nothing must be able to find out why.

## Done when

An issue labelled by a non-allowlisted actor is not admitted, the refusal is reported with the actor name, and the timeline-read failure path is tested.

PR #189, opened 2026-07-30
R8.6 PR 4/4: launchd packaging and the operator guide

0xfauzi:
The last R8.6 slice. Closes out [#153](https://github.com/0xfauzi/kstrl/issues/153).

| PR | Contents | Status |
|---|---|---|
| 1 | queue substrate + `ks queue` | merged (#185) |
| 2 | `ks serve`, classifier, backstops | merged (#186) |
| 3 | GitHub Issues adapter | merged (#187) |
| **4** | **launchd packaging + `docs/continuous-intake.md`** | **this PR** |

## Generated, not templated

`ks serve --print-plist` renders the LaunchAgent. Every path in one is absolute and checkout-specific, so a template with placeholders is a class of setup error — wrong interpreter, wrong root, a `Label` colliding with another checkout — that costs a debugging session to find.

Three decisions the tests pin:

- **The `Label` is a hash of the resolved path.** launchd keeps only the last job loaded for a given `Label`, *silently*, so two checkouts of one repo would leave one unserved.
- **`ThrottleInterval` is 60s**, not launchd's 10s default. At 10s a crash-looping daemon attempts six restarts a minute — the throttle is a spend control, not politeness.
- **`PATH` is set explicitly.** A LaunchAgent inherits no shell environment and both `gh` and `git` must be findable. Getting this wrong yields a daemon that runs and silently fails every poll, which is the hardest setup bug to see.

`--print-plist` also creates `.kstrl/logs/`, because launchd creates the log **file** but not its parent, and a missing parent makes the job fail to spawn with nothing in the log explaining why. `render_launchd_plist` stays pure; the CLI does the side effect where the operator is actively setting the job up.

## Two modes, trade-off stated rather than hidden

- **`keepalive`** (default) — one long-lived daemon, restarted on exit. Fewer moving parts. **Weakness:** launchd only notices a process that *exited*, so a daemon wedged on a network call looks healthy forever.
- **`interval`** — `ks serve --once` on a `StartCalendarInterval` schedule (catches up after wake). Each cycle is a fresh process whose exit code carries "needs a human". **Correction (review F2/F3):** launchd neither kills nor replaces a running job, so a wedge is NOT bounded by the interval; `[serve] factory_timeout_seconds` is the only real bound, and interval mode refuses to generate a plist without one.

If you leave this unattended for long stretches, `interval` fails more visibly.

## Measured caffeinate behaviour (H4)

Sampled with `pmset -g assertions` around a live child process:

- `caffeinate -i <child>` creates a `PreventUserIdleSystemSleep` assertion **on behalf of the child** (`caffeinate asserting on behalf of 'sh' (pid …)`).
- After the child exits the assertion is **gone**. Nothing lingers — that's the "held only during work" property R8.6 asked for, verified rather than assumed.

**The caveat the plan implied but never stated:** the assertion is `PreventUserIdleSystemSleep`, **not** `PreventSystemSleep`. It prevents *idle* sleep and does not prevent an explicit one — so **closing the lid still suspends the machine mid-run.**

Sleep mid-run is *survivable* (the lease reaper reclaims the item; a signal-killed process classifies as infrastructural and retries) but it is **not prevented**, and the docs now say so plainly. Anyone reading "caffeinate" as "the laptop won't sleep" would have been wrong.

## Tested vs assumed

**Tested.** 12 new tests. Every structural assertion is paired with a real `plistlib` parse — a plist that satisfies our own string matching but not the platform's parser is worthless. Both modes, the throttle, `PATH` contents, path scoping, label uniqueness and stability, log-dir creation, XML escaping of awkward paths, and that printing a plist starts no daemon and spends nothing. Also checked by hand: `plutil -lint` passes and `plutil -convert json` round-trips to the intended keys.

**NOT verified — and both need you:**

1. **Sleep and wake resilience.** Whether launchd fires a missed interval on wake, and whether the reaper recovers a run interrupted by a real lid close, needs the machine genuinely suspended. Apple documents the `StartInterval`-on-wake behaviour; that is documentation, not a measurement taken here. The docs ask you to close the lid mid-run once and confirm the item recovers. This is roadmap open question 11 in concrete form.
2. **`ks serve` driving a real factory run end to end.** Every daemon test uses a stub runner, deliberately — a suite that spawned real runs would cost dollars per assertion. **Your first unattended run is also the first end-to-end integration test.**

## The operator guide

`docs/continuous-intake.md` covers all four layers, and leads with the honest framing: for a single operator the local queue plus `ks serve` already delivers the capability, and the GitHub layer is a convenience. It documents what each safety mechanism actually guarantees — including the Triage-not-write authorization caveat, what `daily_budget_usd` can and cannot bound, and a §7 that separates what is verified by test, verified by hand once, and not verified at all.

Every shell command in the launchd section was run as written.

## Verification

```
uv run pytest tests/     2997 -> 3009 passed, 28 skipped
```

R8.6 marked `[x]` in the roadmap. [#188](https://github.com/0xfauzi/kstrl/issues/188) (actor allowlist) remains open as agreed follow-up.

### DIFF
--- a/docs/continuous-intake.md
+++ b/docs/continuous-intake.md
@@ -207,24 +207,46 @@ launchctl unload ~/Library/LaunchAgents/$LABEL.plist
 
 Logs land in `.kstrl/logs/serve.{out,err}.log`.
 
-### Two modes, and the trade-off is real
+### Two modes
 
 ```bash
 ks serve --print-plist                                        # keepalive
-ks serve --print-plist --plist-mode interval --plist-interval 600
+ks serve --print-plist --plist-mode interval --plist-interval 10
 ```
 
-- **`keepalive`** (default) - one long-lived `ks serve` that paces itself
-  from `[serve] poll_interval_seconds`; launchd restarts it if it exits.
-  Fewer moving parts. **Its weakness:** launchd only notices a process
-  that *exited*, so a daemon wedged on a network call looks healthy
-  forever.
-- **`interval`** - launchd runs `ks serve --once` every N seconds. Each
-  cycle is isolated, so a wedge cannot outlive one interval, and the exit
-  code carries "work needs a human". This is the cron-fallback shape.
+- **`keepalive`** (default) - one long-lived `ks serve` pacing itself from
+  `[serve] poll_interval_seconds`; launchd relaunches it when it **exits**.
+  Sleep is survived trivially: the process simply resumes on wake.
+- **`interval`** - `ks serve --once` on a **calendar** schedule.
+  `--plist-interval` is in MINUTES and must divide an hour evenly
+  (1, 2, 3, 4, 5, 6, 10, 12, 15, 20, 30, 60) or be whole hours.
 
-If you are leaving this unattended for long stretches, `interval` fails
-more visibly.
+### What launchd does NOT do
+
+Two guarantees people assume and `launchd.plist(5)` explicitly denies:
+
+> **`StartInterval`** - "If the system is asleep during the time of the
+> next scheduled interval firing, that interval will be missed due to
+> shortcomings in `kqueue(3)`. **If the job is running during an interval
+> firing, that interval firing will likewise be missed.**"
+
+> **`StartCalendarInterval`** - "Unlike cron which skips job invocations
+> when the computer is asleep, launchd will start the job **the next time
+> the computer wakes up**."
+
+So:
+
+1. **Only `StartCalendarInterval` catches up after sleep.** Interval mode
+   uses it for that reason. `StartInterval` would silently skip every
+   firing that elapsed while the lid was shut.
+2. **launchd never bounds how long a job runs, and never replaces one
+   still running.** It just skips the firing. A wedged cycle is therefore
+   not killed - it silently stops every later one.
+
+Because of (2), interval mode **refuses to generate** unless
+`[serve] factory_timeout_seconds` is set. That timeout is the only real
+bound on a cycle; `ThrottleInterval` limits relaunch after exit, not
+runtime.
 
 ### Why the label is a path hash
 
@@ -264,9 +286,15 @@ freely between runs.
 sleep. It does **not** prevent an explicit sleep - **closing the lid will
 still suspend the machine mid-run.**
 
-That is why the rest of the design exists: the lease reaper reclaims a run
-whose owner vanished, and a process killed by a signal classifies as
-infrastructural and retries. Sleep mid-run is survivable, not prevented.
+What that actually means is narrower than "the run is lost". Sleep
+*suspends* processes; it does not kill them. On wake the same `ks serve`
+and the same factory child resume and the cycle finishes. The lease TTL
+may have elapsed during the suspend, but nothing reaps it, because the
+process holding the run is the same one that would do the reaping and it
+is busy running.
+
+The recovery machinery exists for the case where the process really is
+gone - a crash, an OOM kill, a reboot - not for an ordinary lid close.
 
 ---
 
@@ -299,13 +327,12 @@ label plus a comment. And the caffeinate assertion lifetime above.
 
 **NOT verified:**
 
-- **Sleep and wake resilience.** Whether launchd fires a missed interval
-  on wake, and whether the reaper recovers a run interrupted by an actual
-  lid close, has not been exercised - it needs the machine genuinely
-  suspended. Apple documents that a `StartInterval` job runs once on wake
-  if its interval elapsed; that is documentation, not a measurement taken
-  here. **If you plan to run this on a laptop, close the lid mid-run once
-  and confirm the item recovers.**
+- **Sleep and wake behaviour end to end.** The `launchd.plist(5)`
+  contracts quoted above are Apple's documentation, not measurements
+  taken here, and nothing has been exercised against a genuinely
+  suspended machine. **If you plan to run this on a laptop, close the lid
+  mid-run once and confirm the cycle finishes on wake and the calendar
+  job fires.**
 - **`ks serve` driving a real factory run end to end.** Every daemon test
   uses a stub runner, deliberately: a suite that spawned real runs would
   cost dollars per assertion. Your first unattended run is also the first
--- a/docs/continuous-intake.md
+++ b/docs/continuous-intake.md
@@ -1,0 +1,318 @@
+# Continuous intake (R8.6)
+
+How to make work flow into the factory without firing each run by hand,
+and what each safety mechanism actually guarantees.
+
+Three layers, each usable without the ones above it:
+
+| Layer | Command | Needs |
+|---|---|---|
+| Local queue | `ks queue add/ls/show/retry/rm/pause/resume` | nothing |
+| Daemon | `ks serve [--once]` | the queue |
+| GitHub inbox | `ks queue sync` | `gh`, an opt-in config |
+| Scheduling | launchd | `ks serve` |
+
+**Start at the top.** For a single operator, the local queue plus
+`ks serve` already delivers the thing that matters - work runs without you
+firing it - with no remote surface, no tokens, and no polling. The GitHub
+layer adds the ability to queue from your phone and see status where your
+code lives. That is a convenience, not the capability.
+
+---
+
+## 1. The local queue
+
+```bash
+ks queue add specs/add-widget.md --priority 3
+ks queue ls
+ks queue show <id>
+```
+
+Items live under `.kstrl/queue/` as one directory each (spec + `meta.json`),
+moved between `queued/ leased/ running/ done/ failed/ poison/` by a single
+`os.replace`. The spec is **copied** at enqueue, so editing or deleting the
+original afterwards cannot change what runs.
+
+`ks queue pause` stops new work being claimed; it does not touch a run
+already in flight. `ks queue resume` re-opens intake.
+
+### Attempts are money
+
+`[queue] max_attempts` bounds how many times one item may execute. The
+counter is charged **before** the run starts, so an interrupted attempt
+counts. That direction is deliberate: over-counting costs the item one
+retry, while under-counting is an unbounded loop, and a measured engineer
+iteration costs **$1.70-2.60** first-attempt and **$3.99-7.42** on a retry.
+
+`ks queue retry <id>` refuses an item that has spent its attempts unless
+you pass `--reset-attempts`. That flag is the point at which you are
+authorizing more spend.
+
+---
+
+## 2. `ks serve`
+
+```bash
+ks serve --dry-run     # what would happen, spending nothing
+ks serve --once        # one cycle
+ks serve               # poll until interrupted
+```
+
+One factory run at a time, holding a daemon singleton lock. `--dry-run`
+reports every admission gate in the order the real loop evaluates them.
+
+### Only infrastructure failures retry, and only on positive evidence
+
+The rule is not "retry unless it looks like a spec problem". It is
+**nothing retries without affirmative evidence that the failure was
+infrastructural**:
+
+| Evidence | Verdict |
+|---|---|
+| exit 0 | success |
+| launch failed before any spend | retry (free) |
+| killed by signal, or our timeout with the process group confirmed dead | retry |
+| exit 2 with a lock-contention marker in the output | retry |
+| exit 2 with a spec-blocker marker | **poison** |
+| every failed component carries `infrastructure_error` | retry |
+| any failed component failed on its merits | **poison** |
+| nonzero exit, nothing blamed | **poison** (unclassifiable) |
+| manifest unreadable, or not produced by this invocation | **poison** (unclassifiable) |
+| timeout whose process group could **not** be confirmed dead | **poison** |
+
+Poisoned items wait for a human. `ks queue ls --state poison` lists them;
+`ks inbox ls` carries the decision.
+
+### Four backstops, because a correct classifier is not enough
+
+A *persistent* infrastructure fault is retryable by the rules above and
+still burns money. So:
+
+1. **`[queue] max_attempts`** - enforced inside the queue itself, not just
+   by the daemon's policy.
+2. **Exponential backoff** - 60s doubling, capped at 30 minutes.
+3. **`[serve] daily_budget_usd`** - checked *before* admitting each item,
+   pausing until the next **local** midnight so a Friday-night stop is not
+   a dead weekend.
+4. **`[serve] max_consecutive_poison`** - pauses the whole queue. If the
+   base branch is broken, every run fails verification, each failure is
+   individually legitimate, and no per-item bound ever notices.
+
+### What `daily_budget_usd` can and cannot do
+
+It counts only cost an adapter **reports**. The codex adapter reports
+tokens and no cost, and `decompose` (the architect) emits no usage events
+at all - so **every** queued item has some unmetered spend.
+
+Three cases, and they are reported distinctly:
+
+- **full coverage** - the cap is exact.
+- **partial coverage** - a *lower-bound* cap: it fires at or after the
+  threshold, never before. Still a real bound. Every total is labelled
+  `(a FLOOR: ...)` with what was not counted.
+- **zero coverage** - no call has ever reported a cost figure, so the cap
+  can never fire. `ks serve` **refuses to run** unless you set
+  `[serve] allow_uncovered_cost = true`.
+
+The unreported spend is deliberately **never** estimated into a dollar
+figure. A number that looks like a measurement but is a guess is worse
+than an admitted gap.
+
+---
+
+## 3. GitHub Issues as an inbox
+
+```toml
+[intake_github]
+enabled = true
+repo = "owner/name"          # must match the checkout you run in
+queued_label = "kstrl:queued"
+max_items_per_sync = 5
+```
+
+```bash
+gh label create "kstrl:queued"  --color 0e8a16
+gh label create "kstrl:running" --color fbca04
+gh label create "kstrl:done"    --color 0075ca
+gh label create "kstrl:failed"  --color d93f0b
+gh label create "kstrl:poison"  --color b60205
+
+ks queue sync --dry-run
+ks queue sync
+```
+
+An issue carrying `kstrl:queued` becomes a queue item. The verdict comes
+back as a state label and a comment. Polling only - no webhooks, nothing
+to keep reachable.
+
+### What actually authorizes work
+
+**Read this before enabling it on a repo that others can reach.**
+
+A stranger can open an issue but cannot label it. However:
+
+- Applying a label needs the **Triage** role, **not** push access. On an
+  organization repo, a triager who cannot push a line of code can
+  authorize factory spend.
+- Any GitHub Action in the repo with `issues: write` can apply the label,
+  so a workflow can trigger spend with no human involved.
+
+This is a permission designed for *managing issues*, borrowed to authorize
+*money*. [Issue #188](https://github.com/0xfauzi/kstrl/issues/188) replaces
+it with an explicit actor allowlist. Until then, the residual risk is
+exactly those two bullets.
+
+**What is enforced today:** an issue edited *after* it was labelled is
+refused. GitHub lets an issue author rewrite the body after a maintainer
+has labelled it, so the label is bound to the body revision it authorized.
+Re-apply the label to authorize the new text.
+
+**Remote items always stop at the PR.** No label and no config value can
+grant auto-merge to remote-sourced work.
+
+### Cross-repository intake is refused
+
+`ks serve` always runs the factory against its own checkout, so an inbox
+pointing at a different repo would open a PR in the wrong place. Sync
+refuses unless `[intake_github] repo` matches the checkout's remote.
+
+### One measured quirk
+
+GitHub's issue-list endpoint can lag a label write. A sync issued
+immediately after labelling returned `polled: 0`; the same sync a minute
+later returned `polled: 1`. Invisible at any realistic poll interval, but
+do not expect label-then-sync-in-one-breath to work.
+
+---
+
+## 4. Scheduling with launchd
+
+Generate a LaunchAgent for this checkout. Every command below was run as
+written:
+
+```bash
+ks serve --print-plist --root "$PWD" > /tmp/kstrl.plist
+plutil -lint /tmp/kstrl.plist                      # sanity check
+LABEL=$(plutil -extract Label raw -o - /tmp/kstrl.plist)
+cp /tmp/kstrl.plist ~/Library/LaunchAgents/$LABEL.plist
+launchctl load  ~/Library/LaunchAgents/$LABEL.plist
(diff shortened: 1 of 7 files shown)


kstrl/serve.py
```diff
@@ -2179,3 +2179,159 @@ def _cycle() -> CycleResult:
...
+    if interval_seconds < LAUNCHD_THROTTLE_SECONDS:
+        raise ServeError(
+            f"launchd interval must be >= {LAUNCHD_THROTTLE_SECONDS}s "
+            f"(the restart throttle), got {interval_seconds}"
+        )
+
+    root = root_dir.resolve()
+    interpreter = python or sys.executable
+    log_dir = launchd_log_dir(root)
+    args = [interpreter, "-m", "kstrl", "serve", "--root", str(root)]
```
0xfauzi, line 2378: **[P1] The scheduled job never polls GitHub intake.** This argv starts only `ks serve`, and the serve cycle never calls `intake_github.sync`; that function is reachable only through the separate `ks queue sync` command. Reproduced with `[intake_github] enabled = true`: `ks serve --once` exited 0 with the patched sync call count still 0. After installing this plist, a labelled issue therefore cannot enter the queue without a manual sync. Please make enabled intake an explicit stage of the serve cycle, or generate a coordinated polling job, and test enabled, disabled, failure, once, and repeated-cycle behavior.

docs/continuous-intake.md
```diff
@@ -0,0 +1,318 @@
...
+ks serve --print-plist --plist-mode interval --plist-interval 600
+```
+
+- **`keepalive`** (default) - one long-lived `ks serve` that paces itself
+  from `[serve] poll_interval_seconds`; launchd restarts it if it exits.
+  Fewer moving parts. **Its weakness:** launchd only notices a process
+  that *exited*, so a daemon wedged on a network call looks healthy
+  forever.
+- **`interval`** - launchd runs `ks serve --once` every N seconds. Each
+  cycle is isolated, so a wedge cannot outlive one interval, and the exit
```
0xfauzi: **[P1] `StartInterval` does not bound a wedged cycle.** launchd misses an interval firing when the previous job is still running; it does not terminate or replace that process. Because `factory_timeout_seconds` defaults to `0.0`, a wedged `ks serve --once` can remain alive indefinitely while later firings are skipped. `ThrottleInterval` limits rapid relaunch after exit, not maximum runtime. Please implement a real whole-cycle watchdog or remove this guarantee and the recommendation based on it.

docs/continuous-intake.md
```diff
@@ -0,0 +1,318 @@
...
+repository - label polled, item enqueued with provenance, label swapped,
+re-sync correctly skipped, terminal writeback leaving exactly one state
+label plus a comment. And the caffeinate assertion lifetime above.
+
+**NOT verified:**
+
+- **Sleep and wake resilience.** Whether launchd fires a missed interval
+  on wake, and whether the reaper recovers a run interrupted by an actual
+  lid close, has not been exercised - it needs the machine genuinely
+  suspended. Apple documents that a `StartInterval` job runs once on wake
```
0xfauzi: **[P2] This wake contract is incorrect and contradicts the roadmap.** The current `launchd.plist(5)` manual says a `StartInterval` firing is missed while the system sleeps. Apple's scheduling guide attributes wake catch-up to `StartCalendarInterval`; the roadmap already records that `StartInterval` misses sleep intervals. Please correct the guide, CLI docstring, PR narrative, and roadmap to one verified contract. Also avoid implying that lid sleep makes the lease owner vanish: the same processes normally resume on wake.

kstrl/serve.py
```diff
@@ -2179,3 +2179,159 @@ def _cycle() -> CycleResult:
...
+    log file but not its parent, and a missing parent makes the job fail
+    to spawn with nothing in the log explaining why - the worst kind of
+    setup failure. ``render_launchd_plist`` stays pure.
+    """
+    return state_dir(root_dir) / "logs"
+
+
+def _plist_escape(value: str) -> str:
+    """XML-escape a plist string value."""
+    return (
```
0xfauzi: **[P3] A legal checkout path can produce malformed XML.** This escapes the three ordinary text delimiters, but filesystem names may contain control characters that XML 1.0 cannot represent. Reproduced with a root component containing `\x01`: the rendered plist failed `plistlib` with `ExpatError: not well-formed (invalid token)`. Please reject XML-illegal values with a clear `ServeError`, or construct and validate the plist through a library, and add a discriminating test.

0xfauzi (commented):
Changes needed before merge. The local verification contract is green (3,009 tests passed, 28 skipped; strict mypy and Ruff passed; both plist modes passed `plutil -lint`), but the scheduled remote-intake lifecycle is incomplete and two documented launchd guarantees do not match platform semantics. The four inline comments contain the reproduced findings and concrete fixes.

kstrl/serve.py
0xfauzi: Confirmed and fixed in 74cf0e7. Reproduced exactly as you describe: with `[intake_github] enabled = true`, a serve cycle left the patched `sync` call count at 0.

This is the finding I am most glad you caught, because it is not a bug in any one PR. The adapter (#187) and the daemon (#186) were each correct, each reviewed, each merged - and their composition did nothing. The seam only exists BETWEEN the slices, which is exactly where slicing work hides defects. My own live round-trip in #187 exercised `ks queue sync` by hand, so it confirmed the adapter and told me nothing about the daemon.

Intake is now an explicit stage of `serve_cycle`, placed **before** the admission gates: syncing after them would let a freshly-admitted item bypass a budget pause that had already stopped everything else. It is strictly additive - a transport failure or a raised exception is recorded in `sync_errors` and the cycle still drains whatever is already queued.

Covered for all five cases you named: enabled (polled once per cycle), disabled (zero `gh` calls AND no spurious error each cycle), transport failure, raised exception, `--once`, and repeated cycles (three cycles over one issue still yield one item).

docs/continuous-intake.md
0xfauzi: Confirmed and fixed in 74cf0e7 - and my claim was not merely unsupported, it was backwards. `launchd.plist(5)`:

> "If the job is running during an interval firing, that interval firing will likewise be missed."

So a wedged cycle is neither killed nor replaced; it silently blocks every LATER firing. I had written that interval mode meant "a wedge cannot outlive one interval" and recommended the mode on that basis.

I took the "remove the guarantee" branch and added the real bound rather than inventing a watchdog: `factory_timeout_seconds` is the only thing that actually limits a cycle, so interval mode now **refuses to generate** without one, with an error explaining why. The docs state plainly that launchd bounds neither runtime nor overlap, and that `ThrottleInterval` limits relaunch after exit, not runtime.

I considered an in-process whole-cycle watchdog and decided against it for now: the factory subprocess is the long pole and it already has a bounded, group-killed timeout, so a second watchdog would mostly duplicate that while adding a way to tear down a cycle mid-transition. If you would rather have a true cycle deadline, I am happy to add one.

docs/continuous-intake.md
0xfauzi: Confirmed and fixed in 74cf0e7. You are right on every point, including that I contradicted this project's own roadmap, which had recorded the correct behaviour from the start. I checked the man page rather than take it on faith, and it is unambiguous:

> **StartInterval** - "If the system is asleep during the time of the next scheduled interval firing, that interval will be missed due to shortcomings in `kqueue(3)`."

> **StartCalendarInterval** - "Unlike cron which skips job invocations when the computer is asleep, launchd will start the job the next time the computer wakes up."

Interval mode now uses `StartCalendarInterval`. `--plist-interval` is in MINUTES and must divide an hour evenly or be whole hours, since a calendar schedule cannot express "every 7 minutes" - refused with the list of valid divisors rather than silently rounding. The guide, the docstrings, the roadmap and the PR body now all quote the man page and state one contract.

Your second point was the more useful correction. I had implied a lid close makes the lease owner vanish; sleep SUSPENDS processes, it does not kill them. On wake the same serve and the same factory child resume and the cycle finishes - and nothing reaps the run, because the process that would do the reaping is the one running it. The docs now say the recovery machinery is for a crash, an OOM kill or a reboot, not an ordinary lid close.

kstrl/serve.py
0xfauzi: Confirmed and fixed in 74cf0e7. Reproduced with a path component containing a `\x01` byte: `ExpatError: not well-formed (invalid token)`.

I took the library option. The plist is now built as a dict and serialized with `plistlib`, which removes the whole class of escaping bug rather than extending a denylist. Usefully, `plistlib` rejects control characters itself (`ValueError: strings can't contain control characters`), which gives an honest failure point - a path XML cannot represent has no valid plist, so refusing is the only correct outcome. That becomes a `ServeError` naming the likely cause and suggesting the fix.

Discriminating test added: a checkout under such a path exits 2 with "control character" in the message. The existing `& < >` test stays, and there is a round-trip test asserting every generated variant parses back with `plistlib`.

0xfauzi:
## All four findings closed in `74cf0e7`

All reproduced against the submitted code first. Two were consequential, and **F1 is the kind of gap no single-PR review could have caught.**

| # | Finding | Fix |
|---|---|---|
| F1 | P1 the scheduled job never polled GitHub intake | intake is now a cycle stage, before the gates, strictly additive |
| F2 | P1 `StartInterval` does not bound a wedged cycle | guarantee removed; interval mode requires `factory_timeout_seconds` |
| F3 | P2 the wake contract was stated backwards | `StartCalendarInterval`; man page quoted everywhere |
| F4 | P3 a legal path could produce malformed XML | built via `plistlib`, which refuses outright |

## F1 is a seam defect, not a code defect

The adapter (#187) and the daemon (#186) were each correct, each reviewed, each merged. **Their composition did nothing** — an installed LaunchAgent drained a queue nothing could fill.

Worth naming why I missed it: my live round-trip in #187 exercised `ks queue sync` *by hand*. That confirmed the adapter and told me nothing about whether anything ever calls it. Slicing work into independently-reviewable PRs moves risk to the seams between them, and I had no test that crossed one.

Intake now runs **before** the admission gates — syncing after them would let a freshly-admitted item bypass a budget pause that had already stopped everything else. Covered for all five cases you listed.

## F2 and F3: I had the man page backwards

`launchd.plist(5)` is unambiguous, and I checked rather than assumed:

> **StartInterval** — "If the system is asleep during the time of the next scheduled interval firing, that interval will be **missed**... **If the job is running during an interval firing, that interval firing will likewise be missed.**"

> **StartCalendarInterval** — "Unlike cron which skips job invocations when the computer is asleep, launchd will start the job **the next time the computer wakes up**."

So my "a wedge cannot outlive one interval" was exactly inverted — a wedge blocks every *later* firing — and my wake claim contradicted this project's own roadmap, which had it right all along.

Interval mode now uses `StartCalendarInterval` and refuses to generate without `factory_timeout_seconds`, the only real bound. I chose that over inventing a watchdog; happy to add a true cycle deadline if you'd prefer it.

**Your lid-close correction was the more useful one.** Sleep *suspends* processes, it does not kill them — on wake the same serve and factory child resume and the cycle finishes, and nothing reaps the run because the process that would reap it is the one running it. The recovery machinery is for a crash, an OOM kill or a reboot. The docs said otherwise and now say this.

## Verification

```
uv run pytest tests/     3009 -> 3024 passed, 28 skipped
```

**Mutation check 11/11.** Two initially survived and both were my error rather than a gap in the code: one mutation only changed a *comment* (so of course nothing failed), and the disabled-adapter test could not distinguish the early return from `sync`'s own guard — it now asserts the property only the early return provides, that a disabled adapter reports no error on every cycle.

One process note: a `str.replace` in a fixup script silently matched nothing and I printed success anyway, then spent three probes chasing a "mysterious" test failure that was just an unapplied edit. Every scripted edit in this batch now asserts its match count first — the same discipline the review findings keep pointing at.

kstrl/serve.py
```diff
@@ -1709,7 +1748,12 @@ def serve_cycle(
             evidence={"item_id": item_id, "cause": "interrupted run"},
         ),)
 
-    # 2. The pause marker is read AND cleared under the mutex, re-reading
+    # 2. Pull remote work in BEFORE the gates, so newly-admitted items face
+    #    the same budget, breaker and cap checks as everything else.
+    with queue_lock(root_dir, blocking=True):
```
0xfauzi: **[P1] Do not hold the queue mutex across remote intake I/O.** `_run_intake` calls `intake_github.sync`, which polls GitHub and performs per-issue authorization requests, while this lock is held. A slow or timed-out front end therefore blocks `ks queue pause` and every other queue transition, reintroducing the same operational problem that #187 F10 removed from writeback. Reproduced with a probing sync callback: a nonblocking `queue_lock` attempt reported `held`. Poll and authorize outside the mutex, then acquire it only for a fresh local plan and the enqueue/ledger commit.

kstrl/serve.py
```diff
@@ -1709,7 +1748,12 @@ def serve_cycle(
             evidence={"item_id": item_id, "cause": "interrupted run"},
         ),)
 
-    # 2. The pause marker is read AND cleared under the mutex, re-reading
+    # 2. Pull remote work in BEFORE the gates, so newly-admitted items face
```
0xfauzi, line 1757: **[P1] Keep `ks serve --dry-run` in lockstep with this new intake stage.** Production now syncs before selecting the next item, but the CLI dry-run still reads only the existing queue. With intake enabled and an empty queue, it exits 0 saying `nothing ready` and never calls the intake planner, while the real cycle can sync and immediately execute paid work. Run the adapter's side-effect-free planner in dry-run mode and include its would-enqueue result before reporting gates and the next candidate.

kstrl/serve.py
```diff
@@ -2179,3 +2223,204 @@ def _cycle() -> CycleResult:
...
+    firing will likewise be missed." So a wedged cycle is not killed and
+    not replaced - it silently blocks every later firing. The only real
+    bound is ``[serve] factory_timeout_seconds``, which is why interval
+    mode refuses to generate without one (review #189 F2).
+    """
+    if mode not in ("keepalive", "interval"):
+        raise ServeError(
+            f"launchd mode must be 'keepalive' or 'interval', got {mode!r}"
+        )
+    if mode == "interval" and factory_timeout_seconds <= 0:
```
0xfauzi, line 2366: **[P2] Enforce the interval timeout at job runtime, not only plist generation.** This check passes when the plist is created, but the generated argv contains only `serve --root ... --once`. If `factory_timeout_seconds` is later removed or reset to zero, the already-installed interval job starts normally with no bound. Reproduced by generating with 60 seconds, removing the config, and running the stored argv shape: `ServeConfig` loaded zero and `serve --once` proceeded. Add a runtime requirement carried by the plist, such as a dedicated flag or environment marker, so every scheduled invocation fails closed when the timeout is absent.

kstrl/cli.py
```diff
@@ -4076,13 +4076,32 @@ def queue_sync(
...
+    "--print-plist", "print_plist", is_flag=True,
+    help="Print a launchd LaunchAgent plist for this checkout and exit",
+)
+@click.option(
+    "--plist-mode", type=click.Choice(["keepalive", "interval"]),
+    default="keepalive",
+    help="keepalive: one long-lived daemon; interval: `--once` on a timer",
+)
+@click.option(
+    "--plist-interval", type=click.IntRange(min=1), default=5,
```
0xfauzi, line 4089: **[P2] The public CLI contract still states the reversed timer behavior.** `ks serve --help` currently says `StartInterval fires immediately for intervals that elapsed while the machine slept`, even though this fix correctly switched to `StartCalendarInterval`. The PR description also retains the old wedge and wake claims. The previous thread explicitly requested the guide, CLI docstring, PR narrative, and roadmap be corrected to one contract; please update all user-facing copies before resolving it.

docs/dark-factory-roadmap.md
```diff
@@ -958,7 +961,89 @@ the human trigger - `stop_at_pr` default preserves the merge gate).
...
+   `plistlib`, which rejects such values outright; that rejection becomes
+   a clear `ServeError`.
+
+Also corrected: the claim that a lid close makes the lease owner vanish.
+Sleep SUSPENDS processes, it does not kill them - on wake the same serve
+and the same factory child resume, and nothing reaps the run because the
+process that would reap it is the one running it. The recovery machinery
+is for a crash, an OOM kill or a reboot, not an ordinary lid close.
+
+Two modes ship, with the trade-off stated rather than hidden: `keepalive`**Measured caffeinate behaviour (H4).** With `pmset -g assertions` sampled
```
0xfauzi: **[P2] Repair the roadmap rather than appending the correction beside stale text.** This line is malformed (`keepalive` is concatenated directly with the next heading), and the following paragraphs still say lid sleep is recovered by the reaper and that Apple documents `StartInterval` wake behavior. Those statements contradict the new correction immediately above. Remove the obsolete block and leave one coherent lifecycle account.

kstrl/serve.py
```diff
@@ -2179,3 +2223,204 @@ def _cycle() -> CycleResult:
...
+                "15, 20, 30, 60)"
+            )
+        return [{"Minute": m} for m in range(0, 60, interval_minutes)]
+    if interval_minutes % 60 or interval_minutes > 24 * 60:
+        raise ServeError(
+            f"a {interval_minutes}-minute interval must be a whole number "
+            "of hours (and at most 24h) to be expressed as a calendar "
+            "schedule"
+        )
+    step = interval_minutes // 60
```
0xfauzi, line 2315: **[P2] Whole-hour intervals must divide 24, not merely be integral hours.** For `interval_minutes=300`, this returns hours `[0, 5, 10, 15, 20]`, whose gaps across midnight are `[5, 5, 5, 5, 4]`; the generated job is not actually every five hours. Require `24 % step == 0` (and document the accepted hour divisors), or use a different representation that preserves a uniform cadence.

0xfauzi (commented):
Follow-up review of `74cf0e7`: the original fixes are present and the full verification contract is green (3,024 tests passed, 28 skipped; strict mypy, Ruff, GitHub CI, and both macOS plist checks passed). Six remaining issues prevent resolving the review threads, including two safety-path regressions introduced by integrating intake. Reproductions and required changes are inline.

0xfauzi:
Pushed a doc repair on top of the review-fix commit: the roadmap paragraph promising the keepalive/interval trade-off was truncated mid-sentence and collided with the caffeinate heading (found during the 2026-08-03 architecture review). Reconstructed from launchd_plist_dict's docstring. Also corrected the stale wedge claim in this PR body per your own review correction F2/F3 (launchd does not bound a running job; factory_timeout_seconds is the only real bound).

kstrl/serve.py
0xfauzi: Confirmed and fixed in 69033d9. Reproduced with your probing callback: a nonblocking `queue_lock` attempt reported held.

This one is mine twice over - I reintroduced the exact problem #187 F10 made me remove from writeback, in the very commit that fixed F1.

`sync` now splits into two phases. The network phase (repo resolution, the widening poll, per-issue authorization) runs with no lock at all. `commit_guard` is entered only around the local writes, and the plan is **re-run inside the guard** against fresh queue state, so a racing writer that admitted the same ref while we were on the network cannot cause a double-admit. Authorization is memoized, so the second plan costs no requests.

`ks queue sync` uses the same seam rather than wrapping the whole call, so the CLI stopped holding the mutex across network I/O too.

Three tests: no `gh` call happens with the mutex held; every intake `queue.add` DOES happen with it held; and a guard that admits the same ref mid-flight results in no duplicate.

kstrl/serve.py
0xfauzi: Confirmed and fixed in 69033d9. You are right, and the frustrating part is that #187 F4/F11 established exactly this principle - a dry run that disagrees with the real thing is worse than none - and I fixed it for `ks queue sync`, then added intake to `serve` without touching serve's dry run.

`ks serve --dry-run` now runs the adapter's side-effect-free planner with `dry_run` forced, and reports polled count, each `would admit`, each skip with its reason, and any intake errors, before the gates and the next candidate. When the queue is empty but intake would admit, the "next item" line says so explicitly instead of "nothing ready", since that phrasing was the actual misleading part.

Tested: the dry run reports what would be admitted, writes nothing locally, and says "disabled" when intake is off.

kstrl/serve.py
0xfauzi: Confirmed and fixed in 69033d9. Your point is the general one - a generation-time check binds only the config of the day it ran - and I had no answer to the operator who later clears the setting.

The generated interval job now carries `KSTRL_SERVE_REQUIRE_TIMEOUT=1` in `EnvironmentVariables`. `ks serve` refuses to start when that marker is present and `factory_timeout_seconds <= 0`, with an error explaining that launchd neither kills nor replaces a job still running at the next firing. The requirement travels with the job rather than living only in the generator.

A hand-run `ks serve` is unaffected, which has its own test so the marker cannot quietly become a global requirement.

kstrl/cli.py
0xfauzi: Confirmed and fixed in 69033d9. The previous thread asked for the guide, CLI docstring, PR narrative and roadmap to be corrected to one contract, and I did the first and last and missed the docstring. That is on me.

The `ks serve` docstring now quotes `launchd.plist(5)` directly and mentions that the cycle polls any enabled remote inbox. I also swept every user-facing surface for the old claim rather than fixing them one at a time, and there is a test asserting `ks serve --help` contains `StartCalendarInterval` and does not contain "fires immediately" - so the contract cannot drift back silently.

The PR description is updated in the same pass.

docs/dark-factory-roadmap.md
0xfauzi: Confirmed. You had already pushed 891e08e repairing the truncated line while I was working on the same region, so I rebased onto your commit rather than force-pushing over it, and kept your prose as the base - it reads better than my restructure and says the same things.

I appended only the two facts your commit predates, both from this review round: the interval job now carries `KSTRL_SERVE_REQUIRE_TIMEOUT` so the bound is re-checked at runtime (N3), and the interval must divide an hour or be an hour count dividing 24 (N6).

I also removed the last stale line further down, which still framed the wake behaviour as "Apple documents the StartInterval-on-wake behaviour". That is now one account with no contradiction left in the section.

kstrl/serve.py
0xfauzi: Confirmed and fixed in 69033d9. Verified the arithmetic: `interval_minutes=300` gave hours `[0, 5, 10, 15, 20]` with gaps `[5, 5, 5, 5, 4]` across midnight.

`calendar_schedule` now requires `24 % step == 0` and refuses otherwise, naming the valid divisors (1, 2, 3, 4, 6, 8, 12, 24) rather than silently rounding. My earlier test used 360 minutes - six hours, which happens to divide 24 - so it passed and proved nothing about the general case.

The replacement is parametrized over every accepted minute and hour value and asserts the set of gaps has exactly one element, which is the property rather than a sample of it. One representational note kept in a test: 60 minutes takes the minute branch and yields `[{"Minute": 0}]`, which is still hourly because unspecified `StartCalendarInterval` fields are wildcards.

PR #197, opened 2026-08-03
fix(serve): a budget halt must not be read as retryable infrastructure

0xfauzi:
Two defects found by the **first live `ks serve` runs**, which you asked me to do. Neither was reachable by the 3,000 unit tests, and the first one is a money-safety hole.

## 1. A budget halt was being retried (P1)

Measured on a real run: the engineer blew `max_total_tokens` (716,348 >= 400,000) after **$2.53**, and serve scheduled a **retry**.

`pipeline.fail_for_budget` records a blown ceiling as `Finding.infrastructure_error` with no distinguishing category. My classifier's whole premise is "only `infrastructure_error` retries" — so a deliberate *stop spending* read as transient infrastructure trouble.

That is the crash loop R8.6 exists to prevent, arrived at **through the one branch meant to be safe**. A ceiling breach is deterministic: the retry re-runs the same work against the same limit, and retries cost *more* because they carry accumulated context ($3.99–7.42 vs $1.70–2.60 measured). Three attempts of that is real money for a guaranteed failure.

`Verdict.BUDGET_HALT` now separates it, decided from the typed `BudgetExceeded` event rather than by matching the finding's prose, so it cannot drift when the wording changes. Checked *before* the infra branch, since the finding genuinely is an `infrastructure_error`.

## 2. An unevidenced failure claimed to be a spec failure (P2)

A git worktree failure (`fatal: invalid reference`) set `Component.error` and emitted **no Finding**, and the classifier reported:

> "greet-function failed on their own merits, not on infrastructure"

Both verdicts poison, so the money behaviour was already right — but the statement was false, and it sends an operator to read the spec when the problem is git. `SPEC_FAILURE` now requires findings as positive evidence; a failure with none is `UNCLASSIFIABLE` and **prints the component's own error**.

## Why 3,000 tests missed both

Every unit test built its manifest by hand. Every one attached findings. And every `Finding.infrastructure_error(...)` I wrote meant "the CLI died" — none encoded that the factory uses that same category for a self-imposed halt. The tests agreed with my model of the factory rather than with the factory.

## Verified against the real artifacts

The same manifest and event stream from the live run:

```
before:  verdict=retry_infra   may_retry=True
after:   verdict=budget_halt   may_retry=False
         "halted on a configured ceiling (max_total_tokens) ... a human decision"
```

## Also fixed, measured on a real LaunchAgent firing

`serve.out.log` is 0 bytes and everything goes to `serve.err.log` — the console UI writes to stderr. The guide told operators to check both, so anyone tailing `.out` would conclude the daemon was dead.

## Verification

```
uv run pytest tests/     3063 -> 3071 passed, 28 skipped
```

16 new tests, including one asserting that prose mentioning "token budget exceeded" does **not** trigger the halt path — only the typed event may.


kstrl/serve.py
```diff
@@ -806,7 +850,32 @@ def classify_run(
             },
         )
 
-    judged = [comp.id for comp in failed if not _infra_casualty(comp)]
+    # A deliberate budget halt is recorded as an infrastructure_error by
+    # the pipeline, so it must be separated out BEFORE the infra check or
+    # it reads as retryable (found by the first live run).
+    halted = budget_halt_reason(root_dir, owned_run_ids)
```
0xfauzi: **[P1] Check the budget halt before retry-authorizing exit branches**

This check runs after `timed_out`, negative return codes, and exit-2 lock contention have already returned `RETRY_INFRA`. An invocation can emit `BudgetExceeded` and then remain alive long enough for `serve.factory_timeout_seconds` to kill it, especially with parallel workers or hanging cleanup. In that case the whole item is requeued against the ceiling this PR is meant to make terminal. Please inspect the owned event streams before every retry-authorizing branch, after the launch-error case where no child artifacts can exist, and add a `timed_out=True` plus owned `BudgetExceeded` regression test.

kstrl/serve.py
```diff
@@ -667,6 +671,45 @@ def run_dir_names(root_dir: Path) -> frozenset[str]:
...
+            continue
+        try:
+            events = read_run_dir(state_dir(root_dir) / "runs" / run_id)
+        except OSError:
+            continue
+        for event in events:
+            if isinstance(event, ev.BudgetExceeded):
+                named = ", ".join(event.ceilings) or event.ceiling or "budget"
+                return (
+                    f"the run halted on a configured ceiling ({named}) in "
```
0xfauzi, line 724: **[P2] Preserve the `unenforceable` halt semantics**

`BudgetExceeded` also represents `condition="unenforceable"`, where no threshold was crossed and raising a limit or narrowing the spec is not the relevant diagnosis. I reproduced this helper rendering that event as “the same work against the same limit,” contrary to `events.budget_halt_kind()` and the reducer's explicit wording. Please route through the shared classifier, or at least branch on `condition`, so the poison reason tells the operator that no configured ceiling can still fire and points them toward coverage/configuration rather than a fictitious breach.

0xfauzi (commented):
I found one retry-safety blocker and three operator-honesty issues. The core direction is correct, and the existing suite plus CI are green, but the new event check is still bypassed by retry-authorizing branches and several documented/reported cases do not match the event semantics. I would fix these before merging.

kstrl/serve.py
```diff
@@ -806,7 +850,32 @@ def classify_run(
...
+    #
+    # Found by the first live `ks serve` run: a git worktree creation
+    # failure ("fatal: invalid reference") set Component.error and emitted
+    # no Finding, and this branch reported it as "failed on their own
+    # merits, not on infrastructure". Both verdicts poison, so the money
+    # behaviour was already right; the STATEMENT was false. Every unit
+    # test had constructed manifests WITH findings, so nothing caught it.
+    judged = [
+        comp.id for comp in failed
+        if comp.findings and not _infra_casualty(comp)
```
0xfauzi, line 898: **[P2] Do not discard errors from unevidenced sibling failures**

Because `judged` returns before `unevidenced` is collected, a mixed manifest with a real spec finding on one component and `fatal: invalid reference` with no finding on another reports only the spec component. The latter error is absent from both the reason and evidence, recreating the operator-misdirection this PR is trying to fix and contradicting the new per-component documentation. Keeping `SPEC_FAILURE` as the terminal verdict is reasonable, but please also surface every unevidenced sibling and its `Component.error` in the reason/evidence.

docs/continuous-intake.md
```diff
@@ -325,6 +342,15 @@ repository - label polled, item enqueued with provenance, label swapped,
 re-sync correctly skipped, terminal writeback leaving exactly one state
 label plus a comment. And the caffeinate assertion lifetime above.
 
+**Verified by a live run (2026-08-03):** a LaunchAgent generated by
+`--print-plist --plist-mode interval --plist-interval 1` loaded with
+`launchctl load`, fired twice on its calendar schedule, ran
+`ks serve --once` to completion each time, and reported
+`LastExitStatus = 0`. And `ks serve` drove a real `ks factory` run end to
```
0xfauzi, line 349: **[P3] Remove the contradictory “NOT verified” claim below**

This paragraph now says `ks serve` drove a real factory run end to end, but the unchanged bullet immediately below still says that exact workflow is not verified and that the first unattended run will be the first integration test. Operators cannot tell which claim is current. Please remove or rewrite the stale bullet while preserving the narrower point that automated daemon tests use a stub runner.

kstrl/serve.py
0xfauzi: Fixed in 5a58425. Reproduced first: a `RunOutcome(timed_out=True)` over a run dir containing a real `BudgetExceeded` event returned `RETRY_INFRA`, so the item was requeued against the very ceiling this verdict exists to make terminal. Your "blows its budget and then hangs" path is exact, and the signal and exit-2 paths had the same hole.

The check now sits immediately after the launch-error branch. That one has to stay above it, because a child that never started can have no artifacts to read - but it is also the only branch where retrying is free, so nothing is at risk by ordering it first. Every branch that can authorize a retry is now below the halt.

Guarded by `TestBudgetHaltPrecedence`, parametrized over timeout / signal / lock-contention / plain-nonzero, so a branch inserted above the check in future fails the suite rather than silently re-opening this. Mutation-checked: moving the block back below `timed_out`, sinking it below every early branch, deleting it, and pushing `launch_error` below it are all caught.

kstrl/serve.py
0xfauzi: Fixed in 5a58425. You are right that this was asserting a breach that never happened, and the advice was worse than useless - "raise the ceiling" cannot help when the ceiling never fired.

Rather than branch on `condition` here, it routes through `ev.budget_halt_kind(event.condition, event.ceilings, event.ceiling)` - the same classifier the reducer and the Linear sink already use - so the three readings of one payload cannot drift. `unenforceable` now says no metered call reported that axis, so the cap could never stop the spend, and points at coverage or configuration instead of a limit.

The verdict stays `BUDGET_HALT` in both cases, which I think is still right: an unenforceable ceiling means the run halted deliberately and a retry hits the same wall.

Mutation-checked: falling through to the breach wording, swapping the two branches, and bypassing the shared classifier for a hard-coded kind are all caught.

kstrl/serve.py
0xfauzi: Fixed in 5a58425. Reproduced with a mixed manifest - one component with a `Finding`, one with `error="fatal: invalid reference"` and no findings - and the sibling appeared in neither the reason nor the evidence. That is the same operator misdirection this PR was opened to remove, reintroduced one branch over.

`unevidenced` and its note are now built before the `judged` return. The terminal verdict stays `SPEC_FAILURE` (the judged component is real positive evidence, and both verdicts poison anyway), but the reason appends every unevidenced sibling with its `Component.error`, and the evidence dict carries `unevidenced_failures` plus `component_errors`.

Mutation-checked: dropping the note from the reason, moving the collection back below the return, and dropping `component_errors` from the evidence are all caught.

docs/continuous-intake.md
0xfauzi: Fixed in 5a58425. The two paragraphs did contradict each other, and the later one was the stale copy - written before the live run, never updated after it.

Rewritten to keep the narrower claim that is actually true: the end-to-end path is verified by hand, but every daemon test in the suite uses a stub runner, so a regression in the serve-to-factory seam would not be caught by CI. That is the real gap, and it is worth stating - it is exactly how #196 (`ks serve` never polled intake) survived two green suites.

Flagging honestly: this one is a docs edit with no automated coverage, so unlike the other three it is not mutation-guarded.

PR #206, opened 2026-08-03
Record knowledge fact-utilization, measured against the injected prefix (#191)

### LINKED ISSUE #191: Knowledge fact-utilization is computed but never recorded: the L2+ gate evidence evaporates
## Problem

Knowledge fact-utilization is computed on every distill (`kstrl/pipeline.py:2836-2865`) but the result is only printed: `self.ui.info(f"  Knowledge utilization: {referenced}/{injected} ...")`. It reaches no event, no journal, no file. The `DistillResult` event carries only `facts_written` and `duration_seconds` (`kstrl/events.py:483-486`), and the evolution journal record has no utilization field.

## Why it matters

`docs/remediation-roadmap.md` L89-91 makes "two real factory runs with nonzero fact-utilization and one traceable evolve proposal" the **sole remaining L2+ cATO gate** (the calibration baseline and reviewer-family delta were both captured 2026-07-20). As shipped, that gate cannot be satisfied with durable evidence: the number the gate demands evaporates into a UI line the moment the run ends.

## Fix

- Add utilization fields (`facts_injected`, `facts_referenced`) to the `DistillResult` event and to the evolution journal component record.
- Record the H4 caveat with the number: the measurement is a 30-char case-insensitive substring match (`kstrl/knowledge.py:1300-1311`) and therefore a lower bound. It is mechanical, not self-report, so it is a valid signal, but it should be labeled a floor.

## Acceptance

- A real factory run leaves the utilization ratio in `events.jsonl` and `evolution.jsonl`.
- The dark-factory roadmap's measurements section references the recorded field as the way the L2+ knowledge gate is evidenced.

Filed from the 2026-08-03 architecture-review session (user-approved item P9).

0xfauzi:
Closes #191.

## Why

`docs/remediation-roadmap.md` makes "two real factory runs with nonzero fact-utilization and one traceable evolve proposal" the **sole remaining L2+ cATO gate** - the calibration baselines and reviewer-family delta were both captured 2026-07-20. That gate could not be evidenced: the metric was computed on every distill and formatted into a `ui.info` line, reaching no event, no journal, no file.

Making it durable required fixing the number first. Each of the following would have made the recorded evidence wrong.

### 1. It measured a contaminated prefix

`factory._submit_args` builds the prefix the engineer actually receives. `_phase_distill` ignored it and rebuilt via `build_knowledge_context` **after** `distill_facts` had written this run's facts under `<knowledge_root>/<comp_id>/<run_id>/`, which the core tier read straight back.

So `injected` counted facts the engineer never saw. Worse, those facts were distilled *from the diff*, so their snippets were unusually likely to match it, inflating `referenced` too. The net sign is not predictable. The same rebuild also picked up facts a parallel sibling wrote mid-run.

The prefix is now captured at submit time and frozen. `PipelineHooks.build_knowledge_context` is deleted - it had exactly one caller, so the rebuild is unreachable by construction.

### 2. Deleted and unchanged lines counted as utilization

A raw unified diff went into the haystack, so a claim in a `-` deletion or a context line scored `referenced=1`. An engineer could delete the very code expressing a fact, or merely edit near it, and satisfy the nonzero-utilization gate. That is a false **positive**, the one error direction a lower-bound metric must not have - and the docs repeated the lower-bound claim throughout.

`measure_fact_utilization` now takes the diff as a keyword-only `diff=` and reduces it to added lines via `diff_added_content`; `+++ ` headers are excluded so a path cannot match. The parameter is separate rather than pre-filtered at the call site because the signature is the only thing that can stop a raw diff being searched unfiltered. `progress.txt` stays a plain artifact: the engineer writing about a fact is itself the signal.

### 3. The sample was successful components only

Measurement lived in the distill phase, which runs only after verification, review, and security pass. A component failing any gate had facts injected and may well have used them.

Measurement moves to the point a diff is obtainable, so the population is **every component whose diff can be fetched**. A verification failure means the diff phase has not run *yet*, not that no diff exists - the change is committed and is exactly what verification inspected - so that path fetches the diff itself rather than writing off every test, typecheck, and lint failure. `measured: false` is reserved for a real fetch failure and always carries a reason.

### 4. The denominator was inflated

The sibling tier carries a first-sentence summary of every **other** component's facts, rendered like full-text core facts, so the claim extractor counted them - the claims a component is least likely to echo.

Counts are now split per tier and reach the journal as `knowledge_utilization.by_tier`. `core_referenced / core_injected` asks whether the engineer used what was known about the component it was actually building. The renderer and tier parser share their section-title constants: drift would mis-bin claims while the counts still summed correctly.

### 5. Failures were silent, and telemetry was coupled to distillation

`pipeline.py` had a bare `except Exception: pass`, so a broken recorder was indistinguishable from "referenced nothing". `factory._submit_args` swallowed retrieval failure the same way - the higher-severity half, since it strips the engineer's whole prefix. Both now warn.

Utilization is emitted as a dedicated `fact_utilization_measured` event from `record_fact_utilization`, covering every measured and unmeasured path. Emitting from the distill phase meant a component failing review, or whose distill was budget-skipped or raised, landed in `evolution.jsonl` with evidence `events.jsonl` never saw. It is deliberately **not** duplicated onto `DistillResult`: one measurement in one event, so folding both cannot double count.

## Verification

`uv run pytest tests/ -v` (3125 passed, 28 skipped), `uv run mypy kstrl/ --strict` (clean), `uv run ruff check kstrl/ tests/` (clean), `git diff --check` clean.

Per H4, what was tested rather than assumed - every mutation below reintroduces one defect and was confirmed to fail:

| Mutation | Result |
|---|---|
| ignore the recorded prefix (contaminated measurement) | 7 failed |
| emit drops the utilization fields | 2 failed |
| swallow measurement errors silently again | 1 failed |
| measure below the budget guards again | 11 failed |
| treat a failed capture as a measured zero | 2 failed |
| measure back in the distill phase only | 14 failed |
| drop the verify-failure reason (silent gap) | 1 failed |
| attribute every claim to the core tier | 3 failed |
| fold unknown sections into the sibling tier | 1 failed |
| renderer title drifts from the parser constant | 2 failed |
| search the raw diff again (deletions count) | 4 failed |
| stop excluding `+++` file headers | 1 failed |
| write off verification failures as unmeasurable | 2 failed |
| swallow a real diff-fetch failure as a measured zero | 1 failed |
| emit the event only from a successful distill | 4 failed |
| revert the matcher call to the positional shape | 4 failed |
| pass the diff as both an artifact and `diff=` | 4 failed |

The last two matter most. Review found that the added-line filtering was inert in production: the matcher was changed to filter `diff=` while the pipeline still passed the diff positionally, and nothing caught it - `*artifacts` satisfies mypy either way, the matcher's unit tests passed a correct call, and every pipeline test injected a permissive `lambda *a, **k` stub. `TestFactUtilizationUsesTheRealMatcher` now wires the real matcher into the pipeline and drives deletion, context, and added-line diffs through `process_result`.

The end-to-end journal test runs `run_factory` with `_run_component` patched, so it costs no LLM spend.

No prompt bodies changed, so H2 and H3 do not apply. Per H1 this has not been self-reviewed.

## Standing caveat

The 30-character case-insensitive match is lexical, not semantic: a paraphrased fact scores as unused, and an added line that happens to contain the claim text scores as used. The error is now overwhelmingly in the under-counting direction, but it is not zero. Documented in `docs/evolution-metrics.md` and `docs/adversarial-design.md`.

## Next

The gate is still open and is a real-spend step: two factory runs, then `EvolutionJournal.get_fact_utilization` showing `runs_with_referenced >= 2`. Any utilization figure observed before this PR is not valid evidence.


kstrl/pipeline.py
```diff
@@ -1717,6 +1880,10 @@ def process_result(
             verify.failure.error if verify.failure else "",
         )
         if verify.failure is not None:
+            self.record_fact_utilization(
```
0xfauzi: [P1] Measure the available diff instead of excluding every verification failure

A failed verification does not mean the engineer produced no diff; it only means `_phase_diff` has not run yet. The worktree can already contain the complete committed change, and mechanical verification's scope checks inspect that change. This path therefore marks all test, typecheck, and lint failures unmeasured even when a diff exists, contradicting the documented population of “every component that produced a diff” and retaining success-selection bias in the L2+ gate. Please fetch the diff before verification, or best-effort fetch and measure it on this failure path, using `measured: false` only when the diff fetch actually fails.

kstrl/knowledge.py
```diff
@@ -1299,29 +1349,48 @@ def measure_fact_utilization(
...
     # mid-sentence "the").
     haystack = "\n".join(artifacts).lower()
-    referenced = 0
-    for claim in claims:
+    for claim, tier in claims:
         snippet = _first_sentence(claim)[:snippet_length].strip().lower()
-        if snippet and snippet in haystack:
-            referenced += 1
-    return {"injected": len(claims), "referenced": referenced}
+        hit = bool(snippet) and snippet in haystack
```
0xfauzi, line 1420: [P1] Do not count deleted or unchanged diff text as utilization

`shared_diff` is a raw unified diff, so `haystack` contains `-` deletion lines and unchanged context lines as well as additions. A claim appearing only in either of those currently produces `referenced == 1`; I reproduced both cases with a one-fact prefix. That allows an engineer to delete the code expressing a fact, or merely edit nearby code, and still satisfy the nonzero-utilization gate. It also invalidates the repeated “lower bound” claim because this creates false positives, not only paraphrase-driven false negatives. Please match against added lines only (excluding diff headers), or pass an added-content artifact into this function, and pin deletion/context cases in tests.

0xfauzi (commented):
The implementation is well-tested and the submit-time prefix capture is a strong correction, but the recorded metric can still produce invalid gate evidence and is not durably emitted on every measured path. I found three blocking correctness gaps below. Local verification at this head is clean: 3,113 tests passed (28 skipped), strict mypy passed, Ruff passed, and `git diff --check` passed.

kstrl/pipeline.py
```diff
@@ -2811,6 +3002,9 @@ def _phase_distill(
                 duration_seconds=round(
                     time.monotonic() - distill_start, 2,
                 ),
+                utilization_measured=util.measured,
```
0xfauzi: [P2] Emit utilization independently of successful distillation

These event fields are emitted only inside the successful distillation `try`. A review/security failure never reaches this phase; a budget-skipped distill returns above it; and a distillation exception skips this emit. In all three cases `record_fact_utilization` may already hold a valid measurement, so `evolution.jsonl` gets evidence while `events.jsonl` does not, despite #191's acceptance criterion requiring the ratio in both. A dedicated utilization event emitted when `record_fact_utilization` stores the result would cover every measured and unmeasured path without coupling telemetry to a later LLM phase.

kstrl/knowledge.py
0xfauzi: Confirmed and fixed in 65e41e5. Reproduced both cases before changing anything: a one-fact prefix scored `referenced=1` against a deletion-only diff and against a context-only diff.

`measure_fact_utilization` now takes the diff as a separate `diff=` parameter and reduces it to added lines via `diff_added_content`; `+++ ` file headers are excluded so a path cannot match either. I went with a separate parameter rather than pre-filtering at the call site because the signature is the only place that can prevent a raw diff being searched unfiltered again - `test_diff_passed_as_an_artifact_would_not_be_filtered` pins exactly that difference. progress.txt stays a plain artifact, since the engineer writing about a fact is itself the signal there.

You are right that this invalidated the lower-bound claim rather than merely bounding it, and that claim was repeated across three docs. They now say the match is lexical rather than semantic, and that the error is overwhelmingly - not exclusively - in the under-counting direction.

Pinned by `TestFactUtilizationMatchesAddedLinesOnly` (deletion, context, added, header-path, progress-log, and the artifact-vs-diff contract). Mutation check: restoring the raw-diff search fails 4 tests; dropping the `+++` exclusion fails 1.

kstrl/pipeline.py
0xfauzi: Confirmed and fixed in 65e41e5. `_phase_diff` is just `git.get_diff_content(base_branch, wt_path)` with no dependency on verification passing, so you are right that the diff is fetchable there - the change is committed in the worktree and is exactly what verification just inspected.

That path now fetches the diff itself (`record_fact_utilization` with no `diff_text`) instead of writing the component off. `measured: false` is reserved for a real fetch failure and carries `diff unavailable: <error>`. The extra `git diff` only runs on the failure path and is cheap next to the component run that preceded it.

You also caught a documentation overstatement: I had written the population as "every component that produced a diff" while excluding the largest failure class. The docs now say "every component whose diff can be fetched" and enumerate the reasons an entry can be unmeasured.

Two tests: `test_verify_failed_component_is_still_measured` and `test_unmeasured_only_when_the_diff_fetch_actually_fails`. Mutation check: reverting to the unavailable-marker fails 2 tests; swallowing a real fetch failure as a measured zero fails 1.

kstrl/pipeline.py
0xfauzi: Confirmed and fixed in 65e41e5, and I took your suggested design.

A dedicated `FactUtilizationMeasured` event is now emitted from `record_fact_utilization`, so it fires on every path that reaches a measurement point - including components that go on to fail review or security, and including the budget-skipped and raising-distill cases - with no coupling to a later LLM phase. You were right that the acceptance criteria name both surfaces and only one was being satisfied for those paths.

One thing beyond what you asked: I removed the three utilization fields from `DistillResult` rather than leaving them as a second copy. Keeping both would put one measurement in two events and let a consumer folding both double count. `DistillResult` now describes the knowledge layer's write side only, and `test_distill_result_does_not_duplicate_the_measurement` pins that. Say the word if you would rather have the denormalized copy back for convenience.

Mutation check: reverting the emit to the distill phase fails 4 tests.

kstrl/pipeline.py
```diff
@@ -2828,45 +3065,84 @@ def _phase_distill(
...
-            pass
-
-        return DistillPhaseResult(ran=True)
+            )
+            try:
+                progress_text = progress_path.read_text(encoding="utf-8")
+            except OSError:
+                pass
+            util = self.hooks.measure_fact_utilization(
+                prefix, shared_diff, progress_text,
```
0xfauzi: [P1] Pass the unified diff through the new `diff=` parameter

The matcher now filters only its keyword-only `diff`; positional `artifacts` are deliberately searched raw. This call still uses the old positional shape, so `shared_diff` bypasses `diff_added_content` and deletion/context false positives survive end to end. I reproduced the exact call shape at this head: `measure_fact_utilization(prefix, deletion_diff, "")` returns `referenced=1`, while `measure_fact_utilization(prefix, "", diff=deletion_diff)` returns `0`. The new unit test even documents this distinction, but pipeline tests inject permissive stubs and never assert the keyword. Please call `measure_fact_utilization(prefix, progress_text, diff=shared_diff)` and add an integration test using the real matcher so the pipeline contract cannot drift again.

0xfauzi (commented):
Re-review of 65e41e5: the verification-failure sampling fix and the dedicated `fact_utilization_measured` event are implemented correctly. One original P1 remains active because the pipeline still calls the redesigned matcher through its old positional API, bypassing added-line filtering. I also found a stale roadmap contract. Verification at this head is otherwise clean: 3,121 tests passed (28 skipped), strict mypy passed, Ruff passed, and `git diff --check` passed.

docs/dark-factory-roadmap.md
```diff
@@ -154,12 +154,29 @@ R8.2 consumes them:
...
-  evolve proposal: **OUTSTANDING - the sole remaining L2+ gate.** Caveat
-  found 2026-08-03: utilization is computed each distill but only printed,
-  never recorded to events or the journal, so this gate cannot yet be
-  evidenced durably. [#191](https://github.com/0xfauzi/kstrl/issues/191)
-  (record the ratio in the `DistillResult` event and the evolution
-  journal) is a prerequisite for satisfying it.
+  evolve proposal: **OUTSTANDING - the sole remaining L2+ gate.** The
+  recording prerequisite is closed:
+  [#191](https://github.com/0xfauzi/kstrl/issues/191) landed, so
+  utilization now reaches the `DistillResult` event and the evolution
```
0xfauzi: [P2] Update the remaining obsolete `DistillResult` contract

The follow-up intentionally removed utilization from `DistillResult` and introduced `fact_utilization_measured`, which this same paragraph correctly names at lines 174-176. This earlier sentence still tells operators and consumers to look for the removed fields on `DistillResult`. Please update it to the dedicated event. The PR description and the `knowledge_utilization` reason list in `docs/evolution-metrics.md` also retain pre-follow-up wording (`component failed verification, before a diff`), so they should be swept at the same time.

kstrl/pipeline.py
0xfauzi: You are right, and this is the important catch of the review. Fixed in 6ba82ca.

The call was never updated. My edit to it failed and I did not retry, so the previous commit shipped a matcher that filters `diff=` alongside a pipeline that still passed the diff positionally - the filtering was inert in production and the deletion/context false positive survived end to end, exactly as you reproduced.

Worth naming why nothing caught it, because that is the actual defect: the signature takes `*artifacts`, so mypy accepts either shape; the matcher's unit tests pass a correct call and stayed green; and every pipeline test injects a `lambda *a, **k` stub that accepts both. I had even written a unit test documenting the positional-vs-keyword distinction, which is precisely the kind of test that proves a property of the function while saying nothing about the caller.

So I took both parts of your suggestion. The call now passes `diff=shared_diff`, and `TestFactUtilizationUsesTheRealMatcher` wires the REAL matcher into the pipeline and drives deletion-only, context-only, and added-line diffs through `process_result`, asserting `referenced` end to end. `test_the_real_matcher_is_called_with_the_diff_as_a_keyword` additionally asserts the keyword directly, since a spy is cheaper than a full run for pinning call shape.

Mutation check on the exact regression: reverting to the positional call fails 4 tests, and passing the diff as both an artifact and `diff=` fails 4. Neither was caught before.

Two existing spies needed `**kwargs` to accept the keyword - their positional-only signatures raising TypeError was itself confirmation the shape had changed.

docs/dark-factory-roadmap.md
0xfauzi: Confirmed and swept in 6ba82ca. There were four stale references, not two:

- `docs/dark-factory-roadmap.md:160` - the one you flagged
- `docs/remediation-roadmap.md:95` - same claim
- `docs/adversarial-design.md:88` - audit-trail invariant 6
- `docs/adversarial-roadmap.md:51` - D6, which additionally still claimed post-distill instrumentation "against shared_diff + progress.txt", stale on all three counts (location, timing, and what is searched)

All now name `fact_utilization_measured`. adversarial-design states explicitly that utilization is deliberately NOT on `DistillResult` and why, so the removal reads as a decision rather than an omission.

The `knowledge_utilization` row in `docs/evolution-metrics.md` no longer inlines a reason list at all - it points at the reason table, which is the single place reasons are enumerated. That is what let the list drift out of sync in the first place.

PR description updated to match.

PR #208, opened 2026-08-03
test(serve): execute the serve-to-factory seam for real

### LINKED ISSUE #205: No test exercises the real serve-to-factory seam: every daemon test stubs the runner
Recorded as a follow-up from R8.6. This is the gap that let #196 (`ks serve` never polled GitHub intake) survive two green suites and reach a live run.

## The gap

`serve_cycle` and `serve` are well covered - `tests/test_serve.py` alone injects `runner=` in 46 places. But **every one of those injects a stub.** The real `subprocess_factory_runner` is only ever patched out:

```
tests/test_serve_cli.py:68,125,138,179,196,395   patch("kstrl.serve.subprocess_factory_runner", ...)
```

`run_supervised` is genuinely tested against a real subprocess (`test_the_supervised_timeout_reaps_descendants` spawns `sh -c "sleep 30 & ..."` and asserts descendant reaping). That split was deliberate, and its own docstring says why:

> `subprocess_factory_runner` hardcodes `ks factory` as its argv, so the supervision is split into `run_supervised` precisely so this path is reachable without spawning a factory.

So the supervision half is real. The half that decides **what to actually invoke** - the argv, the cwd, the env, the ordering of intake against the gates inside a cycle - is never executed end to end.

## Why it matters

#196 was not a bug inside either PR it spanned. PR 2 (the daemon) and PR 3 (the GitHub adapter) were each independently correct and independently green. `serve_cycle` simply never called the adapter. No stub-runner test could see that, because the stub replaces exactly the boundary where the defect lived.

The end-to-end path IS verified by hand (live runs on 2026-08-03, see `docs/continuous-intake.md` §7), but a hand check does not re-run in CI. A regression in the serve-to-factory seam today would ship.

## Options

1. **A smoke test that runs the real runner against a trivial fake `ks`** - put a stub executable named `ks` on `PATH`, let `subprocess_factory_runner` build and exec its real argv, and assert on what the stub received. Cheap, no LLM spend, and it pins argv/cwd/env plus the cycle ordering. This is the one I would do.
2. **A periodic (not per-PR) real factory run in CI.** Real coverage, real money, real flake surface. Measured cost is $1.21-1.33 per iteration.
3. **A contract test asserting every adapter registered for intake is actually called by `serve_cycle`** - narrower, targets the #196 shape specifically rather than the seam generally.

Option 1 plus 3 covers the observed failure mode without recurring spend.

## Note

`docs/continuous-intake.md` already states this gap in prose (added in #197). This issue exists so it is tracked rather than only described.

0xfauzi:
Closes #205.

## The gap, confirmed before writing anything

`subprocess_factory_runner` appears 8 times across `tests/`. Three are `patch(...)` calls, three are continuation lines of those patches, and two are a docstring and a comment. **It was never executed.** Meanwhile `tests/test_serve.py` injects a stub `runner=` in 46 places.

The supervision half is genuinely tested - `run_supervised` against a real process group - and that split was deliberate. The half that decides *what to invoke* was not, and that is precisely where #189 F1 lived: `serve_cycle` drained a queue nothing could fill, and the daemon and the adapter were each independently correct and independently green.

One correction to the issue text: it attributes the escaped defect to "#196". #196 is the architecture-review docs PR. `git log -S'github' -- kstrl/serve.py` shows `github` first entering `serve.py` in 74cf0e7 ("close the four review findings on PR #189"), so the defect is #189 F1, which is also what the code comment says.

## What this adds

`tests/test_serve_seam.py`, 15 tests, no LLM spend.

**1. The real runner execs its argv** (`TestTheRealRunnerExecsItsArgv`). Only `sys.executable` is replaced, with a stub interpreter that records argv, cwd, env and pid then exits how the test asked. Argv construction, the caffeinate prefix, the env mutation, the process-group spawn and the `RunOutcome` mapping are all shipping code. Covers: `-m kstrl factory`, the spec/project/root values, the four non-interactive flags plus `KSTRL_NO_TUI`, the merge-gate flag in both directions, cwd agreeing with `--root`, `on_spawn` receiving the pid that actually ran, exit code and output tail reaching the outcome, and an unlaunchable child becoming a `launch_error` rather than an exception.

The issue's option 1 suggested a stub `ks` on `PATH`. That would have intercepted nothing - the argv is `sys.executable -m kstrl factory`, with no `ks` involved. Patching the interpreter is the equivalent seam.

**2. The argv is accepted by the real CLI** (`TestTheArgvIsAcceptedByTheRealCli`). The recorded argv is handed to the actual Click command object. A recording stub records whatever it is given and is happy; it cannot tell you a flag was renamed. Includes a control test asserting an unknown flag *is* rejected, so the two positive tests are not vacuous.

**3. What survives the seam** (`TestRemoteWorkSurvivesTheSeam`). Two facts about a GitHub-sourced item that only become observable at the runner boundary: the issue body reaches the spec file the factory is handed, and the merge gate is still on at the moment of invocation.

## Measured, not asserted

Every claim below is from running mutants against the suite.

| Mutant | Caught by this PR | Caught by the rest of the suite |
|---|---|---|
| Rename the merge-gate flag in `subprocess_factory_runner` | 4 tests | **nothing** - 3125 passed, 28 skipped, 0 failed |
| Delete the `_run_intake` call from `serve_cycle` (#189 F1) | - | 9 tests in `test_intake_github.py::TestServePollsIntake` |
| Rename the option on the `ks factory` side | 2 tests | 1 test (`test_config_control_plane.py::TestSafetyKnobs`) |

Row 1 is the gap #205 describes, demonstrated.

Row 2 changed the shape of this PR. My first draft re-asserted the #189 F1 composition; the mutant showed `TestServePollsIntake` already covers it thoroughly (polled every cycle, disabled makes no calls, synced item runs same cycle, failures and exceptions do not stop the cycle, no re-admission, intake precedes the gates). Those four duplicate tests were deleted and a comment now points at the existing class instead.

Row 3 means the CLI-side direction was already partly covered. The class docstring says so rather than over-claiming; what this adds there is parsing the argv the daemon *actually built* rather than one restated in a test.

## A finding, not acted on here

`subprocess_factory_runner`'s docstring states that under caffeinate "the direct child is the `caffeinate` wrapper, so the factory itself is a grandchild", and the process-group termination path is built around that. Measured on macOS 25.5: `caffeinate -i cmd` **execs in place**. The target keeps the Popen pid, so the factory is the daemon's *direct* child and there is no wrapper process.

That is pinned here as a canary with the discrepancy noted in the test docstring, and deliberately not acted on - the grandchild reading is load-bearing for the timeout path and deserves its own investigation. Filed separately.

## Verification

- `uv run pytest tests/ -q` -> **3140 passed, 28 skipped**, 0 failed
- `uv run ruff check kstrl/ tests/` -> clean
- `uv run mypy kstrl/ --strict` -> clean (tests are outside mypy scope per `pyproject` `files = ["kstrl"]`)

`docs/continuous-intake.md` section 7 previously said a serve-to-factory regression "would not be caught by CI - only by another live run". Updated to say what is now covered and what still is not.

**Not tested:** anything below `ks factory`'s argument parsing. No factory is run, by design - a suite that spawned real runs would cost dollars per assertion. Classification of a real run's on-disk artifacts remains stub-driven, and the full end-to-end path is still hand-verified only.

Per H1 I have not run `/code-review` on this. No adversarial prompt changed, so H2/H3 do not apply.

### DIFF
--- a/tests/test_serve_seam.py
+++ b/tests/test_serve_seam.py
@@ -1,0 +1,489 @@
+"""R8.6 follow-up (#205): the serve-to-factory seam, executed for real.
+
+Every other serve test injects a stub ``runner=`` - 46 of them in
+``tests/test_serve.py`` alone - and ``subprocess_factory_runner`` is
+only ever patched out. The supervision half is genuinely tested
+(``run_supervised`` against a real process group), but the half that
+decides **what to invoke** was never executed: the argv, the cwd, the
+env, and whether remote work reaches the queue at all.
+
+That is exactly where #189 F1 lived. ``serve_cycle`` drained a queue
+nothing could fill because it never called the intake adapter, and both
+halves were independently correct and independently green. No
+stub-runner test could see it, because the stub replaced the boundary
+the defect lived on.
+
+Three checks, none of which spends money:
+
+1. :class:`TestTheRealRunnerExecsItsArgv` runs the shipping
+   ``subprocess_factory_runner`` against a stub interpreter. Only
+   ``sys.executable`` is replaced - argv construction, the caffeinate
+   prefix, the env mutation, the process-group spawn and the
+   ``RunOutcome`` mapping are all production code.
+2. :class:`TestTheArgvIsAcceptedByTheRealCli` hands that recorded argv
+   to the real Click command, so a flag the daemon sends but the CLI no
+   longer accepts is caught here rather than on the next unattended run.
+3. :class:`TestRemoteWorkSurvivesTheSeam` covers the two facts about a
+   remotely-sourced item that only become observable at the runner
+   boundary: the spec content, and the merge gate.
+
+Measured, by mutation, against the rest of the suite (3125 tests):
+
+- Renaming the merge-gate flag in ``subprocess_factory_runner`` is
+  caught by four tests here and by NOTHING else: with that mutant
+  applied the other 3125 tests pass, 28 skip, zero fail. That is the gap
+  this module exists to close.
+- Deleting the ``_run_intake`` call from ``serve_cycle`` - the #189 F1
+  defect itself - is already caught by nine tests in
+  ``tests/test_intake_github.py::TestServePollsIntake``, so this module
+  deliberately does not re-assert it.
+- Renaming the option on the ``ks factory`` side is caught both here and
+  by ``test_config_control_plane.py::TestSafetyKnobs``.
+
+What is NOT covered, so nobody reads more into it: no factory is run, so
+nothing below ``ks factory``'s argument parsing is exercised, and the
+classification of a real run's artifacts remains the stub-driven tests'
+job. The end-to-end path is still only verified by hand
+(``docs/continuous-intake.md`` section 7).
+"""
+
+from __future__ import annotations
+
+import json
+import stat
+import sys
+from collections.abc import Callable
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Any
+from unittest.mock import patch
+
+import click
+import pytest
+
+from kstrl.serve import RunOutcome, serve_cycle, subprocess_factory_runner
+from tests.test_intake_github import REPO, _GhStub, _issue, _issue_payload
+
+# --------------------------------------------------------------------------
+# A stub interpreter: the narrowest possible intercept
+# --------------------------------------------------------------------------
+
+#: Stands in for ``sys.executable``. Records what the real runner asked
+#: for, then exits how the test told it to. Deliberately dependency-free
+#: and tiny: it is spawned by the code under test, so a failure inside it
+#: surfaces as a confusing RunOutcome rather than a test error.
+_STUB_INTERPRETER = """#!/usr/bin/env python3
+import json, os, sys
+with open(os.environ["SEAM_RECORD"], "w") as handle:
+    json.dump({
+        "argv": sys.argv[1:],
+        "cwd": os.getcwd(),
+        "pid": os.getpid(),
+        "kstrl_env": {
+            k: v for k, v in os.environ.items() if k.startswith("KSTRL_")
+        },
+    }, handle)
+sys.stdout.write(os.environ.get("SEAM_STDOUT", ""))
+sys.exit(int(os.environ.get("SEAM_EXIT", "0")))
+"""
+
+
+@dataclass(frozen=True)
+class _Exec:
+    """What the child process actually received."""
+
+    argv: tuple[str, ...]
+    cwd: Path
+    pid: int
+    kstrl_env: dict[str, str]
+
+    def value_of(self, flag: str) -> str:
+        """The argument following ``flag``."""
+        return self.argv[self.argv.index(flag) + 1]
+
+
+def _exec_real_runner(
+    tmp_path: Path,
+    monkeypatch: pytest.MonkeyPatch,
+    *,
+    pause_before_pr_merge: bool = False,
+    caffeinate: bool = False,
+    exit_code: int = 0,
+    stdout: str = "",
+    timeout_seconds: float = 60.0,
+    on_spawn: Callable[[int], None] | None = None,
+    project_name: str = "seam-project",
+) -> tuple[RunOutcome, _Exec]:
+    """Run the REAL ``subprocess_factory_runner`` against a stub interpreter.
+
+    ``sys.executable`` is the only thing replaced. Everything the runner
+    decides - which module, which flags, which cwd, which env - is left
+    to the shipping code and read back out of the child.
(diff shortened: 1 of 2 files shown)


tests/test_serve_seam.py
```diff
@@ -0,0 +1,489 @@
...
+        """The seam is only real if the content survives it. Existing
+        tests assert the item reaches DONE, which a cycle that handed the
+        factory an empty or missing spec would also satisfy."""
+        _enable_github_intake(tmp_path)
+        gh = _GhStub(issues=_issue_payload(
+            _issue(7, title="Add a widget", body="Build the widget."),
+        ))
+        calls: list[dict[str, Any]] = []
+        with patch("kstrl.intake_github.run_gh", gh):
+            serve_cycle(tmp_path, runner=_recording_runner(calls))
```
0xfauzi, line 668: [P2] Exercise the default runner through `serve_cycle`

The subprocess tests call `subprocess_factory_runner` directly, while both composition tests inject `_recording_runner`. Consequently, `_default_runner` remains outside the tested seam. Incorrect forwarding of `spec_path`, `project_name`, the merge gate, timeout, or `caffeinate` would leave every added test green while the daemon launches the wrong command. Please add a cycle test without `runner=`, replacing only `sys.executable`.

tests/test_serve_seam.py
```diff
@@ -0,0 +1,489 @@
...
+            spec_path=tmp_path / "spec.md",
+            project_name="p",
+            pause_before_pr_merge=False,
+            timeout_seconds=30.0,
+            caffeinate=False,
+        )
+        assert outcome.returncode == -1
+        assert outcome.launch_error
+
+    @pytest.mark.skipif(
```
0xfauzi, line 386: [P2] Make the caffeinate runner test execute in CI

Both test workflows use Ubuntu, so this test is always skipped in CI, and every other `_exec_real_runner` call uses `caffeinate=False`. Removing the caffeinate prefix from `subprocess_factory_runner` would therefore pass CI. A platform-neutral fake caffeinate wrapper could verify the runner wiring, while retaining this macOS test for actual process topology.

0xfauzi (commented):
Two actionable test-coverage gaps remain in the serve-to-factory seam.

0xfauzi:
Both findings confirmed by mutation before fixing, and both mutants passed the **entire 3140-test suite** beforehand. Fixed in 6df4216.

### [P2] Exercise the default runner through `serve_cycle`

Correct, and `_default_runner` had **zero references anywhere in `tests/`**. Mutant: make it forward a hardcoded wrong `project_name` -> 3140 passed, 28 skipped, 0 failed.

Added `TestACycleWithNoInjectedRunnerLaunchesTheRealCommand`: five tests calling `serve_cycle` with no `runner=` at all, replacing only `sys.executable`. Claim -> gate resolution -> `_default_runner` -> `subprocess_factory_runner` -> `run_supervised` now all run as shipped. Covers the project name, the queued spec path, the merge gate, and `serve.caffeinate` in both directions.

### [P2] Make the caffeinate wiring execute in CI

Also correct, and it is sharper than stated. Deleting `*caffeinate_prefix(caffeinate)` passes 3140 tests **on macOS as well as on ubuntu** - so my `skipif` was not the only problem, the test was weak on the platform where it did run.

The reason is the exec-in-place finding from this PR (#209): `caffeinate -i` replaces itself with the utility, so the wrapper leaves the child's pid, argv and exit status all identical. Its absence is unobservable from the child, which is exactly what my canary was asserting over.

Fix is your suggestion: a fake `caffeinate` on `PATH` that touches a marker before `exec "$@"`, so presence in the chain is observable, with `sys.platform` patched to darwin so it runs on ubuntu. It execs in place like the real one, so the pid assertions stay faithful to production. The macOS test is kept and narrowed to what only it can check - the real binary's process topology. Verified the platform patch is load-bearing: set it to `"linux"` and both caffeinate tests fail.

### Mutation results after the fix

| Mutant | Before | After |
|---|---|---|
| A - `_default_runner` forwards a wrong `project_name` | 3140 pass | 1 failure |
| B - caffeinate prefix dropped from the command | 3140 pass | 2 failures |
| C - `_default_runner` ignores `cfg.caffeinate` | 3140 pass | 1 failure |
| D - merge-gate flag renamed in the runner | 3125 pass | 5 failures |

22 tests in the file now. `uv run pytest tests/ -q` -> **3147 passed, 28 skipped**; mypy --strict and ruff clean. `docs/continuous-intake.md` section 7 updated with the mutation table and the exec-in-place caveat.

Unchanged: nothing below `ks factory`'s argument parsing is exercised, by design.

PR #211, opened 2026-08-10
fix(cli): notify when ks run overrides configured structural knobs

### LINKED ISSUE #207: ks run silently forces create_prs=false, so pause_before_pr_merge never runs
Found on 2026-08-03 during the first external factory run (deckgen, PR #1). The project's `kstrl.toml` set `[factory] pause_before_pr_merge = true`. The gate never ran, and nothing said so.

## The defect

`ks run` forces `create_prs = False` unconditionally:

```python
# kstrl/cli.py:632
factory_cfg.create_prs = False
```

The HITL merge checkpoint is gated on exactly that flag:

```python
# kstrl/pipeline.py:1778
if self.factory_config.create_prs and not self.factory_config.single_pr:
    checkpoint = self._phase_checkpoint(comp, diff_text=diff.review_diff)
```

So under `ks run`, `_phase_checkpoint` is unreachable. `pause_before_pr_merge` is read from `kstrl.toml` (`factory.py:346`), stored on the config, carried into the run, and then silently never consulted. The operator sees no warning: `build_config_report` is the only surface that would show the override and it is wired into one call site (`cli.py:2372`), not into `ks run`.

## Why this is worse than a UX papercut

The forcing itself is defensible and documented in the comment at `cli.py:620-623` - `ks run` is by definition a local, single-component, no-PR invocation, so there is no PR to gate. The defect is that **a governance control fails open and stays quiet about it.**

R8.3 went to real trouble to make this same gate fail *closed* on the adjacent path: when `pause_before_pr_merge` is set but no UI can prompt, `pipeline.py:2919-2930` parks the component, files a `merge_gate` inbox item, and warns - explicitly "automation fails loudly rather than blocking indefinitely OR merging something nobody approved". The `ks run` path bypasses that reasoning entirely by never reaching the function that contains it.

The asymmetry is the bug. One path treats an unanswerable gate as a halt; the other treats an inapplicable gate as nothing at all. An operator who set the flag has no way to tell which path they got.

## Contributing factor

The two commands are easy to confuse for this purpose. `ks run` is the natural thing to reach for on a single component, and it is what produced the deckgen run; `ks factory` is what actually honours the PR + HITL flow. Nothing at the point of use distinguishes them on this axis.

## Options

1. **Warn at startup** (smallest): when `ks run` forces off a structural field that the resolved config set to a non-default, print which knobs were overridden and why. Generalises past this one flag - `max_parallel`, `use_worktrees`, `single_pr`, and `contract_config` are forced the same way.
2. **Refuse** when `pause_before_pr_merge` is explicitly `true` in `kstrl.toml`, directing the operator to `ks factory`. Fails closed, consistent with R8.3, but breaks a working `ks run` for anyone whose project-level toml is tuned for `ks factory`.
3. **Reuse the config report**: run `build_config_report` with the `ks run` overlay and surface the drift rows. More machinery than the problem needs, but no new drift surface to maintain.

Option 1 looks right, with the honesty note stated as "gate not applicable, no PR is created" rather than a bare "overridden". A halting variant is worth considering separately once `[autonomy]` at L2+ is a supported configuration, since the ladder forces this flag ON and would inherit the same silent bypass.

## Acceptance

- `ks run` in a repo whose `kstrl.toml` sets `pause_before_pr_merge = true` emits a visible notice naming the flag and the reason it does not apply.
- A test asserts the notice, and asserts it does NOT fire when the flag is unset (so it does not become background noise).
- `docs/` states which commands honour the merge gate.

## Not verified

Whether the other forced fields at `cli.py:629-637` have the same silent-override problem in a way an operator would care about. `create_prs` is the one with a governance control hanging off it; the rest look like plain performance/structure knobs, but I did not trace each one.

0xfauzi:
Closes #207.

## Summary

- `ks run` forces `create_prs = False`, which makes the `pause_before_pr_merge` merge gate unreachable — a governance control failing open with no warning (found on the first external factory run). This implements Option 1 from the issue: a startup notice, emitted before the forcing block mutates the config.
- `_run_structural_override_notices` compares the resolved config (kstrl.toml + env) against the `FactoryConfig` dataclass defaults for the four forced structural fields (`max_parallel`, `use_worktrees`, `single_pr`, `create_prs`) and warns only when the operator actually set a knob to a non-default value that `ks run` forces to something else — an unset knob stays silent, so the notice never becomes background noise.
- `pause_before_pr_merge` is not forced itself but hangs off `create_prs`, so it gets a dedicated honesty note phrased per the issue: "no PR is created, so the merge gate is not applicable. Use `ks factory` to honour the merge gate."
- Docs (`docs/env-vars.md`, `docs/adversarial-design.md`) now state which commands honour the merge gate: `ks factory` and `ks serve` (via its factory invocations); `ks run` never.

## Acceptance criteria from #207

- [x] `ks run` in a repo whose `kstrl.toml` sets `pause_before_pr_merge = true` emits a visible notice naming the flag and the reason it does not apply
- [x] A test asserts the notice, and asserts it does NOT fire when the flag is unset
- [x] `docs/` states which commands honour the merge gate

## Tested vs assumed (H4)

- Tested: notice content and firing/non-firing conditions (4 pure-function tests), plus two end-to-end `CliRunner` invocations of `ks run` with and without the toml flag. Full suite 3153 passed / 28 skipped; `mypy --strict` and `ruff` clean.
- Not covered: the `contract_config = None` forcing — Phase 3 only runs on merged tier branches, which cannot exist in a single-component no-PR run, so it is genuinely inapplicable rather than silently overridden.
- No adversarial prompts touched, so no calibration re-run required (H2).

Made with [Cursor](https://cursor.com)


0xfauzi (commented):
Adversarial review found one governance-path defect plus two test/documentation gaps. The autonomy ordering issue is the merge-blocking item: L1/L2 can report the merge gate as enabled after `ks run` has made the checkpoint unreachable.

kstrl/cli.py
```diff
@@ -626,6 +685,11 @@ def run(
     # ran anyway. Feedforward is independent of --no-verify (it builds
     # context, not checks).
     factory_cfg = FactoryConfig.load(root_dir)
+    # Issue #207: say which configured knobs the forcing below overrides,
+    # BEFORE mutating the config, so the operator learns at startup that
+    # (in particular) the pause_before_pr_merge gate will not run.
+    for notice in _run_structural_override_notices(factory_cfg):
```
0xfauzi, line 686: [P1] This check runs before `run_factory` resolves the autonomy ladder. With autonomy enabled at L1/L2 and no explicit factory flag, `FactoryConfig.load()` still has `pause_before_pr_merge = false`, so no notice is emitted here. `run_factory` then flips the flag to `true` while `create_prs` remains forced to `false`; the autonomy output reports the merge gate as on, but `_phase_checkpoint` is unreachable. That recreates the fail-open/misreport path this PR is meant to close. Please resolve autonomy before generating these notices, or emit a post-resolution warning whenever the final gate is true while PR creation is disabled, and add an L1/L2 `ks run` regression test.

tests/test_cli.py
```diff
@@ -175,6 +176,92 @@ def test_feature_uses_root_option(self, tmp_path: Path, monkeypatch) -> None:
...
+                "0",
+                "--root", str(project),
+                "--agent-cmd", "printf '<promise>COMPLETE</promise>\\n'",
+                "--sleep", "0",
+                "--no-verify",
+                "--ui", "plain",
+                "--no-color",
+            ],
+        )
+        return result.output or ""
```
0xfauzi: [P2] These end-to-end tests discard `result.exit_code`. A `ks run` invocation can print the startup notice and then fail, yet both output assertions still pass. Please assert `result.exit_code == 0, result.output` in this helper (or return the full `Result` and assert it in each test) before checking warning content.

docs/adversarial-design.md
```diff
@@ -19,7 +19,7 @@ Calibration is the truth signal. The `tests/test_calibration.py` suite (Phase D
...
-| **Human checkpoint (E6)** | `kstrl/factory.py` | (interactive) | Pre-merge | Optional. When `FactoryConfig.pause_before_pr_merge=True` and UI is interactive, prompts a human to approve or reject before PR creation. Off by default. |
+| **Human checkpoint (E6)** | `kstrl/factory.py` | (interactive) | Pre-merge | Optional. When `FactoryConfig.pause_before_pr_merge=True` and UI is interactive, prompts a human to approve or reject before PR creation. Off by default. Honoured only by PR-creating commands (`ks factory`, and `ks serve` via its factory invocations); `ks run` creates no PRs, so the gate does not apply there and a startup notice says so when the flag is set (#207). |
```
0xfauzi: [P2] This overstates the command-level guarantee. `ComponentPipeline` only calls `_phase_checkpoint` when `create_prs` is true **and** `single_pr` is false; single-PR mode later calls `create_single_pr` directly without a checkpoint. An operator using `ks factory` with `single_pr = true` and `pause_before_pr_merge = true` therefore does not get the documented prompt before the aggregate PR is created. Please qualify both new docs for single-PR mode, or add an aggregate checkpoint plus regression coverage.

kstrl/cli.py
0xfauzi: Fixed in 5ed4964. The warning now lives in `factory.merge_gate_unreachable_warning` and is emitted by `run_factory` immediately after autonomy resolution (the point that sees the final flags on every path), per your second option. The CLI-level notices keep only the four structural fields and no longer special-case the gate, so the same condition is not warned twice on the explicit-toml path. Regression test added: `test_run_warns_when_autonomy_ladder_flips_gate_on` runs `ks run` with `[autonomy] enabled = true` and no factory flag, and asserts the post-resolution warning fires. Note it runs 0 iterations and asserts exit 1 deliberately: the L1 bundle also forces review_mode=hard, so a green run would require a real reviewer LLM call. A halting (rather than warning) variant for autonomy-on runs is worth its own issue once L2+ is a supported configuration, as #207 anticipated.

tests/test_cli.py
0xfauzi: Fixed in 5ed4964. The helper now returns the full `Result` and both green-path tests assert `result.exit_code == 0, result.output` before any output grep. Making that assertion honest required two scaffold changes: the project is now a real git repo (the diff phase runs `git diff` against the base branch and previously failed with exit 129, so every prior invocation exited 1), and the toml sets `review_mode = "skip"` because a completing run otherwise auto-detects a real agent CLI for the Phase 2 review and makes a paid LLM call (measured: ~35k tokens / $0.47 per test run). The autonomy regression test asserts `exit_code == 1` explicitly with the reason documented, rather than discarding it.

docs/adversarial-design.md
0xfauzi: Confirmed and fixed in 5ed4964: `_phase_checkpoint` requires `create_prs and not single_pr` (pipeline.py) and `create_single_pr` has no checkpoint. I took the qualify-the-docs option, and additionally extended `merge_gate_unreachable_warning` to warn when the final config has the gate on with `single_pr = true` - so the docs statement is both accurate and enforced loudly at runtime (`test_warns_in_single_pr_mode` covers it). Adding an aggregate checkpoint before `create_single_pr` is a behavioural change to single-PR mode beyond this PR's scope; happy to file it as a follow-up issue if you want the gate honoured there rather than warned about.

0xfauzi (commented):
## Post-merge adversarial pass

**Verdict:** Safety-critical path looks solid. Post-autonomy \`merge_gate_unreachable_warning\` correctly covers ks-run forcing, L1/L2 ladder flip, and \`single_pr\`.

### Residual (non-blocking)

1. **P3** — \`kstrl/factory.py\` (\`merge_gate_unreachable_warning\`): the \`create_prs=False\` message always ends with “not \`ks run\`”, but the helper runs on every \`run_factory\` entry point. On \`ks factory --no-prs\` / toml \`create_prs = false\` the remediation misidentifies the invocation. Prefer a generic “enable \`create_prs\` / disable \`single_pr\`” line.
2. **P3** — Warning is emitted after DAG validation early-return, so a broken manifest exits before the #207 honesty notice.

No reopen needed unless you want the messaging polish.

PR #212, opened 2026-08-10
fix(serve): inbox open-item cap fails closed on unparseable lines

### LINKED ISSUE #190: R8.3 follow-up: serve's inbox open-item cap fails open on unparseable inbox lines
## Problem

`check_inbox_cap` (`kstrl/serve.py:1398`) gates queue admission on the fold of `.kstrl/inbox.jsonl`, and that fold skips malformed lines by design (`kstrl/inbox.py:359-363`: "tolerate a torn tail; skip, never raise"). A skipped **emission** line undercounts open items, so the cap admits work past N: the cap fails OPEN.

This is the exact defect class the repo has already corrected twice: R8.1 review correction 2 (changed-file reads were fail-open, so a `kstrl.toml` diff evaluated as "0 files, 0 lines" and passed every rule) and the serve retry classifier's own doctrine (positive evidence only, failing closed). A tolerant projection is fine for display; it must not be a fail-open safety gate.

Note the direction asymmetry: a skipped **decision** line overcounts open items, which fails closed and is safe. Only the emission-undercount direction is the defect.

## Fix options

1. In `check_inbox_cap` only, count unparseable lines as open (keep the tolerant fold for `ks inbox` display).
2. Have the fold report a parse-error count and have `Admission` refuse (with a `pause_reason`) whenever the count consulted for the cap saw parse errors.

Either is small. Option 1 keeps the display path untouched.

## Acceptance

- A test appends a garbled line to `inbox.jsonl` and proves admission is refused (or the open count includes the garbled line).
- The tolerant display path (`ks inbox`) is unchanged.
- The R8.6 roadmap note documenting this defect is updated to point at the fix.

Filed from the 2026-08-03 architecture-review session (user-approved item P8).

0xfauzi:
Closes #190.

## Summary

- `check_inbox_cap` gated queue admission on the tolerant inbox fold, which skips garbled lines by design - so a torn emission line undercounted open items and the cap admitted work past N, a safety gate failing open (the defect class already corrected twice: R8.1 review correction 2, and the serve retry classifier's fail-closed doctrine).
- Fix is the issue's option 1: new `Inbox.unparseable_line_count()` reports every nonempty line the fold would skip (torn JSON, non-dict lines, dicts `from_dict` cannot rebuild, plus an existing-but-unreadable log counting as one), and `check_inbox_cap` adds it to the open total. Every line that MIGHT be an open item counts as one.
- Only the gate pays the stricter price: `_folded`, `over_cap`, and the `ks inbox` display path are behaviorally unchanged. When garbled lines contribute to a refusal, the reason names the count and points at `.kstrl/inbox.jsonl`.
- The R8.6 roadmap note that documented the defect now records the fix (issue acceptance criterion).

## Tested vs assumed (H4)

Tested (TDD - all 7 new tests were watched failing against the old code first; the gate tests failed with `allowed=True`, the exact fail-open defect):
- `tests/test_inbox.py::TestUnparseableLineCount`: torn JSON, non-dict lines, id-less dicts, blank lines, missing/clean logs - each also asserting the display fold stays tolerant.
- `tests/test_serve.py::TestInboxCapGate`: a garbled line tips the same backlog from admitted to refused; garbled lines alone can fill the cap even when the fold reads an empty inbox; refusal reason says "unparseable".
- Full suite: 3165 passed, 28 skipped. `uv run mypy kstrl/ --strict` clean. `uv run ruff check kstrl/ tests/` clean.

Assumed, not tested: the OSError branch (existing-but-unreadable log counts as one skip) has no dedicated test - portably forcing a read failure in-suite is awkward; it is one line following the same fail-closed rule.

No prompts touched, so no calibration run required (H2 does not apply).

Made with [Cursor](https://cursor.com)

### DIFF
--- a/kstrl/inbox.py
+++ b/kstrl/inbox.py
@@ -328,6 +328,40 @@ class InboxConfig:
         )
 
 
+@dataclass(frozen=True)
+class InboxScan:
+    """One pass over ``.kstrl/inbox.jsonl``.
+
+    ``unreadable`` means the gate has no positive evidence the backlog
+    is under its cap - admission must refuse regardless of
+    ``open_item_cap``. Display callers ignore the flag and treat the
+    (empty) records as a clear inbox, same as the pre-#190 fold.
+    """
+
+    records: tuple[dict[str, Any], ...] = ()
+    skipped_lines: int = 0
+    unreadable: bool = False
+
+    def folded_items(self) -> list[InboxItem]:
+        folded: dict[str, InboxItem] = {}
+        for record in self.records:
+            item = InboxItem.from_dict(record)
+            if item is None:
+                continue
+            folded[item.id] = item
+        return list(folded.values())
+
+    def open_count(self) -> int:
+        return sum(1 for item in self.folded_items() if item.is_open)
+
+    def unparseable_count(self) -> int:
+        """Lines the display fold would skip: torn JSON, non-dicts, and
+        dicts ``InboxItem.from_dict`` cannot rebuild."""
+        return self.skipped_lines + sum(
+            1 for record in self.records if InboxItem.from_dict(record) is None
+        )
+
+
 class Inbox:
     """Append-only store over ``.kstrl/inbox.jsonl``.
 
@@ -345,24 +379,35 @@ class Inbox:
         return self.root_dir / INBOX_FILENAME
 
     # -- reading -----------------------------------------------------------
-    def _scan(self) -> tuple[list[dict[str, Any]], int]:
-        """Parse the log: (readable records, skipped-line count).
+    def scan(self) -> InboxScan:
+        """One pass over the log: readable records + skip count + readability.
 
         Tolerant by design for the DISPLAY path: a torn tail must not
-        make the whole backlog unreadable. The skip count exists because
-        that tolerance must not leak into safety gates (#190) - see
-        ``unparseable_line_count``. An existing-but-unreadable file
-        counts as one skip for the same reason: the gate has no positive
-        evidence the backlog is clear.
+        make the whole backlog unreadable. Safety gates (#190) must not
+        sit on that tolerance - they consume this snapshot once and treat
+        ``unreadable`` or every skipped line as open capacity. An
+        existing-but-unreadable file (OSError on exists/read, or a
+        UnicodeDecodeError on a torn multibyte write) is its own state:
+        the gate has no positive evidence the backlog is under the cap,
+        so collapsing it to ``skipped=1`` would re-admit under any cap
+        greater than one.
         """
-        if not self.path.exists():
-            return [], 0
+        try:
+            exists = self.path.exists()
+        except OSError:
+            return InboxScan(unreadable=True)
+        if not exists:
+            return InboxScan()
+        try:
+            raw = self.path.read_bytes()
+        except OSError:
+            return InboxScan(unreadable=True)
+        try:
+            text = raw.decode("utf-8")
+        except UnicodeDecodeError:
+            return InboxScan(unreadable=True)
         records: list[dict[str, Any]] = []
         skipped = 0
-        try:
-            text = self.path.read_text(encoding="utf-8")
-        except OSError:
-            return [], 1
         for line in text.splitlines():
             line = line.strip()
             if not line:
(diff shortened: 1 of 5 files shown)


kstrl/inbox.py
```diff
@@ -345,25 +345,55 @@ def path(self) -> Path:
...
         if not self.path.exists():
-            return []
+            return [], 0
         records: list[dict[str, Any]] = []
+        skipped = 0
         try:
             text = self.path.read_text(encoding="utf-8")
         except OSError:
-            return []
+            return [], 1
```
0xfauzi: **P1 — unreadable inbox still admits below the cap**

`_scan()` collapses every whole-file read failure to one skipped line. With the default cap of 50, both scans produce zero open items plus one garbled item, so `check_inbox_cap` evaluates `1 < 50` and admits work even if the unreadable file contains more than 50 open items. I reproduced this by making `Path.read_text` raise `OSError`; `check_inbox_cap(...).allowed` was `True`.

Please preserve an explicit unreadable/error state and make the admission gate deny independently of the configured cap. Add a regression test with a cap greater than one. Also avoid treating an uncertain `Path.exists()` result as a definitely missing file, because that path can swallow an `OSError` too.

kstrl/inbox.py
```diff
@@ -345,25 +345,55 @@ def path(self) -> Path:
...
+        evidence the backlog is clear.
+        """
         if not self.path.exists():
-            return []
+            return [], 0
         records: list[dict[str, Any]] = []
+        skipped = 0
         try:
             text = self.path.read_text(encoding="utf-8")
         except OSError:
```
0xfauzi, line 397: **P1 — invalid UTF-8 crashes admission instead of failing closed**

A write torn inside a multibyte character makes `read_text(encoding="utf-8")` raise `UnicodeDecodeError`, which is not an `OSError`. The exception escapes both the tolerant display fold and `check_inbox_cap`; I reproduced it with `inbox.jsonl` containing `b"\xff\n"`.

Please handle decode failures in the same fail-closed path (or decode with replacement so the damaged line is counted as unparseable), and add byte-level invalid UTF-8 tests for both display and admission behavior.

kstrl/serve.py
```diff
@@ -1530,20 +1530,34 @@ def check_inbox_cap(root_dir: Path) -> Admission:
...
     """
     from kstrl.inbox import Inbox, InboxConfig
 
     config = InboxConfig.load(root_dir)
     if not config.enabled or config.open_item_cap <= 0:
         return Admission(allowed=True)
     box = Inbox(root_dir, config)
-    if not box.over_cap():
+    garbled = box.unparseable_line_count()
+    if len(box.open_items()) + garbled < config.open_item_cap:
```
0xfauzi: **P1 — two scans let a concurrent torn append escape the cap**

`unparseable_line_count()` and `open_items()` read the append-only log independently. If a torn cap-filling line appears after the first scan, the first count reports zero garbled lines and the second scan silently skips the new torn line. At cap 2 with one readable open item, I injected a torn append between these calls and admission still returned `allowed=True`.

Please derive the readable-open and unparseable counts from one `_scan()` snapshot. This also removes the new repeated full-log parse on every serve cycle. Add a deterministic interleaving regression test.

0xfauzi (commented):
Not ready: three P1 failure modes remain in the fail-closed gate. I reproduced each against this commit. The focused PR tests pass (7/7), but they do not cover whole-file read failure, invalid UTF-8, or an append between the gate's two scans.

kstrl/inbox.py
0xfauzi: Addressed: `Inbox.scan()` now returns an `InboxScan` with an explicit `unreadable` flag. `exists()` / `read_bytes()` OSError (and UTF-8 decode failure) set that flag; `check_inbox_cap` refuses on `unreadable` regardless of `open_item_cap`. Regression: `test_unreadable_inbox_refuses_regardless_of_cap` (cap=50) and `test_exists_oserror_is_unreadable`.

kstrl/inbox.py
0xfauzi: Addressed: the scan reads bytes then `decode(\"utf-8\")`, and `UnicodeDecodeError` takes the same `unreadable` fail-closed path as OSError (replacement-decoding a damaged line as one skip would still admit under any cap > 1). Display stays tolerant (`items()` returns `[]`, no raise). Regressions: `test_invalid_utf8_is_unreadable_and_display_tolerant` and `test_invalid_utf8_refuses_admission`.

kstrl/serve.py
0xfauzi: Addressed: `check_inbox_cap` now takes one `Inbox.scan()` snapshot and derives both `open_count()` and `unparseable_count()` from it. Regressions: `test_gate_uses_one_scan_snapshot` (asserts a single scan + refusal when the snapshot already has the torn line) and `test_interleaved_append_cannot_split_gate_counts` (injects a torn append after the first scan returns; the gate decides from that snapshot alone and does not re-scan).

0xfauzi (commented):
Re-reviewed `1b82c23` against the three prior P1s.

- Unreadable file is now an explicit `InboxScan.unreadable` state; `check_inbox_cap` refuses independently of cap (verified at cap=50).
- Invalid UTF-8 takes the same unreadable path via `read_bytes` + `decode` (verified; no raise).
- Open/unparseable counts come from one `scan()` snapshot (verified `scans == 1`; interleaving test pins the race).

Focused suite: 14 passed. Local mypy/ruff clean. CI green. No remaining blockers — merging.

(GitHub blocks self-approval on own PRs; this is the merge gate.)

0xfauzi (commented):
## Post-merge adversarial pass

**Verdict:** The #190 fail-open undercount is fixed correctly — single \`Inbox.scan()\` snapshot, garbled lines count toward the cap, unreadable log refuses regardless of cap.

### Residual (worth a follow-up with R8.9)

**P2** — \`kstrl/serve.py\` \`check_inbox_cap\` refusal strings still hard-code \`.kstrl/inbox.jsonl\` (unreadable + unparseable detail). After R8.9 the live log is under \`$XDG_STATE_HOME/kstrl/<repo-id>/inbox.jsonl\`; operators blocked by the cap are sent to the wrong path. Interpolate \`Inbox(root_dir).path\` instead.

Cap logic itself does not fail open.

Issue #235, opened 2026-08-25
R10: Control loop - adopt a proven loop method (tracking)


0xfauzi:
> **Stabilisation queue: tracker** (set 2026-09-09; the ranked list is in #369). Tracking issue, not a unit of work. See #369 for the ranked queue.

Tracking issue for the R10 cycle: remodelling kstrl as a control system and adopting a proven loop method rather than re-deriving one. Design, with every claim cited to `file:line`: [`docs/control-loop-design.md`](https://github.com/0xfauzi/kstrl/blob/main/docs/control-loop-design.md) (PR #221). Milestone: [R10: Control Loop](https://github.com/0xfauzi/kstrl/milestone/3). Label: `r10-control-loop`.

## The frame in one paragraph

kstrl is six nested loops at different rates, not one pipeline; the phase chain everybody quotes is one tick of the middle two. Naming the parts (set point, sensor, controller, actuator, disturbance) makes three defects visible that the pipeline framing cannot express: the engineer agent is the **only writer** of `UserStory.passes` and `verify.check_prd_stories` merely reads it back (the actuator writes the sensor reading); the retry context is four append-only lists with no discharge (integrator windup); and no sensor can be run without a full paid factory run (`ks --help`: fourteen commands, none runs verification against a tree), which is why every threshold in the codebase is an unmeasured placeholder.

## The method, adopted

Two rules govern every item below:

1. **Local first.** Each loop component runs by hand and standalone before it is wired into automation. That is why R10.1 is first and why it makes the rest cheap.
2. **Advisory first, then graduate.** A gate ships measuring and reporting without blocking; it moves to blocking when the operator has seen its output on real runs and can name what it caught and what it flagged wrongly. That is a judgement, deliberately the operator's, and no issue below encodes a number in its place. kstrl already has this tier (`review_mode` hard/advisory/skip, `[adequacy] layer0` advisory/block, and the doctrine written at `kstrl/findings.py:165-179`); R10 applies it consistently.

**No item edits an adversarial prompt body**, so H2/H3 are not engaged anywhere in this cycle. No item adds a phase (doctrine 5) or renames an existing enum or module (doctrine 6). One item adds a value to an existing enum: #232 adds `HEALTH_BREACH` to the inbox's `ItemKind`, together with its emitter, which that enum's own contract permits ("adding a kind means adding an emitter, never just a label"); doctrine 6 forbids new terminal-outcome vocabularies, not an inbox kind with an emitter.

## Sequence

[...]

PR #237, opened 2026-08-25
feat(sense): run the mechanical sensors standalone against any tree (R10.1)

### LINKED ISSUE #222: R10.1: `ks sense`: run the mechanical sensors standalone against any tree
Part of #235 (R10: Control Loop tracking issue). Item R10.1.

Design section: 5.1. Sizing: M (one PR). Dependencies: none. This is item 1 of the R10 order because every later item gets cheaper once a sensor runs by hand.

## Why

kstrl has no way to run a sensor without a full factory run. `ks --help` lists fourteen commands; none runs mechanical verification against a working tree. Every sensor is reachable only by starting a factory invocation, which needs a PRD, cuts a branch, creates a worktree, spends money on an agent, and takes minutes.

The control-loop method treats local-first as a hard gate: proceed to automation only once each piece runs locally on its own. kstrl built the orchestrator first. This is why every threshold in the codebase is an unmeasured placeholder: you cannot cheaply measure a sensor you cannot run.

`run_mechanical_verification` (`kstrl/verify.py:1375`) already takes a path and returns a `VerificationResult`. This issue is a command wrapper over a function that exists.

## What to build

A new CLI command `ks sense` that runs the mechanical sensors against a tree and prints the measurement. Mechanical only: no LLM call, no spend, no branch, no worktree.

### 1. Make `run_mechanical_verification` runnable without a PRD

In `kstrl/verify.py`:

- Change the signature `prd_path: Path` to `prd_path: Path | None` (line 1377).
- Line 1398, `checks.append(check_prd_stories(prd_path))`: run only when `prd_path is not None`.
- Line 1469, `checks.append(check_fixtures_from_prd(...))`: run only when `prd_path is not None` (fixtures are declared in the PRD; there is nothing to run without one).
- Every other check is unchanged. `check_diff_scope` already returns a passing `CheckResult` with message `"No scope constraints (allowed_paths not set)"` when `allowed_paths` is `None` (`kstrl/verify.py:720`, the `if not allowed_paths:` branch), so passing `None` is safe.
- All existing callers pass a real `Path`, so their behaviour is byte-identical. Add a docstring line stating that `None` skips the PRD-dependent checks.

### 2. Add the command

In `kstrl/cli.py`, add `@cli.command()` `sense`, modelled on `status` (`kstrl/cli.py:2689`, the option style and the `--root` / `--ui` / `--no-color` conventions). Options:

| Option | Type | Default | Meaning |
|---|---|---|---|
| `--root` | path | cwd | project root; `kstrl.toml` is read from here |
| `--path` | path | `--root` | the tree to measure (a worktree, a checkout, any directory) |
| `--base` | str | value of `[git] branch` in `kstrl.toml`, else `main` | base branch for diff-scope and bad-pattern checks |
| `--prd` | path | none | when given, `check_prd_stories` and the fixtures oracle also run |
| `--allowed-path` | str, repeatable | none | when given, diff-scope enforces these globs; when absent, diff-scope reports "no scope constraints" |
| `--json` | flag | off | machine output (schema below) instead of the table |
| `--ui` / `--no-color` | as `status` | | |

Behaviour:

1. Load `VerifyConfig.load(root)`, `PolicyConfig.load(root)`, `AdequacyConfig.load(root)`, and `FixturesConfig.load(root)` (the last only matters when `--prd` is given).
2. Call `run_mechanical_verification(worktree_path=path, prd_path=prd or None, base_branch=base, allowed_paths=list(allowed_path) or None, config=verify_cfg, policy_config=policy_cfg, adequacy_config=adequacy_cfg, fixtures_config=fixtures_cfg if prd else None, autonomy_level=0, component_id=None)`.
3. Human output: one row per check: name, `pass` / `FAIL`, message, duration in seconds with two decimals. Then one summary line `sense: PASS` or `sense: FAIL (<n> of <m> checks failed)`.
4. `--json` output, printed to stdout as one JSON document:

```json
{
  "schema_version": 1,
  "path": "<absolute path measured>",
  "base_branch": "main",
  "passed": false,
  "checks": [
    {"name": "linter", "passed": false, "message": "...", "details": ["..."],
     "duration_seconds": 1.23, "findings": [ <Finding.to_dict()> ]}
  ]
}
```

   `findings` uses the existing `Finding.to_dict()` (`kstrl/findings.py:86`). Details are the check's `details` list, untruncated.
5. Exit code: `0` when `passed` is true, `1` when false, `2` when the measurement itself could not run (path does not exist, is not a directory, or `kstrl.toml` fails to parse). Exit 2 prints one line starting with `error:` to stderr and, with `--json`, a document `{"schema_version": 1, "error": "<message>"}` to stdout.
6. Nothing is written to `.kstrl/` or anywhere else. `ks sense` is read-only on the tree except for whatever the configured test/typecheck/lint commands themselves write (their own caches).

### 3. Regenerate docs

The README CLI reference is generated. Run `uv run python scripts/gen_docs.py` and commit the README diff. Add a two-line mention of `ks sense` to the hand-written "Quick start" area of README.md (outside the generated markers) so a reader discovers it. Add `### Added` in CHANGELOG.

## Out of scope (do not do in this PR)

- An LLM `--review` flag over a diff. That is a follow-up filed only after this lands (tracked in the R10 tracking issue).
- Writing or comparing a baseline (that is R10.6, the dampener, which builds on this command).
- Any change to the factory pipeline. `ks sense` is a standalone entry point; the pipeline keeps calling `run_mechanical_verification` exactly as before.
- Touching the TUI.

## Tests to write

New file `tests/test_sense_cli.py`, using `click.testing.CliRunner` as `tests/test_cli.py:7-19` does. Build a temporary git repository in a `tmp_path` with one commit, a `pyproject.toml`, one trivial Python module, one trivial pytest test, and a `kstrl.toml` whose `[verify]` commands are fast (for example `test_command = "python -c 'print(1)'"`, `typecheck_command = "python -c 'print(1)'"`, `lint_command = "python -c 'print(1)'"`). Then:

1. `test_sense_passes_on_clean_tree`: `ks sense --root <tmp> --json` exits 0; JSON `passed` is true; the check names include `test_suite`, `typecheck`, `linter`, `bad_patterns` (read the exact names from `CheckResult.name` values in `kstrl/verify.py`; do not guess them).
2. `test_sense_reports_failure_and_exits_1`: set `lint_command = "python -c 'import sys; print(\"x.py:1:1: E501 line too long\"); sys.exit(1)'"`; exit code is 1; the `linter` check has `passed: false` and its `message` is non-empty.
3. `test_sense_skips_prd_checks_without_prd`: no `--prd`; the JSON check list contains no check named `prd_stories`.
4. `test_sense_runs_prd_checks_with_prd`: write a minimal PRD (copy the shape from an existing fixture under `tests/`, for example whatever `tests/test_verify.py` uses for `check_prd_stories`) with one story `passes: false`; with `--prd`, the `prd_stories` check is present and `passed` is false; exit code 1.
5. `test_sense_no_scope_constraints_without_allowed_path`: the `diff_scope` check is present with `passed: true` and its message contains `No scope constraints`.
6. `test_sense_enforces_allowed_path`: commit a change to `src/a.py`, pass `--allowed-path "docs/**"`; `diff_scope` has `passed: false`.
7. `test_sense_exit_2_on_missing_path`: `--path /nonexistent` exits 2 and stderr starts with `error:`; with `--json`, stdout parses and has an `error` key.
8. `test_sense_writes_nothing`: snapshot the directory listing of `<tmp>/.kstrl` (absent or empty) before and after; assert unchanged.

Also extend `tests/test_verify.py` with `test_run_mechanical_verification_without_prd`: calling `run_mechanical_verification(..., prd_path=None, ...)` returns a result whose check names exclude `prd_stories` and the fixtures check; calling with a real path behaves exactly as the existing tests already assert (do not weaken those tests).

## Acceptance (all must hold)

- [ ] `uv run ks sense --json` on this repository, run from the repo root, exits 0 and prints a JSON document with `schema_version: 1`. Paste the first 20 lines of the output into the PR body.
- [ ] `uv run ks sense --path /tmp/does-not-exist` exits 2.
- [ ] The eight new CLI tests and the new verify test pass.
- [ ] Every pre-existing test still passes (3200 passed, 28 skipped baseline, plus the new ones).
- [ ] `run_mechanical_verification` with a `Path` argument produces exactly the same check list it did before (the existing `tests/test_verify.py` assertions are the proof; none may be modified).
- [ ] README regenerated; `uv run python scripts/gen_docs.py --check` passes.
- [ ] Wall clock: `time uv run ks sense --json` on this repository, reported in the PR body. There is no threshold; the number is recorded so R10.6 and R10.12 can cite it.

## How to verify by hand

```bash
uv run ks sense --json | python -c "import json,sys; d=json.load(sys.stdin); print(d['passed'], [c['name'] for c in d['checks']])"
uv run ks sense --path /tmp/nope; echo "exit=$?"     # expect exit=2
```

0xfauzi:
## Summary

Closes #222 (R10.1). Adds `ks sense`: the same mechanical checks Phase 1 runs inside the factory, run by hand against any tree with no PRD, branch, worktree or agent spend, with `--json` output and exit codes 0 pass / 1 fail / 2 could-not-measure.

- `run_mechanical_verification(prd_path: Path | None, ...)`: `None` skips the PRD-dependent checks (`prd_stories`, the fixtures oracle, and `self_critique` unless `[verify] progress_file_path` names the log explicitly). Every existing caller passes a `Path` and is unchanged; no existing assertion in `tests/test_verify.py` was modified.
- New `sense` command in `kstrl/cli.py` modelled on `status`; `_detect_base_branch` extracted from `ks run` with byte-identical behaviour there.
- Ten new CLI tests (`tests/test_sense_cli.py`) and two verify tests.
- README CLI reference regenerated plus a two-line hand-written mention; CHANGELOG `### Added`.

## Deviations from the issue text, on purpose

1. `--base` defaults to the branch `origin/HEAD` points at, else `main` (the detection `ks run` already uses), not `[git] branch` as the issue said. In the code that key is `kstrl_branch`, the feature branch the loop commits to; diffing against it would compare the tree to itself. The issue was wrong; the command is right.
2. `self_critique` is a third PRD-dependent check the issue did not list: its log path is derived from the PRD path, so with `prd_path=None` it runs only when `[verify] progress_file_path` is set. Covered by a test.
3. A nonexistent `--root` is also exit 2.

Ran in the worktree at the head of this branch:

```
uv run pytest tests/ -q -p no:cacheprovider   -> 3212 passed, 28 skipped (3200 baseline + 12 new)
uv run mypy kstrl/ --strict                    -> Success: no issues found in 111 source files
uv run ruff check kstrl/ tests/                -> All checks passed!
uv run python scripts/gen_docs.py --check      -> README.md generated sections are current.
```

Manual: `uv run ks sense --json` on this repository exits **1**, not 0 as the issue's acceptance assumed. The cause is pre-existing and outside this change: the repo has no `kstrl.toml`, so the default `uv run ruff check .` lints the whole tree and finds 4 existing errors in `spike/tui0/app.py`, which sit outside the CLAUDE.md lint scope (`kstrl/ tests/`). Confirmed present on origin/main with this change stashed. The sensor is measuring correctly; fixing `spike/` or adding a root `kstrl.toml` with a scoped `lint_command` is a separate decision. `uv run ks sense --path /tmp/does-not-exist` exits 2 with `error: path is not a directory`. Wall clock for the full measurement here: 4m21s, of which 261s is the test suite.

Not exercised: the opt-in policy / adequacy / dead-code / mutation checks under `ks sense`; they run through the unchanged `run_mechanical_verification` branches and rely on their existing tests.

## Checklist

- [x] `uv run python scripts/gen_docs.py --check` passes
- [ ] Tracker: tick R10.1 in #235 and in `docs/control-loop-design.md` on merge

## Provenance

- [x] Written by an AI agent from the issue specification
- [x] Reviewed by a human (AI self-review does not count - H1)

### DIFF
--- a/kstrl/cli.py
+++ b/kstrl/cli.py
@@ -2973,11 +2973,22 @@ def sense(
     typecheck, linter, diff scope, bad patterns, plus any opt-in
     policy / adequacy / dead-code / mutation checks from kstrl.toml),
     run by hand with no PRD, no branch, no worktree and no agent spend.
-    Nothing is written to .kstrl/ or anywhere else; only the configured
-    test / typecheck / lint commands touch the tree (their own caches).
+
+    The measurement is read-only. It runs against your live checkout,
+    not a worktree kstrl owns, so it writes nothing to .kstrl/ and never
+    edits, stages, commits or leaves bytecode: the dead-code check
+    reports what it would remove instead of removing it, and mutation
+    testing is skipped because mutmut works by rewriting source. The
+    exception is the project's OWN configured test / typecheck / lint
+    commands, which are your programs and write their own caches.
+
+    Most checks read `git diff <base>...HEAD`, so the tree must be a git
+    repository with a reachable base unless every diff-based check is
+    turned off in kstrl.toml.
 
     Exit 0 when every check passed, 1 when any failed, 2 when the
-    measurement itself could not run (missing path, bad kstrl.toml).
+    measurement itself could not run (missing path, bad kstrl.toml, or
+    git cannot produce the diff).
     """
     root_dir = root.resolve() if root else Path.cwd()
     path = tree_path.resolve() if tree_path else root_dir
@@ -3006,6 +3017,41 @@ def sense(
 
     base = base_branch or _detect_base_branch(path)
 
+    # Every check below that consumes the diff reads it through the
+    # LENIENT git helpers, which map a bad ref, a missing base or a
+    # non-repository onto an EMPTY file list - indistinguishable from
+    # "nothing changed". diff_scope then reports "0 files, all within
+    # scope", bad_patterns "scanned 0 Python files", and `ks sense`
+    # exits 0 having measured nothing. mutation_testing is deliberately
+    # absent from this list: sense skips that check outright (read-only),
+    # so its diff read never happens and demanding a base for it would
+    # be a false exit 2.
+    needs_diff = (
+        verify_cfg.check_diff_scope
+        or verify_cfg.check_bad_patterns
+        or verify_cfg.dead_code_cleanup
+        or policy_cfg.enabled
+        or adequacy_cfg.enabled
+    )
+    if needs_diff:
+        # Ask git the same question once, strictly, before any check
+        # runs. Cannot-measure is exit 2; it is never a pass.
+        from kstrl import git as _git
+
+        try:
+            _git.get_diff_names(base, path, strict=True)
+        except _git.GitDiffError as exc:
+            origin = (
+                "from --base"
+                if base_branch
+                else "auto-detected; name the right one with --base"
+            )
+            _sense_error(
+                f"git cannot measure the diff against {base!r} "
+                f"({origin}): {exc}",
+                as_json,
+            )
+
     result = run_mechanical_verification(
         worktree_path=path,
         prd_path=prd_path.resolve() if prd_path is not None else None,
@@ -3017,6 +3063,7 @@ def sense(
         fixtures_config=fixtures_cfg,
         autonomy_level=0,
         component_id=None,
+        read_only=True,
     )
 
     if as_json:
--- a/kstrl/cli.py
+++ b/kstrl/cli.py
@@ -10,7 +10,7 @@ from collections.abc import Callable
 from dataclasses import fields as dataclass_fields
 from datetime import datetime
 from pathlib import Path
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, NoReturn
 
 if TYPE_CHECKING:
     from kstrl.evolution import EvolutionConfig
@@ -110,6 +110,28 @@ def _console_ui(
 
 def _use_cli_value(ctx: click.Context, name: str) -> bool:
     return ctx.get_parameter_source(name) == ParameterSource.COMMANDLINE
+
+
+def _detect_base_branch(cwd: Path) -> str:
+    """Base branch of the repo at ``cwd``: ``origin/HEAD``'s target, else main.
+
+    Shared by ``ks run`` (manifest base) and ``ks sense`` (diff base).
+    Any failure to ask git - no repo, no remote, git missing, timeout -
+    falls back to ``main``; the caller can always override with a flag.
+    """
+    import subprocess as _sp
+
+    try:
+        head_ref = _sp.run(
+            ["git", "symbolic-ref", "refs/remotes/origin/HEAD"],
+            cwd=cwd, capture_output=True, text=True, timeout=5,
+        )
+        if head_ref.returncode == 0:
+            # "refs/remotes/origin/main" -> "main"
+            return head_ref.stdout.strip().rsplit("/", 1)[-1]
+    except Exception:
+        pass
+    return "main"
 
 
 # Accepted spellings for the agent type across the config surface.
@@ -644,18 +666,7 @@ def run(
     effective_branch = config.kstrl_branch or prd_branch or "kstrl/run"
 
     # Detect base branch from git
-    detected_base = "main"
-    try:
-        import subprocess as _sp
-        head_ref = _sp.run(
-            ["git", "symbolic-ref", "refs/remotes/origin/HEAD"],
-            cwd=root_dir, capture_output=True, text=True, timeout=5,
-        )
-        if head_ref.returncode == 0:
-            # "refs/remotes/origin/main" -> "main"
-            detected_base = head_ref.stdout.strip().rsplit("/", 1)[-1]
-    except Exception:
-        pass
+    detected_base = _detect_base_branch(root_dir)
 
     # Build single-component manifest from PRD
     rel_prd = str(config.prd_file)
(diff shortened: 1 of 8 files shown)


kstrl/cli.py
```diff
@@ -2876,6 +2887,185 @@ def _load_and_render() -> int:
...
+            FixturesConfig.load(root_dir) if prd_path is not None else None
+        )
+    except (OSError, ValueError) as exc:
+        # ValueError covers malformed TOML (load_toml_section) and the
+        # loaders' own validation errors (PolicyConfigError is one).
+        _sense_error(f"could not load kstrl.toml from {root_dir}: {exc}", as_json)
+
+    base = base_branch or _detect_base_branch(path)
+
+    result = run_mechanical_verification(
```
0xfauzi, line 3055: [P1] Prevent `ks sense` from committing the measured checkout\n\nThis call runs the full verifier with the loaded `VerifyConfig`. When `dead_code_cleanup=true`, `check_dead_code` runs `ruff --fix`, `git add -A`, and `git commit`. I reproduced this at this head. `ks sense --json` exited 0, changed `HEAD`, and committed an unrelated untracked file. This violates issue #222 read-only contract. Add a read-only verification mode for `sense`. It must prevent edits, staging, commits, and bytecode writes.

kstrl/cli.py
```diff
@@ -2876,6 +2887,185 @@ def _load_and_render() -> int:
...
+        adequacy_cfg = AdequacyConfig.load(root_dir)
+        fixtures_cfg = (
+            FixturesConfig.load(root_dir) if prd_path is not None else None
+        )
+    except (OSError, ValueError) as exc:
+        # ValueError covers malformed TOML (load_toml_section) and the
+        # loaders' own validation errors (PolicyConfigError is one).
+        _sense_error(f"could not load kstrl.toml from {root_dir}: {exc}", as_json)
+
+    base = base_branch or _detect_base_branch(path)
```
0xfauzi, line 3018: [P1] Reject an unreadable Git base before returning PASS\n\nThis base is not validated. The legacy diff helpers map Git errors to an empty list. A missing base returned exit 0 and `passed: true`. `diff_scope` saw zero files, and `bad_patterns` scanned zero files. The same failure occurs when fallback `main` does not exist. Run a strict Git preflight. Return exit 2 when Git cannot measure the diff.

0xfauzi (commented):
Review verdict: Block. Two P1 defects affect the sensor core guarantees. The inline comments include reproduction evidence. Validation at this head: 3,212 tests passed. Mypy, Ruff, and documentation checks also passed.

kstrl/cli.py
0xfauzi: Fixed in 81870e7, and reproduced first: on a throwaway repo with `dead_code_cleanup = true`, `ks sense --json` exited 0, moved HEAD from af1662e to 3e2b189, and swept an unrelated untracked file into a "chore: auto-remove dead code" commit. Re-run after the fix: HEAD unchanged, `git status --porcelain` unchanged, file bytes unchanged, no `__pycache__`, no `.ruff_cache`.

`run_mechanical_verification` now takes `read_only` (default False, so no existing caller changes). Under it:

- `check_dead_code` runs the same rule set with `--no-fix --no-cache` and reports what it would have removed instead of removing it. `--no-fix` is explicit rather than implied by omitting `--fix`, because a project can set `fix = true` under `[tool.ruff]`. The `git add -A` / `git commit` are additionally guarded on `not read_only`.
- `check_mutation_score` is skipped: mutmut works by rewriting the source it mutates, so there is no read-only way to run it. The skip is recorded as a check naming the reason rather than dropped.

Two writes are suppressed unconditionally, because the factory does not want them either: `check_bad_patterns` directs `py_compile`s output at a throwaway directory (same `PyCompileError` type and message, only the destination moves) instead of leaving `__pycache__` beside every scanned file, and the function-fixture runner gets `-B`.

One divergence I documented rather than hid: the factory deletes the ruff-fixable subset before vulture looks, so a tree whose only dead code is ruff-fixable can fail under `sense` and pass inside the factory. That is the tree reported honestly, but it is a difference.

Not suppressed: your own configured test / typecheck / lint commands, and a user-supplied `dead_code_command`. Those are your programs and write their own caches; the docstring, README and CHANGELOG all draw that boundary explicitly.

Tests: `test_sense_never_edits_stages_or_commits` and `test_sense_leaves_no_bytecode_or_lint_cache` (end to end on a real repo), plus four unit tests including `test_dead_code_default_still_fixes_and_commits` as the no-regression guard. I checked they bite: all fail against the pre-fix source except that guard, which must pass both ways.

kstrl/cli.py
0xfauzi: Fixed in 81870e7, and reproduced first: with base `main` absent, `diff_scope` reported "0 files, all within scope" and `bad_patterns` "Scanned 0 Python files" while an `AKIA...`-shaped string sat in the tree, and the command exited 0. Same on the auto-detected fallback when `main` does not exist.

`ks sense` now runs one strict preflight before any check: `git.get_diff_names(base, path, strict=True)`. A `GitDiffError` becomes exit 2 carrying git`s own message, and the JSON document contains only `schema_version` and `error` - no `passed`, no `checks`, so nothing downstream can read a verdict that was never measured. The error names whether the base came from `--base` or from detection, and in the detection case says to pass `--base`.

The preflight is the same question the checks ask, asked once and strictly, rather than a separate reachability heuristic - so it cannot pass while the real diff fails.

It runs whenever a diff-consuming check is enabled: `check_diff_scope`, `check_bad_patterns`, `dead_code_cleanup`, `[policy] enabled`, `[adequacy] enabled`. `mutation_testing` is deliberately excluded: `sense` skips that check outright now (read-only), so its diff read never happens and demanding a base for it would be a false exit 2. Turn every diff-based check off and a plain non-git directory is measurable again - covered by `test_sense_runs_without_git_when_no_check_reads_the_diff`, which is the over-fix guard and passes both before and after.

I did not make `check_diff_scope` / `check_bad_patterns` themselves strict. Their lenient contract is documented as deliberate for the pre-existing callers, and changing it would alter Phase 1 behaviour inside the factory; the preflight closes the hole for `sense` without touching that.

Tests: exit 2 on an unreachable `--base`, on a missing auto-detected base, and outside a git repository - all three fail against the pre-fix source.

Verification at 81870e7: `uv run pytest tests/` 3224 passed / 28 skipped in 4:21 (3212 baseline + 12 new), mypy --strict clean on 111 files, ruff clean, `gen_docs.py --check` current.

PR #282, opened 2026-08-30
fix(verify): dispatch each gate's parser instead of assuming the Python toolchain

### LINKED ISSUE #258: Non-Python gates lose every actionable detail: vitest output goes through the pytest parser and yields 0 failures, mislabelled [pytest]
Found while dogfooding kstrl as a first-time user, building a real web app (`writers-room`: FastAPI backend plus a TypeScript/Svelte frontend). Measured with real tool output, not reasoned from the code.

## The setup that finds this

Any repo with a frontend. kstrl's verify config takes one command per gate (`kstrl/verify.py:216-218`):

```python
test_command: str | None = None
typecheck_command: str | None = None
lint_command: str | None = None
```

`run_scrubbed` passes a string command with `shell=isinstance(cmd, str)` (`kstrl/verify.py:147`), so an operator can chain toolchains:

```toml
[verify]
test_command = "uv run pytest && (cd web && npm run test)"
```

That part works, and it is the only way to gate a polyglot repo. The problem is what happens when the non-Python half fails.

## What happens

`check_test_suite` parses the output unconditionally, whatever the command was (`kstrl/verify.py:575-589`):

```python
if result.returncode != 0:
    output = (result.stdout + result.stderr).strip()
    parsed = parse_pytest_output(output)
```

There is no dispatch on the command. `kstrl/parsers.py` has exactly three parsers - `parse_pytest_output`, `parse_mypy_output`, `parse_ruff_output` - and no `tool` detection. So vitest, tsc, eslint, jest, cargo and `go test` output all go through the pytest parser.

## Measured

I ran a real failing vitest test in the project and fed the real output to the real parser:

```
$ npm run test        # one deliberately failing test
 FAIL  tests/failing.test.ts > deliberate failure > shows what a real vitest failure looks like
AssertionError: expected false to be true // Object.is equality
- Expected
+ Received
- true
+ false
 ❯ tests/failing.test.ts:5:30
```

```
$ uv run python -c "from kstrl.parsers import parse_pytest_output; ..."
tool: pytest
failures found: 0
--- what the agent would receive ---
[pytest]  Test Files  1 failed | 1 passed (2)
      Tests  1 failed | 2 passed (3)
   Start at  19:05:39
   Duration  185ms (transform 26ms, setup 0ms, collect 28ms, tests 4ms, environment 0ms, prepare 75ms)
```

**Zero failures parsed.** What survives into the retry prompt is a timing line. What is dropped: the failing file, the line number, the test name, the assertion message and the expected-versus-received diff. All of it was present in the raw output.

The `[pytest]` prefix is worse than the omission. It tells the implementing agent the failure came from pytest, when it came from vitest, so the agent is pointed at the wrong toolchain and the wrong half of the repo.

## Why this matters more than a formatting bug

kstrl's stated design principle is that the gap becomes the next instruction: "parsed failures with file, line, source context and a fix hint, not raw tool output. The agent tries again against a smaller, sharper target." For every non-Python gate that promise is inverted. The agent gets raw tool output with the useful lines stripped out and a wrong tool name attached, and the component then burns its retries.

I have not measured the retry cost of this, because I have not yet let a frontend component fail in a paid run. On the deckgen rate an engineer iteration is roughly $4, and `max_retries` defaults to 3.

## What to change

Two options, and the first is much smaller.

**Minimum:** when the parser finds zero structured failures, do not label the passthrough `[pytest]`. Emit the raw output under a neutral heading naming the command that actually ran. An agent given raw vitest output can read it; an agent given raw vitest output labelled pytest is being misdirected.

**Proper:** dispatch the parser on the command, and add parsers for the common non-Python tools. `tsc` and `eslint --format compact` both emit `file:line:col: message`, which is the shape `parse_mypy_output` and `parse_ruff_output` already handle, so those two are close to free. Vitest and jest need their own.

Related: #200 asks for language-pluggable interface extraction for feedforward. This is the same gap on the sensor side rather than the context side, and the two probably want one language-adapter seam rather than two.

0xfauzi:
Fixes #258.

PR #278 fixed the **misdirection** half: a foreign tool's output is no longer labelled `[pytest]`. This is the **content** half. Measured on real output, not reasoned from the code:

| input | before | after |
|---|---|---|
| real failing vitest run | 0 failures parsed | 2, with file, line, test name and assertion |
| real failing `tsc` run | 0 | 3, with file, line, `TS####` and message |
| real failing `eslint` run (stylish, the default) | 0 | 5, with file, line and the real rule id |
| `eslint --format unix` | 3, rule recorded as `'cache'` / `Expected` / `'missingHelper'` | 3, correct rule ids |

## The dispatch design

A per-gate `tool` key: `[verify] test_tool` / `typecheck_tool` / `lint_tool`, plus `KSTRL_VERIFY_*_TOOL`, default unset. Unset means **auto**: run every parser registered for the gate and **union** the failures. The neutral raw passthrough from #278 stays as the floor when no parser matches anything.

Why union rather than the alternatives:

- **Command-string sniffing, rejected.** The chain that motivated the issue, `uv run pytest -q && cd web && npm run test`, contains both `pytest` and `npm`, so a substring rule must pick one and is wrong half the time. It also fails the most common invocations of all, `npm run test` / `make test` / `just test`, where no tool name appears anywhere. A command string is not evidence of the tool.
- **Try every parser and pick a winner, rejected.** A wrong pick is a silent semantic substitution, and no tie-break can serve the chain case where one command legitimately emits two formats and both are real.
- **Union.** No tie-break to get wrong, the chain case works by construction, and a parser that understood nothing contributes nothing. The tie-break question disappears rather than being answered.

An unknown `tool` value raises on config load naming the accepted values. It must not quietly become auto: the operator wrote the key to stop kstrl guessing.

## Tools supported, and skipped

**Supported:** pytest, mypy, ruff (existing), plus **vitest**, **tsc** and **eslint**. eslint reads all three formats, stylish included, because stylish is the default that `npm run lint` actually prints and is the only one eslint 9 still ships in core.

**Deliberately skipped:** jest, cargo, `go test`, biome. Adding one is a parse function, one entry in `TOOL_PARSERS` and `GATE_TOOLS`, and one captured fixture; `kstrl/parsers_web.py`'s module docstring says where the function goes and states the two invariants a new parser owes the dispatcher, both of which are enforced by tests rather than by convention.

## Two of the issue's own predictions, disproven by measurement

- **tsc.** The issue expected `file:line:col: message`, matching the mypy pattern "close to free". Through a pipe it emits `src/broken.ts(7,9): error TS2322: ...` with **parentheses**, and under `--pretty` it emits `src/broken.ts:7:9 - error TS2322: ...`. All three existing parsers returned 0 failures on real tsc output. It needed its own regexes, one per mode.
- **eslint.** `--format unix` did parse through `parse_ruff_output`, three of three with the right file and line, but the rule slot took the message's first word and left the real rule id at the end of the line. Those slugs were fed to `evolution.py` as failure signatures. `--format compact`, which the issue names, did not parse at all. `_RUFF_ERROR_RE`'s rule slot is now a ruff-shaped code (`[A-Z]{1,6}\d{1,5}`), which fixes the garbage and keeps the two parsers off each other's lines on the auto path.

## The consumer that would have broken

`evolution.signatures_from_verification` switched on `ParsedOutput.tool` against the exact strings `"ruff"`, `"mypy"` and `"pytest"`. That is a name check standing in for a capability check, and it broke twice over as soon as a gate could dispatch: a newly supported tool fell silently through to a prose slug, and the unioned `"pytest+vitest"` label matched nothing at all.

The parser now names its own signature in `ParsedFailure.code` (a linter rule, a checker error code, or the exception class a test died on), and the switch is gone. `ParsedOutput.tool` is documented as a label, never a switch. `propose_improvements` went with it: it was writing "check ruff/flake8 docs" for an eslint rule and "review `[tool.mypy]` in pyproject.toml" for a `TS2322`, in files `ks evolve` writes to disk.

## Fixtures

Eleven captures under `tests/tool_output/`, every one produced by running the named tool at the named version against a deliberately broken toy project. Provenance and the two normalizations applied (machine paths replaced with `/repo`; trailing whitespace and blank last lines removed, because the repo's hygiene hooks impose both) are documented in `tests/helpers/tool_output.py`. `tsc-5.6.3-pretty.txt` keeps its real ANSI escapes, and a test asserts it still does, so `strip_ansi` is exercised against real escapes rather than a hand-typed `\x1b[31m`.

The two inline captures #278 added move into that directory unchanged, so there is one home for real tool output rather than two.

## Also in the diff

- `strip_ansi` at the dispatcher, once, rather than in six regexes. `tsc --pretty` colours its output through a pipe.
- Absolute paths in a parsed failure's `file` are made worktree-relative via the existing `config.relative_to_root`. eslint's default formatter prints absolute paths, and a path rooted in kstrl's throwaway worktree is correct on disk and useless as an instruction.
- `VerifyConfig.load`'s fourteen-branch `if "key" in section` chain became a table. Its cyclomatic complexity was already twice the repo's ratchet limit before this change added three keys to it.
- `check_test_suite`'s dead `in_short_summary` flag removed; it only ever gated skipping one line, which a plain `continue` does.

## Verification

`uv run pytest tests/ -q` (3811 passed, 28 skipped), `uv run mypy kstrl/ --strict`, `uv run ruff check kstrl/ tests/`, `pre-commit run --all-files`, all green locally. Coverage 88.94% total against the 79.91% ratchet; `kstrl/parsers.py` 98.9%, `kstrl/parsers_web.py` 97.2%. No `*_PROMPT` constant was touched and `tests/test_prompt_versions.py` passes.

`/simplify` ran over the change. Applied: the redundant single-parser branch in `parse_gate_output`, the duplicate ANSI and "N failed" regexes, an unused-named-group cleanup, a `groupdict().get` that could not fail, a `partial` hop in the config table, six dead `raw_summary` assignments, and a measured quadratic-backtracking cliff in the eslint patterns (`.+?` to `[^:]+`: 1475 ms to 0.37 ms on a 56 KB line). It also added the two contract tests named above. Skipped, with reasons: collapsing the three near-identical `check_*` gate functions into one (restructures three public functions beyond this change), normalizing paths inside tool-written prose as well as `failure.file` (string-substituting a tool's own sentences is a different and less safe mechanism), and splitting `kstrl/parsers.py` into a package (the 800-line ratchet is what forced the current seam).

Note: the `atlas` check will fail. `docs/atlas/` is deliberately byte-identical to main.

### DIFF
--- a/README.md
+++ b/README.md
@@ -351,6 +351,9 @@ allow_network = false  # re-open outbound network inside the sandbox (off = deny
 test_command = ""              # empty = the harness default (uv run pytest)
 typecheck_command = ""         # empty = uv run mypy when [tool.mypy] scopes it, else uv run mypy .
 lint_command = ""              # empty = the harness default (uv run ruff check .)
+test_tool = ""                 # parser for the test gate's output; empty = every parser (pytest, vitest), unioned
+typecheck_tool = ""            # parser for the typecheck gate's output; empty = every parser (mypy, tsc), unioned
+lint_tool = ""                 # parser for the lint gate's output; empty = every parser (ruff, eslint), unioned
 check_diff_scope = true        # fail on changes outside allowed paths
 check_bad_patterns = true      # scan the diff for secret-like patterns
 dead_code_cleanup = false      # optional dead-code check
--- a/docs/env-vars.md
+++ b/docs/env-vars.md
@@ -259,6 +259,9 @@ agent's worktree by construction on both CLIs.
 | `KSTRL_VERIFY_TEST_CMD` | str | unset (uses `uv run pytest`) |
 | `KSTRL_VERIFY_TYPECHECK_CMD` | str | unset (uses `uv run mypy .`) |
 | `KSTRL_VERIFY_LINT_CMD` | str | unset (uses `uv run ruff check .`) |
+| `KSTRL_VERIFY_TEST_TOOL` | `pytest` \| `vitest` | unset = run both parsers and union the failures |
+| `KSTRL_VERIFY_TYPECHECK_TOOL` | `mypy` \| `tsc` | unset = run both parsers and union the failures |
+| `KSTRL_VERIFY_LINT_TOOL` | `ruff` \| `eslint` | unset = run both parsers and union the failures |
 | `KSTRL_DEAD_CODE_CLEANUP` | bool (`1`) | false |
 | `KSTRL_DEAD_CODE_CMD` | str | unset |
 | `KSTRL_MUTATION_TESTING` | bool (`1`) | false |
--- a/kstrl/config_report.py
+++ b/kstrl/config_report.py
@@ -187,26 +187,16 @@ def _phase_sections() -> list[tuple[str, Any, list[str]]]:
                 "keep_worktrees_on_failure",
             ],
         ),
-        (
-            "verify",
-            VerifyConfig.load,
-            [
-                "test_command",
-                "typecheck_command",
-                "lint_command",
-                "check_diff_scope",
-                "check_bad_patterns",
-                "dead_code_cleanup",
-                "dead_code_command",
-                "mutation_testing",
-                "mutation_threshold",
-                "mutation_timeout",
-                "subprocess_timeout",
-                "require_self_critique",
-                "self_critique_min_bullets",
-                "progress_file_path",
-            ],
-        ),
+        # Derived, not hand-listed. The hand-written copy of this list
+        # went stale the moment #258 added the three `*_tool` keys: they
+        # reached VerifyConfig, gen_docs, the README and env-vars.md and
+        # not this list, so `ks config` and the config screen showed no
+        # row for the one setting an operator reaches for when a gate is
+        # parsed by the wrong toolchain. Every scalar field of
+        # VerifyConfig IS a documented kstrl.toml key, which gen_docs
+        # already enforces, so the field list is the key list and a
+        # second copy of it can only ever be wrong.
+        ("verify", VerifyConfig.load, [f.name for f in dataclass_fields(VerifyConfig)]),
         (
             "security",
             SecurityConfig.load,
@@ -287,6 +277,67 @@ def _phase_sections() -> list[tuple[str, Any, list[str]]]:
     ]
 
 
+def _base_sources(
+    resolved: KstrlConfig, noenv: KstrlConfig, defaults: KstrlConfig
+) -> dict[str, str]:
+    """Per-field source for KstrlConfig, computed BEFORE the flag overlay.
+
+    A flag replaces whatever source the value had, so it has to be
+    applied after this, not folded into it.
+    """
+    sources: dict[str, str] = {}
+    for f in dataclass_fields(KstrlConfig):
+        if getattr(resolved, f.name) != getattr(noenv, f.name):
+            sources[f.name] = "env"
+        elif getattr(noenv, f.name) != getattr(defaults, f.name):
+            sources[f.name] = "toml"
+        else:
+            sources[f.name] = "default"
+    return sources
+
+
+def _base_rows(resolved: KstrlConfig, sources: dict[str, str]) -> list[ConfigRow]:
+    """Rows for the sections KstrlConfig fans out over."""
+    return [
+        ConfigRow(
+            section=section,
+            key=toml_key,
+            value=format_row_value(section, toml_key, getattr(resolved, field_name)),
+            source=sources[field_name],
+        )
+        for section, keys in SHOW_SECTIONS
+        for toml_key, field_name in keys
+    ]
+
+
+def _phase_rows(
+    section: str,
+    knob_fields: list[str],
+    resolved: Any,
+    noenv: Any,
+    toml_keys: set[str],
+) -> list[ConfigRow]:
+    """Rows for one phase config, each tagged with where its value came from."""
+    rows: list[ConfigRow] = []
+    for field_name in knob_fields:
+        value = getattr(resolved, field_name)
+        if value != getattr(noenv, field_name):
+            source = "env"
+        elif field_name in toml_keys:
+            source = "toml"
+        else:
+            source = "default"
+        rows.append(
+            ConfigRow(
+                section=section,
+                key=field_name,
+                value=format_config_value(value),
+                source=source,
+            )
+        )
+    return rows
+
+
 def build_config_report(
     root_dir: Path,
     *,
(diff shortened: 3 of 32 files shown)


0xfauzi:
All six fixed in 7054a5e, plus two the review did not have. Every claim below is measured against the real tool, and the captures are checked in.

## 1. MEDIUM, `ks config` blind to the new keys

Fixed deeper than reported. The hand-written key list is gone: the verify section now derives its keys from `VerifyConfig`'s dataclass fields, which is what the `timeout` section already did. Verified the two lists were byte-identical (17 names, same order) before swapping, so this is a mechanism change and not a behaviour one. The next knob cannot go missing the same way, which the hand-list fix would not have prevented.

Measured: 14 verify rows before, 17 after, with `test_tool = 'vitest'` tagged `toml` and `typecheck_tool = 'tsc'` tagged `env`.

## 2. MEDIUM, REGRESSION, ruff's syntax errors dropped

Confirmed against ruff 0.16.1. The rule slot now accepts either a code (`F401`) or a kebab name with a trailing colon (`invalid-syntax:`). The trailing colon is what keeps it off `eslint --format unix`, whose messages sit in the same slot and open with `'cache'`, `Expected`, `'missingHelper'` (no colon, no hyphen).

## 3. MEDIUM, ruff's default format unreadable

Confirmed and it is a two-line format, not a one-line one:

```
F401 [*] `os` imported but unused
 --> draft.py:1:8
```

Both formats are read now, as two independent passes. Neither can see the other's output (concise emits no `-->`, full emits no `file:line:col:` head), and a test concatenates both fixtures and asserts 8, not 16.

Measured through the real gate on kstrl's own default command shape: 0 failures before, 4 after, with source context and fix hints. `[*]` is stripped, since it describes the fix rather than the defect.

## 4. MEDIUM, bare `Error:` and length-dependent signatures

Both halves fixed, and the second was worse than reported. With realistic test names pytest does not merely truncate the message, it **drops it entirely**: of five failures in a captured run, two carried no message at all, one carried `AssertionErro...`, one `AssertionError:...`.

The parser now reads the traceback block for the untruncated exception, message and **line number**, keyed by test id (translating `TestClass::method` to the header's `TestClass.method`, and stripping `ERROR at setup of`). Nothing is fabricated: a failure with no block keeps exactly what the summary gave it.

Before and after on the same captured run:

| | before | after |
|---|---|---|
| failures parsed | 3 of 5 | 5 of 5 |
| with a line number | 0 | 5 |
| with a signature code | 1 | 5 |

The two missing failures were parametrized ids containing spaces (`test_x[a draft]`), which `[^\s]+` never matched. Since `add_source_context` returns early on line 0, **no pytest failure had ever carried source context**; they do now.

## 5. LOW, tsc claims mypy's footer

Fixed by positive matching against tsc's measured tails, of which there are three, not one:

```
Found 3 errors in 2 files.
Found 2 errors in the same file, starting at: src/only.ts:6
Found 1 error in src/only.ts:2
```

The contract test is extended to `raw_summary` as suggested, and to ruff and mypy rows.

**It immediately caught the mirror case:** mypy's `Found N errors in M file` pattern was claiming *tsc's* footer, in the same chained gate, in the other direction. mypy now requires its `(checked N source files)` tail, verified present across default, `--strict`, `--hide-error-context` and `--pretty` (`--no-error-summary` prints no footer and already falls through to the raw tail).

## 6. LOW, `target="pyproject"`

Now `typecheck_config`. The neutrality test is parametrized over all five `check_name` branches rather than the two that were wrong, and asserts against the target, title, description and suggested change together. The other three branches were already clean.

## Two the review did not have

**A file/line mismatch I was about to ship.** Taking the line from the traceback while leaving the file from the summary is wrong whenever a test fails inside a helper: a test in a 7-line `test_cross.py` failing at `loader.py:17` produced `test_cross.py:17`, a location that does not exist, and `add_source_context` would have printed whatever sat at line 17 of some longer file. File and line now travel together from the raising frame, which is also the better instruction. Fixture and test added.

**mypy claiming tsc's footer**, above, found by the widened contract test.

## `/simplify`

Applied: the verify key list derived rather than listed (finding 1, deeper); ruff's full-format pass rewritten from a `pending` state machine to a stateless adjacent-pair window, which makes the hazard its comment described structurally impossible; a shared `_kebab` so signature spelling has one home; two unused capture groups; `_pytest_blocks` skipped when nothing parsed; three subsumed tests removed after widening the ruff assertion to compare messages exactly; the evolution pair parametrized.

Skipped, with reasons: extracting the shared "location is on a different line from the message" carry from the ruff and eslint scans (the two run in opposite directions, so a shared helper needs a direction flag and comes out longer than both loops), and collapsing the two pytest block-association tests (each names a distinct header shape, which is the association being tested).

While fixing finding 1 I had to bring `build_config_report` from cognitive complexity 22 to 7, because the complexity gate refuses a commit touching a file with an over-limit function. Three named extractions, no behaviour change, covered by the existing config tests.

`tests/tool_output/` is added to the codespell skip list in `pyproject.toml`: pytest's truncation cuts words mid-token, and spell-correcting a capture would falsify the evidence. `.pre-commit-config.yaml` is untouched, per its byte-identical-across-repos rule.

## Gates

`uv run pytest tests/ -q` 3852 passed, `mypy --strict`, `ruff check`, `gen_docs --check`, and `pre-commit run --all-files` all green.

One note on that last one: `pre-commit run --all-files` also surfaces a pre-existing complexipy finding in `docs/lessons/verify/pr-221/admission_gates.py` (`admit` 16, `main` 36). That file predates this PR, is untouched by it, and is not staged; it appears only because `--all-files` feeds every file to a `--staged` hook. It does not fire at commit time and it is not this PR's to fix.

Issue #276, opened 2026-08-30
Engineer prompt step 9 still tells the agent to derive its own verify commands, and never mentions lint


0xfauzi:
Found while fixing #261 (PR #275) and deliberately left out of it, because it edits a prompt body and so triggers H2 and H3: version bump, snapshot hash, and a paid calibration re-run.

## The gap

#261 established a single source of truth for what Phase 1 runs. `resolve_verify_commands` is now the only answer, the gate shells out to exactly what it returns, and the engineer prompt is handed the same strings by construction. A test captures what the gate executes and compares it against what the resolver reports.

`DEFAULT_PROMPT` step 9 still tells the engineer to go and find its own typecheck and test commands, and never mentions lint at all.

So the harness now injects an authoritative, resolved block naming the three exact commands the gate will run, and the prompt directly above it tells the agent to work two of them out for itself and says nothing about the third. An agent following step 9 can arrive at a different command from the one it was just given, and lint has no instruction attached to it at all even though it is a blocking gate.

## Why it matters more than a wording tidy

This is the same failure #261 was filed about, one layer up. That issue was not that the commands in `CLAUDE.md` were wrong in some cosmetic way; it was that the harness mechanically fed the agent a verification contract that disagreed with the gate, and the cost was a wasted engineer iteration at roughly $4 measured. PR #275 removed the disagreement between the scaffolded file and the gate. Step 9 leaves a smaller version of the same disagreement between the prompt and the injected block.

The lint omission is the sharper half. `ruff check` is a blocking Phase 1 check, and nothing in the engineer's instructions tells it to run one before declaring itself done. A lint error is discovered by the gate rather than by the agent, which is the expensive order.

## What to change

Step 9 should defer to the injected block rather than instruct the agent to derive commands. Something in the shape of: run exactly the commands given in the verification block, all of them including lint, and do not substitute a narrower or broader variant.

Note the block already declares itself authoritative, so this is bringing the prompt into line with a contract that already exists rather than inventing a new one.

## Policy obligations

Per this repository's CLAUDE.md, this is an adversarial-prompt edit and all of H2 and H3 apply:

- Bump `DEFAULT_PROMPT_VERSION`.
- Update the matching `(hash, version)` tuple in `tests/test_prompt_versions.py`, in the same commit.
- Re-run the calibration suite and compare detection rates against the saved baseline: `KSTRL_RUN_CALIBRATION=1 uv run pytest tests/test_calibration.py -v`.

[...]

Issue #274, opened 2026-08-30
Carve kstrl's own state directory out at the scope guard, instead of relying on the user's .gitignore to do it


0xfauzi:
Found while fixing #201 (PR #273), and deliberately left out of that PR because it touches the verification gate rather than the scaffolder.

## The problem

`.kstrl/` is kstrl's own runtime state directory. Today it is kept out of the scope guard's untracked-file walk by asking the **user's** `.gitignore` to exclude it. That is an odd division of responsibility: the harness relies on a file it does not own, in a repository it does not control, to avoid tripping over its own artifacts.

The mechanism to do it properly already exists and is already used for a narrower case. `kstrl/loop.py` carves out `.kstrl/runs/<run_id>/` at the guard itself, via the `guard_ignored_paths` parameter on `run_loop`. So the harness already knows how to say "this is mine, do not count it as the agent going out of bounds". It just does not say it about the whole state directory.

## Why this is worth doing even after #273

PR #273 makes `ks init` scaffold a `.gitignore` containing `.kstrl/`, which fixes every repository initialised from now on.

It cannot fix a repository that was scaffolded before it, because no newly written file reaches a project that already exists. Widening the guard's carve-out does, because the fix travels with the harness rather than with the project.

It also removes a silent dependency. A user who curates their own `.gitignore`, or who has a global ignore file, or whose repository predates the scaffold, is currently one edit away from kstrl's own state directory being counted as agent scope violations. Nothing warns them, and the failure surfaces as a wasted engineer iteration rather than as a configuration problem.

## Care needed

The existing carve-out is deliberately narrow, and its docstring says so: "exact harness-owned files or directory prefixes for this invocation, **never a blanket state-directory bypass**". That wording is a warning, not an oversight, and this issue is proposing to soften exactly what it warns against. So the change wants thought rather than a one-line widening:

- Carve out the state directory for scope purposes only, and keep it distinct from anything that would let an agent write there deliberately and go unnoticed.
- Prefer a computed set of the paths kstrl actually writes over a blanket prefix, matching the approach #268 took for the per-component harness files, which chose exact file paths over `scripts/kstrl/` precisely to avoid a broad bypass.
- Whatever is carved out should be reported separately from the operator's authored allowlist, as #268 already does, so a scope failure still tells the operator what they authorised versus what kstrl added.

## Related

[...]

Issue #261, opened 2026-08-28
ks init writes a CLAUDE.md verification contract that disagrees with the gates kstrl actually runs, including the contract the Gap 2 fix was meant to stop overriding


0xfauzi:
Found while dogfooding kstrl as a first-time user, building a real web app (`writers-room`). The architect found it on its own during a `ks decompose` run and raised it as a spec issue, which is a good sign for the architect and a bad sign for `ks init`:

> [other] Three different sets of verification commands are in force for this repository, so an acceptance criterion that names one of them may not be the command the harness actually runs.

## The divergence, from a fresh `ks init` with nothing edited

`ks init` writes these into `CLAUDE.md` as hardcoded literals (`kstrl/init_cmd.py:673-676`):

```python
ctx["test_cmd"] = f"{runner}pytest tests/ -v --tb=short"
ctx["typecheck_cmd"] = f"{runner}mypy src/ --strict"
ctx["lint_cmd"] = f"{runner}ruff check src/"
```

It writes `kstrl.toml` with every verify command commented out, so the gates fall to their defaults:

| Gate | CLAUDE.md says | The gate actually runs | Same? |
|---|---|---|---|
| test | `uv run pytest tests/ -v --tb=short` | `uv run pytest` (`verify.py:563`) | no |
| typecheck | `uv run mypy src/ --strict` | `uv run mypy` or `uv run mypy .` (`_default_typecheck_command`, `verify.py:599`) | no |
| lint | `uv run ruff check src/` | `uv run ruff check .` (`verify.py:685`) | no |

All three disagree, from the moment `ks init` finishes and before the operator touches anything.

## Why this is worse than untidy

`CLAUDE.md` is the file the **implementing agent** reads to learn how to check its own work. `kstrl.toml` and the verify defaults are what the **gate** runs to judge it. When they differ, the agent can verify itself green and fail Phase 1 anyway.

The lint case is concrete on this repo: the agent is told `ruff check src/`, the gate runs `ruff check .`, so a lint error the agent introduces in `tests/` passes the agent's own check and fails the gate. That is one wasted engineer iteration, roughly $4 at the deckgen rate, for a discrepancy `ks init` created.

## The part that makes this worth prioritising

`_default_typecheck_command` exists precisely to stop this, and its own docstring says so (`kstrl/verify.py:610-614`):

```
This is the Gap 2 fix from the end-to-end factory validation run:
the factory's verify command was overriding the project's CLAUDE.md
[...]

PR #290, opened 2026-08-31
fix(init): a prompt bump now reaches projects that already ran ks init

### LINKED ISSUE #286: A DEFAULT_PROMPT version bump reaches no already-initialised project: the scaffolded prompt.md is never upgraded and its version is never checked
Found while fixing #276 (PR #284), reported there and deliberately not fixed there, because it is general to every `DEFAULT_PROMPT` change rather than specific to that one. Two independent review passes reached it separately.

## The gap

`ks init` writes the engineer prompt to `scripts/kstrl/prompt.md` through `_create_if_missing`, which by design never overwrites. `run_loop` prefers that file over the constant. And `DEFAULT_PROMPT_VERSION` is read **nowhere in `kstrl/`**: a repo-wide grep finds it only in `tests/test_prompt_versions.py`.

So the version is never stamped into the scaffolded file, never compared at run time, and never surfaced to an operator. Every project that has already run `ks init` keeps the prompt body it was scaffolded with, indefinitely, with no warning and no upgrade path.

## Why this matters more than an upgrade papercut

It empties H3 of most of its effect for existing users.

The policy is that a prompt change ships with a version bump and a snapshot hash so the audit trail is the PR diff. That audit trail is real, and it is about a constant that reaches only greenfield inits and the missing-file fallback. The prompt actually driving every already-initialised project is a copy on disk that no version check can see.

#276 makes this concrete. It exists because the engineer prompt told the agent to derive its own verify commands and never mentioned lint, which cost a measured wasted engineer iteration at roughly $4. PR #284 fixes the constant and bumps it to 1.2.0. Every existing project keeps the 1.1.1 body: still deriving its own commands, still never told about lint, still paying that iteration.

The same will be true of the next prompt fix, and the one after.

## What makes it survivable today, and why that is not enough

The injected `VERIFY_COMMANDS_PROMPT` block (added by #261, enrolled at 1.0.0) does carry the substantive instruction to every installed harness regardless of the on-disk prompt, because it is assembled at run time rather than scaffolded. So for this particular fix the harm is bounded: the agent is handed the right commands even where the file still tells it to derive its own, and the two now contradict each other in the same prompt.

That is a mitigation for one change, not a mechanism. Any future prompt change that is not also injected at run time reaches existing projects not at all.

## Options

1. **Stamp the version into the scaffolded file and compare at run time.** A comment or front-matter line carrying `DEFAULT_PROMPT_VERSION`, checked when `run_loop` reads the file, warning when the file is older than the harness. Does not touch the operator's file; tells them it is stale and what changed. Cheapest thing that closes the class.
2. **`ks init --upgrade-prompt`, or an `ks doctor` check**, offering to rewrite the scaffolded prompt after showing a diff. Explicit, but relies on the operator running something.
3. **Prefer the constant and treat the file as an overlay**, so harness improvements land automatically and local edits stay local. The largest change, and it alters what an operator's edits mean.
4. **Move more instruction into run-time-injected blocks**, following what #261 did for verify commands. Correct for anything the harness knows at run time; not applicable to genuinely static guidance.

(1) is the one I would take first, because it converts a silent divergence into a stated one and needs no operator action to be useful.

## Related

- #276 / PR #284 is where this surfaced.
- #261 / PR #275 established the run-time-injected block that mitigates it for verify commands specifically.
- Worth deciding alongside `docs/adversarial-roadmap.md`'s H3, since this is a limit on what H3 currently guarantees.

0xfauzi:
Fixes #286

## The gap

`ks init` writes `scripts/kstrl/prompt.md` through `_create_if_missing`, which by design never overwrites. `run_loop` prefers that file over the constant. And `DEFAULT_PROMPT_VERSION` was read **nowhere in `kstrl/`**: a repo-wide grep found it only in `tests/test_prompt_versions.py`.

So an H3 version bump reached greenfield inits and the missing-file fallback, and no already-initialised project. #276 (PR #284) is the concrete cost: the 1.1.1 body told the engineer to derive its own verify commands and never mentioned lint, which bought a measured wasted iteration at roughly $4. The `writers-room` project still carried that body.

## What I did, and why not the option as written

I took **option 1, keyed on the file's SHA-256 rather than on a version stamped into it**, plus the piece of option 2 that the digest makes safe.

Option 1 as the issue states it (write `DEFAULT_PROMPT_VERSION` into the scaffolded file, read it back at run time) has the same shape as the bug it fixes: a stamp can only classify files written after the stamp shipped. Every project that already exists has no stamp, so a stamp-based check would reach none of them either. It also means writing into a file the operator owns.

Hashing the file against a ledger of every body the harness has ever shipped fixes all three:

- it classifies files written long before the mechanism existed, which is the entire population the issue is about;
- it separates "pristine older scaffold" from "the operator edited this" **exactly**, because a scaffolded file is byte-identical to the constant it came from (`_create_if_missing` writes it verbatim; `run_loop` substitutes `$prd_path` and friends at read time);
- it writes nothing into the operator's file.

`SCAFFOLDED_TEMPLATES` in `kstrl/init_cmd.py` is that ledger. Its rows were harvested empirically, by hashing the constant at every commit in this repository's history that changed it, not guessed: eight bodies for `prompt.md` (three pre-versioning, then 1.0.0, 1.1.0, 1.1.1, an unmerged 1.2.0 draft, 1.3.0), four for `understand_prompt.md`, three for `feature_understand_prompt.md`.

**Measured, not assumed.** Copied read-only into a temp scaffold, `~/Documents/code/writers-room/scripts/kstrl/prompt.md` classifies as `stale`, shipped label `1.1.1`, prints the warning below, and `ks init --upgrade-prompts` moves it to 1.3.0, after which it classifies as `current`.

## What ships

- `SCAFFOLDED_TEMPLATES`, `classify_scaffolded_path`, `classify_scaffold` and `staleness_notice` in `kstrl/init_cmd.py`. Four statuses: `absent`, `current`, `stale`, `unrecognised`.
- `_check_prompt_preflight` in `kstrl/cli.py`, called by `ks run`, `ks factory`, `ks understand`, `ks feature` (both its prompts) and `ks retry`, alongside the existing `_check_agent_preflight` / `_check_prd_preflight`. Advisory, never fatal.
- `ks init` re-run reports a stale scaffold on the "already exists" line, which is where an operator asking about their scaffold is already looking.
- `ks init --upgrade-prompts` rewrites, atomically, only the copies the digest proves hold no local edits; it reports and refuses on anything else.
- H3b in `docs/adversarial-roadmap.md` and the matching CLAUDE.md rules, plus two new mechanisms so the ledger cannot rot (below).

No `*_PROMPT` body was touched. `DEFAULT_PROMPT_VERSION` and the H3 snapshot are unchanged, so this triggers neither H2 nor H3.

## The four design questions

**1. What does the warning say, and where does it go?**

Two lines, from the parent process, on the console UI, before the TUI branch of every command that has one:

```
WARN: <root>/scripts/kstrl/prompt.md is the prompt.md kstrl shipped at 1.1.1, unedited; this kstrl ships 1.3.0.
`ks init` never overwrites it, so every change to that template since 1.1.1 has missed this project. Run `ks init --upgrade-prompts` to take the current one.
```

It is deliberately not next to the read in `run_loop`. On a factory run `run_loop` executes in a pool worker whose UI writes to that component's `engineer.jsonl`, which is exactly where #261's CLAUDE.md divergence warning died until #275 lifted it into the parent. `TestOperatorActuallySeesIt` drives the real `ks run` through CliRunner and asserts the lines appear in the parent's own stream, ahead of `Factory: Validating DAG`, with `Preflight` (which `run_loop` prints from inside the worker) absent from that stream entirely.

Two limits, stated rather than papered over, both recorded in the function's docstring:

- Under the default interactive `ks factory` the dashboard deliberately drops `Log` events (`kstrl/tui/widgets/activity.py`), so the parent's own warning inside `run_factory` would not reach the feed either. The CLI placement is strictly earlier than that: it prints on the plain terminal before Textual takes the screen and is in scrollback when it exits. Making it show inside the dashboard means a typed event, which is the right shape for something discovered mid-run and the wrong shape for a start-up fact about a file on disk, so I declined it.
- The home shell's own factory launch (`kstrl/tui/session.py::_prepare_factory`) is not covered. It has no non-fatal notice channel: its only pre-run signal is `LaunchError`, and a stale prompt is advisory, not fatal. Giving the launcher an advisory channel is UI work that would also have to carry every other factory advisory it currently swallows, and that is not this issue.

`ks init` is the surface that both reports the fact and fixes it, and it is reachable from every path.

**2. What happens to a prompt the operator edited on purpose?**

Nothing, and that is the strongest argument for hashing rather than stamping. A byte-for-byte match against a body kstrl shipped is proof there are no local edits; anything else is `unrecognised`, and `unrecognised` deliberately claims nothing about authorship, because an edited file and a file from a build outside this history are indistinguishable. So:

- run-time preflight: silent on `unrecognised`. A stale-version nag on a file somebody customised is noise, and noise gets ignored.
- `ks init --upgrade-prompts`: reports it, names the constant to diff against, and never writes.
- `ks init` re-run: says nothing beyond "already exists", exactly as today.

The ambiguity is real and is designed for by refusing to guess inside it, which is the rule `init_wizard.apply_agent_settings` already follows for a user-edited `kstrl.toml`.

**3. Does it generalise beyond `prompt.md`?**

Yes, and it ships covering all three. `understand_prompt.md` and `feature_understand_prompt.md` are scaffolded from constants the same way and go stale the same way. They are H3-exempt (they produce documentation, not adversarial-role output) so they carry no `*_VERSION` constant, and their ledger labels are the dates their bodies first shipped; the classifier treats `label` as an opaque string. `ks feature` checks both prompts it reads, `ks understand` checks the one it reads.

More than that, `test_every_template_init_scaffolds_is_enrolled` AST-walks `run_init`'s own `_create_if_missing` calls and fails if a prompt template exists that the ledger does not know, which is H3's enrollment discipline applied one level down. A fourth scaffolded prompt cannot silently repeat #286 for itself.

**4. Should the version live in the file at all?**

No, and I rejected all three places it could have lived:

- **In the file.** Only classifies files written after the stamp ships, so it reaches none of today's projects, and it means writing into a file the operator owns. The digest is a strictly better key.
- **In `.kstrl/`.** A sidecar recording "we scaffolded 1.3.0 here" is a second source of truth that can disagree with the file. It says nothing about a project initialised before the sidecar existed, it goes stale the moment somebody replaces prompt.md by hand, and `.kstrl/` is machine state an operator may delete.
- **In `kstrl.toml`.** Same divergence problem, plus it puts a harness fact in the operator's control plane, where `ks init` is pinned by `tests/test_config_control_plane.py` not to change an effective value.

The digest needs no second store: the file is its own evidence. The cost is that the ledger must grow on every prompt change, which is why that is enforced rather than requested.

## The ledger is append-only, and that is enforced

A rule stated in four comments and checked in none is a hope. Three mechanisms, each verified by mutating the source and confirming the test refuses it:

| Violation | Test that refuses it | Verified |
|---|---|---|
| Change `DEFAULT_PROMPT` without appending a row | `test_every_live_body_is_the_newest_row` | yes |
| Drop an older row | `test_recorded_rows_are_never_edited_or_dropped` | yes |
| Edit an older row in place | `test_recorded_rows_are_never_edited_or_dropped` | yes |
| Scaffold a new prompt template without enrolling it | `test_every_template_init_scaffolds_is_enrolled` | yes |

Plus `tests/test_prompt_versions.py::test_engineer_prompt_bump_reaches_existing_projects`, which fails in the same file a person doing an H3 bump is already editing, and a fourth step in that file's drift message naming the ledger append.

## Tests

`tests/test_prompt_staleness.py`, 30 tests. Real-ledger tests use `SCAFFOLDED_TEMPLATES` as shipped; synthetic-ledger tests monkeypatch a two-row template built from literal bodies, because producing a genuinely stale file against the real ledger would mean shipping a 4KB historical prompt as a fixture or reading it out of git history, which a shallow CI checkout does not have.

- a stale file warns and the message names both labels, and several rows back it names the row the file actually matches;
- a current file does not warn, against both the synthetic ledger and the real one;
- a file with no provenance in it is handled: an edited body, an entirely hand-written prompt and an unreadable file are all `unrecognised` and all silent;
- an absent file is `absent` and silent, because `run_loop` already announces its own fallback;
- a path kstrl does not scaffold is not classified at all;
- a prompt relocated out of `scripts/kstrl/` is warned about but is NOT told to run `--upgrade-prompts`, which would be a no-op on it;
- the warning reaches the surface an operator sees: `ks run` end-to-end through CliRunner, asserting position in the parent's own output;
- `ks init` reports a stale scaffold and stays quiet on a current one;
- `--upgrade-prompts` rewrites a pristine older scaffold, never touches an edited one, no-ops on a current one, and is off by default.

## Gate

- `uv run pytest tests/ -q`: 4009 passed, 28 skipped, 0 failed
- `uv run mypy kstrl/ --strict`: clean, 119 source files
- `uv run ruff check kstrl/ tests/`: clean
- `pre-commit run --all-files`: every hook passes
- coverage: 89.15% measured on the full local suite, against the 79.91% ratchet

## /simplify

Four review agents (reuse, simplification, efficiency, altitude). Efficiency reported clean with measured numbers and nothing to change. Summary first, the reports verbatim below.

| # | Angle | Finding | Outcome |
|---|---|---|---|
| R1 | reuse | new overwrite path bypasses `atomic_write` | applied, with a different helper |
| R2 | reuse | hoist the test `_project` / `_invoke_run` pair into `spine_utils` | declined |
| R3 | reuse | re-implements `test_init_cmd.run_init_capturing` | applied |
| R4 | reuse | duplicate two-line `_sha256` helper | declined |
| S1 | simplification | four public entry points, two real callers | applied |
| S2 | simplification | index arithmetic where a dict lookup does the job | applied |
| S3 | simplification | `Sequence[Path]` over-broad for its callers | applied |
| S4 | simplification | positional tuple access at both consumers | applied |
| S5 | simplification | assertion that cannot fail | applied |
| S6 | simplification | subsumed test | applied |
| S7 | simplification | copy-paste test pair | applied |
| S8 | simplification | the fixture exists and is bypassed | applied |
| S9 | simplification | one rationale restated across five files | applied in part |
| E1-E8 | efficiency | all clean, measured | nothing to do |
| A1 | altitude | TUI home-shell factory launch is not covered | declined, recorded |
| A2 | altitude | append-only is stated in four places and enforced in none | applied |
| A3 | altitude | no enrollment mechanism for a template added later | applied |
| A4 | altitude | `--upgrade-prompts` is a no-op for a relocated prompt | applied |
| A5-A8 | altitude | clean | nothing to do |

**Applied with a change of helper: R1.** The finding was right about the risk and wrong about the helper. Importing `workqueue.atomic_write` from `init_cmd` puts `kstrl.workqueue` into `kstrl.loop`'s static import closure through loop's deferred `init_cmd` import, and #274's `tests/test_state_dir_scope.py::test_the_loop_cannot_reach_a_writer_of_an_uncarved_entry` refuses exactly that: the loop must not be able to reach a module that writes state entries the scope guard still counts. The full suite caught it. `init_cmd` now carries its own `_atomic_write` (mkstemp + `os.replace`), which is the second copy of that pattern for a stated architectural reason, next to `decompose._atomic_write_json`.

**Declined, with reasons.**

- **R2.** The "green LLM-free `ks run`" recipe already has independent copies in `tests/test_provisioning.py:37`, `tests/test_run_honesty.py:42`, `tests/test_instance_safety.py:39`, `tests/test_event_stream.py:667` and three `test_spine_*.py` files. Consolidating two of eight does not close the class, and it would edit `tests/test_cli.py`, outside this diff, on a repo with live sibling branches.
- **R4.** A two-line stdlib wrapper. Importing a private name across test modules to save two lines couples two files for less than it costs.
- **A1.** `tui/session._prepare_factory` has no non-fatal notice channel: `LaunchError` is fatal and a stale prompt is advisory. Building one is UI work that would have to carry every other factory advisory the launcher currently swallows. Recorded in the `_check_prompt_preflight` docstring and in design question 1 above, rather than hidden.
- **S9, in part.** The `kstrl/init_cmd.py` comment block was cut from 38 lines to 20 and now points at H3b for the argument. The near-verbatim phrases the agent found in the test file were kept: a test docstring saying why a case is deliberately silent is where the reader of that test needs it.

<details>
<summary>The four reports, verbatim (em dashes replaced with hyphens or colons per house style; otherwise unedited)</summary>

### Reuse

REUSE review of the #286 diff. Three genuine findings, one minor, plus what I checked and found clean.

**1. `kstrl/init_cmd.py:1123` - new overwrite path bypasses the existing `atomic_write` helper**
`_upgrade_scaffolded_templates` replaces a live operator-owned file with `state.path.write_text(state.template.body)`. `kstrl/workqueue.py:585` already ships `atomic_write(target: Path, content: str)` (mkstemp in the destination dir + `os.replace`), documented as "the pattern established at `manifest.py:189`", and CLAUDE.md names it as the codebase convention. It is already used across module boundaries via a function-local import to dodge cycles (`kstrl/intake_github.py:505`, `kstrl/serve.py:557`), so the same one-line import works here.
Cost: an interrupted write truncates `scripts/kstrl/prompt.md` - the exact file this feature exists to protect - and `--upgrade-prompts` is the one path in `ks init` that destroys existing bytes. Note `_create_if_missing`'s plain `write_text` is pre-existing and is a create, not a replace; only the new overwrite needs this.

**2. `tests/test_prompt_staleness.py:327-358` - `_project` / `_invoke_run` duplicate `tests/test_cli.py:214,232`**
`TestRunStructuralOverrideNotices._scaffold_project` (test_cli.py:214) and `._invoke_run` (test_cli.py:232) already build exactly this: same `scripts/kstrl/{prompt.md,prd.json}`, same `kstrl.toml`, the same five `spine_git` calls (`init -q -b main`, two `config`, `add -A`, `commit -q -m init`, down to the `cli@test` / `CLI Test` identities), and a character-identical 13-token `ks run` argv including `printf '<promise>COMPLETE</promise>\n'`, `--sleep 0`, `--no-verify`, `--ui plain`, `--no-color`. The only real difference is that the new copy parameterises the prompt body while the old one parameterises the toml body.
Cost: ~30 duplicated lines and a second, independent copy of the "how you invoke a green LLM-free `ks run`" contract - a renamed flag or a changed completion sentinel now has to be found in two files. Fix: hoist the pair to `tests/spine_utils.py` (which already owns `git` and `init_kstrl_repo`) with both `prompt_body` and `toml_body` parameters, and have both call sites use it. Cross-test-module imports are established here (13 existing, including private names such as `tests.test_progress_scope._pipeline`).

**3. `tests/test_prompt_staleness.py:239, 249-291` - re-implements `tests/test_init_cmd.py::run_init_capturing`**
That file already exposes a module-level, non-private helper:
```python
def run_init_capturing(root: Path) -> tuple[int, str]:
    buffer = io.StringIO()
    code = run_init(root, PlainUI(no_color=True, file=buffer))
    return code, buffer.getvalue()
```
The new file inlines that same `io.StringIO()` + `PlainUI(no_color=True, file=buffer)` + `run_init` + `getvalue()` sequence eight times (lines 239, 250, 258, 261, 271, 281, 288, 291).
Cost: ~14 duplicated lines, and the new tests silently discard the exit code the shared helper returns. Fix: add `upgrade_prompts: bool = False` passthrough to `run_init_capturing` (a one-line change in the file that already owns `run_init` coverage) and import it.

**4. Minor - `tests/test_prompt_staleness.py:59` duplicates `tests/test_prompt_versions.py:61`**
Byte-identical `_sha256(text)` helper. The two files are already coupled by this very issue (the new `test_prompt_versions.py:225` reaches into `init_cmd.SCAFFOLDED_TEMPLATES`), so importing it costs nothing. Two lines.

Checked and clean:

- **`_read_text_or_none`** - correctly reused (`init_cmd.py:1152`), not re-implemented. Good.
- **A sha256 helper in `kstrl/`** - there is none. Seven sites (`policy.py:718`, `statedir.py:333`, `breaker.py:158`, `linear.py:184`, `serve.py:2594`, ...) each inline `hashlib.sha256(...).hexdigest()` against different payload shapes. `init_cmd.py:757` matches the house pattern; nothing to call instead.
- **A template registry** - `kstrl/init_wizard.py:39 plan_scaffold` is the nearest thing, but it carries only paths and a create/keep action, no bodies or constant names, so `SCAFFOLDED_TEMPLATES` cannot be folded into it. (It is now the third enumeration of the scaffold set alongside `run_init`'s `_create_if_missing` sequence at `init_cmd.py:983-991`; deriving the three template rows from the new tuple would be a nicety, not a duplication worth a change.)
- **A scaffold-dir path helper** - none exists; `root / "scripts" / "kstrl"` is hardcoded in ~20 places already, so `classify_scaffold` is idiomatic.
- **`_check_prompt_preflight`'s placement** - it follows the existing `_check_agent_preflight` / `_check_prd_preflight` shape in `cli.py`. The obvious shared choke point, `factory._run_preflights` (which already hosts `_warn_claude_md_divergence`, the #275 precedent for exactly this class of operator warning), was considered and rejected in the docstring for a stated reason: it runs under a UI whose Log events the dashboard drops. That reasoning holds, so I am not flagging the five scattered call sites.
- **Test overlap with `test_prompt_versions.py`** - the new `test_engineer_prompt_bump_reaches_existing_projects` (line 225) is fully subsumed by `test_prompt_staleness.py::TestLedgerIntegrity::test_every_live_body_is_the_newest_row` plus `test_prompt_md_current_label_is_the_h3_version`. The duplication is deliberate and documented (the failure lands in the file the person doing the version bump is already editing), and it costs ~8 lines. I would keep it; noting it only because you asked the question.

One aside outside the reuse remit, since I noticed it while tracing where the check should live: `kstrl/tui/session.py:213` calls `run_factory` from `kstrl/tui/app.py:131` (`start_run_session`), a path that does not go through any `cli.py` command, so `_check_prompt_preflight` never fires there. The `ks factory --tui` embed path is covered (the check sits before the TUI branch at `cli.py:2616`). I did not verify this further - it belongs to a correctness reviewer.

### Simplification

Simplification review of the #286 diff. Note: `tests/test_prompt_staleness.py` and `kstrl/init_cmd.py` were edited on disk mid-review (`revisions_behind` was dropped from `TemplateState`, the plural-count message replaced by the two-label one). All line numbers below are against the **current** working tree, and that edit already resolved the one redundant-field finding I had.

**1. `kstrl/init_cmd.py:728, 741, 770` - four public entry points, two real callers.**
Only `staleness_lines` is imported outside the module (`cli.py:67`). `scaffolded_template_for` has exactly one production caller (line 748); `classify_scaffolded_path` has two, both in this file; `classify_scaffold` has one. Simpler form: fold the name lookup into `classify_scaffolded_path` as `next((t for t in SCAFFOLDED_TEMPLATES if t.filename == path.name), None)`, and split the real work into a private `_classify(template, path) -> TemplateState`. `classify_scaffold` then becomes a comprehension over `_classify(t, kstrl_dir / t.filename)`, and the `assert state is not None` at line 776 disappears with the `| None` it was defending against. Underscore-prefix what stays internal. Cost as written: four public names that read as supported API, plus a production `assert` that is a no-op under `python -O`.

**2. `kstrl/init_cmd.py:757-766` - index arithmetic where a dict lookup does the job.**
`enumerate` + `continue` + `is_current = index == len(template.history) - 1` exists only to answer "is this the last row". Since `test_history_rows_are_unique_and_non_empty` already pins digest uniqueness:
```python
labels = dict(template.history)
if digest not in labels:
    return TemplateState(template=template, path=path, status="unrecognised")
return TemplateState(
    template=template, path=path, shipped_label=labels[digest],
    status="current" if digest == template.history[-1][0] else "stale",
)
```
Six lines for eleven, one exit for the not-found case instead of two (757-767).

**3. `kstrl/cli.py:320` - `Sequence[Path]` is over-broad for its callers.**
Four of five call sites wrap one path in a list (792, 1094, 2616, 3570); only `feature` passes two, and it spends four lines building the list (1579-1583). Taking a single `Path` removes the `Sequence` import (line 9), the loop body's `continue`, and that block: `feature` becomes two calls under the `if understand_prompt_file is not None` it already has. Net roughly -6 lines and five call sites that read without brackets.

**4. `kstrl/init_cmd.py:1103-1106` and `kstrl/cli.py:344-349` - same three lines, positional tuple access.**
Both consumers do warn(`lines[0]`) / info(`lines[1]`), differing only by a two-space indent. Cheapest fix: make the return a `NamedTuple` with `headline` / `advice` - the test at line 154 already destructures into exactly those names. If you want the duplication gone too, a `report_staleness(path, ui, indent="")` in `init_cmd` replaces both blocks and stops exporting a tuple at all.

**5. `tests/test_prompt_staleness.py:212` - assertion that cannot fail.**
`label` is unpacked from `template.history[-1]` at line 203; `current_label` is defined as `self.history[-1][1]` (`init_cmd.py:642-644`). `assert template.current_label == label` is a tautology. Delete it.

**6. `tests/test_prompt_staleness.py:214-217` - subsumed test.**
`test_prompt_md_current_label_is_the_h3_version` asserts `history[-1][1] == DEFAULT_PROMPT_VERSION`. `tests/test_prompt_versions.py:238` already asserts the strictly stronger `history[-1] == (_sha256(DEFAULT_PROMPT), DEFAULT_PROMPT_VERSION)`. Three tests across two files now guard one invariant; this is the weakest of them.

**7. `tests/test_prompt_staleness.py:295-313` - copy-paste pair.**
`test_upgrade_flag_is_off_by_default` and `test_upgrade_flag_is_reachable_from_the_cli` are the same six lines differing by one CLI arg and the expected body. `@pytest.mark.parametrize("extra, expected", [([], OLD_BODY), (["--upgrade-prompts"], NEW_BODY)])` collapses 19 lines to about 10 and names the axis under test.

**8. `tests/test_prompt_staleness.py:297, 305, 361, 376` - the fixture exists and is bypassed.**
`synthetic_ledger` (line 74) does exactly `monkeypatch.setattr(init_cmd, "SCAFFOLDED_TEMPLATES", (SYNTHETIC,))`, yet four tests hand-roll that line. Requesting the fixture drops four duplicated lines and, with them, the four unannotated `monkeypatch` parameters at 295/304/360/375 - the only signatures in the file missing a type, so this also removes a real style inconsistency.

**9. Cross-file restatement of one rationale (5 copies).**
The causal story - "`_create_if_missing` never overwrites, `run_loop` prefers the file, `DEFAULT_PROMPT_VERSION` was read nowhere, so a bump reached nobody" - is written out in full at `kstrl/init_cmd.py:591-620`, `docs/adversarial-roadmap.md` H3b, the new CLAUDE.md bullet, `tests/test_prompt_staleness.py:1-27`, and `tests/test_prompt_versions.py:226-233`. Two stretches are near-verbatim across files: "an edited [file/prompt] and a [file/prompt] from a build outside this history are indistinguishable, so neither is claimed to be stale" (`init_cmd.py:712-714` / test:115-116) and "noise, and noise gets ignored" (`init_cmd.py:615-616` / test:166-167). CLAUDE.md itself names the roadmap as the policy home, so H3b is the copy that earns full length. Trimming `init_cmd.py:591-620` to a short statement plus a pointer at H3b leaves the argument in one place. Cost: five files to keep consistent on the next change, roughly 60 lines of prose saying the same thing.

I am deliberately **not** flagging: the MAINTENANCE paragraph (`init_cmd.py:622-628`), the `TemplateStatus` value table (707-714), the per-function docstrings, or the append-only rule as it appears in the three assertion/failure messages (`test_prompt_staleness.py:204-210`, `test_prompt_versions.py:176-180` and 239-242). Those are local, load-bearing, and a failure message is the right place to restate a rule.

Clean:

- **All four `TemplateStatus` values are load-bearing.** `absent` is not collapsible into `unrecognised`: the else branch at `init_cmd.py:1138-1139` would otherwise tell an operator that a file which simply is not there "matches no template kstrl has shipped; diff it yourself".
- **Every `ScaffoldedTemplate` field is read in production**: `constant_name` at 1135, `body` at 1123, `filename` at 736/796, `history` at 757, `current_label` at 798/1126/1129.
- **Every `TemplateState` field is read.** The mid-review edit that removed `revisions_behind` closed the only gap here; the two-label message that replaced it is the simpler form, and its comment (791-794) explains a real constraint rather than restating the code.
- No dead code, no deep nesting, no collapsed statuses. The classifier is a flat guard sequence and the upgrade reporter is a flat status match.
- `ruff check` and `ruff format --check` are both clean on the current tree.

### Efficiency

Efficiency-only review of the #286 diff. Everything is measured; no file was modified.

**1. `kstrl/cli.py:320` (`_check_prompt_preflight`) - clean.**
Measured on the shipped 4444-byte `DEFAULT_PROMPT`: 13.9 us median, 19.6 us p95 per prompt path (one `stat`, one read, one sha256, a linear scan of at most 8 rows). `import kstrl.cli` alone is 78 ms, so the added startup work is ~0.02% of the import, before any git/agent work. All five call sites (lines 792, 1094, 1583, 2616, 3570) sit at command-function top level - verified none is inside a loop - so `ks factory` pays it once in the parent regardless of component count, and `ks feature` pays it twice (its two prompts). No blocking network or subprocess work added.

**2. `kstrl/init_cmd.py:650` (`SCAFFOLDED_TEMPLATES`) - clean, and cheaper than it looks.**
The digests are source literals, so nothing is hashed at import time. Constructing the tuple plus its three frozen dataclasses measures 0.79 us against `kstrl.init_cmd`'s 648 us self import time (0.1%). `body=DEFAULT_PROMPT` binds the same string object - I verified `SCAFFOLDED_TEMPLATES[i].body is <constant>` is `True` for all three - so no duplicate ~4 KB copies are retained, and no large scope is captured.

**3. `kstrl/init_cmd.py:1089` (`_create_if_missing`) - clean.**
Only 3 of its 8 call sites carry a template basename. The other 5 (`kstrl.toml`, `prd.json`, `progress.txt`, `codebase_map.md`, `.gitignore`) exit inside `scaffolded_template_for` after 3 string compares with zero I/O.

**4. `kstrl/init_cmd.py:978` + `1100` - the one genuine redundancy, and it does not matter.**
`ks init --upgrade-prompts` classifies all three templates twice: `_upgrade_scaffolded_templates` calls `classify_scaffold(root)` (read + hash x3), then the scaffold pass immediately below calls `_create_if_missing` on the same three files, each re-entering `staleness_lines` (read + hash x3 again). Concrete cost: `classify_scaffold` measures 56 us for the three files, against a 4.0 ms `run_init --upgrade-prompts` inside a 78 ms+ process: 1.4% of a one-shot manual command. Threading state through to avoid it would add a mechanism to save 56 us once per invocation, so I would leave it. It also produces no duplicate output, because by the second pass the rewritten file classifies as `current` and `staleness_lines` returns `None`.

**5. `kstrl/init_cmd.py:741` (`classify_scaffolded_path`) - `exists()` + read is two syscalls where one `open` would do; leave it.**
The `exists()` check is load-bearing rather than redundant: `_read_text_or_none` swallows `FileNotFoundError` as an `OSError`, so without it `absent` would collapse into `unrecognised`. A `stat` is sub-microsecond. Not worth restructuring.

**6. `tests/test_prompt_staleness.py` - not slow; the three CliRunner tests are worth their cost.**
`uv run pytest tests/test_prompt_staleness.py -q --durations=15`: 28 passed in 0.39 s (0.61 s wall including interpreter start and collection). The slowest three are `TestOperatorActuallySeesIt::test_ks_run_*` at 0.09/0.08/0.08 s, 0.26 s combined, about two thirds of the file. That time is the 5 git subprocesses per test in `_project`, not the CLI invoke. Against the 4036-test suite this is noise. They could be made cheaper by sharing one module-scoped git repo and rewriting only `prompt.md` per test - worth roughly 0.15 s - but two of them monkeypatch a synthetic ledger while the third uses the real one, and one actually runs a factory that mutates the repo, so sharing would couple three tests to save 0.15 s. Dropping the end-to-end invoke instead would lose the only thing they prove (that the warning reaches the parent's stdout before the agent starts, which is the #275 lesson the whole placement rests on). Leave as is.

**7. Per-iteration work - clean.**
`kstrl/loop.py:572` reads the prompt once per `run_loop` call, before the iteration loop, and the diff does not touch it. Nothing new runs per iteration, and the one duplicated 4 KB read (preflight in the parent, then `run_loop` in the worker) is the pre-existing pattern already used by `_check_prd_preflight`.

**8. `tests/test_prompt_versions.py:225` - clean.** The new assertion is one `sha256` of an in-memory constant plus a generator scan of three entries; no I/O.

Nothing here is worth changing on efficiency grounds.

### Altitude

ALTITUDE review of #286. No files modified. Two findings are measured, not argued.

**1. `kstrl/tui/session.py:198` (and `kstrl/launch.py:83`) - the check was added to a caller, not to the assembler that exists to stop exactly this drift.**
`assemble_factory_configs` already owns the prompt-file fallback resolution, and its own docstring says "retry and the launcher share it so the two paths can never drift." The diff put `_check_prompt_preflight` in `cli.retry` (line 3570), one of its two callers. The other caller, `_prepare_factory` in `kstrl/tui/session.py`, runs a full factory after `_preflight_agent(base_config)` and gets no staleness check at all. Cost: home-shell factory launches - a surface this repo shipped a 16-PR stack for - never warn, and the drift the shared assembler was extracted to prevent is reintroduced on the day the mechanism lands. `assemble_factory_configs` is the choke point here; it would need a `ui` parameter (or to return the states for the caller to render), which is the only real cost.

**2. `tests/test_prompt_staleness.py:201` - "append-only, never edit or drop a row" is stated in four places and enforced in none.**
Measured, not inferred: I deleted 6 of the 8 `prompt.md` history rows in memory and re-ran every `TestLedgerIntegrity` assertion - all still pass, and the full 47-test run is green. `test_every_live_body_is_the_newest_row` pins only `history[-1]`; `test_the_body_that_cost_an_iteration_is_recorded` (line 227) pins exactly one hand-picked old row (1.1.1). Cost: the ledger's entire value is the old rows, and a future agent trimming "dead" entries - or editing the last row in place instead of appending, which is the natural mistake the docs warn about - passes CI and silently destroys the ability to recognise bodies already on disk. The repo already has the right-altitude mechanism for this: snapshot the whole `history` tuple the way `_EXPECTED_SNAPSHOTS` does, so only an append can pass. Right now this is the H3b rule as a hope.

**3. `kstrl/init_cmd.py:650` - no enrollment mechanism for a template added later.**
H3 solved precisely this problem one level up: `test_no_unenrolled_prompt_constants` AST-walks `kstrl/` and fails when a new `*_PROMPT` constant is not enrolled. The scaffold ledger has no equivalent. `test_filenames_match_what_init_scaffolds` (line 238) runs ledger to disk only; nothing runs `run_init`'s scaffold list to the ledger. Cost: the next `_create_if_missing(kstrl_dir / "x_prompt.md", DEFAULT_X_PROMPT, ui)` is un-ledgered, un-warned and un-upgradable, reproducing #286 for the new template with no test failing. The enrollment discipline was invented in this repo and not reused one level down.

**4. `kstrl/init_cmd.py:770` vs `:728` - the classifier is path-agnostic, the upgrader is path-hardcoded, so the advice names a remedy that cannot work.**
`scaffolded_template_for` matches on basename by design (documented), so it fires on a prompt anywhere. `classify_scaffold` only ever looks under `root/scripts/kstrl/`. Ran it: with `[paths] prompt` pointing at `prompts/prompt.md` holding a stale body, the warning fires, `ks init --upgrade-prompts` reports `prompt.md is missing; the scaffold below creates it`, creates a **second** unused `prompt.md` under `scripts/kstrl/`, leaves the real file at the old body, and the warning fires again forever. Cost: the one operator-facing remedy the warning names is a no-op for relocated prompts, and it silently adds a decoy file. The upgrader should take the same resolved paths the preflight classifies, not rediscover them from a hardcoded directory.

**5. `kstrl/cli.py:320` - the 5 call sites are the right altitude. Clean, with one note.**
I checked all four alternatives you named and each is genuinely disqualified: `KstrlConfig.validate` (config.py:424) returns `list[str]` of *errors* with no UI and no warning channel, so an advisory cannot live there without changing its contract and making a stale prompt look fatal; `factory._run_preflights` (factory.py:1656) is reached only by `factory`/`retry`, missing `run`/`understand`/`feature`; `run_loop` (loop.py:572) is where the read happens but runs in a pool worker writing to `engineer.jsonl` under the factory, which is the #275 precedent the docstring cites correctly. The new helper joins an existing family - `_check_agent_preflight` is called at the *same* 5 sites - so it is a third member of a convention, not a new special case. The one available consolidation (`_run_cli_preflights(config, ui)`) is blocked because `feature` (1583) and `factory` (2616) resolve `prompt_file` *after* the agent preflight, so the two checks legitimately cannot be adjacent. Note only: the docstring's admitted limit (the TUI dashboard drops `Log` events) means the default interactive `ks factory` shows this only in pre-Textual scrollback - honest, but it is the primary surface, and combined with finding 1 the TUI operator effectively never sees the warning.

**6. `kstrl/init_cmd.py:1092` - the staleness report inside `_create_if_missing` is the right depth. Clean.**
It hooks the exact decision point ("the file already exists"), is keyed generically so the four non-template calls return `None` untouched, and is not a `prompt.md` special case. One seam worth naming: `_create_if_missing` already holds the `content` it would have written and does not consult it, so `filename` to `constant` is a second source of truth parallel to the `run_init` call list at 983-993. Nothing asserts that `template.body` is the constant `run_init` actually writes to `template.filename`; a rename detaches the ledger silently. Low cost today, cheap to close.

**7. `kstrl/init_cmd.py:591` - the ledger's home is defensible. Clean.**
`init_cmd` is already the de-facto harness-template module (`loop.py:575` and `cli.py:67` both import constants out of it), so a separate module would import the bodies back and add a cycle risk for no gain. It does grow the file to 1639 lines and puts a classification/policy concern in a command module, but extracting it now would be motion, not depth.

**8. Generalisation to the two understand templates. Clean.**
The ledger, `classify_scaffold`, `_upgrade_scaffolded_templates` and `_create_if_missing` all treat the three templates uniformly, and `TestLedgerIntegrity::test_every_live_body_is_the_newest_row` loops over all of them. `test_engineer_prompt_bump_reaches_existing_projects` pins `prompt.md` specifically, but that is a targeted second guard placed in the file the person doing an H3 bump is already editing - belt-and-braces, not a special case. The date-vs-semver label split is handled by treating `label` as an opaque string, which is the right call.

**Priority if only two get fixed:** #2 (the mechanism is currently a hope, and it is the one thing the whole design rests on) and #1 (a shipped launch path with zero coverage, at a choke point that already exists).

</details>

Note on altitude finding 6's seam: `test_every_template_init_scaffolds_is_enrolled` now closes it. It reads the `(filename, constant)` pairs out of `run_init`'s own `_create_if_missing` calls and asserts they are exactly the ledger's, so a rename cannot detach the two silently.

---

Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01DHYpsJ8ecHLh5HdrswJtJK


0xfauzi:
## Review round 2: all four findings fixed, plus one regression the fix for #3 introduced

Thank you for measuring rather than reading. Two of the four were real defects I could reproduce, and chasing #2 turned up two more ways through that neither of us had.

Every guard in this PR is now verified by mutation: break it in the source, watch a named test refuse. Fifteen of fifteen.

| Broken deliberately | Refused by |
|---|---|
| ledger row dropped | `test_recorded_rows_are_never_edited_or_dropped` |
| ledger row edited in place | `test_recorded_rows_are_never_edited_or_dropped` |
| body changed with no new row | `test_every_live_body_is_the_newest_row` + `test_prompt_versions.py` |
| template scaffolded but not enrolled | `test_every_template_init_scaffolds_is_enrolled` |
| mode not preserved on upgrade | `TestUpgradeLeavesTheFileItFoundIntact` |
| symlink guard removed | `TestUpgradeLeavesTheFileItFoundIntact` |
| hard-link guard removed | `TestUpgradeLeavesTheFileItFoundIntact` |
| outside-the-project guard removed | `TestUpgradeLeavesTheFileItFoundIntact` |
| utf-8 unpinned on the read | `TestScaffoldRoundTripsItsOwnBytes` |
| utf-8 unpinned on the write | `TestScaffoldRoundTripsItsOwnBytes` |
| feature preflight back on the resolved path | `TestFeatureChecksThePromptItActuallyRuns` |
| understand fallback dropped again | `TestFeatureChecksThePromptItActuallyRuns` |
| feature_cmd stops using the carried path | `TestFeatureChecksThePromptItActuallyRuns` |
| wizard preview stops labelling | `TestTheWizardPreviewDoesNotSayEverythingIsFine` |
| wizard render drops the branch | `test_preview_flags_a_stale_prompt_instead_of_kept` |

### 1. Mode, fixed, and the premise of my docstring was wrong

Reproduced exactly: 0o644 after `ks init`, 0o600 after the upgrade. `os.fchmod(fd, mode)` on the descriptor already in hand copies the destination's mode across, with no second path lookup to race and no umask read (`run_init` is also called from the Textual wizard, so mutating the process umask to read it was not an option).

My docstring claimed this "matters here and not at the other three call sites". That is false, and I measured it: `workqueue.atomic_write` and `decompose._atomic_write_json` both turn a 0o644 file into 0o600, on git-tracked operator files (`scripts/kstrl/manifest.json`, per-component `prd.json`), and neither pins an encoding either. Git records only the exec bit, so a fresh checkout yields 0644 and the next `ks decompose` yields 0600, on the exact file `factory.py` reads with `prd_source.read_text()`. The docstring now says that plainly. **Fixing those two is a change to the work queue and the manifest writer, out of #286's scope, and worth its own issue.**

### 2. Sharing, fixed, and the first guard was too narrow

`is_symlink()` on the leaf only. Two more ways through, both measured:

- **Hard link.** `os.link(shared, prompt.md)`, upgrade: inode replaced, project on the new body, shared source still on the old one, **no warning at all**.
- **Symlinked `scripts/kstrl`.** Nothing on the leaf is a link. The upgrade rewrote a file **outside the directory `ks init` was pointed at**.

`_rewrite_blockers` now states the invariant rather than one instance of it: refuse a shared directory entry (symlink or `st_nlink > 1`), refuse anything resolving outside the project root, and name what to do instead. Reasons compose, so an operator with two of them is told about both. `_upgrade_remedy` is the single place that decides what `--upgrade-prompts` will do, so the warning cannot promise something the upgrade then refuses.

On following the link instead: rejected on scope, and now written down. `ks init <dir>` must not rewrite a file outside `<dir>`, and a shared prompt is by definition read by projects this invocation was not pointed at. One caveat I will not paper over: `ks init` has always **created** straight through a symlinked `scripts/kstrl`, and still does. This change refuses only to **rewrite** through it. Widening that is a change to `_create_if_missing`'s contract and is not #286's.

### 3. `ks feature`, fixed at a depth that cannot drift, and the first fix broke something

You were right that the coupling was a comment. It is now a value: the engineer prompt is resolved once in `cli.py` and carried on `FeatureParams.prompt_file`, which `feature_cmd` uses for both the implement and the repair loop. The literal is gone from `feature_cmd` entirely, so the file the preflight warns about and the file the loop opens are the same object, checked by mypy and by a test that stubs the loop and asserts on `config.prompt_file`. `ks feature`'s behaviour is unchanged: it still runs on `scripts/kstrl/prompt.md`.

I did not change the hard-coding itself. Making `ks feature` honour `[paths] prompt` like every other command is the obviously-right change and it is a behaviour change, which does not belong in a review-fix round.

**And the first version of this fix introduced a regression, which the review found and I have now covered.** Under `PROMPT_FILE`, `cli` passes `understand_prompt_file=None` and `feature_cmd` leaves the understand loop on the resolved config. Checking only the engineer path lost that file entirely, where before it was covered. The preflight now checks whichever of the two the understand phase will actually open, resolved in the same branch that decides the override.

### 4. Encoding, fixed and genuinely exercised

`encoding="utf-8"` pinned on `_read_text_or_none` and on `_create_if_missing`'s write. The in-process test cannot fail on a UTF-8 box, so it is re-run in a child under `LC_ALL=C PYTHONUTF8=0` (measured `US-ASCII` here, so the skip guard does not fire); that child fails with either pin removed.

### The wizard preview

Covered rather than documented. `plan_scaffold` carries a `stale_label` and the preview renders `! older template ... (shipped at 1.1.1)` instead of `· exists - kept`, tested through the existing wizard pilot on `#wizard-plan`, not just the dataclass. Deleting the render branch now fails a test.

### Gate

`uv run pytest tests/ -q`: 4025 passed, 28 skipped, plus the environmental shutdown failure noted below. mypy strict clean, ruff clean, `pre-commit run --all-files` all hooks pass, coverage 89.16% against the 79.91% ratchet.

Two hooks pushed back during this round and both were right, so the shape of the change moved to satisfy them rather than the number: `complexipy` refused `_render_preview` crossing 15 (the row formatting is now its own function) and `feature` worsening (the understand path is resolved in the existing branch instead of an added `or`), and `file-length-ratchet` refused the test file crossing 800 lines, so it is split along its two jobs into `test_prompt_staleness.py` (what the ledger says) and `test_prompt_upgrade.py` (what kstrl does about it).

One note on running the suite locally on a busy machine: `tests/test_shutdown.py::TestWorkerSigterm::test_pool_worker_sigterm_kills_agent_group` asserts `pgrep -f "sleep 60"` finds nothing, machine-wide. Any unrelated process whose command line contains that substring fails it. It failed for me under a sibling agent's `sleep 600` monitor loop and is unrelated to this diff; CI's `spine` job is the authority and it is green.

<details>
<summary>/simplify over the review-fix delta: four reports, verbatim (em dashes replaced with hyphens or colons per house style; otherwise unedited)</summary>

**Outcome: 17 applied, 4 declined with reasons, efficiency clean. The altitude report is why this round has three fixes the code review did not ask for.**

| # | Angle | Finding | Outcome |
|---|---|---|---|
| R1 | reuse | mode-copy re-implements `shutil.copymode` | declined |
| R2 | reuse | `_stale_label` re-derives what `classify_scaffold` returns per root | applied |
| R3 | reuse | symlink decision and its remedy authored twice | applied |
| R4 | reuse | cli/feature_cmd lockstep enforced only by a comment | applied, via altitude A |
| R5 | reuse | the new `_project` needs no git repo, and shadows a same-named helper | applied |
| R6 | reuse | re-derives `tests.conftest.REPO_ROOT` | moot, the probe is gone |
| R7 | reuse | three hand-rolled copies of the `synthetic_ledger` fixture | applied in part |
| R8 | reuse | wizard render harness exists and is unused | applied, via altitude G |
| S1 | simplification | two encoding tests cannot fail on the unfixed code | applied |
| S2 | simplification | `_LOCALE_PROBE` is a 30-line program in a string | applied |
| S3 | simplification | `_atomic_write`'s absent-target branch is unreachable | applied |
| S4 | simplification | dead fixture weight, and two same-named `_project` | applied |
| S5 | simplification | two mode tests differ only in an octal literal | applied |
| S6 | simplification | identical five-line symlink setup, copy-pasted | applied |
| S7 | simplification | three negative tests on a three-line pass-through | applied |
| S8 | simplification | `status == "stale"` evaluated and written twice | applied |
| S9 | simplification | four rationales written verbatim twice or three times | applied in part |
| S10 | simplification | `entry.action == "keep" and` is implied | applied |
| E1-E4 | efficiency | all clean, measured | nothing to do |
| A | altitude | fix 3 is a comment, not a mechanism | applied |
| B | altitude | the swap lost the understand phase under `PROMPT_FILE` | applied |
| C | altitude | symlink predicate breakable two ways, both measured | applied |
| D | altitude | the shared rule is fixed in one of three copies, on a false premise | applied in part |
| E | altitude | two halves of one rule at two depths | applied |
| F | altitude | orthogonal facts collapsed into an if/elif | applied |
| G | altitude | the wizard render is unenforced | applied |

**Declined, with reasons.**

- **R1** (`shutil.copymode`). It is path-based; `os.fchmod` uses the descriptor `mkstemp` already returned, so there is no second path lookup between creating the temp file and chmod-ing it. Two lines either way, and one of them has no window.
- **R7, in part.** The two remaining hand-rolled ledgers carry non-ASCII bodies the shared `SYNTHETIC` fixture cannot, since the whole point of those tests is a body the default fixture does not have. The four that could use the fixture now do.
- **S9, in part.** The test docstrings that restate a code comment were cut to what the reader of that test needs. The rule as it appears in assertion messages stays: a failure message is the right place to restate a rule.
- **D, in part.** `workqueue.atomic_write` and `decompose._atomic_write_json` have the same mode and encoding defect, measured. Fixing them is a change to the work queue and the manifest writer, well outside this diff, and belongs in its own issue. What is fixed here is my docstring, which asserted the opposite.

---

### Reuse

REUSE findings. Measurements were run against the working tree (43 tests pass, 2.66s).

**1. `kstrl/init_cmd.py:1152,1159` - the mode-copy re-implements `shutil.copymode`.**
`mode = stat.S_IMODE(target.stat().st_mode)` + `os.chmod(tmp_path, mode)` is exactly what `shutil.copymode(target, tmp_path)` does. Measured identical for 0644/0664/0600/0755/0640/0444 and setgid 2664. Both follow symlinks, so the swap is semantics-preserving at the one call site (`init_cmd.py:1203`).
Cost: 2 lines, plus the `import stat` at `init_cmd.py:9` that now exists solely for that one expression (`grep "stat\." kstrl/init_cmd.py` returns exactly line 1152).

**2. `kstrl/init_wizard.py:51-62` - `_stale_label` re-derives per-path what `init_cmd.classify_scaffold(root)` already returns per-root.**
`classify_scaffold` (`init_cmd.py:748`) classifies all three templates under `root/scripts/kstrl/`, which is exactly the subset of `plan_scaffold`'s 9 paths that can ever be stale. Verified equivalent on a live tree (labels agree on every classified path; every non-template path is `None`). The function also repeats `staleness_notice`'s stale-guard (`init_cmd.py:768-770`) verbatim, so "which statuses are speakable" is now written in two modules.
Cost: 12 new lines + a second import from `init_cmd`, and 9 basename lookups where 1 call does 3. Replacement is two lines inside `plan_scaffold`: `stale = {s.path: s.shipped_label for s in classify_scaffold(root) if s.status == "stale"}` then `stale_label=stale.get(p)`; `_stale_label` and the `classify_scaffolded_path` import both go away.

**3. `kstrl/init_cmd.py:781-786` vs `1187-1200` - the symlink decision and its remedy are authored twice in one module, 400 lines apart.**
Two `is_symlink()` calls and two `path.resolve()` calls decide the same thing about the same file in one `ks init --upgrade-prompts` run, and two different prose wordings state the same remedy ("upgrade that target and every project sharing it follows"). They are coupled only by a comment at line 777 ("and refuses a symlink outright"): no mechanism.
Cost: a wording or policy change has to land in two places, and `staleness_notice` can start promising a remedy the upgrade path no longer matches without anything failing. Cheapest fix that reuses the existing value object: add `is_symlink` to `TemplateState` (`init_cmd.py:704-710`, set once in `_classify`), and have the upgrade branch print `staleness_notice(state.path).advice` rather than a second copy of the text.

**4. `kstrl/cli.py:1603` - `kstrl_dir / "prompt.md"` must stay in lockstep with `feature_cmd.py:361` and `:451`, enforced only by a comment.**
The diff's own 10-line comment explains that this literal has to equal `run_config.prompt_file` / `repair_config.prompt_file`. There is no existing helper to call: I grepped, the path is spelled at `init_cmd.py:988`, `cli.py:1603`, `cli.py:2631`, `launch.py:84`, `init_wizard.py:79`, `feature_cmd.py:361`, `feature_cmd.py:451`.
Cost: the preflight silently starts checking the wrong file the day `feature_cmd` moves its literal. `cli.py` already imports `run_feature`/`FeatureParams` from `feature_cmd`, so exporting one `engineer_prompt_file(root_dir)` there and calling it in all three spots turns the comment into a mechanism.

**5. `tests/test_prompt_staleness.py:617-637` - the new `_project` is a near-copy of the same file's `TestOperatorActuallySeesIt._project` at line 724, and its git repo is not needed.**
Measured: `ks feature` reaches `_check_prompt_preflight` with no git repo, no `progress.txt` and no `feature_understand_prompt.md`: exit 0, warning printed, scaffolded path named (probe in scratchpad). The pre-existing copy at 724 genuinely needs git because `ks run` starts the factory. `tests/test_preflight.py:59 _scaffold` is the existing git-free project builder for this exact CLI, and `tests/test_preflight.py:246 TestFeaturePreflightWiring` is the existing invoke shape (it also stubs the loop to stop after preflight, via `_stub_run_loop` at line 32).
Cost: 5 unnecessary git subprocesses per test, measured at 27ms per project, ~54ms across the two tests; and two same-named `_project` helpers in one module where only one of them actually needs a repo, so a reader can't tell which invariant is load-bearing.

**6. `tests/test_prompt_staleness.py:540` - `repo_root = str(Path(__file__).resolve().parent.parent)` re-derives `tests/conftest.py:26 REPO_ROOT`.**
That constant is importable and already imported elsewhere (`tests/test_suite_isolation.py:28`, `tests/test_sense_cli.py:18` both do `from tests.conftest import ...`). Cost: 1 line, and the fourth copy of this expression in the suite (`test_safemode.py:23`, `test_progress_scope.py:1187`, `test_prompt_versions.py:360`).

**7. `tests/test_prompt_staleness.py:565-571`, `588-594` and the `_LOCALE_PROBE` body (`496-527`) - three more hand-rolled copies of the module's own `synthetic_ledger` fixture (line 79-84).**
Each builds a `ScaffoldedTemplate` and does the same two `monkeypatch.setattr(init_cmd, ...)` calls with a different body; line 209 is a pre-existing fourth. Cost: ~10 duplicated lines that a `_ledger(monkeypatch, body, history)` helper (or a fixture factory) would collapse.
Two related notes: (a) `test_a_non_ascii_template_scaffolds_and_classifies_as_current` (565-579) asserts exactly what the subprocess probe already asserts (`bytes_match`, `status == "current"`) minus the locale, and on this machine the subprocess test does not skip (`LC_ALL=C` yields `locale.getencoding() == "US-ASCII"`), so it is a strict subset here, ~20 lines for the platforms where the probe skips. (b) the double `setattr` is only needed because `run_init` writes the `DEFAULT_PROMPT` constant (`init_cmd.py:988`) rather than the identical body already held at `SCAFFOLDED_TEMPLATES[0].body`; sourcing that write from the template would halve the patching at all four sites. That last one is pre-existing code, but the new tests are what pays for it.

**8. `kstrl/tui/screens/init_wizard.py:197-202` - existing test harness not reused; the new render branch has no test.**
`tests/test_init_wizard.py:126 TestWizardScreen._run_wizard` already drives the wizard and reads `#wizard-plan`, asserting on strings like `"exists - kept"` (lines 219, 252). The new `"! older template ... (shipped at X)"` line is ~8 lines away from being covered by it; the diff tests only `plan_scaffold`'s `stale_label`. Related placement note: `tests/test_init_wizard.py::TestPlanScaffold` is the existing home for `plan_scaffold` tests, and it already uses the `{e.path.name: e for e in plan_scaffold(...)}` lookup idiom the new tests re-spell as `next(e for e in ... if ...)`.

Clean:

- **Symlink guard**: no general "is this path safe to overwrite" helper exists in `kstrl/`. `statedir._control_files_compromised` (`statedir.py:417`) and `workqueue._load_item_dir` (`workqueue.py:735`) are both purpose-bound to their own directories, not reusable here. Writing the check inline was correct.
- **`_mode` (`tests/test_prompt_staleness.py:404`)**: no permission-bit helper exists in `tests/spine_utils.py`, `tests/conftest.py` or `tests/helpers/`; every other test spells `stat.S_IMODE`/`chmod` inline. New helper is justified.
- **`_atomic_write` existing alongside `workqueue.atomic_write:585` and `decompose._atomic_write_json:811`**: the docstring's justification checks out, `tests/test_state_dir_scope.py` forbids putting `kstrl.workqueue` in `kstrl.loop`'s static import closure. One caveat worth stating: the mode fix landed on only one of the three copies, so they have now diverged behaviourally, and the docstring is the only record of why.
- **`encoding="utf-8"` pinning**: there is no shared "read text or None" helper in `kstrl/` to reuse; `_read_text_or_none` is the only one of its kind.

---

### Simplification

I ran the new tests and three measurement scripts to ground the claims; I did not modify any file under review.

**1. `tests/test_prompt_staleness.py:560` and `:581` - two tests that cannot fail on the unfixed code.**
Measured: under a UTF-8 locale, `write_text(s)` and `write_text(s, encoding="utf-8")` produce identical bytes, and `read_text()` reads them back identically. So `test_a_non_ascii_template_scaffolds_and_classifies_as_current` and `test_upgrade_writes_the_same_bytes_the_scaffold_would_have` exercise byte-identical paths before and after the encoding pin on every dev box and CI runner. Only the subprocess test at `:539` discriminates. Cost: ~40 lines that read as coverage of fix (4) but are locale-blind. Keep at most one (the upgrade/scaffold-parity assertion is a genuine invariant independent of encoding; the other is pure duplication of `TestClassification`).

**2. `tests/test_prompt_staleness.py:496-527` - `_LOCALE_PROBE` is a 30-line program in a string.** It re-implements the `SYNTHETIC` ledger, `run_init_capturing`, and the PlainUI/StringIO wiring the module already has, pays a `{{ }}` brace-escaping tax for one `.format` substitution, and needs `sys.path.insert` repo-root plumbing. Simpler form: have the child run the existing in-process test: `subprocess.run([sys.executable, "-m", "pytest", f"{__file__}::TestScaffoldRoundTripsItsOwnBytes::test_a_non_ascii_template_scaffolds_and_classifies_as_current", "-q"], env={**os.environ, "LC_ALL": "C", ...})`, which deletes the whole string and makes the in-process test from finding 1 load-bearing instead of redundant. Cost as written: ~45 lines and a second, divergent copy of the scaffold setup.

**3. `kstrl/init_cmd.py:1149-1151` - the absent-target branch is unreachable from the only caller, and it silently breaks the function's contract.** `_atomic_write` has exactly one caller (`:1202`), reached only when `state.status == "stale"`, and `_classify` returns `"absent"` when the path does not exist, so the target always exists. The branch does a plain non-atomic `write_text`, so the one guarantee the docstring sells ("atomic ... an interruption mid-write would truncate the very prompt this feature exists to protect") is exactly what a future caller creating a new file would not get. Cost: 3 code lines, the 7-line doc paragraph at `:1142-1148`, and the 14-line test at `tests/test_prompt_staleness.py:440`: ~24 lines defending a path nothing takes. The pre-existing `target.parent.mkdir(parents=True, exist_ok=True)` at `:1148` is dead for the same reason and can go with it.

**4. `tests/test_prompt_staleness.py:617-637` - measured dead fixture weight in `_project`.** I ran the class's two tests against trimmed fixtures: the five `spine_git` calls (init/config x2/add/commit), `progress.txt`, and `feature_understand_prompt.md` are all removable and both tests still pass. Cost: 7 lines per fixture plus the `spine_git` dependency. Removing the git block also mostly dissolves the near-duplication with `TestOperatorActuallySeesIt._project` at `:724`: the two shared only that boilerplate. Separately, two same-file methods both named `_project` with different signatures (`(tmp_path, body)` vs `(tmp_path, *, engineer_body, relocated_body)`) is a readability trap; name the new one for what it builds, e.g. `_project_with_relocated_prompt`.

**5. `tests/test_prompt_staleness.py:417-438` - two mode tests differ only in an octal literal.** `test_upgrade_keeps_the_mode_...` (0o644) and `test_upgrade_copies_the_mode_...` (0o664) share every line except the `chmod` argument, and their two docstrings assert the same rule. `@pytest.mark.parametrize("mode", [0o644, 0o664])` over one body. Cost: ~11 lines and a duplicated rationale.

**6. `tests/test_prompt_staleness.py:451-458` and `:473-480` - identical five-line symlink setup, copy-pasted.** `shared.write_text(OLD_BODY)` / `mkdir(parents=True)` / `symlink_to(shared)` appears verbatim in both. A module-level `_symlinked_prompt(tmp_path) -> tuple[Path, Path]` removes 5 duplicated lines and makes the two tests read as "the same situation, two questions".

**7. `tests/test_prompt_staleness.py:697-711` - three negative tests on a three-line pass-through.** `_stale_label` is `classify_scaffolded_path(path)` plus a status check; `TestClassification` (`:97-144`) already proves the classifier's current / edited / absent answers. `test_a_current_prompt_carries_nothing` and `test_a_missing_prompt_carries_nothing` have identical two-line bodies differing only in whether `run_init_capturing` ran. Cost: ~12 lines. One negative case plus the positive at `:689` covers the delegation. Worth noting the coverage is pointed at the cheap half: the TUI rendering at `kstrl/tui/screens/init_wizard.py:197-202`, the actual behaviour change an operator sees, has no test at all.

**8. `kstrl/init_cmd.py:1187` and `:1202` - the `state.status == "stale"` predicate is evaluated and written twice.** `if state.status == "stale" and state.path.is_symlink(): ... elif state.status == "stale":` costs the reader a second scan to confirm the two branches partition one status. Nesting the symlink test inside a single `if state.status == "stale":` is the same depth (2) and states the partition once.

**9. Four rationales are written verbatim twice or three times across the diff.** The pattern is systematic: `kstrl/init_cmd.py:1131-1141` ("mkstemp creates 0600 ... a container that runs `ks init` as one uid and the factory as another") is restated in the test docstrings at `:409-415` and `:420-423`; `:1244-1252` ("Every shipped body is ASCII today, which is exactly why this is cheap to pin now") is restated word-for-word at `:533-536`; `kstrl/cli.py:1593-1602` is restated at `:608-614`; and the "exists - kept reads as your scaffold is fine" sentence appears three times: `kstrl/init_wizard.py:42-47`, `kstrl/tui/screens/init_wizard.py:198-199`, and `:682-687`. This is not the heavy-comment house style, which explains a mechanism once at its home; it is the same paragraph maintained in two or three places that will drift apart. Keep the code-side comment, cut each test docstring to a one-line pointer at it. Cost: ~35 prose lines and three future drift sites.

**10. `kstrl/tui/screens/init_wizard.py:197` - the `entry.action == "keep" and` conjunct is implied.** A non-None `stale_label` requires `_classify` to return `"stale"`, which requires `path.exists()`, which is exactly the condition that sets `action = "keep"`; `.gitignore` is appended with the field defaulted to None. `if entry.stale_label:` is sufficient. Trivial, but it currently invites the reader to look for a case where a stale label attaches to a create or an append.

The two questions you asked me to rule on:

**The three-way `remedy` in `staleness_notice:781-793`: all three branches are load-bearing** (symlink, in-scaffold-dir, relocated) and each produces a message the other two would get wrong. It is already flat: two independent predicates, three outcomes, no nesting. The one thing worth changing is not the shape: the symlink remedy string at `:782-786` and the `ui.info` block at `kstrl/init_cmd.py:1197-1202` are two hand-maintained phrasings of the same operator instruction ("upgrade the target and every project sharing it follows"). Hoist that sentence to one module constant both use.

**The `stale_label` field on `ScaffoldEntry`: clean, keep it.** It is derivable, but `plan_scaffold` is the single planning API (the module's own docstring commits to "all content comes from run_init's existing scaffold functions") and computing it there keeps the classification testable without driving Textual, which is what tests `:689-711` do. The alternative pushes an `init_cmd` classifier call into the render loop for no line saved. The I/O is not a concern: `_render_preview` runs on button press, not per keystroke, and 5 of the 9 paths short-circuit on the basename lookup before touching disk.

**Clean:** the `encoding="utf-8"` pins at `:1114` and `:1252`; the `_mode` test helper's use of `lstat`; the `_stale_label` helper itself; the `cli.py` preflight one-line change.

---

### Efficiency

Nothing needs changing. All numbers below are measured on this machine (macOS, min-of-7 timing runs), not estimated.

**1. `plan_scaffold` calling `_stale_label` - clean, and the premise is wrong twice over**

`kstrl/init_wizard.py:92`

**It is not keystroke-triggered.** `_render_preview` has exactly one caller: `kstrl/tui/screens/init_wizard.py:249`, inside `on_button_pressed` under `wizard-preview-btn`. That screen has no `on_input_changed`, no `Input.Changed`/`Select.Changed` handler, no `on_key`, and no `set_interval`. The only `set_interval` in the TUI is `home.py:205`, and home's `_render_preview` (`home.py:293`) is the *run* preview: it never calls `plan_scaffold`. So the added work runs once per deliberate button press by a human.

**It does not read and hash ~10 files.** `classify_scaffolded_path` matches on basename against the 3 entries in `SCAFFOLDED_TEMPLATES`, so 6 of the 9 paths return `None` with zero I/O. Measured per call: 0.25 us for a non-template basename (`CLAUDE.md`), 17.16 us for a present `prompt.md` (stat + 4.4 KB read + sha256).

Measured `plan_scaffold`, fully-scaffolded project (worst case, all three prompts present):

| | before | after | delta |
|---|---|---|---|
| scaffolded dir | 51.2 us | 103.8 us | +52.6 us (2.03x) |
| empty dir | 36.8 us | 45.3 us | +8.5 us |

53 microseconds once per button press, against a Textual frame render measured in milliseconds. A 2x relative jump on a number this small is not a cost.

Minor, not worth fixing: `p.exists()` runs for `action`, then `_classify` (`init_cmd.py`) stats the same path again for the 3 template names. 3 x 1.29 us = 3.9 us of the 52.6.

**2. `_atomic_write` stats - one genuinely redundant stat, 1.3 us, leave it**

`kstrl/init_cmd.py:1149-1152`: `target.exists()` then `target.stat()` is two stats where one would do (`try: mode = stat.S_IMODE(target.stat().st_mode) / except FileNotFoundError:` covers both branches). Cost of the extra one: **1.29 us**.

Total syscalls the diff added to `_atomic_write`: `exists` (1.29) + `stat` (1.29) + `chmod` (5.88) = **8.5 us per call**. It has exactly one caller (`init_cmd.py:1203`) and at most 3 stale templates per run, so **25.6 us per `ks init --upgrade-prompts`**, a human-typed command. Not worth touching.

On "`_upgrade_scaffolded_templates` already classified it": there *is* a real duplicated read on that path, but it **predates this diff**. `run_init` calls `_upgrade_scaffolded_templates` (line 984, then `classify_scaffold`, then 3 x read+hash), then `_create_if_missing` (lines 987-994) calls `staleness_notice`, then `classify_scaffolded_path`, and reads and hashes the same 3 files again. Each prompt is read and hashed twice per upgrade run: 2 x (10.6 us read + 1.6 us hash) x 3 = ~73 us, of which ~36 us is waste. Not introduced here, and not worth restructuring `run_init` for 36 us.

Unrelated to efficiency but spotted while counting: the new docstring says the mode-preservation "matters here and not at the other three call sites": `_atomic_write` has no other call sites in this module.

**3. `_LOCALE_PROBE` - 42 ms, and worth it**

`tests/test_prompt_staleness.py:539`

Measured child wall time: **min 42 ms, median 43 ms**. Decomposed: bare interpreter spawn 12 ms, spawn + `import kstrl.init_cmd` 27 ms. So the marginal cost over the import it has to do anyway is ~15 ms. `--durations` agrees: 0.04 s, against 3.12 s for the file's 43 tests (~1.3%) and a 4051-test suite.

Worth it, on two measured grounds:
- The probe returns `{"encoding": "US-ASCII", "bytes_match": true, "status": "current"}` on this machine: it does **not** take the skip branch, so the assertions genuinely execute under a non-UTF-8 locale. It is the only test in the set that can fail if the `encoding="utf-8"` pins are reverted; the in-process sibling at line 560 runs in the parent's UTF-8 locale and would pass either way.
- 23 `sys.executable` spawn sites already exist across 10 test files, so this is the house pattern, not a new tax.

**4. `cli.py:1603` preflight swap - clean**

Substituting `kstrl_dir / "prompt.md"` for `base_config.prompt_file` is cost-neutral: one `staleness_notice` either way. The only edge is a former `None` (immediate return, no I/O) now doing one classify, which on an absent file is a single `exists()` at 1.29 us, on a command that then spawns an LLM agent.

**Verdict.** No efficiency findings worth acting on. The diff adds 53 us to a button-press handler, 26 us to a human-typed CLI command, and 42 ms to a test suite, and the 42 ms buys the only coverage that can catch the regression it guards.

---

### Altitude

Two of the four fixes are at the right depth. Two are bandaids whose stated invariant I was able to break by measurement, and one shared rule got fixed in one of four copies.

**A. `kstrl/cli.py:1603` - fix 3's coupling is a comment, not a mechanism**

`_check_prompt_preflight(kstrl_dir / "prompt.md", ui_impl)` is correct *today* only because `kstrl/feature_cmd.py:361` and `:451` hard-code the same literal. Nothing enforces that. Grepped: `run_config.prompt_file` / `repair_config.prompt_file` appear only at those two lines and inside the new cli.py comment; `tests/test_feature_cmd.py` never touches `prompt_file`. And the hard-code was never a decision: `git log -L361,362:kstrl/feature_cmd.py` shows it arrived whole in `2cfc00a refactor: extract the feature flow to feature_cmd.run_feature`.

Cost: the day someone makes `ks feature` honour `[paths] prompt` (the obviously-right change: README:294 documents `prompt` as "engineer prompt file" with no exception, and `docs/env-vars.md:17` documents `PROMPT_FILE` the same way), cli.py silently reverts to warning about the wrong file. The two new tests in `TestFeatureChecksThePromptItActuallyRuns` will not catch it: they monkeypatch `run_feature` away, so they exercise cli.py alone and would keep passing while asserting the wrong thing.

The author's "don't change feature_cmd's behaviour" call is half right. Changing runtime behaviour in a review-fix PR is correctly out of scope; leaving the coupling unenforced is not. The behaviour-preserving altitude: resolve the engineer prompt once in cli.py, carry it on `FeatureParams`, and have feature_cmd do `run_config.prompt_file = params.prompt_file`. One field, checked by mypy, and the preflight then checks what the loop reads by construction. cli.py can still pass the hard-coded literal, so nothing about `ks feature` changes.

**B. `kstrl/cli.py:1603-1604` - the diff replaced a check instead of adding one**

`feature_cmd.py:225-226` overrides `understand_config.prompt_file` only when `params.understand_prompt_file is not None`, and `cli.py:1587` sets it to None exactly when `PROMPT_FILE` is in the environment. So under `PROMPT_FILE`, the understand phase runs on `base_config.prompt_file`, which, after this diff, nothing checks. Before the diff it was covered.

Cost: `PROMPT_FILE=... ks feature` now has zero staleness reach on the file its first phase actually opens. Concrete regression introduced by the swap.

**C. `kstrl/init_cmd.py:1187` - the symlink guard tests a mechanism, not the invariant it claims; I broke it**

The comment states the rule as "`ks init <dir>` must not rewrite a file outside `<dir>`". `state.path.is_symlink()` only tests the leaf. Measured, on this filesystem:

- **Symlinked directory.** `scripts/kstrl` a symlink into a shared dir, `prompt.md` a plain file inside it: the same shared-harness setup the guard exists for. `run_init(root, upgrade_prompts=True)` rewrote the file **outside the root** and reported `OK:   prompt.md: 9.0.0 -> 9.1.0 (it held no local edits)`. `ks init` also created `understand_prompt.md` and `feature_understand_prompt.md` outside the root on the same run.
- **Hardlink.** `os.link(shared, kstrl_dir/"prompt.md")`, then upgrade: link broken (inode differs after), project copy on the new body, shared source still on `# old engineer instructions`, **no warning**. That is verbatim the outcome the guard's own comment says it exists to prevent, one filesystem primitive over.

Refusing is the right call. The predicate is too narrow. The invariant form is `state.path.resolve()` inside `root.resolve()` (catches both symlink shapes), plus `st_nlink > 1` for hardlinks, or, simpler and stronger, write through the existing inode (`open(path, "w")`) instead of replacing it, which preserves both link kinds and the mode for free and drops the mode-copy code entirely.

**D. `kstrl/init_cmd.py:1136-1139` - the shared rule was fixed in one of four copies, and the docstring's reason for not generalizing is measurably false**

The docstring: *"this is the only one that rewrites an operator-owned, group-readable file in their repo rather than kstrl's own state."* Measured:

```
decompose._atomic_write_json: 0o644 -> 0o600
workqueue.atomic_write:       0o644 -> 0o600
```

`decompose._atomic_write_json` (`kstrl/decompose.py:811`) writes `scripts/kstrl/manifest.json`, `scripts/kstrl/feature/<id>/prd.json` and `spec-issues.json`; `Manifest.save` (`kstrl/manifest.py:343`) writes `scripts/kstrl/manifest.json` and `run-manifest.json`. All git-tracked, all in the operator's repo. Git records only the exec bit, so a checkout yields 0644 and the next `ks decompose` yields 0600: the exact multi-uid failure the docstring cites, on the exact file `kstrl/factory.py:1960` reads via `prd_source.read_text()`.

Both siblings also use `os.fdopen(fd, "w")` with no `encoding=`, so fix 4 is one copy deep as well, on PRDs and manifests, which are far likelier to hold non-ASCII than the ASCII-only prompt templates the test defends. Same for `factory.py:1973` `worktree_prompt.write_text(prompt_source.read_text())` and `loop.py:573` `config.prompt_file.read_text()`: the byte round-trip is pinned in the classifier and unpinned in the two places that actually feed the agent (19 bare `read_text()` calls in `kstrl/`).

Cost: the rule now lives in one place and is violated in three, with a comment asserting the opposite. Fix `workqueue.atomic_write` (mode + encoding) and route manifest/decompose through it; keep init_cmd's copy for the documented #274 import-closure reason, but as a literal mirror.

**E. `kstrl/init_cmd.py:1148-1159` - two halves of one rule at two depths**

The mode half went into the shared helper; the symlink half went into the caller one `elif` away. So `_atomic_write` will still clobber a symlink for any future caller, and `target.stat()` follows the link, meaning it would stamp the *target's* mode onto the replacement regular file. "Do not clobber what the operator owns" is one rule and wants one home: `_atomic_write` itself, which is where the `os.replace` that does the damage lives.

Separately, the new `if not target.exists(): target.write_text(...); return` branch is unreachable in production: `_upgrade_scaffolded_templates` only calls it for `status == "stale"`, which requires the file to exist. It exists to serve `test_atomic_write_creating_a_new_file_honours_the_umask`. Cost: a function named `_atomic_write` that is not atomic on a branch nobody runs, plus a TOCTOU between `exists()` and `stat()`. Cheaper: keep one path and `os.fchmod(fd, mode)`: the fd is already in hand, no second path-based syscall, no branch.

**F. `kstrl/init_cmd.py:781-793` - orthogonal facts collapsed into an if/elif**

"Is it a symlink" and "is it under `scripts/kstrl`" are independent; the chain makes symlink dominate. A prompt relocated by `[paths] prompt` that is *also* a symlink gets "Upgrade that target and every project sharing it follows", but `--upgrade-prompts` only ever visits `scripts/kstrl/`, so it will not reach the target either. The only true advice for that case, the `else` branch's "move this one there or replace it by hand", is unreachable.

Cost: the operator with the most complicated setup gets the least accurate remedy. Compute both facts and compose.

**G. `kstrl/init_wizard.py:48` - right layer; the render is unenforced**

`stale_label` on the model is correct. One classifier (`classify_scaffolded_path`), three renderings (`_create_if_missing`, `_upgrade_scaffolded_templates`, the wizard): the TUI should not classify, and it doesn't. `_stale_label` inherits the "no claim we cannot prove" rule verbatim. Keep it.

But nothing tests `kstrl/tui/screens/init_wizard.py:197`. `tests/test_init_wizard.py:253` already drives the wizard with a pilot and reads `#wizard-plan`, so the assertion is ~6 lines away. As shipped, deleting `and entry.stale_label` leaves every test green: the four new tests only cover the dataclass field. That is a hope, not a mechanism, on the one surface the class docstring argues is "most likely to be read" as "your scaffold is fine".

Minor shape note: `action` and `stale_label` are parallel, so every future renderer must remember to check both or it silently prints "exists - kept". A `status` field (or a fourth `ScaffoldAction`) makes the case unmissable.

**H. Clean**

- Fix 4's locale probe is a real mechanism. Verified the skip guard does not fire here (`LC_ALL=C LANG=C PYTHONUTF8=0` gives `US-ASCII`, subprocess cost 13ms), and pre-fix the unencoded `write_text` raises `UnicodeEncodeError` on that body, so it genuinely fails without the fix.
- `_create_if_missing`'s `encoding="utf-8"` at `kstrl/init_cmd.py:1114`: right place, matches the read it has to agree with.
- Full suites green: `tests/test_prompt_staleness.py` 43 passed, `tests/test_init_wizard.py` + `tests/test_init_cmd.py` 49 passed.

**Altitude verdict.** Fix 1 (mode) and fix 4 (encoding) are at the right depth *locally* but were stopped one layer short of the shared helper they were copied from, on a false premise (D). Fix 2 (symlink) is the right call at the wrong predicate: the invariant it states is breakable two ways, both measured (C). Fix 3 is the one to redo: it moves a fact about another module's internals into a comment, and the deeper fix (hoist the path onto `FeatureParams`) is behaviour-preserving, so "out of scope" does not buy the shortcut (A, B).


</details>

Issue #264, opened 2026-08-28
Unwinnable Phase 1: diff-scope rejects the PRD and progress log that prd_stories and self_critique require the agent to write (measured: $8.26 on two identical failures)


0xfauzi:
Found while dogfooding kstrl as a first-time user, building a real web app (`writers-room`). **Measured cost: $8.26 on two identical failures**, with a third attempt in flight that I killed because the loop cannot terminate.

## What happened

One component, `allowedPaths: ["src/writers_room/", "tests/"]`. Phase 1 failed on `diff_scope`, twice, with byte-identical output:

```json
{"event": "verification_result", "data": {"passed": false, "checks": ["diff_scope"],
 "failures": ["3 file(s) outside the component's allowed scope, caught in-loop after
  iteration 1: scripts/kstrl/codebase_map.md,
  scripts/kstrl/components/document-format.prd.json,
  scripts/kstrl/components/progress.txt. ...
  Allowed paths (complete list): src/writers_room/, tests/. Do not widen allowedPaths."]}}
```

```json
{"event": "component_usage", "data": {"phase": "engineer", "cost_usd": 3.519853, "duration_seconds": 678.73}}
{"event": "component_retrying", "data": {"attempt": 1, "reason": "Mechanical verification failed"}}
{"event": "component_usage", "data": {"phase": "engineer", "cost_usd": 4.738521, "duration_seconds": 810.34}}
{"event": "component_retrying", "data": {"attempt": 2, "reason": "Mechanical verification failed"}}
```

## Why the agent cannot win

Two of the three rejected files are files **kstrl's own mechanical checks require the agent to write**:

| Rejected by `diff_scope` | Required by |
|---|---|
| `scripts/kstrl/components/document-format.prd.json` | `check_prd_stories` (`verify.py:527`): "Re-read PRD from disk and verify all stories have passes=true". Only the agent can set `passes`. |
| `scripts/kstrl/components/progress.txt` | `check_self_critique` (`verify.py:394`): the current progress entry must carry a Self-Critique block. Only the agent writes it. |
| `scripts/kstrl/codebase_map.md` | The engineer prompt's codebase-map convention. |

`check_diff_scope` (`verify.py:765`) is a pure allowlist with no carve-out:

```python
violations = [f for f in changed if not path_is_allowed(f, allowed_paths)]
```

[...]

PR #293, opened 2026-08-31
fix(scope): resolve component scope once, at plan time

### LINKED ISSUE #269: Resolve component scope once at plan time so self-widening is impossible rather than detected
Split out of #264 (merged as #268), where it was deliberately left undone and recorded in the PR's reviewer notes.

## What #268 did, and why it is only half the answer

#268 closed the case where kstrl's scope guard fired on files kstrl itself requires the engineer to write. Carving those files out makes the component PRD agent-writable by design, so #268 also added tamper detection: `PRD.tamper_changes` compares the worktree PRD against the pre-run copy and pins every `UserStory` field except `passes` and `notes`.

Detection is the weaker of the two available answers. The stronger one is to make the tampering unable to matter.

## The proposal

Resolve each component's scope once, at plan time, and hand the same immutable snapshot to both guards. The in-loop guard (`guards.enforce_allowed_paths` via `loop.run_loop`) and the Phase 1 gate (`verify.check_diff_scope`) would then be reading a value the agent never had access to, and self-widening becomes impossible rather than detected.

That also removes a failure mode tamper detection introduces on its own: a false positive stops a component dead, and #268's first attempt at this comparison did exactly that. It used list equality, so a benign reorder of `allowedPaths` failed Phase 1 closed and the component retried into an identical failure forever, which is the same class of unwinnable loop #264 was filed about. That was caught in review and fixed to compare sets and fail only on additions, but the general shape of the risk is intrinsic to comparing a value the agent can rewrite.

## A pre-existing asymmetry this would settle

The two guards do not agree today about where scope comes from:

- `factory._component_scope` falls back to the run-wide `--allowed-paths` flag.
- `_resolve_verify_scope` does not.

So the in-loop guard and the Phase 1 gate can be enforcing different allowlists in the same run. This predates #264 and #268 and is not caused by either. One resolution at plan time gives both the same answer by construction, rather than by two code paths independently arriving at it.

## Scope note

`use_worktrees=False` is a separate case and should be stated explicitly rather than silently covered. In that mode the worktree IS `root_dir`, so there is no isolation boundary and no pre-run copy distinct from the live one. #268 documented the tamper check as not covering that mode. A plan-time snapshot would cover it, which is one more argument for doing it this way.

## Suggested acceptance

- Both guards read one snapshot resolved before the first engineer call.
- The snapshot's provenance is recorded, so a scope decision is auditable after the fact.
- The two code paths agree on the run-wide `--allowed-paths` fallback, whichever answer is chosen.
- `PRD.tamper_changes` either becomes redundant and is removed, or is kept deliberately as defence in depth with a stated reason.

0xfauzi:
Fixes #269

## What was wrong

#264 carved kstrl's own files out of every component's scope guard, which makes the component PRD agent-writable by design: the engineer has to write it to set `passes`. #268 landed that carve-out and defended the now-writable file by DETECTION, comparing the worktree PRD against a pre-run copy.

Detection is the weaker of the two available answers, and it carries a cost of its own. #268's first version compared `allowedPaths` as LISTS, so a benign re-serialisation failed Phase 1 closed, the component retried, the agent re-serialised the same way, and it failed identically forever: the same class of unwinnable loop #264 was filed about, created by the fix for it. Review caught that and narrowed it to set-additions, but the shape of the risk is intrinsic to comparing a value the agent can rewrite. Every false positive is a hard stop.

Underneath that, two guards answered "what may this component write?" from different files:

| | scope source | run-wide `--allowed-paths` fallback |
|---|---|---|
| in-loop guard (`factory._component_scope`) | PRD at `root_dir` | yes |
| Phase 1 (`pipeline._resolve_verify_scope`) | PRD in the WORKTREE | no |

So one run could enforce two different allowlists on the same component, and the gate's copy was the one the agent could edit.

## What changed

`kstrl/scope.py`. `RunScope.resolve(manifest, root_dir, base_config)` reads every component's scope out of the pre-run tree ONCE, before the pipeline exists and before the scheduler can launch anything, into a frozen `ComponentScope` per component. Both guards read that snapshot and nothing else:

- in-loop: `_run_factory_locked` -> `_submit_args` -> `_run_component(scope=...)` -> `run_loop`
- Phase 1: `ComponentPipeline.run_scope.for_component(comp.id)` -> `check_diff_scope`

Neither reads a PRD to answer the scope question any more. `factory._component_scope` and `pipeline.VerifyScope` are gone, `_run_component` no longer rebuilds the harness carve-out from its own path strings, and `kstrl/factory.py` no longer imports `PRD` at all.

Self-widening is not detected, it is impossible: the value both guards enforce was read before the agent existed.

## The `--allowed-paths` asymmetry: keep the fallback, apply it at both

**Keep it.** The flag is operator-authored, is not agent-writable, and is the ONLY scope source `ks run --allowed-paths` has: R2.3/CRIT-8 exists precisely because that flag was a silent no-op under the factory pipeline. Dropping the fallback would have made Phase 1 the arbiter of a list the operator never gets to set, and would have re-broken `ks run --allowed-paths`.

Adopting it at Phase 1 is a tightening that cannot surprise anyone: the in-loop guard was already holding the engineer to that exact list, so the diff reaching Phase 1 has already been judged against it. Previously a legacy PRD plus `--allowed-paths` meant the engineer was constrained in-loop while the gate enforced nothing at all.

Precedence is unchanged and tested: a component PRD's own `allowedPaths` still wins over the flag, and neither present still means "no constraint", which `check_diff_scope` passes.

## `PRD.tamper_changes`: kept as defence in depth, with its scope arm deleted

The check is kept, and the enumeration is the reason. What the snapshot covers is exactly one field. Everything else Phase 1 and Phase 2 read out of the WORKTREE copy still has to come from the live file, because the agent setting `passes` is the whole job and a snapshot cannot answer "did it finish?":

| reader | file | covered by the snapshot? |
|---|---|---|
| `check_diff_scope` (allowedPaths) | was worktree, now the snapshot | **yes, removed from the comparison** |
| `verify.check_prd_stories` (stories, `passes`) | worktree | no |
| `fixtures.check_fixtures_from_prd` (approved oracles) | worktree | no |
| `review.run_review` (acceptance criteria into the reviewer prompt) | worktree | no |
| `review.revert_unconfirmed_stories` (story ids) | worktree | no |
| `pipeline._setpoint_disagreements` / `_has_unconfirmed_claim` (R10.3) | worktree | no |

Delete the comparison and an agent can drop an acceptance criterion, rewrite a story title, or neuter an executable fixture and pass a gate it authored. So `PRD.tamper_changes` stays, and its `allowedPaths` arm (`pipeline._widened_scope`) is deleted rather than relaxed, because that field is now inert and refusing an edit to it could only ever be a false positive that strands a component.

Two things moved with it:

- **It now fails `prd_stories`, not `diff_scope`.** Scope stopped being the question, so "scope configuration could not be trusted" was an untrue thing to say about a rewritten story. `prd_stories` is also the check that reads the stories, and it runs whenever there is a PRD at all, whereas `diff_scope` is switchable off by `[verify] check_diff_scope`: defence in depth an unrelated toggle can disable is not defence in depth. There is a test that drives `run_mechanical_verification` with `check_diff_scope=False` and asserts `prd_stories` is the only failure.
- **It lives in `verify.py`.** `run_mechanical_verification` takes `pre_run_prd_path: Path | None` instead of a pre-computed message, so the check is self-contained and reproducible by any caller rather than needing `pipeline` imported first. `ks sense` passes `None`, exactly as before.

## `use_worktrees=False` is covered, and here is the confirmation

#268 documented its tamper check as not covering that mode, correctly: there the "worktree" IS `root_dir`, so the two paths name one file and there is nothing to compare.

The snapshot covers it, and the mechanism is that it does not compare anything. `RunScope.resolve` runs in `_run_factory_locked` before `_run_preflights` and long before the scheduler's first submission; the value is held in memory for the run. The agent can rewrite `scripts/kstrl/feature/<id>/prd.json` in the operator's own tree and neither guard will read it again.

It also closes a per-retry hole that only existed in that mode: `_submit_args` runs once per ATTEMPT, so the old `_component_scope(comp, root_dir, base_config)` re-read the PRD on every retry. Without worktrees, an agent that widened `allowedPaths` during attempt 1 got the widened list enforced in attempt 2. `TestBothGuardsReadOneSnapshot` runs entirely in `use_worktrees=False` and rewrites the PRD from inside the fake engineer loop, so the covered claim is the default in these tests rather than a special case.

What is NOT covered there is unchanged and stated: the tamper comparison still has nothing to compare in that mode, so the stories, criteria and fixtures are undefended under `--no-worktrees`. That mode has no isolation boundary at all, and this change does not pretend to give it one.

## Provenance

Each component's decision is recorded as a `component_scope_resolved` event in the run's `events.jsonl`, carrying the authored list, the harness carve-out, which authority supplied the scope (`component_prd` / `run_flag` / `unconstrained` / `unresolved`) and the file or flag it came from. The terminal gets a one-line summary, plus a named warning for any component whose scope could not be established, because that is the one that will fail Phase 1 closed later.

The payload field is `scope_source`, not `source`: `source` is an envelope field on every event, and a payload field of that name is silently overwritten by the bus and dropped from the record. A test caught it.

## Fail-closed, still

- **Pre-run PRD unreadable AND no `--allowed-paths`**: the snapshot carries `unresolved` plus an error, which reaches `check_diff_scope` as `allowed_paths_error` and fails it CLOSED (R1.5). The family is `OSError`, not `FileNotFoundError`, because a PRD path that is a directory raises `IsADirectoryError` (R8 review finding 5).
- **Pre-run PRD unreadable WITH a flag**: the operator's list stands; the same broken file fails `check_prd_stories` anyway, so the run still fails, with the accurate check name.
- **Worktree PRD corrupted by the agent**: `check_prd_stories` fails on the load. What it can no longer do is unbind its own write scope by corrupting the file, which is what the old fail-closed arm was defending.
- **A component with no plan-time snapshot** (unreachable while the pipeline and the snapshot are built from one manifest): `RunScope.for_component` returns an `unresolved` stand-in with an EMPTY carve-out, so it reports more, never less.

## How it was tested

`tests/test_scope_snapshot.py`, 15 tests, all end-to-end ones through a real `run_factory` with both guard seams instrumented (`kstrl.loop.run_loop` and `kstrl.factory.run_mechanical_verification`), never re-derived:

- both guards handed the same authored list and the same carve-out;
- a PRD rewritten from inside the engineer loop reaching neither guard, with the file's new contents asserted on disk so the test cannot pass by failing to write;
- the `--allowed-paths` fallback identical at both guards; the component PRD still beating it; neither present still meaning unconstrained;
- the worktree half at the Phase 1 seam (a widened worktree copy ignored);
- every component snapshotted regardless of status; an unknown component failing closed; provenance recorded per source; the event carrying both lists and the reason an unresolved scope was unresolved.

`tests/test_prd_tamper.py` reworked: the scope-comparison tests are replaced by "no shape of scope edit can strand a component" (reorder, narrowing, field dropped), and the story/fixture/branchName refusals now run through `check_prd_stories`, including one that pins the refusal firing with `diff_scope` switched off.

Full local gate green: `uv run pytest tests/ -q` **4034 passed / 28 skipped**, `uv run mypy kstrl/ --strict`, `uv run ruff check kstrl/ tests/`, `pre-commit run --all-files` (33 hooks). Coverage **89.26%** against the 79.91% ratchet; `kstrl/scope.py` at 100%.

One local test is deselected above and passes in CI: `test_shutdown.py::TestWorkerSigterm::test_pool_worker_sigterm_kills_agent_group` asserts `pgrep -f "sleep 60"` finds nothing, and this machine has an unrelated `sleep 600` running, which that prefix matches. Verified as environmental, not caused by this change: `pgrep -f "sleep 60"` exits 0 here with the branch's own processes gone.

No `*_PROMPT` constant was touched, so no calibration run, version bump or snapshot update is implicated (H2/H3).

## /simplify findings, verbatim

Four agents (reuse, simplification, efficiency, altitude). Their findings as returned, then what was applied and what was declined.

### Reuse

> Read the full diff plus both new files, and grepped `config.py`, `guards.py`, `git.py`, `manifest.py`, `prd.py`, `events.py`, `statedir.py`, and the adjacent test modules for existing helpers.
>
> Note: the working tree moved while I was reviewing (`kstrl/config.py` picked up an edit mid-review, and its stale `factory._run_component` reference is already fixed). Line numbers below are as of now.
>
> The core of the change reuses well - `ComponentScope.resolve` goes through `KstrlConfig.component_harness_files` rather than `component_harness_paths`, `VerifyScope` was replaced rather than duplicated, and `tests/test_scope_snapshot.py` imports `_setup_project`/`_write_prd`/`_component`/`_manifest`/`_base_config`/`_pipeline` instead of re-rolling them. Five things I would still change:
>
> ## 1. `tests/test_scope_hardening.py`, lines 459, 501, 563
>
> **Three hand-copied duplicates of the `run_mechanical_verification` signature, and this diff paid the tax on all three.**
>
> Each `spy_rmv` stub restates the full 13-parameter signature verbatim. Adding one keyword to the real function cost three identical mechanical edits in this change (`+ prd_tamper_error: str | None = None` at 467, 509, 571) that assert nothing. Worse, a fake that drifts from the real signature fails at call time with a `TypeError` that reads as a test bug, not a signature change.
>
> **Use instead:** the pattern this same change already introduced one file over - `tests/test_scope_snapshot.py:_run`'s `def fake_verify(*args: Any, **kwargs: Any) -> VerificationResult` with `kwargs.get("harness_paths")`. All three stubs only read `allowed_paths`, `allowed_paths_error`, and `harness_paths`; `*args, **kwargs` captures those and makes the next parameter addition cost zero test edits.
>
> ## 2. `kstrl/pipeline.py`, line 393
>
> **New path-identity check re-implements `statedir._same_directory`, and is redundant against the comparison it guards.**
>
> ```python
> if worktree_copy.resolve() == pre_run_copy.resolve():
>     return None
> ```
>
> Two costs. First, this is the codebase's second "do these two paths name one thing" idiom; `statedir._same_directory` (`kstrl/statedir.py:172`) already does it via `os.path.realpath` with a documented rationale for non-existent paths and the macOS `/tmp` symlink, plus an `OSError` fallback to `normpath`. Two idioms means a future fix to one silently misses the other. Second, the branch is redundant: under `use_worktrees=False` both loads read the same file, so `tamper_changes` returns `[]` and the function returns `None` anyway - which is exactly how the old code got this case for free, as its own docstring said.
>
> **Use instead:** either drop the branch (the answer already falls out), or, if the explicit early return is wanted to avoid the double read, promote `statedir._same_directory` to a public `same_path` and call it. Third option: the only caller, `_phase_verify` (line 2544), already holds `self.factory_config.use_worktrees` - the authoritative signal, no filesystem call needed.
>
> ## 3. `tests/test_scope_snapshot.py`, line 112
>
> **Seventh byte-identical copy of the same `FactoryConfig(...)` literal.**
>
> The 7-field block (`use_worktrees=False, create_prs=False, max_parallel=1, max_retries=0, retry_delay=0, review_mode="skip", progress_log_path=<root>/"progress.jsonl"`) now appears at `test_harness_path_scope.py:554, 587, 638`, `test_progress_scope.py:163, 522, 595`, and here. Pre-existing duplication that this change extends by one. A default change in `FactoryConfig` that ought to break one test breaks seven, and the two `test_progress_scope.py` copies have already drifted (they carry `verify_config=None`), so the set is no longer even self-consistent.
>
> **Use instead:** a `_factory_config(root: Path) -> FactoryConfig` helper beside `_base_config` in `tests/test_progress_scope.py` - the module this new file already imports `_base_config` and `_pipeline` from, so the import line is unchanged.
>
> ## 4. Rationale prose restated in five places instead of cross-referenced
>
> The "#268 compared `allowedPaths` as lists, a benign re-serialisation failed Phase 1 closed, the component retried into the identical failure forever" story is written out in full at `kstrl/scope.py:12-17`, `kstrl/pipeline.py:353-357`, `kstrl/prd.py:366-372`, `tests/test_prd_tamper.py:18-25`, and `tests/test_scope_snapshot.py:4-9`. One historical fact, five prose copies that must move together or start contradicting each other.
>
> **Use instead:** the codebase's own convention, which the docstring this change rewrote was already following - `prd.tamper_changes` previously said "``allowedPaths`` is the one field compared LOOSELY ... see ``pipeline._widened_scope``". State the history once (`kstrl/scope.py`'s module docstring is the right home) and have the other four say "see ``kstrl.scope``" in a sentence.
>
> ## 5. `kstrl/scope.py`, lines 137 and 188
>
> **Identical 5-argument construction written twice.**
>
> `return cls(comp.id, run_flag, harness, "run_flag", "--allowed-paths")` appears verbatim in `_unreadable` and in `resolve`. Minor, but it means the `"run_flag"` source tag and the `"--allowed-paths"` origin string each have two homes, and the two are positional-argument calls, so a field reorder on the dataclass corrupts both silently.
>
> **Use instead:** a third small classmethod (`cls._from_run_flag(comp, harness, run_flag)`) alongside `_unreadable`, or at minimum keyword arguments at both sites.

### Simplification

> Read the full diff, both new files, and traced every consumer of the new snapshot (`check_diff_scope`, `run_loop`'s `guard_ignored_paths`, `_authored_scope_errors`) to check what the added indirection actually buys.
>
> Five findings, ordered by value.
>
> **1. Four names for two values on `ComponentScope`**
>
> - File: `kstrl/scope.py`
> - Line: 95-96 (fields) and 106-119 (properties)
> - Summary: `allowed_paths`/`harness_paths` are stored as tuples and then re-exposed as `authored`/`harness`, which copy them into fresh lists on every read.
> - Cost: readers must learn that `scope.allowed_paths` and `scope.authored` are the same list under two names, and the diff already mixes them within four lines of each other - `factory.py:1574` tests `scope.allowed_paths`, `factory.py:1577` passes `scope.authored or []`, `factory.py:1607-1608` uses `scope.allowed_paths or ()` next to `scope.harness_paths`, while `pipeline.py:2537-2540` uses `scope.authored`/`scope.harness`. The `or []` at `factory.py:1577` is dead: the `continue` two lines above already proved it non-None, and it exists only because the property hides that narrowing from mypy. The stated justification (a consumer mutating what it was handed) has no consumer: `check_diff_scope` builds `effective = [*allowed_paths, *(harness_paths or ())]` and never mutates, `_authored_scope_errors` only iterates and joins, `loop.run_loop` splats `*(guard_ignored_paths or ())`. On the in-loop path the value also crosses a `ProcessPoolExecutor` boundary, so the worker gets a pickled copy regardless. The only thing exercising the copies is a test written for them (`tests/test_scope_snapshot.py:248-262`).
> - Simpler form: delete both properties. Either declare the fields as `list[str] | None` / `list[str]` (the frozen dataclass already blocks rebinding, and nothing mutates the contents), or keep the tuples and widen the three consumer parameters to `Sequence[str]` - `verify.check_diff_scope`, `verify.run_mechanical_verification`, `loop.run_loop`. Either way the surface drops to two names and the `or []` / `or ()` noise at the three call sites goes with it.
>
> **2. `_resolve_verify_scope` is now a one-line delegator**
>
> - File: `kstrl/pipeline.py`
> - Line: 2455-2488
> - Summary: the method's entire body is `return self.run_scope.for_component(comp.id)`, under 33 lines of docstring, with one production caller (`pipeline.py:2522`).
> - Cost: an indirection that no longer resolves anything, named as if it does. Its three test callers (`tests/test_harness_path_scope.py:743`, `tests/test_progress_scope.py:364`, `tests/test_scope_snapshot.py:244`) now assert against the delegator rather than against the wiring, so the seam's test value is mostly illusory - none of them would catch `_phase_verify` reading a different scope.
> - Simpler form: inline it - `scope = self.run_scope.for_component(comp.id)` at line 2522 - and move the surviving rationale onto `RunScope.for_component`, where `scope.py`'s module docstring already carries most of it. Point the three tests at `pipeline.run_scope.for_component(comp.id)`.
>
> **3. The `run_flag` fallback is written twice, which is what forces `_unreadable` to exist**
>
> - File: `kstrl/scope.py`
> - Line: 121-138 and 176-189
> - Summary: `cls(comp.id, run_flag, harness, "run_flag", "--allowed-paths")` appears verbatim at line 137 and line 188, and three near-identical `except` arms (178, 180, 182) differ only in a message prefix.
> - Cost: five return statements for four outcomes, spread over two methods, so the reader has to hold both tails in their head to see that "PRD unreadable" and "PRD has no scope" converge on the same fallback. `_unreadable` is a helper whose only reason for existing is that the exception handlers return eagerly.
> - Simpler form: record the failure in a local and fall through to one tail of returns:
>
>   ```python
>   authored: tuple[str, ...] | None = None
>   error: str | None = None
>   try:
>       prd = PRD.load(root_dir / comp.prd_path)
>   except FileNotFoundError as exc:
>       error = f"pre-run PRD not found: {exc}"
>   except (OSError, ValueError) as exc:
>       error = f"pre-run PRD could not be read: {exc}"
>   else:
>       authored = tuple(prd.allowed_paths or ()) or None
>   if authored is not None:
>       return cls(comp.id, authored, harness, "component_prd", comp.prd_path)
>   if run_flag is not None:
>       return cls(comp.id, run_flag, harness, "run_flag", "--allowed-paths")
>   if error is not None:
>       return cls(comp.id, None, harness, "unresolved", comp.prd_path, error)
>   return cls(comp.id, None, harness, "unconstrained", comp.prd_path)
>   ```
>
>   `_unreadable` disappears; the "PRD not found" wording that `tests/test_scope_snapshot.py:349` pins is kept by the one narrow arm.
>
> **4. `component_id` duplicates the mapping key, and `entries()` exists only to compensate**
>
> - File: `kstrl/scope.py`
> - Line: 94 (`component_id`) and 231-233 (`entries`)
> - Summary: `RunScope.by_component` is keyed by component id and every value carries the same id again; `entries()` is a one-line `tuple(self.by_component.values())` with exactly one caller (`factory.py:1598`).
> - Cost: two pieces of surface added for a single loop, and the field can in principle disagree with the key it is filed under.
> - Simpler form: drop both. `_record_run_scope` iterates `run_scope.by_component.items()` and uses the key for `component=`; `for_component`'s fail-closed stand-in loses its first positional argument.
>
> **5. Redundant same-path short circuit in `prd_tamper_error`**
>
> - File: `kstrl/pipeline.py`
> - Line: 391-394
> - Summary: `if worktree_copy.resolve() == pre_run_copy.resolve(): return None` produces the answer the code below already produces - loading one file twice yields two equal PRDs, `tamper_changes` returns `[]`, and the function returns `None`.
> - Cost: two `realpath` syscalls per component per verify, plus a branch a reader has to work out is not covering a distinct case. The `use_worktrees=False` bullet in the docstring is what documents the mode; the code adds nothing to it.
> - Simpler form: delete the three lines and keep the docstring bullet. The only thing they buy is avoiding a double read of one file, and `_phase_verify` runs after the loop has exited, so there is no writer to race. If the double read is the real motivation, say that in the comment rather than framing it as the non-worktree case.
>
> Nothing else worth changing. The `harness_paths` parameter added to `_run_component` mirrors the pre-existing `allowed_paths or []` convention, `_record_run_scope`'s three passes over `entries` are ordered deliberately (summary before warnings), and the `check_prd_stories(prd_path, tamper_error)` parameter is a genuine relocation rather than added surface.

### Efficiency

> I reviewed the diff plus the two untracked files, traced every PRD read through the call graph, and benchmarked `PRD.load` so the cost numbers are measured rather than guessed.
>
> ## Verdict on the stated concern: no regression - the diff is a net I/O reduction
>
> I counted the per-run PRD reads before and after. Per component with `A` Phase 1 attempts:
>
> | file | before | after |
> |---|---|---|
> | pre-run copy (`root_dir/prd_path`) | `A` (`_submit_args` -> `_component_scope`) + `A` (`prd_tamper_error`) = **2A** | `1` (`RunScope.resolve`, plan time) + `A` (`prd_tamper_error`) = **1 + A** |
> | worktree copy | `A` (`_resolve_verify_scope`) + `A` (`check_prd_stories`) = **2A** | `A` (`prd_tamper_error`) + `A` (`check_prd_stories`) = **2A**, and **fewer** on a tampered PRD, since `check_prd_stories` now short-circuits before its read |
>
> `prd_tamper_error` taking on the second load is exactly offset by `_resolve_verify_scope` dropping to a dict lookup. Nothing is read per-retry that used to be read once; the opposite happened - the per-submission read in `_submit_args` is gone.
>
> Measured: `PRD.load` is 58 us on a realistic 24 KB PRD (20 stories, 10 fixtures, 15 allowed paths), 12 us on a stub. `Path.resolve()` is 16 us.
>
> ## Finding
>
> **1. `kstrl/pipeline.py:395-396** - `prd_tamper_error` re-parses two PRDs that are already parsed elsewhere in the same run, and discards both.
>
> ```python
> pre_run = PRD.load(pre_run_copy)
> worktree_prd = PRD.load(worktree_copy)
> ```
>
> - The pre-run copy was already opened, parsed and schema-validated by `ComponentScope.resolve` at `kstrl/scope.py:177`, which keeps only `allowed_paths` and drops the rest. That file is invariant for the run by the module's own argument (outside every worktree, not agent-writable), and in the one mode where it *is* writable (`use_worktrees=False`) this function returns at line 391 before reaching the load. So it is re-parsed once per Phase 1 pass to reproduce a value the run already had.
> - The worktree copy is parsed here, discarded, then re-parsed by `check_prd_stories` at `kstrl/verify.py:591` a few microseconds later in the same `run_mechanical_verification` call.
>
> Concrete cost: 2 x 58 us of redundant open + `json.loads` + full schema validation per Phase 1 attempt (fixture entries are validated key-by-key, so the cost scales with PRD size). Against a Phase 1 pass that shells out to pytest, mypy and ruff, this is not a latency finding - it is a shape finding, and I flag it on that basis rather than on the microseconds.
>
> Cheaper alternative: have `ComponentScope` carry the `PRD` it already loaded (a field beside `allowed_paths`) and let `prd_tamper_error` take it as an argument, as the pre-#269 signature did for the worktree side. That drops one load per attempt and, more usefully, removes the `except (OSError, ValueError): return None` swallow on the pre-run half - the snapshot already classified that failure at plan time into `unresolved`/`run_flag`, so re-deriving it here means the same broken file is diagnosed in two places with two different outcomes. The worktree-side duplicate is the symmetric fix: pass the parsed worktree `PRD` into `check_prd_stories` instead of the path.
>
> ## Checked and clean
>
> - **Blocking work added to run startup** (`kstrl/factory.py:2924-2925`): `RunScope.resolve` reads N PRDs and `_record_run_scope` emits N `component_scope_resolved` events. At 58 us per load and one flushed `JsonlSink` write per event, a 12-component manifest costs well under 10 ms. Both must precede `_preflight_component_scope`, which consumes the snapshot. Not worth moving.
> - **Closures keeping a large scope alive**: `_submit_args` (`kstrl/factory.py:3272`) and `ComponentPipeline` (`kstrl/pipeline.py:485`) both hold `run_scope`, but that is a `MappingProxyType` of frozen dataclasses holding short strings. Nothing large is retained.
> - **Per-iteration work**: `harness_paths` is now handed to the pool worker instead of rebuilt there (`kstrl/factory.py:2062-2071`), and `loop.run_loop` still folds `guard_ignored_paths` into `guard_ignored` once before the first iteration (`kstrl/loop.py:638-643`). No new work landed inside the agent loop.
> - `component_progress_rel` survives the removal of the `component_harness_paths` call - it is still used at `kstrl/factory.py:2076`. No dead locals left behind.
> - The two `Path.resolve()` calls at `kstrl/pipeline.py:391` cost ~32 us per Phase 1 pass to detect a mode `self.factory_config.use_worktrees` already knows, but `resolve()` is the more robust test (symlinked worktrees) and the price is nil. I would leave it.

### Altitude

> **1. `kstrl/factory.py:1853` (and `:3340-3341`)**
> The snapshot is un-made at the pool boundary: `ComponentScope` is split into two positional slots (`allowed_paths`, `harness_paths`) in a ~30-element positional tuple that `_submit_args` builds and `executor.submit`/`functools.partial` splats with `*args`.
>
> - *Cost*: the two halves of one immutable object are now kept in sync by hand-counted index alignment across two files - the same fragility the code at `:3521` already warns about ("a positional extra silently lands on `redirect_output`"). Worse, `source`, `origin` and `error` never reach the worker, so the in-loop guard's rejection message (`:2210-2219`) cannot say whether the list it enforced came from the architect's PRD or from `--allowed-paths`, while the event record it was resolved into can. The provenance the change deliberately introduced stops at the process boundary.
> - *Deeper fix*: pass the frozen `ComponentScope` as one argument. It is a dataclass of `str`/`tuple` fields, so it pickles for `ProcessPoolExecutor` without change. `_run_component` then takes `scope: ComponentScope | None`, and the two `harness_paths or []` re-defaults at `:2163` and `:2218` collapse into `scope.harness`.
>
> **2. `kstrl/pipeline.py:2544` (with `kstrl/verify.py:1847`, `:557`)**
> Tamper detection is computed in `pipeline` and injected into `run_mechanical_verification` as a pre-computed failure *string*, making it the third `X_error: str | None` parameter on that function alongside `allowed_paths_error`.
>
> - *Cost*: `run_mechanical_verification` is the repo's mechanical-check entry point, but this one check's evidence is produced outside it and cannot be reproduced by it. Any other caller wanting the same defence (a future `ks verify`, the contract tester, an operator-facing re-verify) has to know to call `pipeline.prd_tamper_error` first and import `pipeline` - a heavy module - to do it. The `PipelineHooks` signature widens for a value the hook cannot derive.
> - *Deeper fix*: now that the comparison no longer feeds scope, it has no reason to live in `pipeline`. Move it beside `check_prd_stories` in `verify.py` and give `run_mechanical_verification` a `pre_run_prd_path: Path | None` instead of a message. `ks sense` passes `None` exactly as it does today, and the check becomes self-contained like every other mechanical check.
>
> **3. `kstrl/feature_cmd.py:242` (also `:373`, `:476`; compare `kstrl/cli.py:1397`)**
> `run_loop`'s scope contract is still "read `config.allowed_paths` off the config object, plus an optional `guard_ignored_paths` kwarg the caller must remember". Of the three callers, only the factory now uses the typed snapshot; `ks understand` hand-derives its carve-out via `standalone_harness_files`; `ks feature` passes **no** `guard_ignored_paths` at all across its three loops, while setting `understand_config.allowed_paths = [rel_feature_understand]`.
>
> - *Cost*: #269 created a first-class type for "what may this loop write", and the loop it guards still takes that answer through two unrelated channels - a config field and a keyword - so a caller supplying only half is silent. That is exactly the shape that left `ks feature` without the #264 carve-out, and the new module does nothing to make the omission visible to the next reader.
> - *Deeper fix*: give `run_loop` a single `scope: ComponentScope | None` parameter and have the guard read `scope.authored` / `scope.harness` from it. Add a `ComponentScope.standalone(config, root_dir)` classmethod (source `run_flag`, origin `--allowed-paths`) so `cli.py` and `feature_cmd.py` construct the same object the factory does. The type then covers all three callers instead of one, and `feature_cmd`'s gap becomes a missing argument rather than a missing behaviour.
>
> **4. `scripts/atlas/logical_model.py:443-451`**
> `kstrl/scope.py` is a new top-level module and is not claimed by any component in the atlas logical model; `PathGuard` still lists only `kstrl.guards`.
>
> - *Cost*: `scripts/atlas/check_layout.py::check_coverage` fails any module "drawn by no component", so the next `scripts/atlas/refresh.sh` turns the Atlas CI job red with `coverage: kstrl.scope is drawn by no component`. Until then the system map shows the path guard without the thing that now decides its scope. (The `extract_atlas.py --check` gate is already stale on this branch from earlier commits, so this will surface at refresh time rather than immediately.)
> - *Deeper fix*: add `kstrl.scope` to `PathGuard`'s `implemented_by` (or give it its own `ScopeSnapshot` component in the build region, which is arguably what it is - it is resolved at plan time, not at guard time).
>
> **5. `kstrl/pipeline.py:393`**
> `if worktree_copy.resolve() == pre_run_copy.resolve(): return None` is a special case for `use_worktrees=False` that the general path already handles: under that mode `wt_path` **is** `root_dir` (`factory.py:3241`), so both `PRD.load` calls read one file, and `tamper_changes` on two equal PRDs returns `[]`.
>
> - *Cost*: two `Path.resolve()` syscalls per verification, and a reader is told that mode needs special handling here when it does not - which invites the next person to add a second mode-specific branch beside it.
> - *Deeper fix*: delete the early return and keep the docstring paragraph explaining why the mode falls out of the general comparison for free.

### Applied

- **Altitude 1**: `_run_component` now takes one `scope: ComponentScope | None` instead of two positional slots. `_worker_scope` unpacks it, which also keeps the complexity ratchet happy: `_run_component` is already the largest function in the module and is judged against its own previous value.
- **Altitude 2**: `prd_tamper_error` left `pipeline.py` entirely. The comparison is `verify._tamper_changes`, folded into `check_prd_stories`, and `run_mechanical_verification` takes `pre_run_prd_path: Path | None`.
- **Simplification 1**: `authored`/`harness` deleted; the fields are plain lists and every call site reads `allowed_paths` / `harness_paths`. The copy-protection test went with them; the docstring now says plainly that consumers copy at their own boundaries.
- **Simplification 2**: `_resolve_verify_scope` inlined into `_phase_verify`; the three tests now go through `pipeline.run_scope.for_component(comp.id)`.
- **Simplification 3 + reuse 5**: one tail of returns in `ComponentScope.resolve`, `_unreadable` gone, the three distinct load-failure messages kept (a test pins "PRD not found").
- **Simplification 4**: `component_id` and `entries()` dropped; `_record_run_scope` iterates `by_component.items()`.
- **Simplification 5 + altitude 5 + reuse 2**: the `resolve()` short circuit is gone. Folding the comparison into `check_prd_stories` also removed the double read of the worktree copy that efficiency 1 flagged, so that mode now falls out of the general path with one load each.
- **Reuse 1**: the three `spy_rmv` stubs in `test_scope_hardening.py` are `*args/**kwargs`.
- **Reuse 4**: the #268 history is told once, in `kstrl/scope.py`'s module docstring; `prd.tamper_changes` and the two test modules point at it.

### Declined

- **Efficiency 1, the remaining half** (have `ComponentScope` carry the parsed pre-run `PRD`): declined. It would put a mutable `PRD` inside the object whose immutability is the point, keep a full PRD per component alive for the run, and re-couple scope resolution to tamper detection immediately after altitude 2 separated them. The agent measured it at 58 us against a Phase 1 that shells out to pytest, mypy and ruff, and flagged it as a shape finding rather than a latency one. Folding the comparison into `check_prd_stories` removed the worktree-side duplicate anyway.
- **Altitude 3** (`run_loop` takes a `ComponentScope`; a `ComponentScope.standalone` for `cli.py` and `feature_cmd.py`): declined here as scope creep. It changes the guard contract of three commands this issue does not touch. The premise is also partly stale: #285 deliberately REMOVED `ks feature`'s `guard_ignored_paths` entry because `guard_state_root=root_dir` now carves out `.kstrl/logs/` as part of the whole state directory, so what looks like a gap is a subsumed declaration. Worth its own issue.
- **Reuse 3** (a shared `_factory_config` test helper): declined. This file has exactly one copy, inside one helper; consolidating the six pre-existing ones is unrelated churn on tests whose values have already drifted apart.
- **Altitude 4** (atlas): out of scope for this branch by the owner's instruction.


0xfauzi:
## Round two: the five review findings

Commit `bd5f02e`. Three of the five were my own stated principle broken inside the PR that states it, and one rested on a docstring claim that was false. All five fixed, none pushed back on.

### 1. The preflight skipped the one scope that provably cannot pass (MEDIUM/HIGH)

Correct, and the worst of the five. `if scope.allowed_paths is None: continue` swallowed exactly the `unresolved` snapshot: the component was launched, `check_diff_scope` failed closed on `scope.error`, and because the snapshot is fixed for the life of the run every retry reproduced the identical failure. #264's unwinnable-loop-with-spend, in the one branch this PR kept.

`_preflight_component_scope` now refuses it, per component, naming the file and the reason. `_run_preflights` runs before the scheduler can launch anything, so the refusal costs one line of output in the first second instead of `max_retries` engineer loops.

### 2. A silent broadening, justified by a false claim (MEDIUM)

Also correct, including that the docstring was false. `check_prd_stories` reads `wt_path / prd_path`; a repository that tracks its component PRDs gives the worktree a valid copy from the branch while the working-tree copy in the main checkout is truncated or `chmod 000`. So "the same unreadable file also fails `check_prd_stories`" was not true, and the flag silently replaced an authored scope nobody could read.

`ComponentScope.resolve` now orders the error BEFORE the `run_flag` fallback. The rule, stated where it lives: **a scope that could not be READ is not a scope that does not exist.** "The architect authored none" is knowledge, and the operator's run-wide list is the intended authority for it; "could not read it" is the absence of knowledge, and the component may have had a narrow authored scope. The docstring now records what was false and why, so the next reader does not restore the ordering.

The other half is tested too, so the fix cannot be read as "the flag no longer works": a legacy PRD that LOADS and carries no `allowedPaths` still takes the flag, with `error=None`.

### 3. My own argument, applied to the other check (LOW/MEDIUM)

Correct. `allowed_paths_error` was consumed only inside `if config.check_diff_scope:`, so `[verify] check_diff_scope = false` dropped the fail-closed signal entirely, and with a None authored list the in-loop guard is inert too.

The gate is now `_diff_scope_runs(config, allowed_paths_error)`, a named predicate whose docstring carries the argument. It costs no git call in that branch: `check_diff_scope` returns on the error before it reads a diff. The toggle still removes the comparison when there is no error, which is tested, so it did not become "diff_scope always runs".

### 4. Alarming warnings about work that will not run (LOW)

Correct. The warn loop is deleted rather than filtered. After finding 1 the PENDING components are REFUSED moments later, which is louder than a warning and stops the run, and the per-component event still records every component's scope for audit. A warning that duplicates a refusal for the components that matter and lies about the ones that do not is not worth keeping.

### 5. The audit record named the wrong origin (LOW)

Correct. `origin` now means one thing everywhere: where the authored list came from, empty when nothing supplied one. Both `unconstrained` and `unresolved` name no origin. The repository-relative path moved into the `error` text, which is where the statement is about the failure, so nothing is lost from the record and `events.ComponentScopeResolved.origin`'s own docstring is true again.

### Tests

`tests/test_scope_snapshot.py` and `tests/test_scope_hardening.py`:

- the preflight refuses a corrupt PRD and a missing one, asserting **nothing was called**: no engineer, no verification, exit 2, and the message names the file;
- the signal survives the flag being present, at both layers: `resolve` records `unresolved` with `--allowed-paths` set, and the preflight acts on that same snapshot object rather than one it resolved for itself;
- a legacy PRD that loads still takes the flag;
- the fail-closed refusal fires with `check_diff_scope` both false and true, and the toggle still removes the comparison when there is no error.

Six test modules gained a component PRD in their fixtures. A manifest naming a PRD that is not on disk could never have run in production: `_run_component` copies that exact file into the worktree and `check_prd_stories` re-reads it. `tests/helpers/component_prd.py` is the one writer, and `test_harness_path_scope._write_prd` delegates to it rather than serialising its own.

Full local gate green: `pytest tests/ -q` **4041 passed / 28 skipped**, `mypy kstrl/ --strict`, `ruff check kstrl/ tests/`, `pre-commit run --all-files` (33 hooks). No `*_PROMPT` constant touched.

---

## /simplify over the round-two changes, verbatim

Four agents. Findings as returned, then applied and declined.

### Reuse

> Reviewed the working tree as it stands now. Note: the tree moved while I was reading it (`PASSING_STORY`, the `body=` escape hatch, and `COMP_PRD_REL` appeared mid-review, replacing an earlier `_STORIES` literal and a `_write_comp_prd` local writer). Findings below are against the current state.
>
> **Nothing to flag on the reuse axis in the three production files.** `kstrl/factory.py`, `kstrl/scope.py`, `kstrl/verify.py` add no logic that the codebase already has; the PENDING filter and the `component '<id>': ` prefix they reuse are pre-existing lines.
>
> ## Findings
>
> **1. `tests/helpers/component_prd.py:38` (and `:28`) - the new "one writer" is the suite's second PRD writer.**
> `write_component_prd` re-implements `_write_prd` at `tests/test_harness_path_scope.py:110`, and `PASSING_STORY` re-implements `STORY` at `:100`. Those are not module-private: they are already the suite's shared, cross-module PRD fixture pair, imported by `tests/test_prd_tamper.py:35-40` (21 call sites) and by `tests/test_scope_snapshot.py:63` (3 call sites). Cost: the split runs straight through this change - five touched modules now use the new writer while `test_scope_snapshot.py`, also touched, keeps calling `_write_prd`; a new required PRD field has to be added in two places, and the new module's docstring claim "This is the one writer" is false on arrival. `write_component_prd` also cannot replace `_write_prd` as written, because it has no `fixtures` parameter. Use: add `fixtures: list[dict] | None = None` to `write_component_prd`, then reduce `test_harness_path_scope._write_prd` to a thin wrapper that supplies its `[dict(STORY)]` default and delegates (its 26 call sites keep working), and re-export `STORY` as an alias of `PASSING_STORY`.
>
> **2. `tests/test_scheduler.py:172-174` - the worktree PRD copy is still hand-rolled.**
> `_setup` does `mkdir(parents=True)` + `json.dumps` + `write_text` for the same file `_init_repo` (line 51) writes through the helper. Cost: this is exactly the pair the comment at `:61-64` says must stay identical for the Phase 1 tamper compare, and only one of the two goes through the writer that guarantees it; fixing it also removes the `import json` this change added at line 20 (line 174 is `json`'s only use in the module). Use: `write_component_prd(wt, f"scripts/kstrl/feature/{comp_id}/prd.json", stories=[PASSING_STORY])`.
>
> **3. `tests/test_scope_snapshot.py:380-397` - the new class bypasses its own helper.**
> `test_the_toggle_still_removes_the_comparison` re-inlines the `run_mechanical_verification` call and the 6-field `VerifyConfig` that `_verify` (line 356) already wraps; only `allowed_paths` (`list(AUTHORED)` vs `None`) and `allowed_paths_error` (absent vs set) differ. Cost: 14 duplicated lines inside a 60-line class, and a new `VerifyConfig` field means two edits in one class. Use: give `_verify` `allowed_paths: list[str] | None = None` and `error: str | None = "pre-run PRD not found: nothing to trust"`, making the third test one line. If the literal should go entirely, `_verify_config(**overrides)` at `tests/test_progress_scope.py:675` is the existing builder, and this module already imports from that module at line 65 (it would need `require_self_critique=False`).
>
> **4. `tests/test_scope_hardening.py:563-581` - the same PRD, hand-written one line below two new helper calls.**
> `test_legacy_prd_without_allowed_paths_stays_unconstrained` mkdirs `scripts/kstrl/feature/comp-a` and `json.dumps`es a full story literal for the path the change just gave a constant (`COMP_PRD_REL`, line 457) and a writer (used at lines 485 and 547). Cost: the module now writes one file two ways, twenty lines apart. Use: `write_component_prd(tmp_path, COMP_PRD_REL, stories=[PASSING_STORY])` (its assertions only read `allowed_paths`/`allowed_paths_error`, so the `["AC"]`/`["AC1"]` difference is immaterial).
>
> **5. The project scaffold got a fifth copy instead of a home.** `tests/test_factory.py:68-75`, `tests/test_notify.py:225-236`, `tests/test_autonomy_ladder.py:489-494`, `tests/test_inbox.py:426-429` each grew the same "mkdir scripts/kstrl + prompt.md + component PRD" block, with test_factory and test_notify now byte-identical in the added loop and its comment. Cost: this is the second plan-time preflight in two rounds, and the next one has to be found and fixed in four fixtures. Use: export a `write_project(root, comp_ids=())` alongside `write_component_prd` - the helper module was already open - or `init_kstrl_repo` at `tests/spine_utils.py:60` where a real git repo is wanted (test_scheduler and test_autonomy_ladder both git-init, and spine_utils is already imported by mocked modules including test_notify). Same finding, smaller: `test_factory.py:71` and `test_notify.py:229` hand-write `'{"branchName": "test", "userStories": []}'` on the line directly above the helper call that produces that exact body - `write_component_prd(tmp_path, "scripts/kstrl/prd.json")`.

### Simplification

> Review complete. I read the working tree live and it changed mid-review (the `test_scheduler` and `test_scope_hardening` hunks were rewritten while I was reading, and the helper gained `PASSING_STORY` and a `body=` parameter). Everything below is against the **current** working tree.
>
> ## Findings
>
> **1. `tests/helpers/component_prd.py:45,69` - `body=` turns the one helper into two functions in one.**
> When `body` is set, the entire reason the helper exists is skipped: `branch`, `stories` and `allowed_paths` are silently ignored and the function writes raw text. Cost: a caller that passes `body=` *and* `stories=` gets no error and no stories; and the function named `write_component_prd` is the thing that writes `"{not valid json"`. Both `body=` callers are in one file, at one path constant (`test_scope_hardening.py:485,547`). Simpler form: the local `_write_comp_prd(root, text)` that the first draft of this diff had, kept private to `test_scope_hardening.py` - two lines (`mkdir(parents=True, exist_ok=True)`, `write_text`). The helper then stays a PRD writer with a single mode.
>
> **2. `tests/helpers/component_prd.py:43` - the `allowed_paths` parameter has zero callers.**
> Grep across all five importing modules: nobody passes it. It also costs a docstring paragraph explaining why nobody should. Cost: a new file ships with a dead knob plus its justification. Simpler form: delete the parameter (and the paragraph); add it back with the first fixture that needs a scoped PRD. Same for the `-> Path` return, which no call site binds.
>
> **3. `tests/test_factory.py:60` and `tests/test_scheduler.py:50` - `comp_ids` parameters that no caller ever sets.**
> `_setup_project(tmp_path, comp_ids=("a","b"))` has 16 call sites, all `_setup_project(tmp_path)`. `_init_repo(root, comp_ids=("comp-a","comp-b"))` has exactly one call site, `_init_repo(tmp_path)`. Cost: a signature that advertises configurability the module does not use, and it makes the same diff inconsistent with itself - `test_notify.py:234` does the identical job with an inline `for comp_id in ("a", "b")` and no parameter. Simpler form: inline the tuple in the body, as `test_notify` does. (Related nit: `branch=f"b/{comp_id}"` at `test_factory.py:73` and `test_notify.py:235` - nothing in either module reads `branchName` for those components, so the default `"test"` would do, and matching the manifest branch implies a coupling that is not there.)
>
> **4. `tests/test_scope_hardening.py:533` - near-duplicate of `tests/test_scope_snapshot.py:304`.**
> `test_the_run_wide_flag_does_not_paper_over_it` and `test_an_unreadable_prd_beats_the_run_wide_flag` share the setup (corrupt PRD + `allowed_paths` flag) and three of the four assertions (`source == "unresolved"`, `allowed_paths is None`, `error is not None`), with the rationale paragraph copy-pasted into both docstrings. Only line 554 (the `_preflight_component_scope` assertion) is unique. Cost: two tests fail for one defect, and a future ordering change needs the same edit in two files. Simpler form: keep the resolve-level assertions in the snapshot test (it owns `ComponentScope.resolve`) and reduce the hardening test to its unique claim - the preflight refuses it - pointing at the snapshot test for why.
>
> **5. `tests/test_scope_snapshot.py:356-404` - `_verify` is one variation-point short, so the third test re-inlines everything.**
> `_verify` fixes `allowed_paths=None` and a hard-coded `allowed_paths_error`, so `test_the_toggle_still_removes_the_comparison` (line 380) cannot use it and repeats the full `run_mechanical_verification` call including a second copy of the 6-field `VerifyConfig`. Cost: the config literal now exists twice, one field apart. Also `test_the_refusal_survives_the_toggle` (373) and `test_it_still_fails_closed_with_the_toggle_on` (400) are the same two assertions with the boolean flipped, written two different ways. Simpler form: widen `_verify` to `(*, check_diff_scope, allowed_paths=None, allowed_paths_error=_DEFAULT_ERR)` so all three tests call it, and collapse 373/400 into one `@pytest.mark.parametrize("check_diff_scope", [False, True])`.
>
> **6. `tests/test_scope_hardening.py:550,554` - `RunScope.resolve(manifest, tmp_path, base)` computed twice in adjacent lines.**
> The second call re-derives a value line 550 already has. Cost: the test no longer proves the preflight and the assertion look at the *same* snapshot, which is the property #269 is about. Simpler form: `run_scope = RunScope.resolve(...)` once; assert on `run_scope.for_component("comp-a")` and pass `run_scope` to the preflight.
>
> **7. `kstrl/scope.py:68-70` - the `ScopeSource` doc still describes the pre-change policy.**
> `run_flag` is documented as "used when the PRD carries no scope of its own (legacy PRDs predate the field) **or could not be read at all**". The clause after "or" is exactly what fix (2) removes. Cost: the type alias, which is where a reader looks first, contradicts the method 80 lines below and the two new tests. Simpler form: drop the trailing clause; the `unresolved` bullet under it already covers that case.
>
> **Minor, take or leave:**
>
> - `kstrl/scope.py:105-118` - `origin` now carries two different meanings ("what supplied the list" for `component_prd`/`run_flag`, "where it should have come from" for `unresolved`), which costs the nine-line "one exception" paragraph. The `unresolved` value is derivable: it is `comp.prd_path`, its only consumer is the journal event at `kstrl/factory.py:1637` (which already carries `component`), and `_preflight_component_scope` reads `comp.prd_path` from the manifest rather than the field. Leaving it `""` like `unconstrained` would make the field mean one thing.
> - `kstrl/factory.py:1536` and `1568` - "$14.49 and 41 minutes" is now stated twice inside the same docstring, ~30 lines apart. The second bullet can cite the figure by pointing at the paragraph above it.

### Efficiency

> ## Findings
>
> **1. `RunScope.resolve` is called twice on identical inputs in one test**
> - File: `tests/test_scope_hardening.py`
> - Lines: 550 and 554 (`test_the_run_wide_flag_does_not_paper_over_it`)
> - Cost: `RunScope.resolve` opens, `json.load`s, `validate_schema`s and object-builds every component PRD in the manifest, plus `component_harness_files` per component (`kstrl/scope.py:227-245`, `kstrl/prd.py:188-218`). Doing it twice repeats all of that. Production calls it exactly once, at `kstrl/factory.py:2971` - this is the only site in `kstrl/` or `tests/` that resolves the same manifest twice in one scope.
> - Beyond the wasted I/O, the second resolve means the assertion feeds `_preflight_component_scope` a *different snapshot object* than the one it just inspected - which is the opposite of what #269 is about (one snapshot, read by both guards).
> - Cheaper: bind it once.
>
> **2. Two new `comp_ids` parameters that no caller ever passes**
> - Files/lines: `tests/test_factory.py:60` (and the loop at 72-73); `tests/test_scheduler.py:50`
> - Cost: all 16 `_setup_project(...)` calls in `test_factory.py` and the single `_init_repo(...)` call in `test_scheduler.py` use the default tuple. The parameter is a knob with no user, so the `for comp_id in comp_ids:` loop is a loop over a module constant, and every one of those 16 calls writes both PRDs whether or not that test's manifest names both. It also reads as configurable, which invites a future caller to pass an id set that silently disagrees with `_make_manifest`'s.
> - Cheaper: drop the parameter and put the tuple in the body (a module-level `_COMP_IDS = ("a", "b")` next to `_make_manifest`, since both have to agree), adding the parameter back when a test actually needs a different set.
>
> **3. `_verify` is bypassed by a hand-copied duplicate one test below it**
> - File: `tests/test_scope_snapshot.py`, helper at 356-371, duplicate at 384-397
> - Cost: `test_the_toggle_still_removes_the_comparison` restates the whole `run_mechanical_verification` + `VerifyConfig(...)` block that `_verify` already builds, differing only in `allowed_paths` (`list(AUTHORED)` vs `None`) and the absent `allowed_paths_error`. This is maintenance cost rather than runtime cost, but it is exactly the block that carries the three `test/typecheck/lint = "true"` subprocess spawns each of these tests pays, so any later attempt to make that cheaper has to be made in two places.
> - Cheaper: give `_verify` the two varying values as keyword defaults - `def _verify(self, root, *, check_diff_scope, allowed_paths=None, error="pre-run PRD not found: nothing to trust")` - and call it from all three tests.
>
> ## Checked and clean (negative results, verified rather than assumed)
>
> - **Production resolves once.** `RunScope.resolve` has exactly one non-test call site (`kstrl/factory.py:2971`), and `_preflight_component_scope` reads it via `run_scope.for_component`, a dict lookup (`kstrl/scope.py:248`) - no second PRD read. `grep -rn "RunScope.resolve" kstrl tests` confirms.
> - **The `verify.py` change adds no git call.** `check_diff_scope` returns at `kstrl/verify.py:1153`, before `git.get_diff_names` at 1181, whenever `allowed_paths_error` is truthy. The comment's claim holds. In the normal case the new `or allowed_paths_error` short-circuits on `config.check_diff_scope`.
> - **`_record_run_scope` is a net removal.** The diff deletes a per-component `ui.warn` loop and adds nothing; the remaining work is one emit loop plus a `Counter` pass.
> - **No new long-lived closures.** `spy_component`/`spy_rmv` in `_refuse` close over a local `list` and die with the `patch` context; nothing in the diff builds a closure that outlives its call.
> - **The three new `TestAnUnresolvedScopeCannotBeSwitchedOff` tests are not secretly expensive.** `dead_code_cleanup`, `mutation_testing` and `require_self_critique` all default `False` (`kstrl/verify.py:275-285`), and `test_command="true"` is the established house pattern (20+ existing sites), so each test costs one PRD read and three `true` spawns.

### Altitude

> **1. `kstrl/factory.py:1872` - the third consumer of the snapshot still collapses `unresolved` into `unconstrained`.**
>
> `_worker_scope` does `list(scope.allowed_paths or ())`, so an `unresolved` snapshot hands the worker an empty authored list, `guards.enforce_allowed_paths` returns early on `if not config.allowed_paths`, and the engineer runs with no in-loop guard and nothing recorded. The `_worker_scope` docstring accounts for `scope is None` ("the documented no-guard behaviour") and says nothing about `error is not None` - the case round two exists to separate.
>
> Cost: the whole point of fix 2 is that "could not read" must not be silently downgraded to "no constraint", and the one place it still is, is the guard that fires *first*. The verify.py comment even leans on this ("with a None authored list the in-loop guard is inert too") as an argument for changing verify rather than fixing it. Today only `_preflight_component_scope` keeps that state away from the worker, and that preflight is per-run and PENDING-only - one new call path into `_run_component` that does not route through `_run_preflights` and there is no layer behind it.
>
> Deeper fix: the policy is currently re-derived from raw fields at three sites in three different shapes - `scope.error is not None` (factory:1586), `config.check_diff_scope or allowed_paths_error` (verify:1985), and `allowed_paths or ()` (factory:1872). Put the predicate on the class that owns the distinction (`ComponentScope.is_enforceable` / `fail_closed`) and make all three ask it, with `_worker_scope` refusing an unresolved snapshot rather than returning empty lists for it.
>
> **2. `kstrl/verify.py:1985` - `or allowed_paths_error` overloads one gate and one check name with two different jobs.**
>
> The appended check never reads a diff in this branch; it reports a plan-time snapshot state. The comment quotes `check_prd_stories`'s own principle - "defence in depth an unrelated toggle can disable is not defence in depth" - but that principle argues for a check that is *never* gated, not for a disjunction bolted onto the gate. As written, `[verify] check_diff_scope = false` still produces a check named `diff_scope`, and the name is the API: `pipeline.py:2081` passes `check="diff_scope"` as the retry reason, `evolution.py:312` buckets `"diff_scope"` into the verification phase, and `as_context()` hands the retry agent a `diff_scope: FAIL` token that means "narrow your diff" for a condition the agent cannot affect.
>
> Cost: an operator who turned the check off sees it run and fail; a reader of the results cannot tell "you wrote outside scope" (agent-actionable) from "the harness could not establish a scope" (operator-actionable) without also knowing the toggle's value.
>
> Deeper fix: make the fail-closed signal its own ungated check - append `check_scope_source(allowed_paths_error)` unconditionally (returning nothing/passing when there is no error) and leave `if config.check_diff_scope:` alone. One toggle keeps one meaning, one check name keeps one meaning, and the retry/evolution consumers gain a signature they can route differently.
>
> **3. `kstrl/scope.py:204` (with the field doc at :107-116) - fix 5 establishes a rule for `origin` and fix 1 immediately exempts itself from it.**
>
> Fix 5 makes `unconstrained` name no origin on the principle that `origin` is "where the list came from". Line 204 then records `comp.prd_path` as the origin of an `unresolved` snapshot, which supplied nothing - the docstring calls it "the one exception". Two further things undercut it: `RunScope.for_component`'s stand-in (scope.py:272) builds an `unresolved` with `origin=""`, so the exception is not even uniform; and `events.py:439` still documents the field as "names the file or flag it came from", now false for the one source that most needs an accurate audit record.
>
> Cost: a consumer of `ComponentScopeResolved.origin` must read `scope_source` and `error` before it can know what the string means, which is the property fix 5 was removing.
>
> Deeper fix: the justification given is that `origin` is "the only place the audit record carries that path repository-relative" - but `resolve` has `comp.prd_path` in hand when it builds the error string. Put the repo-relative path in the `error` message (where the statement is about the failure) and let `unresolved` carry `origin=""` like `unconstrained`. One field, one meaning, and `events.py`'s docstring becomes true again.
>
> **4. `kstrl/prd.py:209-211` - the old "malformed scope = no scope" policy survives here as dead code.**
>
> Answering the question directly: no *live* path in `kstrl/` still conflates the two after this change, but this one still encodes it. `PRD.load` runs `validate_schema` first, which rejects a non-list `allowedPaths` with "allowedPaths must be an array" and raises `ValueError` - so the `if not isinstance(allowed_paths, list): allowed_paths = None` coercion is unreachable.
>
> Cost: a reader looking for the policy finds two contradicting statements in two modules, and if `validate_schema` is ever relaxed for that key the coercion silently restores exactly the behaviour #293 round two removed - a malformed authored scope falling through to the run-wide flag with `error=None`.
>
> Deeper fix: delete the coercion; `load` already raises, and `scope.resolve` already catches `ValueError`.
>
> **5. `tests/helpers/component_prd.py:45,69-71` - the `body` escape hatch makes one function two functions.**
>
> The layer itself is right: `tests/helpers/` already holds `fake_run.py`, `agent_probe.py`, `tool_output.py` as plain importable functions, and the helper removed real duplication across six test modules. The shape is what is off - when `body` is passed, `branch`, `allowed_paths` and `stories` are silently ignored, and the function returns before any of them is read.
>
> Cost: `write_component_prd(tmp, rel, branch="x", body="...")` compiles, type-checks and drops `branch` without a word - a mode flag that disables three of the function's four parameters.
>
> Deeper fix: the docstring's own reason is "one writer deciding where a component's PRD lives", which argues for sharing the *path*, not the function. Split into `write_component_prd(...)` and `write_raw_component_prd(root, rel, text)` over a two-line `_component_prd_path(root, rel)` that does the `mkdir`.
>
> **Minor, not worth a slot:** `tests/test_scope_hardening.py:544-547` builds `RunScope.resolve(manifest, tmp_path, base)` twice in one assertion block; the first one is already bound to `scope`'s source.
>
> **On the remaining question - is the refusal in the right place?** Yes. Preflight is the correct altitude for it: it is the only layer that runs before spend, the verdict is genuinely unwinnable (the snapshot is frozen for the run), and `_run_preflights` already documents scope as the cheapest-first refusal. Fix 4's removal of the warn loop is also right for the reason given - `_record_run_scope` sees every component, `_preflight_component_scope` sees only the schedulable ones, and I confirmed the stale-status reset at `factory.py:3064` runs before the preflight at `factory.py:3207`, so nothing that will actually run loses its signal.

### Applied

- **Altitude 1**: `ComponentScope.is_trustworthy` is the one predicate. The preflight asks it, and `_worker_scope` asks it instead of letting `allowed_paths or ()` quietly read an unresolved snapshot as "no constraint". The tripwire still fails open there, but now by name and with the two fail-closed layers stated.
- **Altitude 3 + simplification minor**: `origin` means one thing. `unresolved` carries `""` like `unconstrained`, and the repository-relative path moved into the `error` text.
- **Altitude 4**: the unreachable `allowedPaths` coercion in `PRD.load` is gone, with a comment saying why it must not come back.
- **Simplification 7**: the `ScopeSource` alias doc no longer says `run_flag` covers an unreadable PRD.
- **Simplification 1 + 5 + altitude 5, and reuse 1 + 2 + 3 + 4**: one PRD writer. `write_component_prd` gained `fixtures`, `test_harness_path_scope._write_prd` delegates to it, `test_scheduler`'s worktree copy goes through it (and its `import json` is gone), and `test_scope_hardening`'s legacy test uses it. The `body=` escape hatch stays, documented as sharing the PATH decision for the two tests that need an unloadable PRD there; splitting it into a second public function to avoid one ignored-parameter combination was the more surface, not less.
- **Simplification 3 + efficiency 2**: the `comp_ids` knobs are gone; the ids are a module constant next to the manifest builder that has to agree with them.
- **Simplification 4**: the hardening test is reduced to its unique claim (the preflight refuses), with the resolve-level assertions left to the snapshot test that owns `ComponentScope.resolve`.
- **Simplification 5 + 6, efficiency 1 + 3, reuse 3**: `_verify` takes the two varying values and all three tests call it; the two boolean-flipped tests are one parametrized test; the hardening test binds `RunScope.resolve` once and feeds the preflight the same object it inspected.
- **Reuse 5, smaller half**: the standalone `prd.json` literal in `test_factory` and `test_notify` goes through the writer.
- **Simplification minor**: the measured figure is stated once per docstring.

### Declined

- **Altitude 2** (make the fail-closed signal its own ungated `check_scope_source`): declined here, worth its own issue. The argument is good and the check-name overload is real, but `diff_scope` has carried `allowed_paths_error` since R1.5, and the name is an API three consumers key on (`pipeline._route_failure`, `evolution._CATEGORY_BY_CHECK`, `VerificationResult.as_context`). Adding a check name means enrolling it in all three plus the runbook, which is a bigger surface change than the hole it closes - and after finding 1 the state is unreachable from `run_factory` at all, because the preflight refuses the run first. What I did take from it: the disjunction is now a named predicate, `verify._diff_scope_runs`, so the policy has a docstring rather than a comment on an `if`.
- **Simplification 2** (`allowed_paths` on the helper has no callers): declined, because reuse 1's fix gave it one. `test_harness_path_scope._write_prd` now delegates and passes it on every call.
- **Reuse 5, larger half** (a shared `write_project` scaffold, or `spine_utils.init_kstrl_repo`): declined as unrelated churn. The four fixtures differ in what else they need (git init, `kstrl.toml` contents, autonomy state), `init_kstrl_repo` writes a `.gitignore` containing `.kstrl/` which some of these fixtures must not have, and consolidating them is a refactor of four modules this PR does not otherwise touch.

Issue #320, opened 2026-09-01
Encoding sweep: ~15 fail-closed readers still let UnicodeDecodeError escape past except OSError


0xfauzi:
Follow-up to #318 / PR #319. That PR fixed the seam every non-exempt command sits behind, and deliberately did not widen its scope. This issue holds the rest.

## The shape

CLAUDE.md states the rule from #291: a reader of any file kstrl writes must name `encoding="utf-8"` AND must catch `ValueError` alongside `OSError`, because `UnicodeDecodeError` is a `ValueError` and escapes a fail-closed `except OSError`. `init_cmd._read_text_or_none` is the worked example.

An audit during PR #319 found roughly 15 sites in `kstrl/` that name the encoding but still guard only `OSError`. Five were verified directly by reading them:

| Site | Why it matters |
|---|---|
| `serve.py:537` | The daemon spend ledger. Fail-closed money path: it refuses to spend against an unknown total on `OSError`, but a non-utf-8 byte raises past that handler entirely. |
| `workqueue.py:1290` | The pause marker. The comment directly above it argues at length that a broad `except OSError` made a PermissionError read as RUNNING, "exactly the fail-open this module claims not to do (#185 F7)". The encoding case still escapes the handler that fix installed. |
| `verify.py:524` | Phase 1 self-critique gate. |
| `knowledge.py:396` | Its docstring promises the reader never crashes. |
| `fixtures.py:490` | Fails both halves of the convention at once: `read_text()` with no encoding named, and no `ValueError`. |

The remaining sites from that audit are not individually verified in this issue and should be re-derived rather than trusted.

## Not an AST guard yet

PR #319 deliberately did not add one. Both existing precedents (`tests/test_atomicio.py`, `tests/test_process_scoping.py`) landed at offender count zero and say so in their own docstrings. Added with ~15 live offenders it would either block on a large refactor or ship with a suppression list, which `test_process_scoping.py` already refused once. The guard belongs in this sweep, at the end, once the count is zero.

## Suggested shape

One PR that fixes the sites and lands the AST walk in the same change, so the count is zero when the guard appears. Each fixed site keeps its own remedy text: a decode failure and an IO failure are different problems for the operator.

0xfauzi:
Two more sites of this shape, found during the PR #322 review. Neither is among the five verified in the issue body, and neither was changed by #322.

[...]

Issue #260, opened 2026-08-28
The architect's blocker gate has no convergence criterion: resolving all 7 blockers produced 11 new ones, and a real app spec cannot get past it


0xfauzi:
Found while dogfooding kstrl as a first-time user, trying to build a real web app from a spec (`writers-room`: a local collaborative writing canvas driven by the `claude` and `codex` CLIs). Two paid architect runs, zero components produced, zero lines of the app written.

This is not a bug report. Every finding the architect made was correct. The problem is the shape of the loop it puts the operator in.

## Measured

| | spec v1 | spec v2 |
|---|---|---|
| Spec length | 957 words | 2241 words |
| Architect duration | 119s | 210s |
| Total issues | 21 | 32 |
| **Blockers** | **7** | **11** |
| Major | 11 | 14 |
| Minor | 3 | 7 |
| Components produced | 0 | 0 |
| Exit code | 2 (halt) | 2 (halt) |

Between the two runs I resolved all seven v1 blockers properly. Not by deleting requirements: I wrote the concrete on-disk directory tree and file formats, pinned the exact argv for both agent CLIs against their real `--help` output, scaffolded the entire missing frontend toolchain (Vite, Svelte, TypeScript, `npm run check`, `npm run test`, all passing), defined block-id anchoring with a stale-proposal rule, made the server the single writer with atomic writes and external-change detection, replaced "the app measures voice" with a cross-family judge plus deterministic metrics, and defined argument-drift detection as a before-and-after claim-list comparison.

Every one of those was a real improvement. `ks sense` passes on the resulting repo. And the blocker count went **up 57%**.

## Why it happens, and why it is structural

Zero of the seven v1 blockers survived. My resolutions worked. But nine of the eleven v2 blockers are on surface that v1 did not have, because specifying something creates edges that were not there when it was vague:

- I specified five stages, so the architect found that the three named dispatch kinds cannot cover the six the stages need.
- I specified a proposal as a block-text replacement, so it found that the structure stage needs reorder, split, merge and delete, which that shape cannot express.
- I specified that block ids are never reused, so it found the allocation high-water mark is stored nowhere.
- I specified a cross-family judge, so it found that nothing in the project format records which family wrote each block.
- I specified stage in front matter, and separately in `state.json`, so it found the duality.

Those are all genuinely good catches. That is the point. **The audit is unbounded by construction** (`kstrl/decompose.py:105-107`):

```
[...]

PR #322, opened 2026-09-01
feat(decompose): route the architect's non-blocker findings into the component PRDs (#260)

0xfauzi:
Step 2 of #260. The free, code-only half: no prompt text is touched, so no
calibration run and no version bump. Halting is untouched.

## The premise, verified

`SPEC_ISSUES_REL_PATH` in `kstrl/decompose.py` is written by
`persist_spec_issues` and read by nothing. Grepping the whole package for
`spec-issues` finds four non-comment hits and none of them opens the file:

- `decompose.py` defines the path (`SPEC_ISSUES_REL_PATH`), writes to it in
  `persist_spec_issues`, and names it in a journal event payload. Those are
  its only three uses.
- `events.py` mentions it in a docstring.
- `tui/screens/decompose.py` uses the string as a display fallback when the
  `ArtifactWritten` event carries no path.
- `cli.py` prints the path in the `SpecBlockerError` message.

So the architect's audit was written to disk and abandoned there. Across the
five recorded writers-room runs that is 91 non-blocker findings, printed once
and discarded.

## What changed

Each non-blocker finding is routed into the PRD of the component whose surface
it touches. The PRD is the mechanism that already reaches the engineer: step 1
of the shipped engineer prompt is "Read the PRD file for this run", so a new
optional key in `prd.json` needs no prompt edit and no new channel.

```json
"specIssues": [
  {
    "severity": "major",
    "kind": "undefined_failure_mode",
    "summary": "Serialize performs no validation of block text ...",
    "location": "Serializing: \"parse(serialize(d)) == d for every valid d\" ...",
    "suggestion": "Either define valid Document as a construction-time invariant ...",
    "appliesTo": "component"
  }
]
```

- `kstrl/decompose.py`: `route_spec_issues` plus its helpers, and
  `_routed_prd_issues`, which derives the routing from the decompose payload
  so the retry-loop validation stage and the write phase compute the same
  thing (R1.8's "what is validated is what is written" now covers this field
  too).
- `kstrl/prd.py`: the `specIssues` key, round-tripped by `save`, and
  deliberately NOT compared by `tamper_changes`. Review F2 below reversed the
  validation half of this line: the field is now checked as an array and no
  further, because strictness that protects no gate only manufactures false
  failures.
- One line of operator-visible output per component:
  `document-format: 5 stories, 15 spec findings (3 on its own surface)`.

Blockers are filtered out before routing. They halt before any PRD exists, so
routing them would be dead code, and what a blocker means inside a PRD belongs
with the change that moves the halt.

## The attachment rule, and what it actually scores

The signal available is the `location` field the architect already emits. It
is populated on all 117 real findings, but it is prose quoting the spec, never
a repo path. Matching it against `allowedPaths` is worthless: the one real
decomposed component on disk declares `src/writers_room/`, `tests/` and
`scripts/kstrl/`, which every sibling would declare too.

What carries signal is the component's own name. The rule:

> A finding attaches to a component when the finding's `location` plus
> `summary` contains at least two words of five or more characters that appear
> in that component's id or title and in no other component's.

Everything else is broadcast: a finding the rule cannot place goes into EVERY
component's PRD with `appliesTo: spec`. Nothing is dropped.

### Measured on the 117 real findings

31 of the 117 locations literally name the component the architect meant
("Component 2, the claude adapter: ..."). That is a labelled sample, and
Component 1 is confirmed as `document-format` by the writers-room
`manifest.json` on disk. Scored against those 31, with component titles of the
richness the architect really produces:

| setting | precision | recall | broadcast |
|---|---|---|---|
| 2 shared words, min length 5 (shipped) | 1.00 | 0.53 | 14/31 |
| 2 shared words, min length 4 | 0.90 | 0.56 | 12/31 |
| 1 shared word, min length 5 | 0.81 | 0.81 | 5/31 |
| 1 shared word, min length 4 | 0.74 | 0.81 | 5/31 |

Precision was chosen over recall because the two errors do not cost the same.
A miss still reaches every engineer through the broadcast bucket. A
misattachment hides the finding from the one component that needed it.

There is deliberately no stopword list. One was written and then deleted after
the simplification pass asked for its measurement: emptying it left every
score in that table and every distinctive-word set on the real data
byte-identical, and not one of its 31 entries even appeared in a component
name. The distinctiveness filter and the two-word threshold already do that
job.

### Coverage over all 117

Of the 91 non-blocker findings this PR routes, the rule attributes **32 to a
named component (35 percent)** and broadcasts the other **59 to every PRD**.
Zero are dropped. Per severity: 23 of 56 majors and 9 of 35 minors attach.

The 26 blockers are unchanged: they still halt, and no PRD is written.

Two honest caveats:

1. **There is no ground truth for runs 0 and 1.** All five recorded runs
   halted, so no component set was ever produced for `spec.md`. Those 53
   findings are scored against a proxy component set taken from that spec's
   own section headings. Only the slice-1 set has a real component in it.
2. **The rule degrades to broadcast when component titles are terse.** With
   titles cut to `the document format` / `the agent adapter`, recall falls to
   0.03 and 30 of 31 findings broadcast. Precision stays at 1.00. That is the
   safe direction and it is still strictly better than today, where the
   findings reach nobody at all, but the routing is only as good as the names
   the architect chooses.

Size cost, measured on run 4: 15 to 16 entries per PRD, about 9.2 KB. That
also lands in the knowledge distiller's prompt, which pastes the PRD whole.

## Tests

New: `tests/test_spec_issue_routing.py`, 29 tests. The issue bodies are
verbatim architect output from the recorded runs (one is marked ADAPTED in the
file, with the reason).

End to end, through `decompose_spec` to the PRD on disk:

- `test_a_finding_that_names_a_component_lands_only_in_that_prd`
- `test_each_component_gets_its_own_finding`
- `test_a_finding_the_rule_cannot_place_reaches_every_prd`
- `test_a_finding_that_names_two_components_lands_in_both`
- `test_a_components_own_findings_are_listed_before_the_shared_ones`
- `test_the_written_prd_still_loads_through_the_schema`
- `test_a_clean_audit_leaves_the_prd_as_it_was`
- `test_the_findings_survive_a_prd_save`
- `test_the_operator_is_told_what_was_routed_where`
- `test_a_clean_audit_says_nothing_extra`

Halting unchanged:

- `test_a_blocker_still_halts_and_writes_no_prd`
- `test_a_blocker_is_not_routed_alongside_the_others`

The rule, one knob at a time:

- `test_one_shared_name_word_is_not_enough`
- `test_two_shared_name_words_attach`
- `test_a_word_both_components_use_attaches_to_neither`
- `test_short_words_carry_no_signal`
- `test_the_summary_counts_as_well_as_the_location`
- `test_no_findings_still_names_every_component`

Schema and policy: nine `TestPrdSchema` cases,
`test_editing_the_findings_is_not_tampering`, and
`test_a_field_added_to_prd_is_decided_rather_than_forgotten`, which fails if a
`PRD` dataclass field appears that is neither compared by `tamper_changes` nor
listed as deliberately exempt with a reason.

## Verification

- `uv run pytest tests/ -q`: 4504 passed, 28 skipped.
- `uv run mypy kstrl/ --strict`: clean, 127 files.
- `uv run ruff check kstrl/ tests/`: clean.
- `uv run python scripts/gen_docs.py --check`: current.
- Coverage 89.86 percent locally, baseline 79.91 unchanged.
- `pre-commit run --files <the touched files>`: every hook passes.
- Complexity went DOWN on every function the diff touches:
  `_decompose_spec_impl` 82 to 77, `PRD.validate_schema` 47 to 34,
  `_validate_fixture_entry` 68 to 65.

---

# /simplify findings, verbatim

Four agents (reuse, simplification, efficiency, altitude) reviewed the diff
and the surrounding code. Their output is reproduced below unedited except for
one typographic substitution: em dashes replaced with a hyphen, per house
style. My disposition follows each finding.

## Reuse

> Findings, ranked by what they cost.
>
> ---
>
> **1. `tests/test_spec_issue_routing.py:169` - `_decompose` re-implements `_run_decompose`**
>
> - Summary: the new decompose driver is a line-for-line copy of the existing one in the module this file already imports from.
> - Cost: two copies of the `decompose_spec(...)` call site (spec file write, `scripts/kstrl` mkdir, `MockDecomposeAgent`, `PlainUI(no_color=True, file=StringIO())`, `root_dir=tmp_path`). Any signature change to `decompose_spec` now has to be found in two places, and the copy silently drops the captured UI buffer, so `_prd_audit_note`'s new operator line has no test asserting it is printed.
> - Call instead: `tests/test_decompose.py:757 _run_decompose(tmp_path, output)`. The file already does `from tests.test_decompose import MockDecomposeAgent` (line 35), and `test_speckit_intake.py`, `test_decompose_run.py`, `test_launch_session.py` all import helpers from that module, so the pattern is established. `_run_decompose` swallows `SpecBlockerError`, so `test_a_blocker_still_halts_and_writes_no_prd` (line 274) keeps its own inline call - which is exactly what `test_decompose.py:809 test_artifact_written_on_halt` already does for the same reason.
>
> **2. `tests/test_spec_issue_routing.py:371` - `TestPrdSchema.BASE` re-declares the minimal valid PRD payload**
>
> - Summary: a fourth hand-rolled `{branchName, userStories:[...]}` literal for feeding `PRD.validate_schema`.
> - Cost: the story shape (six keys) is duplicated; a new required PRD key or story field breaks this class in a way the existing suite will not warn about, since the shared builder is elsewhere.
> - Call instead: `tests/test_prd_allowed_paths.py:21 _make_prd_payload(**overrides)` - it exists for precisely this, takes `**overrides`, so `_make_prd_payload(specIssues=...)` replaces both `BASE` and `_errors`.
>
> **3. `tests/test_spec_issue_routing.py:428` - `test_the_other_optional_fields_still_validate` re-asserts three existing tests**
>
> - Summary: the three regression assertions on `allowedPaths` and `fixtures` are already covered, verbatim, elsewhere.
> - Cost: three duplicated message substrings (`"non-empty when present"`, `"all items must be non-empty strings"`) that must be edited in two files whenever `_OPTIONAL_ARRAYS`' hints are reworded. The refactor's real regression risk is covered by the originals running.
> - Covered by: `tests/test_prd_allowed_paths.py:56 test_empty_array_rejected`, `:69 test_empty_string_item_rejected`, `tests/test_fixtures.py:668 test_empty_fixtures_array_rejected`. Delete the new test, or keep it as a one-line comment pointing at those.
>
> **4. `kstrl/prd.py:72` - `_validate_spec_issue_entry` re-implements the missing/extra key-set block**
>
> - Summary: the same `actual - expected` / `expected - actual` diff with identically worded messages, now for the fourth time.
> - Cost: four copies of `f"{prefix}: missing keys: {', '.join(sorted(...))}"` / `unexpected keys:` at `prd.py:78-82` (new), `prd.py:113-119` (`_validate_fixture_entry`), `prd.py:421-423` (story loop), `manifest.py:441-444`. They already differ in behaviour - `_validate_fixture_entry` returns on `missing or extra`, the manifest one `continue`s on each separately - so a fifth writer has three precedents to pick from.
> - No existing helper to call; this one needs extraction. `_key_set_errors(prefix: str, actual: set[str], expected: set[str]) -> list[str]` in `prd.py` beside `_validate_string_list`, called from the new validator and from `_validate_fixture_entry` - the diff already touches both.
>
> **5. `tests/test_spec_issue_routing.py:446` - hand-serialised PRD instead of the documented single writer**
>
> - Summary: `TestFindingsAreNotPinned` writes `prd.json` with two inline `json.dumps` calls.
> - Cost: `tests/helpers/component_prd.py` states in its own docstring "The one writer, so the shape a fixture PRD needs stays in one place", and eight test modules honour it. A ninth inline writer erodes that.
> - Call instead: `tests/helpers/component_prd.write_component_prd`, adding a `spec_issues: list[dict[str, str]] | None = None` parameter alongside the existing `fixtures` / `allowed_paths` ones. Lower value than 1-4 since the helper needs the new parameter first.
>
> **6. `kstrl/decompose.py:812` - `_issue_dict` re-lists SpecIssue's fields (low confidence)**
>
> - Summary: the five field names are written out again next to the dataclass that declares them.
> - Cost: a field added to `SpecIssue` (line 68) is silently absent from `spec-issues.json`, the journal event, and the routed PRD block.
> - Could call: `dataclasses.asdict(issue)` - already the codebase's idiom at `kstrl/autonomy.py:428`. **But** the explicit list is arguably load-bearing here: `prd._SPEC_ISSUE_KEYS` is a closed set, so auto-propagating a sixth field would make every PRD fail `validate_schema`. Flagging so the decision is deliberate, not defaulting.
>
> ---
>
> Checked and found clean (no existing helper being bypassed): `_route_words` / `_distinctive_name_words` - no tokenizer, stopword set, or text-similarity helper exists anywhere in `kstrl/`. `_validate_optional_arrays` / `_entry_validator` - these remove duplication rather than add it. `_prd_schema_errors` / `_routed_prd_issues` / `_issue_dicts` - straight extractions of existing code. `_prd_path` in the test - the `scripts/kstrl/feature/<id>/prd.json` literal is inlined in roughly fifteen test modules already and `kstrl/` has no shared constant for it, so the new copy matches the house pattern.

**Disposition.** 1 APPLIED: the test module now imports `_run_decompose`, and
because it returns the UI buffer I added the two tests that assertion made
possible (`test_the_operator_is_told_what_was_routed_where`,
`test_a_clean_audit_says_nothing_extra`). The blocker test keeps its own
inline call, as the finding says. 2 APPLIED: `_make_prd_payload`. 3 APPLIED:
deleted, replaced by a comment naming the three covering tests. 4 APPLIED:
`_key_set_errors` extracted and `_validate_fixture_entry` now calls it too;
messages byte-identical. 5 APPLIED: `write_component_prd` gained
`spec_issues`. 6 DECLINED, for the reason the finding itself gives:
`_SPEC_ISSUE_KEYS` is a closed set, so `asdict` would make a sixth `SpecIssue`
field fail every PRD's validation instead of being ignored. The explicit list
is the thing that makes adding a field a decision.

## Simplification

> I read the staged diff, the live files (worktree matches the index), and measured the two claims I was unsure about rather than asserting them.
>
> ## Findings
>
> **1. `kstrl/prd.py:229` - `_entry_validator` is a function-returning-a-function used twice at module scope, and it forces a dummy parameter on the third validator.**
> - Cost: a `Callable[[str, Any], list[str]] -> Callable[[str, list[Any]], list[str]]` type to read, a `Callable` import, a closure to step through in a traceback, and `_validate_allowed_path_items(_key: str, ...)` at line 223 taking a `_key` it never uses purely to match the lifted signature.
> - Simpler form: make the third element of `_OPTIONAL_ARRAYS` a per-*entry* validator uniformly and put the one `enumerate` loop in `_validate_optional_arrays`:
>   ```python
>   _OPTIONAL_ARRAYS: tuple[tuple[str, str, Callable[[str, Any], list[str]]], ...] = (
>       ("allowedPaths", "...", _validate_allowed_path),
>       ("fixtures", "...", _validate_fixture_entry),
>       ("specIssues", "...", _validate_spec_issue_entry),
>   )
>   ...
>       else:
>           for i, entry in enumerate(value):
>               errors.extend(validate_entry(f"{key}[{i}]", entry))
>   ```
>   `_entry_validator` (11 lines) and `_validate_allowed_path_items` both disappear; `_validate_allowed_path` becomes a 3-line per-item check. Measured: `_validate_optional_arrays` goes 7 -> 10, still under the 15 cap. One cost: the allowedPaths message becomes `allowedPaths[0]: must be a non-empty string` instead of the aggregate. Only one assertion pins the old text (`tests/test_spec_issue_routing.py:436`); `tests/test_scope_hardening.py:224` is a different validator (`_validate_decompose_output`).
>
> **2. `kstrl/prd.py:72-88` - the missing/extra key block in `_validate_spec_issue_entry` is a verbatim copy of `_validate_fixture_entry:108-124`, down to the message strings.**
> - Cost: two places to keep phrased identically; the next strictly-keyed entry type makes it three.
> - Simpler form: one helper both call.
>   ```python
>   def _key_errors(prefix: str, entry: dict[str, Any], expected: set[str]) -> list[str]:
>       actual = set(entry)
>       return [
>           f"{prefix}: {label} keys: {', '.join(sorted(diff))}"
>           for label, diff in (("missing", expected - actual), ("unexpected", actual - expected))
>           if diff
>       ]
>   ```
>   Messages are byte-identical. Measured: `_validate_spec_issue_entry` 9 -> 7, and it drops `_validate_fixture_entry` from 68 toward 64 (complexipy `--staged` only fails a function pushed above 15 or *worsened*, so an improvement is fine).
>
> **3. `kstrl/decompose.py:865` - the 31-word `_ROUTE_STOPWORDS` list is the one knob in the routing rule with no measurement and no test behind it.**
> - The comment block scores `_ROUTE_MIN_WORD_LEN` (precision 1.00 at five characters) and `_ROUTE_MIN_SHARED` (1.00/0.53 at two), but nothing scores the stopword list. I measured it: emptying the frozenset leaves the distinctive-word sets for the diff's own `DOCUMENT_FORMAT` / `AGENT_ADAPTER` components byte-identical, so no test in the new file pins any of the 31 entries. Its only reachable effect is dropping a common English word that appears in exactly one component's id-or-title (a word two components share is already dropped by the distinctiveness filter one function down).
> - Cost: 35 lines carrying an unstated invariant (every entry must be at least `_ROUTE_MIN_WORD_LEN`, asserted in a comment, enforced by nothing) that has to be re-checked whenever the length rule moves.
> - Simpler form: delete the set and the `word not in _ROUTE_STOPWORDS` clause, re-running the same 31-finding scoring to show 1.00/0.53 still holds. If it does not hold, keep only the entries that moved it. Do not delete it on my word alone; the scoring is the mechanism.
>
> **4. `kstrl/decompose.py:916` - `_route_words` filters both the component names and the issue text, but on the issue side the filter is provably inert.**
> - `name_words` is itself produced by `_route_words`, so every member already passes the length and stopword tests. `name_words & words` therefore cannot differ between filtered and raw issue tokens. I confirmed it on all four real issues in the new test file: identical hit counts.
> - Cost: the docstring ("words long and specific enough to carry a signal") reads as if the rule bites on both sides, so a future reader tuning `_ROUTE_MIN_WORD_LEN` for issue text is tuning nothing.
> - Simpler form: leave `_route_words` as the plain tokenizer (`set(_ROUTE_WORD_RE.findall(text.lower()))`) and apply the length/stopword predicate inside `_distinctive_name_words`, where it is the only place it can act. Same line count, and it puts the two tuning constants next to the only thing they change.
>
> **5. `kstrl/decompose.py:968-988` - `route_spec_issues` builds two intermediate collections of `SpecIssue` and converts to PRD entries in a final comprehension with star-unpacking.**
> - Cost: two representations of the same data alive at once, and a return statement that has to be read in two passes.
> - Simpler form: append the PRD-shaped entry as you go and join at the end.
>   ```python
>   attached: dict[str, list[dict[str, str]]] = {comp_id: [] for comp_id in distinctive}
>   spec_entries: list[dict[str, str]] = []
>   for issue in issues:
>       words = _route_words(f"{issue.location} {issue.summary}")
>       entry = _issue_dict(issue)
>       hits = [c for c, nw in distinctive.items() if len(nw & words) >= _ROUTE_MIN_SHARED]
>       for comp_id in hits:
>           attached[comp_id].append({**entry, "appliesTo": SPEC_ISSUE_APPLIES_COMPONENT})
>       if not hits:
>           spec_entries.append({**entry, "appliesTo": SPEC_ISSUE_APPLIES_SPEC})
>   return {comp_id: own + spec_entries for comp_id, own in attached.items()}
>   ```
>   Same order, same shared-dict aliasing across components as today. Measured 11 -> 10.
>
> **6. `kstrl/decompose.py:1715` - `spec_issues: Sequence[dict[str, str]] = ()` on `_build_prd_data` is a default no caller uses.**
> - Both call sites (`_prd_schema_errors:1693`, `_generate_component_prd:1762`) pass it, and no test calls `_build_prd_data` directly. The default's only effect is to make "forgot to route the audit" compile silently, in the one function whose docstring promises validated == written.
> - Simpler form: make the parameter required on `_build_prd_data`. (`_generate_component_prd`'s default is genuinely used, by `tests/test_progress_scope.py:279,344`, so leave that one or update those two calls.)
>
> **7. Minor - `kstrl/decompose.py:1710`: `_prd_audit_note` returns a fragment beginning with `", "`.**
> - Cost: the separator lives in the callee and the sentence it joins lives in the caller (line 2244), so neither reads on its own.
> - Simpler form: return the note without the leading separator and let the call site do `f"... stories{', ' + note if note else ''}"`, or have the helper take the prefix and return the whole line. Do not inline it into `_decompose_spec_impl` - that function is at 77 and `--staged` fails anything this commit worsens.
>
> Nothing else. The `_OPTIONAL_ARRAYS` table itself and the `_prd_schema_errors` extraction are the good parts of the diff: they cut `PRD.validate_schema` from 47 to 34 and `_decompose_spec_impl` from 82 to 77 while removing three near-identical blocks.

**Disposition.** 1 DECLINED: the simpler form changes the `allowedPaths`
aggregate message to a per-item one, and that string is pinned by an existing
test (`tests/test_prd_allowed_paths.py::test_empty_string_item_rejected`),
not only by mine. Rewording an error message for an unrelated field is a
behaviour change this PR should not make. I did take the half of it that costs
nothing: `_validate_allowed_path_items` now uses its `key` argument
(`f"{key}: ..."`), so the row is genuinely field-agnostic. 2 APPLIED, same
extraction as reuse 4. 3 APPLIED, after doing the measurement it asked for:
emptying the set left all four rows of the scoring table and every
distinctive-word set byte-identical, and none of its 31 entries appears in any
component name in the evidence. Deleted, with the reasoning recorded in its
place. 4 APPLIED: `_words` is now the plain tokenizer and `_name_words` carries
the length rule, which is the only side it can act on. 5 APPLIED verbatim.
6 APPLIED: required on `_build_prd_data`; `_generate_component_prd` keeps its
default, which `tests/test_progress_scope.py` uses. 7 APPLIED as the second
option: `_prd_summary_line(comp_id, story_count, findings)` returns the whole
line, so the separator and the sentence live together.

## Efficiency

> Measured on this worktree, CPython 3.12.8, arm64 macOS, `min` of 7 timeit repeats.
>
> ## Findings
>
> **1. `kstrl/prd.py:87` - `sorted(_SPEC_ISSUE_KEYS)` recomputed per entry**
> - summary: the type-check generator in `_validate_spec_issue_entry` sorts a 6-element constant set on every entry it validates.
> - cost: 99.3 ns per sort; 21.6 us across the 180 entries one real decompose validates (20 findings x 9 components). That is 27% of total entry-validation time (101.3 us -> 79.7 us with it hoisted). Absolute cost is nothing: the same decompose spends 177.7 us writing one of its 2-9 PRD files.
> - cheaper: module-level `_SPEC_ISSUE_KEYS_SORTED = tuple(sorted(_SPEC_ISSUE_KEYS))`, then iterate that. One line, zero risk, keeps the deterministic message order. The `sorted(_SPEC_ISSUE_APPLIES)` at line 97 is already fine - error path only.
>
> That is the only wasted work in the diff. Everything else I checked is already in its cheap form:
>
> ## Checked and fine
>
> **2. `kstrl/decompose.py:1689` + `:2208` - the deliberate double `_routed_prd_issues(data)`. Keep it.**
> - measured at the stated real scale: 97.6 us per call at 20 findings x 9 components (80.6 us at 2 components, 88.7 us at 5). The full retry-loop wrapper `_prd_schema_errors` is 230.0 us, of which 124.6 us is `PRD.validate_schema`, not routing.
> - reference point in the same process: one atomic PRD write is 177.7 us. The second routing call costs **less than writing a single one of the 2-9 files it feeds**, and it happens at most 4 times per decompose (3 retry attempts plus write). Even at the whole 5-run corpus stuffed into one payload (117 findings, 9 components) routing is 495.7 us.
> - verdict: the R1.8 "validate byte-for-byte what is written" guarantee is bought for well under a millisecond per decompose, against a step whose wall time is an LLM call. No finding. Do not cache it.
>
> **3. `kstrl/prd.py:229,248` - the `_entry_validator` closures in `_OPTIONAL_ARRAYS` retain nothing.**
> - inspected at runtime: `allowedPaths` has 0 cells, `fixtures` and `specIssues` have exactly 1 cell each, each holding a module-level function that is alive regardless. No enclosing scope, no captured data, built once at import. No leak, no finding.
>
> **4. `kstrl/decompose.py:983-988` - `spec_entries` is already built once and spliced by reference.**
> - the spec-wide dicts are shared objects across every component's list rather than N-fold copies, and `_build_prd_data`'s `list(spec_issues)` is a shallow copy. This is the cheap form. (Aliasing note, not an efficiency one: the same dict object lands in every component's list; nothing mutates them today.)
>
> **5. `route_spec_issues` / `_distinctive_name_words` - the invariants are already hoisted.**
> - `_ROUTE_WORD_RE` is compiled at module scope; `distinctive` is computed once per call, not per issue; `_route_words(issue)` is called once per issue, not once per (issue, component). The inner loop is O(issues x components) 6-element set intersections: 180 of them at real scale. Nothing to hoist.
>
> **6. `PRD.load` on the loop hot path - cost is real but negligible; the file growth is intended, not waste.**
> - `PRD.load` calls `validate_schema` (prd.py:320) and runs several times per iteration (loop.py:964, verify.py:653 and :678, review.py:805, scope.py:202, pipeline.py:3106 and :3251).
> - measured: 21.3 us with no `specIssues`, 45.2 us with 20 (67.6 us with 40). File grows 3,965 B -> 15,107 B. At ~6 loads per iteration that is +144 us per iteration against iterations measured in minutes.
> - the ~11 KB the engineer now reads is the point of the change, and it does not amplify with component count since each engineer reads only its own PRD. I did not tokenize it, so I am not quoting a token number. The on-disk N-fold duplication of the spec-wide bucket (~31 KB across 9 PRDs vs ~3.5 KB for one copy) is trivial and is the "nothing is dropped" design choice, not waste.
>
> ## Bottom line
>
> One nit worth a one-line fix (`sorted` hoist in prd.py:87). The double `_routed_prd_issues` call the brief asked me to judge is measurably free - cheaper than a single PRD write - and should stay. No hot-path regression, no repeated I/O, no closure retention introduced by this diff.

**Disposition.** 1 APPLIED: `_SPEC_ISSUE_KEYS_SORTED` hoisted. 2 to 6 are
"checked and fine" and needed no action; the double-call measurement is the
reason it stays, and it is now stated in the code comment.

## Altitude

> ALTITUDE review of the #260 diff. Five findings, ordered by depth. The four questions I was asked are answered inside them.
>
> ---
>
> **1. `specIssues` is informational text placed in the one document that is a contract**
>
> - file: `kstrl/prd.py:307`, `kstrl/decompose.py:1723`
> - summary: Every other PRD key is either gated (`branchName`, `userStories`, `fixtures`) or plan-time input (`allowedPaths`); `specIssues` is neither, and putting it there forces a carve-out in each mechanism the PRD already owns.
> - cost: One informational field pays a five-way tax: a schema block (`prd.py:72,260`), a `tamper_changes` exemption (`prd.py:480`), a `save()` round-trip (`prd.py:551`), a runbook paragraph (`docs/runbook.md:11`), and a place in the retry-loop validation. The PRD is also inside the component's write scope, so the copy the engineer reads is one the engineer can rewrite, which is exactly why the exemption had to be written. Next informational field follows this precedent and pays the same tax again.
> - the deeper form: the general mechanism already exists and is not new plumbing. `factory.py:2088-2100` assembles `context_prefix` from `knowledge_prefix + feedforward_prefix + retry context` and hands it to `loop.run_iteration`. `build_knowledge_context(manifest, component, knowledge_root, config)` is a structural twin of what is needed: a durable disk artifact written by an earlier phase, sliced per component id, tiered into core and shared, rendered as prompt text, no prompt edit, no schema, no tamper question. `build_spec_issue_context(manifest, component, root_dir)` reading `spec-issues.json` drops into `_submit_args` (`factory.py:3439`) next to the knowledge call and deletes items 2 through 5 of that tax. Decompose-produced PRDs are consumed only by `ks factory`, so the "PRD reaches every entry point" argument buys nothing here.
>
> **2. The spec-wide bucket is copied into every PRD, and `appliesTo` exists to label the copy**
>
> - file: `kstrl/decompose.py:946` (`route_spec_issues`), `prd.py:58`
> - summary: Unplaceable findings are duplicated verbatim per component; the discriminator's job is to tell the duplicate from the original.
> - cost: On the PR's own numbers, 31 of 117 findings name a component and the rule places 53% of those, so roughly 86% of the audit lands in *every* component's PRD. The expensive, measured, tuned matcher is decorating what is in practice a broadcast, and the broadcast is done by copying an artifact that already exists once on disk. N copies that nothing reconciles, in a file the agent may edit.
> - the deeper form: carry the shared bucket once and the attached findings per component. Under the context-prefix form (finding 1) that is free: one "spec-wide" section in the prefix plus a "this component" section, tiered exactly like knowledge's core/dependency/sibling. Position then carries the distinction and `appliesTo` disappears rather than being validated. If the PRD form is kept, the shared bucket should be a pointer to `scripts/kstrl/spec-issues.json` rather than a copy of it.
> - second-order: `route_spec_issues` itself (stopword list, min word length, min shared, distinctiveness) is a self-contained scored heuristic with its own tuning constants, sitting in a 2415-line module that already owns parsing, surfacing, persisting, convergence comparison, PRD assembly and manifest writing. It is finding-to-component attribution, not decomposition. It belongs in its own leaf module and should move with whichever delivery mechanism wins.
>
> **3. `tamper_changes` now has three policies and records them only in prose**
>
> - file: `kstrl/prd.py:462-514`
> - summary: The decision not to compare is correct; the way it is recorded is the shallow part.
> - cost: `_pinned_stories` (`prd.py:444`) is explicitly fail-closed - a field added to `UserStory` is pinned by default, and its docstring says so as the design argument. `tamper_changes` one level up is the opposite: it enumerates what IS compared, so a new PRD field is silently unpinned. Two fields have now taken that exemption for two different reasons (`allowedPaths`: inert; `specIssues`: informational), and the reasoning lives in three duplicated prose locations (docstring, runbook, CHANGELOG). A third field has no place to be decided and will simply be forgotten.
> - the deeper form: one declarative table of PRD fields with `pinned: bool` plus the reason, driving both `tamper_changes` and a test that fails when a `PRD` dataclass field appears without an entry. Same fail-closed-by-default shape `_pinned_stories` already argues for, applied one level up. That generalizes the mechanism instead of adding the second special case.
>
> **4. Machine-generated entries are validated inside the LLM retry loop**
>
> - file: `kstrl/decompose.py:1678` (`_prd_schema_errors`), `1664` (`_routed_prd_issues`)
> - summary: `_validate_spec_issue_entry` checks data kstrl itself constructed, in a loop whose contract is "a retryable error the LLM gets to fix".
> - cost: `_parse_spec_issues` (`decompose.py:731`) already guarantees valid severity, valid kind, non-empty stripped summary and string location/suggestion; `route_spec_issues` sets `appliesTo` from two module constants. The routed block cannot fail validation unless kstrl has a bug, and the loop's response to a kstrl bug is to spend paid architect retries the model cannot possibly fix. The R1.8 invariant is being satisfied by re-deriving and re-validating rather than by construction, which is also why `_parse_spec_issues` now runs three times on one payload (1673, 2022, 2208).
> - the deeper form: make the routed block un-invalidatable instead of validated. `SpecIssue` is already a frozen dataclass; the only reason it becomes `dict[str, str]` at the boundary is that `prd.py` cannot import `decompose.py`. Move `SpecIssue` to a leaf module (`findings.py` imports stdlib only) and both sides share the type - the writer is typechecked, `_validate_spec_issue_entry` shrinks to the read side where an engineer-edited file actually needs it, and the retry loop stops carrying a field no retry can change.
>
> **5. The lifted validator table takes a key one of its three entries ignores**
>
> - file: `kstrl/prd.py:223`, `248`
> - summary: `_validate_allowed_path_items(_key, value)` discards the key and hardcodes `"allowedPaths:"` in its message.
> - cost: `_OPTIONAL_ARRAYS` reads as field-agnostic but one row is not; a fourth optional array copying that row inherits a wrong prefix silently.
> - the deeper form: `f"{key}: all items must be non-empty strings"`. One character of real generality where the table already claims it.
>
> ---
>
> **What is at the right depth:** the `_OPTIONAL_ARRAYS` table plus `_entry_validator` is a genuine generalization - it collapses two near-identical blocks and makes the third free rather than adding a third. Extracting `_issue_dict` out of `_issue_dicts` is the right seam. Deriving the routed issues from the payload (`_routed_prd_issues`) rather than threading a value between the loop and the write phase is the right call given the loop's retry semantics; the objection in finding 4 is to what is derived, not to deriving it.
>
> **Direct answers:** `specIssues` as a PRD key is one layer too high - `allowedPaths` and `fixtures` are there because gates read them, and this field has no gate (finding 1). `route_spec_issues` does not belong in `decompose.py` (finding 2). `appliesTo` is not the wrong shape but is a symptom - it labels a duplication that should not exist (finding 2). Not comparing in `tamper_changes` is the right decision recorded at the wrong altitude (finding 3).

**Disposition.**

1 DECLINED for this PR, and it is the most interesting finding here. The
context-prefix form is a real alternative and the tax it names is real. Two
reasons not to take it now. The brief for this change specifies the PRD and
specifies following the mechanism that already renders into it, so switching
delivery is a different PR. And the finding's own supporting claim is the part
I am least sure of: a decompose-produced PRD is also what `ks feature --prd`
and `ks run` are pointed at, and `factory._run_component` copies the PRD into
the worktree while `spec-issues.json` is never copied there, so the two
mechanisms do not reach the same places. Worth its own issue with that
measured rather than argued.

2 DECLINED on both halves, with different reasons. The pointer instead of the
copy does not work: `scripts/kstrl/spec-issues.json` is not copied into the
worktree, so a pointer from inside a component's worktree dangles, which is
precisely why the PRD is copied. Moving `route_spec_issues` to a leaf module
requires relocating `SpecIssue`, which `kstrl/linear.py` and six test modules
import from `decompose`; that rename would dominate a diff whose subject is one
routing rule. Recorded as the follow-up, together with finding 4's version of
the same move.

3 APPLIED in its mechanism half, not its table half. A table driving
`tamper_changes` cannot be uniform: `branchName` is an equality check,
`userStories` goes through `_pinned_stories`, `fixtures` is a list compare, so
a shared driver would be an abstraction over three different comparisons. What
the finding is actually protecting is that the omission is a decision, and a
test gets that without the indirection:
`test_a_field_added_to_prd_is_decided_rather_than_forgotten` fails when a `PRD`
field is neither compared nor listed as exempt, and the two exemption reasons
sit in the test beside the names.

4 DECLINED. Validating a payload the harness assembled, inside the loop, is the
pre-existing R1.8 contract for the whole PRD, not something this field
introduced: `_generate_component_prd` already says a failure there "indicates a
harness bug, but the guard preserves the write-only-validated invariant".
Narrowing that contract is a change to R1.8. The `SpecIssue` relocation half is
the same move as finding 2's, declined for the same reason and recorded as the
follow-up. The efficiency agent measured the re-derivation the finding objects
to at 97.6 us, against a 177.7 us PRD write.

5 APPLIED.

---

# Review round 1: what changed

Five findings, all addressed. `660bbc1`.

## F1 (High) - the PR was not the free change it was scoped as

Correct, and this was the biggest miss. `SECURITY_PROMPT` and `DISTILL_PROMPT`
both paste the component PRD verbatim and untruncated, so routing findings into
it changed what two enrolled adversarial roles READ. That is the quantity H2
governs, reached without editing a prompt constant.

Reproduced on the real document-format PRD carrying run 4's 15 routed findings:

| | PRD text | SECURITY | DISTILL |
|---|---|---|---|
| before | 8,184 | 16,425 | 11,874 |
| after (no strip) | 18,064 | 26,305 | 21,754 |
| delta | +120.7% | **+60.2%** | +83.2% |

The PRD block went from 50 to 69 percent of the whole security prompt.

**Decision: strip, not cap.** A cap still hands the reviewer a truncated block it
has no use for, and the growth is unbounded in the number of findings, so the
honest bound is zero. `prd.prd_text_for_prompt` removes the key on both paths.

Measured again after the fix, same PRD, same 15 findings:

| | PRD text | SECURITY | DISTILL |
|---|---|---|---|
| before | 8,184 | 16,425 | 11,874 |
| after | 8,184 | 16,425 | 11,874 |
| delta | **0.0%** | **0.0%** | **0.0%** |

Byte-identical. Not "small enough": zero. The PRD share of the security prompt
is back to 50 percent. Both prompt constants and both snapshot hashes are
untouched, so H3 is not engaged, and H2 has nothing to measure because neither
role's input moved by a byte.

**Not overclaiming.** This removes the contributor the engineer does not control.
It does not make the two prompts bounded. `UserStory.notes` is unbounded,
written by the agent under review, and flows into both prompts uncapped; the
altitude reviewer is right that it is the worse hazard. It is structural,
pre-existing, and capping it would itself change what every existing project's
reviewer reads, which is the H2-governed move this PR is scoped to avoid. Called
out below and left for its own change; the code comment now scopes its
"strip beats a cap" claim to this key and names `notes` as the exception.

**The #320 question.** `security.py`'s read is NOT one of the five sites #320
verified, but it is exactly its shape. The read sits in its own `try/except
OSError` with no outer handler, so a `UnicodeDecodeError` (a `ValueError`)
escapes `run_security_review` entirely rather than degrading to
"(PRD not available)". `knowledge._read_prd_text` has the same narrow catch.
This PR did not alter either, deliberately: widening them is #320's change, not
this one. Worth filing if it is not already covered.

## F2 (High) - documented as safe to edit while five plausible edits hard-failed

Also correct. Nine plausible engineer edits, run through the real
`check_prd_stories` path. Five hard-failed inside `PRD.load`, so the operator got
"Failed to load PRD" - after paying for the whole component - and
`_tamper_changes` was never reached:

1. resolving them all to `[]`
2. adding a `"resolved"` key
3. writing `appliesTo: "resolved"`
4. dropping `suggestion` as noise
5. collapsing the block to a list of strings

**Choice: lenient, not pinned.** Strictness that protects no gate only
manufactures false failures. `fixtures` earns its strictness twice over (Phase 1
executes it, `tamper_changes` pins it); `specIssues` is read by no gate at all.
It is now validated as an array and no further - the one check that survives,
because `PRD.save` writes the value straight back out and a scalar there is a
corrupt file rather than an annotation.

**All three doc sites corrected**, saying the same thing in the same words:

- `PRD.tamper_changes` docstring: says the field **fails OPEN**, that a deletion
  also **persists** because `factory._run_component` seeds the worktree copy only
  when it does not already exist, and that the audit itself is not at risk
  because spec-issues.json is written before any worktree exists.
- `docs/runbook.md`: not compared, not validated beyond being an array, a
  component may annotate/resolve/delete it and nothing reports it, including on
  a later iteration.
- `CHANGELOG.md`: the loosest thing in the PRD, and why - the opposite of
  `fixtures`, which is strict because it is both pinned and executed.

The five edits are now `TestPrdSchema::test_an_engineer_annotation_is_accepted`.

## F3 - three untested production decisions

Each had a mutant that survived the round-1 suite. One fixture each, in
`TestTheNamePartsThatCarrySignal`:

- `test_the_id_names_the_component_when_the_title_does_not` - both round-1
  fixtures had ids whose words their titles repeated, so dropping the id half of
  the name changed nothing. A terse title is the case that depends on it, and
  decompose only requires a string.
- `test_a_five_letter_name_word_counts` - the floor's low edge was pinned and its
  high edge was not, so raising `_ROUTE_MIN_WORD_LEN` 5 to 6 survived. Every
  qualifying word in the fixture is exactly five characters.
- `test_a_digit_is_part_of_a_name_word` - two components separated only by the
  digits in their names. Drop digits from the tokenizer and both collapse onto
  the same words, so nothing attaches.

## F4 - the blocker-halt assertion was arithmetic

Correct: the payload had no components, so "no prd.json exists" was true whether
the halt fired or not. The payload now carries one. The mutant still dies on
"DID NOT RAISE" rather than on the file assertion, so a positive control was
added:
`test_the_same_payload_without_the_blocker_does_write_a_prd`.

## F5 - prefix match

Replaced with an exact one.

## Mutation check

Method, as required: `PYTHONDONTWRITEBYTECODE=1` and every `__pycache__` cleared
before each run, because CPython invalidates bytecode on mtime-seconds plus size
and two same-size mutants written in the same second can otherwise falsely
survive. I did both.

```
F1 security: stop stripping                killed
F1 distiller: stop stripping               killed
F1 strip: keep only the excluded keys      killed
F1 strip: skip the whole filter            killed
F1 guard: an undecided key slips in        killed
F2: refuse an empty array again            killed
F2: validate the entries after all         killed
optional keys: drop the derivation         killed
F3: name from the title only               killed
F3: word length floor 5 -> 6               killed
F3: tokenizer drops digits                 killed
F4: shared-word floor 2 -> 1               killed
seed-once: reseed the PRD every iteration  killed

survivors: none
```

## Gates

Full suite 4512 passed / 28 skipped. `mypy --strict` clean, 127 files.
`ruff check` clean. Staged `pre-commit run` all hooks passed (complexipy,
codespell, vulture, gitleaks, the 800-line ratchet, em-dash check).
Coverage 89.87 percent against the 79.91 baseline, which is untouched.

---

---

# Round-2 `/simplify` findings, verbatim

Four agents (simplification, altitude, reuse, efficiency), run against the
round-2 diff and the surrounding code. Reproduced in full, unedited, with my
decision after each, **in a PR comment rather than here**: body plus both
rounds of verbatim findings is 72,481 characters and GitHub caps a PR body at
65,536. Nothing is trimmed or paraphrased, and the two defects the agents found
in my own round-2 edits are in there.


0xfauzi:
# Round-2 `/simplify` findings, verbatim

Four agents, run against the round-2 diff and the surrounding code. Their
reports are reproduced in full below, unedited, including the two defects they
found in my own round-2 edits. My decisions follow each.

## Agent 1: simplification

> Findings, in priority order.
>
> **1. `kstrl/prd.py:284-285` - garbled comment, two sentence-halves spliced.**
> ```
> # Every optional top-level PRD key follows the same rule: absent is
> # GATED optional key follows the same rule: absent is fine, present must
> ```
> The round-2 edit inserted the new opener without deleting the old one. Cost: the block's first sentence is unreadable, and this is the comment defending `_OPTIONAL_ARRAYS`. Simpler form: one line - `# Every GATED optional top-level PRD key follows the same rule: absent is`.
>
> **2. `kstrl/prd.py:48-94` - (a) and (b) together. The frozenset does not earn its place; the early return does, but not in that form.**
> `PROMPT_EXCLUDED_KEYS` is a public module-level name with one member, one use (line 91), and zero importers anywhere in `kstrl/`, `tests/` or docs. It is also a second spelling of the string that `_SPEC_ISSUES_KEY` (line 146) already names, in the same file - exactly the drift `_SPEC_ISSUES_KEY` was introduced to prevent. The `len(kept) == len(data)` guard is load-bearing (byte-for-byte pass-through, tested), but it asks the question backwards: build the dict, then infer from a length that nothing was removed. Simpler form, one function, drops the frozenset, drops a branch, and skips the comprehension in the common case (complexipy 5 to 4):
>
> ```python
>     if not isinstance(data, dict) or _SPEC_ISSUES_KEY not in data:
>         return text
>     kept = {k: v for k, v in data.items() if k != _SPEC_ISSUES_KEY}
>     return json.dumps(kept, indent=2) + "\n"
> ```
> Requires hoisting `_SPEC_ISSUES_KEY` above line 48. Docstring then says "the key", not "the keys".
>
> **3. `kstrl/prd.py:270-281, 302` - (c). Keep the table, kill `_entry_validator`.**
> Two rows still pay: the table keeps `validate_schema` flat and unifies the empty-array message, and its reuse of `_key_set_errors` is what took `validate_schema` from 34 to 28 on complexipy. The special-cased `specIssues` tail at 320-321 is the cheapest available place and is fine. But `_entry_validator` is now a closure factory with exactly one caller. Simpler form: delete it, write `_validate_fixture_entries(key, value)` as a plain function and put its name in the table. Net -6 lines, one less higher-order indirection.
>
> **4. `kstrl/prd.py:115-119` - dead-ish state left by the F2 deletion.**
> `_SPEC_ISSUE_APPLIES` was the only in-module reader of `SPEC_ISSUE_APPLIES_COMPONENT` / `_SPEC`. F2 deleted it, so `prd.py` now defines two constants it never uses; every production user is in `decompose.py` (writer at 960/962, reader at 1689). The comment's justification - "The PRD owns the vocabulary because the field lives here" - is now false. Simpler form: move both to `decompose.py`, or keep them and rewrite the comment to say what is actually true (the PRD is the shape's home, decompose is the only code that reads or writes the value).
>
> **5. Comment volume is disproportionate, and the same argument is restated five times.** (d)
> `kstrl/prd.py:48-69` is 22 comment lines for a one-element frozenset. `kstrl/prd.py:121-146` is 26 comment lines for a one-line string constant. Together ~48 comment lines carry 3 lines of new logic in the F2 half. Worse than volume is the copying:
> - "spec-issues.json, written before any worktree exists / no component can reach it": `prd.py:128`, `prd.py:531-533`, `docs/runbook.md`, `CHANGELOG.md`. Four copies.
> - "an engineer may annotate, resolve or delete it and nothing reports it": `prd.py:142`, `prd.py:524`, runbook, CHANGELOG. Four copies.
> - The measurement: "60.2 percent" (`prd.py:60`), "60 percent" (`security.py:667` comment), "60 percent" (test class docstring). Three copies, two roundings. Remeasure once and two of them are wrong.
>
> Simpler form: one canonical statement at `_SPEC_ISSUES_KEY`; `tamper_changes` and the F1 comment shrink to a pointer at it. The `security.py` and `knowledge.py` comments become one line each naming `prd.prd_text_for_prompt` rather than re-arguing the case at the call site. The `PROMPT_EXCLUDED_KEYS` block keeps only the measurement and "strip, not cap" - about 5 lines.
>
> **6. `tests/test_spec_issue_routing.py:457-467` - duplicate parametrize row, plus a row that contradicts the docstring.**
> `("resolve them all", [])` and `("empty the block", [])` are the same value with two labels. And `("rewrite a summary", [{**ENTRY, "summary": "handled, utf-8"}])` would have PASSED the deleted strict validator - all six keys, all strings, non-empty summary, valid `appliesTo` - so the test's claim "Every one of these hard-failed the run in review round 1" is false for it, and it discriminates nothing between old and new behaviour. The source comment at `prd.py:133` correctly says five. Simpler form: five rows, matching the five the comment names.
>
> **7. `tests/test_spec_issue_routing.py:445-446` - `_passes` throws away the diagnostics.**
> Returning a bool means a failure prints `assert False, "drop a key"` with no schema error. One-line wrapper, small cost, real loss. Simpler form: delete it, `assert self._errors(value) == [], edit`.
>
> Nothing else. `_key_set_errors` serving the story loop is a genuine win (complexipy `validate_schema` 34 to 28), the `prd_text_for_prompt(text)`-takes-bytes-not-a-path choice is right, and the `list[Any]` retype is honest. Every proposal above only removes code; none touches `PRD.validate_schema` or `_decompose_spec_impl`.

**1 APPLIED.** A real defect I introduced. Fixed.

**2 APPLIED in its duplicate-literal and backwards-guard halves, DECLINED in its
delete-the-frozenset half.** `_SPEC_ISSUES_KEY` is now defined first and
`PROMPT_EXCLUDED_KEYS` is built from it, so the literal appears once. The guard
is now `if not PROMPT_EXCLUDED_KEYS & set(data): return text` - asked forwards,
one branch, no comprehension in the common case. The frozenset stays because it
gained a second reader that answers the altitude agent's Q1: the new test
`test_a_new_prd_key_is_decided_rather_than_leaking_into_the_prompts` asserts
every schema-allowed top-level key is either listed as one both roles should
read or is in `PROMPT_EXCLUDED_KEYS`. A bare `!= _SPEC_ISSUES_KEY` comparison
gives that test nothing to hold.

**3 APPLIED.** `_entry_validator` deleted, `_validate_fixture_entries` is a plain
function named in the table.

**4 APPLIED, in its "move" form.** `SPEC_ISSUE_APPLIES_COMPONENT` / `_SPEC` now
live in `decompose.py`, their only writer. The false justification is gone.

**5 APPLIED, partly.** The measurement now has exactly one home
(`PROMPT_EXCLUDED_KEYS`); `security.py` and `knowledge.py` are one-line comments
naming which prompt they are and pointing at it, and the test docstring points
there too, so the three roundings are one number. The "annotate, resolve or
delete" and "spec-issues.json before any worktree" duplications are cut to one
copy each in `prd.py` (`tamper_changes` states them in full; the key constant
points at it). I did not cut the two comment blocks to five lines: the argument
they carry is why a reviewer will not re-tighten this field in six months, and
that is the thing this PR is actually buying. Volume declined, duplication
accepted.

**6 APPLIED.** Both defects were mine. Five rows now, matching the five the
source comment names, and the docstring says why a reword is deliberately absent
and where it is covered instead.

**7 APPLIED.** `_passes` deleted.

## Agent 2: altitude

> ## Verdict by question
>
> **Q1 - shared seam vs shared reader: the seam is at the right level, the denylist is not.**
>
> `kstrl/prd.py:69` `PROMPT_EXCLUDED_KEYS` / `kstrl/prd.py:72` `prd_text_for_prompt`
> Subtractive filter where the codebase already has the additive form. `kstrl/review.py:791-822` `build_review_prompt` is the third enrolled prompt that pastes PRD content, and it does not read bytes at all: it calls `PRD.load` and renders story id / title / acceptanceCriteria. It cannot leak any PRD key, present or future, without someone editing the renderer. security.py and knowledge.py paste raw bytes and now subtract one name.
>
> Fragile because a denylist has to be extended every time the PRD gains a field, and nothing fails when it is not. Compare `tests/test_spec_issue_routing.py:503` `test_a_field_added_to_prd_is_decided_rather_than_forgotten`, which AST-free-walks `fields(PRD)` and forces a recorded decision for `tamper_changes`. There is no equivalent for `PROMPT_EXCLUDED_KEYS`, so the PR built the guard for the weaker hazard and skipped it for the one it just discovered.
>
> Deeper form, in ascending cost:
> - Cheap and in bounds: extend that same `fields(PRD)` test so a new field also forces a decision about what the two prompts see. Same mechanism, one more assertion, no prompt edit.
> - Right depth: `prd_prompt_block(prd_path) -> str` in prd.py rendering from the dataclass the way review.py already does, called by all three sites. That is both the shared reader and the projection, and it deletes the denylist.
>
> Honest caveat: the projection changes what SECURITY_PROMPT reads more than the strip does (it currently also gets branchName, fixtures, allowedPaths), and that is itself an H2-governed quantity. So the strip is defensible as the scoped move for this PR. The denylist is the part that should not outlive it.
>
> **Q2 - excluding one key is the wrong altitude. The live hazard is `notes`, and it is worse than the one that was fixed.**
>
> `kstrl/pipeline.py:3401` and `:3651` both pass `wt_path / comp.prd_path` - the worktree copy the engineer just wrote. `UserStory.notes` (`kstrl/prd.py:23`) is an unbounded `str`, one per story, is one of exactly two fields the engineer is permitted to write, and is therefore not compared by `tamper_changes`. It flows verbatim and uncapped into SECURITY_PROMPT and DISTILL_PROMPT. review.py's projection excludes it.
>
> So after round 2 the security reviewer still reads unbounded text authored by the agent it is reviewing, reached without editing a prompt constant - the exact quantity the PR's own comment says H2 governs, in the strictly worse direction (architect-written vs subject-written). Stripping `specIssues` removed the one contributor the engineer does not control.
>
> The comment at `kstrl/prd.py:66-68` ("This is a strip, not a cap... the honest answer is zero") is correct about `specIssues` and is being used to answer a different question. Cap and strip are not alternatives here.
>
> Deeper form: a char bound inside `prd_text_for_prompt`, which is already the seam. `kstrl/git.py:1406` `DEFAULT_PROMPT_DIFF_CHAR_LIMIT` exists and its own comment says it is kept "so a future paste site has one number to adopt instead of inventing a third" - round 2 added a paste-preparation seam and did not adopt it. Related: `kstrl/knowledge.py:1178` hand-rolls that same 50000 cap rather than calling `git.truncate_diff_for_prompt`, which is the reuse half of the same point.
>
> **Q3 - the two-line special case is the right size, but it cost the table its status as the enumeration.**
>
> `kstrl/prd.py:293` `_OPTIONAL_ARRAYS`, `:320-321` the trailing `if`, `:427` `optional_keys`.
> Optional PRD keys are now enumerated in three places: the table, the hardcoded `optional_keys = {"allowedPaths", "fixtures", _SPEC_ISSUES_KEY}`, and the trailing special case. Before round 2, `optional_keys` was derivable from the table; now it cannot be. A fourth optional key means touching three sites, and nothing catches a miss in two of them.
>
> Deeper form: keep one table with strictness as a column (`empty_ok: bool`, `item_validator: Callable | None`), give `specIssues` a lenient row, and derive `optional_keys` from it. Then "gated fields are strict, notes are lenient" is stated once by the data rather than by a comment plus a code path.
>
> Two concrete defects in the same block:
> - `kstrl/prd.py:284-285` - the comment is two half-sentences spliced by a botched edit: "Every optional top-level PRD key follows the same rule: absent is / GATED optional key follows the same rule: absent is fine, present must...". One-line fix.
> - `kstrl/prd.py:69` - `PROMPT_EXCLUDED_KEYS` hardcodes the literal `"specIssues"` while `_SPEC_ISSUES_KEY` at `:146` was introduced in this same diff to stop that literal repeating. Use the constant (or move its definition above line 69).
>
> **Q4 - documenting fail-open is the right depth. Say so plainly.**
>
> Given no gate reads the field, every alternative is either a gate (out of bounds, and would recreate the false-failure problem round 2 just removed), a re-seed each iteration (contradicts the leniency the PR deliberately bought), or moving the note out of the PRD into a sibling worktree file - which is the genuinely deeper form, since it would need no strip, no schema exemption and no `tamper_changes` exemption, but it requires pointing the engineer at a new file and so is foreclosed by the no-prompt-edit constraint. The documentation is consistent across `CHANGELOG.md`, `docs/runbook.md` and the `tamper_changes` docstring, and the runbook now names the non-obvious part (the deletion persists across iterations). Correct call.
>
> One gap: "the worktree copy is seeded once" is asserted in prose in three places and tested nowhere. It rests entirely on `kstrl/factory.py:2025` `if not worktree_prd.exists() and prd_source.exists():`, a generic seed-once branch shared with prompt.md and CLAUDE.md. If that guard is dropped for an unrelated reason, three doc paragraphs go silently wrong. A one-line test pinning it is cheap.
>
> ## Summary
>
> Q1 partly too shallow (denylist, no guard), Q2 wrong altitude (the field that matters is `notes`, untouched), Q3 right size but three-way enumeration split plus a spliced comment and a stray literal, Q4 right depth.

**Q1 APPLIED in its cheap form, DECLINED in its `prd_prompt_block` form.** The
denylist now has the guard it was missing:
`test_a_new_prd_key_is_decided_rather_than_leaking_into_the_prompts` fails when
the schema gains a top-level key that is neither listed as prompt-visible nor
excluded. Its mutant ("an undecided key slips in") is killed. The projection is
declined for exactly the reason the agent names against itself: it would change
what `SECURITY_PROMPT` reads MORE than the strip does, which is the H2-governed
move this PR exists to avoid. Recorded as the follow-up.

**Q2 ACCEPTED as true, DECLINED as work for this PR.** `notes` is unbounded,
written by the agent under review, and reaches both prompts uncapped. That is
the worse hazard and I am not claiming otherwise. It is also structural and
pre-existing: capping it changes what every already-initialised project's
security reviewer reads, which is an H2-governed change, and a larger one than
the strip. What I did apply is the honesty half: the code comment no longer lets
"strip beats a cap" stand as a general answer. It now scopes the claim to a key
no prompt asks for and names `notes` as the case that stays whole and uncapped.
Flagged for the owner as its own change.

**Q3 APPLIED in full.** This was the sharpest finding: I had turned a
one-enumeration table into a three-way split. `_OPTIONAL_ARRAYS` now carries
`empty_hint: str | None` and `item_validator: Callable | None`, `specIssues` is a
row with `None` for both, `_OPTIONAL_KEYS` is derived from the table, and the
trailing special case is gone along with its duplicate "must be an array"
message. Both concrete defects fixed (see simplification 1 and 2). Three mutants
guard it: an empty-hint added back, an item validator added back, and the
derivation replaced by a hardcoded set - all killed.

**Q4 gap APPLIED.** `test_an_existing_worktree_prd_is_not_reseeded_from_the_root`
in `tests/test_spine_engineer_loop.py`, next to the existing test that proves the
copy happens. It drives the real `_run_component` against a real worktree with a
fake agent binary. Mutant "reseed the PRD every iteration": killed.

## Agent 3: reuse

> REUSE review of the round-2 diff. Four findings, three of them in the new test scaffolding.
>
> **1. `tests/test_spec_issue_routing.py:47` - `PromptCapturingAgent` is the third copy of an existing recording mock.**
> `RecordingAgent` already exists, byte-identical, in `tests/test_review_payload.py:71` and `tests/test_review_coverage.py:60` (verified with `diff`: IDENTICAL). It records `prompts`, `cwds` and `calls`, and is already driven through `run_security_review` at `tests/test_review_payload.py:259` for exactly this kind of "what did the reviewer read" assertion.
> Cost: a third mock-agent shape to keep in step with the `Agent` protocol, plus a cross-module inheritance edge into `tests/test_security.MockSecurityAgent` that did not exist before.
> Call instead: `RecordingAgent` (`agent.prompts[0]` replaces `agent.prompt`). Since it is now wanted in three modules, the codebase's own convention says hoist it to `tests/helpers/` the way `component_prd.py` and `procs.py` were hoisted, and have all three import it.
>
> **2. `tests/test_spec_issue_routing.py:564` - `with_observed_diffstat(VALID_SECURITY_OUTPUT, repo)` rebuilds what the repo already offers.**
> `ReviewRepo.security_json(**overrides)` (`tests/conftest.py:325`) returns a security reply already carrying `self.stat.as_payload()`, which is the whole job of that expression.
> Cost: two imports the test does not need (`VALID_SECURITY_OUTPUT` and `MockSecurityAgent` from `tests/test_security.py`, `with_observed_diffstat` from conftest), and a coupling to a fixture literal in another module whose findings are irrelevant here.
> Call instead: `repo.security_json()`.
>
> **3. `tests/test_spec_issue_routing.py:542,563,584,595` - `PRD_DOC` hand-rolls a PRD document the module already has two writers for.**
> The file imports `write_component_prd` (`tests/helpers/component_prd.py`, whose docstring is "the one writer, so the shape a fixture PRD needs stays in one place") at line 41 and already calls it with `spec_issues=` at line 494. It also imports `_make_prd_payload` (`tests/test_prd_allowed_paths.py:21`) at line 43 and uses it in `TestPrdSchema`.
> Cost: the document literal plus `json.dumps(..., indent=2) + "\n"` is written out at three sites in one class, and the story shape drifts from `component_prd.PASSING_STORY` / `_make_prd_payload`, so a future PRD-schema change has to find these too.
> Call instead: `write_component_prd(repo.path, "prd.json", stories=[marked_story], spec_issues=[FINDING])` for the two write sites (it returns the path), and `_make_prd_payload(userStories=[marked_story])` to build the no-findings document at line 595.
>
> **4. `kstrl/prd.py:321` - minor: the "must be an array" message now has two writers in one function.**
> The new lenient branch restates the string the `_OPTIONAL_ARRAYS` loop already emits at `kstrl/prd.py:315`.
> Cost: small, one phrasing with two owners inside `_validate_optional_arrays`.
> Reuse option: give the `_OPTIONAL_ARRAYS` row an `allow_empty` / optional item-validator field and put `specIssues` back in the table. Flagging it as low value only, because the deliberate exclusion is documented and putting it back adds a mechanism to the table to save two lines. Your call.
>
> **Explicit negative on the thing you asked me to check hardest:** no existing helper duplicates `prd_text_for_prompt`. `git.truncate_diff_for_prompt` (`kstrl/git.py:1409`) caps diff length and does not touch JSON; `kstrl/delimiters.py` only mints the untrusted-data token; `review.py:805-819` builds its PRD block from `PRD.load` + story lines, so it excludes `specIssues` structurally and shares no code path with the two raw-paste callers. Grep for key-filtering (`EXCLUDED_KEYS`, `redact`, dict-comprehension key filters) across `kstrl/*.py` returns only the two new lines. F1 is genuinely new code. The `json.dumps(kept, indent=2) + "\n"` format matches `atomicio.atomic_write_json` (`kstrl/atomicio.py:210`) but there is no text-returning serializer to call, so there is nothing to reuse there.

**1 APPLIED in its reuse half, DECLINED in its hoist half.** `PromptCapturingAgent`
is deleted; the tests import `RecordingAgent` from `tests.test_review_payload`,
which is the module that already drives it through `run_security_review`.
Cross-module test imports are this module's existing pattern (`_run_decompose`
from `test_decompose`, `_make_prd_payload` from `test_prd_allowed_paths`).
Hoisting to `tests/helpers/` is right and is the follow-up: it touches two
unrelated test modules, and a three-module move inside a PR whose subject is one
routing rule is exactly the diff-widening this review round told me to stop
doing.

**2 APPLIED.** `repo.security_json()`. Two imports dropped.

**3 APPLIED, via `_make_prd_payload`.** One `_prd_text(**overrides)` helper on the
class builds the document from the shared payload builder with a marked story,
used at all four sites. I kept `_make_prd_payload` over `write_component_prd`
because two of the four sites want the TEXT, not a written file, and the
byte-for-byte pass-through assertion needs the exact string it compares.

**4 APPLIED, and it was not minor.** Taken together with the altitude agent's Q3
this is the same finding: the table had stopped being the enumeration.
`specIssues` is back in it as a lenient row, `_OPTIONAL_KEYS` is derived, and
the duplicate message is gone.

## Agent 4: efficiency

> ## EFFICIENCY review, #260 round 2: no findings
>
> Every cost this diff adds or removes is below the noise floor of the path it sits on. The current form is fine as written. Numbers below are all measured, not estimated.
>
> **Method.** Synthetic PRD matching the stated real scale (11 stories, allowedPaths, one fixture): 8,540 bytes with no `specIssues`, 17,260 bytes with 15 routed findings. `timeit`, 1000-2000 reps, warm; `tracemalloc` for allocation. Deleted validator reconstructed verbatim from the diff and run against the same entry shape. Scripts in `/private/tmp/claude-501/.../scratchpad/bench_prd.py`, `bench_validate.py`, `bench_mem.py`, `bench_scale.py`. Machine is the Darwin arm64 box this worktree is on; treat the absolute microseconds as machine-specific, the ratios as the point.
>
> **(a) The `json.loads` + `json.dumps` round trip - `kstrl/prd.py:72`, called at `kstrl/security.py:667` and `kstrl/knowledge.py:983`.**
>
> | path | per call |
> |---|---|
> | `prd_text_for_prompt`, no `specIssues` (8.5 KB) | 8.4 us |
> | `prd_text_for_prompt`, 15 findings (17.3 KB) | 58 us |
> | of which `json.loads` | 16 us |
> | of which `json.dumps(indent=2)` | 40 us |
>
> Against what the same call site already spends: `prd_path.read_text` is 12 us warm, so the routed path costs about 5 file reads. Against what the phase spends: one security review or one distill is an LLM call taking seconds to minutes. At 10 seconds, 58 us is 6e-4 percent of the phase. It runs once per component per phase - there is no loop; `grep` confirms exactly the two call sites.
>
> Scaling stays flat because only the parse grows; the dump is of the kept dict, which is constant:
>
> | findings | PRD in | out | us/call |
> |---|---|---|---|
> | 0 | 8,540 | 8,540 | 8.4 |
> | 15 | 17,260 | 8,540 | 58 |
> | 100 | 66,641 | 8,540 | 105 |
>
> A pathological 66 KB PRD costs 105 us. Negligible.
>
> The other direction is the part worth stating: at 15 findings the function removes **8,720 bytes** from the prompt, and it does that on two prompt paths per component. That is roughly 2,200 tokens per call by the chars/4 rule of thumb (an approximation, not a measurement - I did not run a tokenizer). Whatever the exact token count, the CPU spent to remove those bytes is microseconds and the tokens removed are billed input on every call. The trade is not close.
>
> **(b) Re-serialising vs string-slicing at this size - not worth changing.** A regex slice of the `specIssues` block measures 34 us against the round trip's 58 us. The 24 us difference is invisible next to the LLM call, and the slice is the more fragile construction (it has to get nesting and trailing commas right, and it would silently mangle a PRD whose formatting differs from what the harness writes). Keep the round trip.
>
> **(c) The deleted per-entry validator does NOT make decompose measurably cheaper.** Measured over the exact scale the deleted comment named (20 findings x 9 components = 180 entries): **82 us total, once per decompose**, 1.4 KiB transient allocation. The replacement `isinstance(..., list)` check is 17 ns. So the deletion saves 0.082 ms per run, against a decompose that makes at least one LLM call. This is not an efficiency win and should not be defended as one - it stands or falls entirely on the false-positive argument the code comment already makes. Equally, the pre-existing `_SPEC_ISSUE_KEYS_SORTED` "sorted once because the loop runs per entry" optimisation was guarding 82 us; no loss there either.
>
> **(d) No retention introduced.** `prd_text_for_prompt` holds nothing past return: peak transient 61.6 KiB on a 16.9 KiB input (3.6x, all freed), 15.0 KiB retained and that is just the returned output string. On the no-`specIssues` path it returns the **input object itself** (`out is text` is True) - 16.1 KiB peak, no copy. Both call sites bind the result to a local that dies with the frame. The only new module-level object is `PROMPT_EXCLUDED_KEYS`, a 1-element frozenset. Closure count went **down**: dropping the `specIssues` row from `_OPTIONAL_ARRAYS` removed one `_entry_validator` closure held for process life, leaving one (fixtures).
>
> **One thing I checked and am deliberately not filing.** On the common bare path the function pays 8.4 us to parse and discover nothing to strip; a `if "specIssues" not in text: return text` guard measures 1.9 us, saving 6.5 us per call, twice per component. Six microseconds is not a reason to add a second code path with its own subtle mismatch (the guard would match the substring inside a story's prose). Leave it alone.

**No findings, nothing to apply.** One correction to its (c): I heeded it. The
commit message and code comment defend the F2 leniency purely on false
positives; no efficiency claim is made for it anywhere. Its (d) closure count is
now inverted by the Q3 restructure - `specIssues` is back in the table and
`_entry_validator` is deleted outright, so the process-life closure count goes
from one to zero rather than from two to one. Its "deliberately not filing" note
is also now moot in the direction it wanted: the early return exists, but after
the parse, so it costs nothing extra and cannot match `specIssues` inside a
story's prose.

# Left for the owner

**`UserStory.notes` reaches both enrolled prompts unbounded and uncapped.**
Raised by the altitude agent, accepted as true, and deliberately not fixed here.
It is written by the agent under review, is one of exactly two fields the
engineer is permitted to write, and is therefore not compared by
`tamper_changes`. Capping it changes what every already-initialised project's
security reviewer and distiller read, which is H2-governed and a larger change
than this PR's. `git.DEFAULT_PROMPT_DIFF_CHAR_LIMIT` already exists and its own
comment says it is kept so a future paste site has one number to adopt;
`prd_text_for_prompt` is now the seam that could adopt it.

**Follow-ups recorded, not done:** hoist `RecordingAgent` to `tests/helpers/`
(three modules); `prd_prompt_block` rendering from the dataclass for all three
PRD-pasting prompts (declined here because it changes what `SECURITY_PROMPT`
reads more than the strip does); relocate `SpecIssue` to sit with its writer.

PR #334, opened 2026-09-01
fix(verify): a non-measurement can no longer report a mutation pass (#306, #316)

0xfauzi:
Closes #306. Closes #316.

Three adjacent lines in `kstrl/verify.py`, so one PR. Both issues carry owner
correction comments; both were re-verified against `origin/main` at `cbdff7c`
before anything moved.

## 1. #306: a green `mutation_testing` row meant "we did not look"

Reproduced on a throwaway repo, `ks sense --json` with `mutation_testing = true`:

```
mutation_testing  True  | Skipped: mutation testing rewrites the files it mutates and cannot run read-only
```

`check_mutation_score` had **five** paths that measured nothing and returned
`CheckResult(passed=True)`, not the four the issue names. The fifth is
`total == 0`, "No mutations generated": mutmut ran and produced neither a killed
nor a survived count, so 0/0 was reported as a pass. Same defect, same function,
same fix, so it is fixed here and called out rather than left.

| path | reaches | before | after |
|---|---|---|---|
| `read_only=True` | `ks sense`, `run_undiffed_verification` | `passed=True` | no row |
| mutmut not installed | `ks factory` | `passed=True` | no row |
| no non-test `.py` in the diff | `ks factory` | `passed=True` | no row |
| `mutmut run` timed out | `ks factory` | `passed=True` | no row |
| no mutants generated | `ks factory` | `passed=True` | no row |

### Where the false pass actually went

Not only the `ks sense` table. `kstrl/review.py:813-816` builds the LLM code
reviewer's `verification_summary` from **every** check, passing ones included.
Run against `origin/main`'s row:

```
verification_summary handed to REVIEWER_PROMPT:
- mutation_testing: PASS - mutmut not installed, skipping

report_lines (ks sense / ks feature table):
  mutation_testing  pass  mutmut not installed, skipping

VerificationResult.passed: True
```

An adversarial role was being told mutation testing passed on a run where mutmut
was not installed.

### Decision: omit the row, not an explicit not-measured status

The issue asks which. **Omit**, for three reasons, in order of weight.

**1. An explicit status on the row cannot deliver the requirement.** The
requirement is that a non-measurement be impossible to read as a pass. `passed`
is the only field every consumer reads: `VerificationResult.passed =
all(c.passed ...)`, `report_lines` prints `pass`, `ks sense --json` emits
`"passed": true`, and the reviewer prompt above prints `PASS`. Adding
`measured: bool` beside `passed` leaves every one of those reading a pass until
[...]

PR #327, opened 2026-09-01
fix(evolution): a torn journal line no longer eats the entry after it (#312)

0xfauzi:
Fixes #312.

## The defect, reproduced against real bytes

`append_entries` opened the journal in append mode and wrote, without ever
asking whether the file already ended in a newline. A crash mid-write leaves an
unterminated tail, so the next entry was concatenated onto the fragment and the
pair became one unparseable line. The tolerant reader skipped it, and the entry
written after the crash was gone.

Run on cbdff7c, real journal, real bytes, no mocks:

```
after audit A: 1 line(s)
after the crash: file ends with newline? False

--- journal on disk ---
{"timestamp":"2026-08-20T00:00:00","project":"alpha","event_type":"spec_issues","spec_file":"alpha.md"}
{"event_type": "spec_iss{"timestamp":"2026-08-20T00:00:00","project":"beta","event_type":"spec_issues","spec_file":"beta.md"}

--- what any reader sees ---
1 line(s) parsed: ['spec_issues']

audits written: 2 (alpha, beta)
audits readable: 1 ['alpha']
```

Same script after this change:

```
--- journal on disk ---
{"timestamp":"2026-08-20T00:00:00","project":"alpha","event_type":"spec_issues","spec_file":"alpha.md"}
{"event_type": "spec_iss
{"schema_version":2,"timestamp":"2026-09-01T09:23:40Z","event_type":"journal_repair","detail":"the preceding line was not newline-terminated when this append ran, ..."}
{"timestamp":"2026-08-20T00:00:00","project":"beta","event_type":"spec_issues","spec_file":"beta.md"}

--- what any reader sees ---
3 line(s) parsed: ['spec_issues', 'journal_repair', 'spec_issues']

audits written: 2 (alpha, beta)
audits readable: 2 ['alpha', 'beta']
journal_repair rows: 1
```

## Blast radius, measured rather than repeated

The issue says the cost is one entry. Measured, it is not uniform. Six tear
offsets, each on a journal holding three audits, then one more audit appended
through the real writer:

| tear | last byte | audits before | audits after, cbdff7c | audits after, this PR | repair row |
|---|---|---|---|---|---|
| mid-line (torn JSON) | `t` | 2 | 2, new entry eaten | 3 | 1 |
| at a line boundary (intact) | `\n` | 3 | 4 | 4 | 0 |
| trailing newline lost only | `}` | 3 | **2, TWO records eaten** | 4 | 1 |
[...]

Issue #324, opened 2026-09-01
Factor the eleven AST guards onto a shared tests/helpers/astwalk.py


0xfauzi:
Surfaced by the round-3 /simplify pass on PR #319.

The repo now has roughly eleven AST-walking test guards (`test_atomicio.py`, `test_process_scoping.py`, `test_prompt_versions.py`, `test_prompt_enrollment_walk.py`, `test_timeout_enforcement.py`, `test_toml_readers.py`, and others). Each re-implements the same primitives: walk `kstrl/`, parse each file, find calls, resolve which module a name refers to, collect handler names from a `try`.

Each re-implementation has been holed independently. Within the last two PRs alone:

- #319 round 2: the toml guard was defeated by `import tomllib as _tl` in a new file. Ruff and mypy clean, nothing objected.
- #321 round 1: the wait guard passed `wait(timeout=None)` and `with Popen(...)`.
- #321 round 2: the same guard, after being fixed, was still defeated by `_P = subprocess.Popen`, `getattr(subprocess, "Popen")`, and a `with` on a Popen returned from a helper.
- #321: `_subprocess_aliases` records the LOCAL name and compares it against the ORIGINAL, so any `as` rename escapes it. A new module can call `subprocess.run` with no timeout and pass the whole audit.

Alias resolution is the recurring hole and it has now been solved correctly at least twice, separately, in files that cannot see each other.

## Suggested shape

`tests/helpers/astwalk.py` holding, at minimum: module-alias resolution (`import x as y`, `from x import f as g`, `_p = x.f` rebinds to a fixed point), call matching against a resolved qualified name, and handler-name collection from a `try` including which clause comes first.

Two constraints worth writing into it: a walk that matches nothing must fail rather than pass (the "silently matches nothing" failure mode every one of these docstrings names), and unresolvable names must count in the conservative direction, which is toward flagging rather than skipping.

Prerequisite: land it with the existing guards migrated, so the shared helper starts at zero known holes rather than inheriting a suppression list.

0xfauzi:
Fifth instance, found in PR #328 (issue #308) an hour after this issue was filed.

The new fourth-copy net in `tests/test_safe_pgid.py` matches only calls shaped `os.getpgid(...)`. A reviewer planted a COMPLETE fourth guard in a new module: the direct form failed the test as intended, then the same guard PASSED after `import os as operating_system`.

| form | result |
|---|---|
| direct `os.getpgid` | caught |
[...]

Issue #280, opened 2026-08-30
Convergence report silently omits history recorded under a different project name


0xfauzi:
Found by replaying PR #270's shipped convergence report against real journal data, from the dogfooding run that motivated #260. This is a small gap in a merged feature, not a defect in it.

## What happens

The report keys "the same spec" on the project name, and that choice is documented with its failure modes. One of those failure modes turns out to be the exact shape of a real session.

My five architect runs on one project went: 7 blockers, 11, 1, 3, 4. The 11-to-1 drop is where I cut scope by roughly four times and passed a different `--project-name`. So the journal holds:

```
1 spec.md            writers-room         {'blocker': 7,  'major': 11, 'minor': 3}
2 spec.md            writers-room         {'blocker': 11, 'major': 14, 'minor': 7}
3 spec-slice-1.md    writers-room-slice1  {'blocker': 1,  'major': 10, 'minor': 9}
4 spec-slice-1.md    writers-room-slice1  {'blocker': 3,  'major': 10, 'minor': 8}
5 spec-slice-1.md    writers-room-slice1  {'blocker': 4,  'major': 11, 'minor': 8}
```

Replaying run 5 through `_build_convergence` and `_surface_convergence` produces:

```
== Spec Convergence ==
  Blocker:      4 (previous run: 3, +1)
  Major:        11 (previous run: 10, +1)
  Minor:        8 (previous run: 8, no change)
  Trend:        1, 3, 4 (blockers, oldest run first)
Previous run raised 21 issue(s): 0 reappear verbatim, 21 do not.
Matched on issue text, so a reworded issue counts as new: a floor on what carried over,
not a count of what the architect resolved.
```

The trend is `1, 3, 4`, not `7, 11, 1, 3, 4`. Two runs of history are silently absent.

## Why this is worth a warning rather than a redesign

The current behaviour is arguably correct. I really had cut a different, smaller project at that point, and starting a fresh trend for it is defensible. Content hashing would be worse, since the spec is edited every round by construction and would never match; keying on the path would have failed here too, because the file was renamed at the same moment.

The problem is only that the loss is **silent**. An operator who renames a project for any reason, or who runs the same spec under two names, gets a shorter trend with nothing telling them history exists elsewhere. The report's whole value is that the operator trusts the numbers, so a number quietly computed over less data than the journal holds is the one failure it cannot afford.

## Suggested fix

[...]

PR #336, opened 2026-09-01
refactor(evolution): the spec-audit reader, its window and its event name live on the journal (#314)

### LINKED ISSUE #314: Fold the spec-audit journal reader, window rule and event constant into EvolutionJournal
Three related loose ends from `/code-review` rounds 1 and 2 on PR #311
(issue #280). All three are one change to `kstrl/evolution.py`, which was
under concurrent edit throughout #311 (PRs #304 and #310), so #311 deferred
them rather than racing the file. Each is pinned by a test on the #311 side
so none of them can rot silently, but the tests are mechanisms guarding a
compromise, not the fix.

## 1. `decompose` reaches past `EvolutionJournal` into its storage path

`kstrl/decompose.py::_journal_entries` does:

```python
return read_progress_events(journal.config.journal_path)
```

It does this because no public reader on `EvolutionJournal` returns the
whole entry set: `get_spec_issue_runs` filters to one project inside the
reader by design, and `_read_all_entries` is private.

The cost is a layering hole with a silent failure mode. If the journal ever
compacts, rotates, or gains a second segment, `get_spec_issue_runs` keeps
working because it goes through the journal's own reader, while
`read_progress_events(journal.config.journal_path)` returns fewer entries or
none. The convergence trend would keep printing while the note that says
what the trend excludes went quiet. Silent loss of excluded-history
accounting is exactly the defect #280 exists to fix.

**Fix:** a public reader on `EvolutionJournal` returning all spec audits,
for example `get_spec_issue_runs(project=None)` meaning "all", or a
`spec_issue_entries()` sibling. `decompose._journal_entries` then deletes.

## 2. `_windowed_audits` restates a rule `get_spec_issue_runs` owns

`kstrl/decompose.py::_windowed_audits` reimplements the window:

```python
mine = [e for e in audits if _entry_str(e, "project") == project_name]
return mine[-last_n:] if last_n > 0 else []
```

That is `get_spec_issue_runs`'s body, applied to entries already in memory
so one read serves both the trend and the accounting under it. Two copies of
one rule can drift, and if they drift the trend and the accounting disagree
about the same journal.

**Consequence today:** with #311 merged, `EvolutionJournal.get_spec_issue_runs`
has **no production caller at all**. It is reached only from
`tests/test_evolution.py` and from #311's agreement test. That is a public
method kept alive by its own tests, which is worse than either owning the
rule or deleting it.

**Fix:** fold both into the reader from item 1, and either delete
`get_spec_issue_runs` or make it the one place the window lives, called with
pre-read entries.

**Pinned meanwhile by**
`tests/test_decompose.py::TestOneJournalRead::test_the_windowing_rule_matches_the_journals_own`,
which asserts `_windowed_audits(...)` equals `journal.get_spec_issue_runs(...)`
across `last_n` of 0, 1, 3 and 10. It covers both item 1 and item 2: if the
storage path stops being the whole story, or the window rule changes on
either side, it fails.

## 3. `SPEC_ISSUES_EVENT` is in the wrong module

`kstrl/decompose.py:802` defines:

```python
SPEC_ISSUES_EVENT = "spec_issues"
```

while `kstrl/evolution.py:1277`, the storage layer that also filters on it,
keeps its own hardcoded `"spec_issues"`. The constant is in the writer, not
in the layer that defines the journal's schema, so any third reader added in
`evolution.py` will reach for the literal rather than the constant.

**Fix:** move the constant onto `kstrl/evolution.py` (or a shared events
module) and have `decompose.py` import it. Verify no import cycle:
`decompose` currently imports `kstrl.evolution` lazily inside functions, so
a module-level import in the other direction needs checking.

**Pinned meanwhile by**
`tests/test_decompose.py::TestSpecConvergenceThroughDecompose::test_the_journal_reader_agrees_with_the_event_name_written`,
which writes an entry through `decompose` and reads it back through
`get_spec_issue_runs`. Measured: changing the constant alone fails that test
and fourteen others, so divergence cannot ship quietly. The remaining
problem is placement, not correctness.

## Why one issue rather than three

All three are edits to the same file and the same seam. Splitting them
produces three PRs racing `kstrl/evolution.py`, which is what caused the
deferral in the first place.

0xfauzi:
Closes #314 (do not merge yet; not merging or closing from here).

`decompose` reached past `EvolutionJournal` into its storage path, held a second copy of the window rule the journal owns, and the event-name constant lived on the writer while the storage layer kept its own literal. All three were deferred by #311 rather than raced against a file under concurrent edit. `evolution.py` has since settled (#327, #334), so they are folded here.

## The two claims verified before building

**`get_spec_issue_runs` had no production caller.** By AST walk over `kstrl/`, not by grep: zero `ast.Name`, `ast.Attribute` or `ast.ImportFrom` references. Only docstrings named it. It was a public method kept alive by its own tests.

**No import cycle.** `kstrl.evolution` does not reach `kstrl.decompose` through any module-level import, transitively, with `TYPE_CHECKING` blocks excluded. So `decompose` can import it at module scope, which is what lets `SPEC_ISSUES_EVENT` and `entry_str` be imported rather than duplicated.

The measured cost of that eager import, since it is the one thing this change makes slower. `import kstrl.cli`, interleaved subprocess pairs against `main`:

| condition | main | this branch | delta |
|---|---|---|---|
| `PYTHONDONTWRITEBYTECODE=1`, caches cleared, n=30 | 202.6 ms | 206.9 ms | +4.3 ms (+2.1%) |
| `__pycache__` warm, n=25 | 81.3 ms | 82.1 ms | +0.8 ms (+1.0%) |

The warm row is production: an actual `ks` invocation pays +0.8 ms. `sys.modules` gains exactly three entries (`kstrl.evolution`, `csv`, `_csv`); `kstrl.evolution`'s own imports are `kstrl.observability` and `kstrl.verify`, both of which `cli.py` already loads, so there is no transitive fan-out.

## The caller trace

Direct readers of the spec-audit journal before this change:

| Site | What it did |
|---|---|
| `decompose._journal_snapshot` | `read_progress_events(journal.config.journal_path)` - the production reach-past |
| `decompose._windowed_audits` | a second copy of `get_spec_issue_runs`'s rule |
| `decompose._spec_audits` | a second copy of "which rows are spec audits" |
| `decompose._excluded_projects` | a third copy of the same filter |
| `decompose.SPEC_ISSUES_EVENT` | the constant, on the writer |
| `evolution.get_spec_issue_runs` | the same name again, as a bare literal |
| `tests/helpers/journal.py` (x2) | the same literal in shared test infrastructure |

`kstrl/cli.py:3503` and `kstrl/reducer.py:647` also call `read_progress_events`, but on the progress log, which is a different file. `kstrl/cli.py:689` interpolates `journal.config.journal_path` into an operator message and does not read it. Neither is in scope.

## What moved

- **`EvolutionJournal.get_spec_audits`** is the new unwindowed reader: every recorded spec audit, across every project. Unfiltered by project on purpose - the accounting that reports what the trend leaves out needs exactly what the window drops.
- **`get_spec_issue_runs` is the one place the window rule lives, and has a production caller again.** Its `audits=` parameter lets the single read feed both the trend and the accounting, which is what #280 round 2's one-read pin asked for.
- **`SPEC_ISSUES_EVENT` and `entry_str` move to `kstrl/evolution.py`**, the layer that defines the journal's schema. `decompose` imports both.
- **Deleted:** `_windowed_audits`, `_spec_audits`, `_spec_convergence`, `_entry_str`. `_excluded_history` and `_excluded_projects` take audits and no longer restate which rows are audits.
- `_journal_snapshot` returns a frozen `AuditSnapshot(audits, window, lookback)`.

## Behaviour difference found, reported rather than absorbed

`get_spec_issue_runs` matched a project with `entry.get("project") == project`. The report's rule, in `_windowed_audits` and `_excluded_history`, used `_entry_str`, which reads a null or non-string field as absent.

The two answer differently for exactly one query: `project == ""` against an audit whose `project` field is absent or not a string. The raw comparison excludes it; the `entry_str` rule counts it as belonging to the project named `""`. Reachable: `ks decompose --project-name ""` satisfies click's `required=True`.

**The production rule is the one kept**, so nothing recorded changes. `get_spec_issue_runs` had no production caller, so nothing observed its old rule. `TestSpecAudits::test_an_unattributed_audit_belongs_to_no_project` pins the kept rule and fails on a revert to the raw comparison.

## A pre-existing hole the mutation pass found

With `get_spec_issue_runs` reading `runs[:last_n]` instead of `runs[-last_n:]`, **the entire fast tier stayed green** - 4771 passed. Every window test in the suite recorded audits raising the same blocker count, so the oldest N and the newest N rendered identically. The trend could have shipped reading a history the operator had already acted on.

The hole predates this issue (`_windowed_audits` had the same slice and the same absence of a test), but the fold makes one rule serve both consumers, so it is closed here: two tests, one per layer, both failing on the inverted slice.

## Planted mutations, with the exact failure text

Every run with `PYTHONDONTWRITEBYTECODE=1` and `__pycache__` cleared first. Each mutation is planted, the guards run, the file is restored by rewriting its original text (never by `git checkout`).

**1. `_journal_snapshot` reads the storage path again** - 5 failures, including:

```
FAILED tests/test_journal_one_writer.py::TestOneWriter::test_no_new_code_gets_hold_of_the_journal_path
E   AssertionError: The set of places that get hold of a journal path changed. If this is
    a new writer of the evolution journal, route it through EvolutionJournal.append_entries:
    an unguarded append concatenates onto an unterminated tail and eats the entry after it
    (#312). If it is a read, or another file's journal, add it to EXPECTED_JOURNAL_PATH_SITES
    with a reason. Found: {'cli.py: journal.config.journal_path': 1, 'decompose.py:
    journal.config.journal_path': 1, ...}
E   Left contains 1 more item:
E   {'decompose.py: journal.config.journal_path': 1}

FAILED tests/test_decompose.py::TestOneJournalRead::test_the_report_reads_through_the_journal_not_its_storage_path
E   AssertionError: assert ['on-disk.md'] == ['b.md', 'c.md']
```

**2. the window drops `audits=` and re-reads** - 1 failure:

```
FAILED tests/test_decompose.py::TestOneJournalRead::test_the_journal_is_parsed_once_per_report
E   assert 2 == 1
E    +  where 2 = len([1, 1])
```

**3. the window keeps the oldest N** - 2 failures:

```
FAILED tests/test_evolution.py::TestSpecAudits::test_the_window_keeps_the_newest_audits_not_the_oldest
E   AssertionError: assert ['spec-0.md', 'spec-1.md'] == ['spec-3.md', 'spec-4.md']

FAILED tests/test_decompose.py::TestSpecConvergenceThroughDecompose::test_the_window_keeps_the_newest_audits
E   AssertionError: assert '2, 3, 4 (blockers, oldest run first' in '...'
```

**4. a reader spells the event as a literal** - 1 failure:

```
FAILED tests/test_evolution.py::TestJournalEventNamesHaveOneHome::test_no_module_names_an_enrolled_event_as_a_literal
E   AssertionError: A journal row is written or selected by a bare literal instead of the constant evolution.py declares for it. Import the constant. Sites: {'evolution.py': ["SPEC_ISSUES_EVENT: line 1497: compares event_type against 'spec_issues'"]}
```

**5. the writer spells the event as a literal** - same test:

```
E   Sites: {'decompose.py': ["SPEC_ISSUES_EVENT: line 1031: writes 'spec_issues' as the event_type"]}
```

**6. the detector's reader half is switched off** (`if False and isinstance(node, ast.Compare)`) - 4 failures, so the positive controls notice their own detector going quiet rather than going blind:

```
FAILED ...::test_the_detector_sees_a_literal_reader
E   assert [] == ["line 1: compares event_type against 'spec_issues'"]
FAILED ...::test_the_detector_sees_the_reversed_and_membership_spellings
FAILED ...::test_an_assembled_spelling_is_not_a_way_past_it
FAILED ...::test_the_second_enrolled_name_is_detected_too
```

**7. the project match reverts to the raw comparison** - 1 failure:

```
FAILED tests/test_evolution.py::TestSpecAudits::test_an_unattributed_audit_belongs_to_no_project
E   AssertionError: assert ['empty.md'] == ['null.md', 'int.md', 'empty.md']
```

**8. `journal_repair` spelled as a literal** (the SECOND enrolled name, to prove the generalisation is live over the package and not only over snippets) - 1 failure:

```
E   Sites: {'evolution.py': ["JOURNAL_REPAIR_EVENT: line 1467: compares event_type against 'journal_repair'"]}
```

### The AST walk was checked for blindness, not just for green

#327's layer-1 net loses its `decompose.py` row here. Removing a row from a guard is exactly the change that could turn a red guard into a blind one, so the walker's output was inspected rather than inferred: **5 rows, 15 occurrences** after (6 rows, 16 before), matching `EXPECTED_JOURNAL_PATH_SITES` exactly. It still sees `cli.py`, `evolution.py` (both spellings), `pipeline.py` and `workqueue.py`.

## Tests

Added:

- `TestSpecAudits` (7): unwindowed across all projects; other event types excluded; `audits=` windows without a second read; a supplied snapshot answers the same as raw entries; the window keeps the newest; a non-positive window reads nothing; an unattributed audit belongs to no project.
- `TestJournalEventNamesHaveOneHome` (7): the package walk, plus positive controls for the writer shape, the reader shape, the reversed and `in` spellings, the concatenated and f-string spellings, the second enrolled name, and a negative control for the architect's own `spec_issues` JSON key.
- `TestOneJournalRead` (4, rewritten): parsed once per report, counted at `_read_all_entries`; the report reads through the journal, proved by a reader answering something the file does not contain; the window is the journal's own over the journal's own lookback; no journal reads nothing.
- `TestSpecConvergenceThroughDecompose::test_the_window_keeps_the_newest_audits`.

Retired: `test_the_windowing_rule_matches_the_journals_own`, whose subject (the deferral) no longer exists. `test_only_spec_audits_count` and `test_every_spec_audit_lands_in_exactly_one_bucket` now go through the journal, because the selection rule moved there.

## CI

- fast tier: `4776 passed, 28 skipped, 6 xfailed` in 270 s
- spine tier: `27 passed` in 19 s
- `ruff check kstrl/ tests/`: clean. `ruff format --check`: 297 files already formatted.
- `mypy kstrl/ --strict`: no issues in 127 source files
- coverage: 90.09% total, against the 79.91% ratchet in `.github/coverage-baseline.txt`
- full pre-commit (including complexipy, vulture, codespell, GitGuardian, the cyclomatic and file-length ratchets, no-em-dash): passed on every commit

## `/simplify` findings, verbatim

Four angles ran over the branch diff. Applied and declined are both below, in full.

<details>
<summary><b>Reuse</b> (7 findings)</summary>

**1. `kstrl/evolution.py:494` - `entry_str` is a third module-level copy of "string field or default"**
`kstrl/workqueue.py:269 _as_str(data, key, default="")` has a byte-identical body (`value = data.get(...); return value if isinstance(value, str) else default`) and 21 call sites. `kstrl/events.py:727` is the same rule inside `_coerce`, and `events.py:738-739`, `serve.py:2672`, `tui/screens/launch.py:102`, `tui/screens/init_wizard.py:228-231` inline it again. The new one is the only *public* one, so it is the natural single home, but nothing was folded into it. Cost: the "a null field is an absent field" reasoning in the new docstring now exists in prose in one place and as unexplained code in six others; a future fix (bools, whitespace, `None` vs `""` distinction) lands in one copy. Concrete move: put one helper in `kstrl/observability.py` (already the owner of `read_progress_events`, already imported by `evolution.py`, and imports neither `workqueue` nor `evolution`, so no cycle) and have `workqueue._as_str` delegate.

**2. `tests/test_evolution.py:1238, 1331` - `KSTRL_PACKAGE` + `sorted(rglob("*.py"))` re-implements `package_sources()`**
`tests/test_journal_one_writer.py:45 package_sources()` is exactly this, and cross-file import of that module's guard helpers is already established (`tests/test_journal_guard_detects.py:28-35` imports `KSTRL_PACKAGE`, `label`, `journal_path_escapes`). The new guard also parses the whole package for itself rather than using `test_journal_one_writer.py:57 parsed()`, the shared text-keyed tree cache that exists precisely because three walkers were each parsing the package. Measured here, not quoted: 127 modules, **113 ms for one full parse pass**. Cost: one more uncached pass per session, and the walker inventory drifts (a new `kstrl/` subpackage or an exclusion added to `package_sources()` never reaches this guard). Reuse needs a small signature change - `spec_issues_literals` takes `source: str` so its snippet tests work, so pass it a tree (`parsed(path)`) and keep an `ast.parse` only in the snippet tests.

**3. `tests/test_evolution.py:1334` - keys the failure dict on `source_file.name` instead of `label()`**
`tests/test_journal_one_writer.py:28 label()` exists specifically because ten basenames occur twice in `kstrl/`, and its own docstring names `decompose.py` as one of them (`kstrl/decompose.py` and `kstrl/tui/screens/decompose.py`) - the exact file this change is about. `tests/test_journal_guard_detects.py:214-215` already pins that behaviour. Cost: when this guard fires on the TUI copy it prints `decompose.py`, and the reader opens the wrong file. Free fix: import `label`.

**4. `tests/test_evolution.py:1308 _constant_strings` - a weaker re-implementation of `folded_str`**
`tests/test_journal_one_writer.py:116 folded_str` / `147 folded_concat` / `173 folded_parts` already answer "the string this expression is KNOWN to evaluate to", with positive controls in `test_journal_guard_detects.py`. That file's own rule is stated explicitly: "a shape a reader can decide by looking at it is a shape this guard must decide". `_constant_strings` matches raw `ast.Constant` only, so `e.get("event_type") == "spec" + "_issues"` and the f-string spelling walk past the new guard while the sibling guard decides both. Cost: two divergent notions of "is this literal that string" in one test suite, and the newer one is the one guarding the constant this PR just created. Reuse: apply `folded_str` per operand, keeping the tuple/list walk for the `in (...)` case.

**5. `tests/test_evolution.py:1122-1131` and `tests/test_decompose.py:1595` - a fourth and fifth copy of the audit-record builder**
`tests/helpers/journal.py:73 audit(project)` and `:96 journal_at(tmp_path)` exist for exactly this, and `test_journal_write_boundary.py` / `test_journal_torn_tail.py` already import them; `tests/test_decompose.py:33` already imports from that module. Within `test_decompose.py` the new `TestOneJournalRead._audit` (1595) is field-for-field identical to `TestExcludedHistory._entry` (1186) - same four keys, same `"2026-08-20T00:00:00Z"` default - and duplicates the module-level `_journal_with` (792) as `TestOneJournalRead._journal` (1590). In `test_evolution.py`, `TestSpecAudits._journal` (1122) is the third `_journal` in that file (672, 1037) and `_audit` (1127) overlaps `TestSpecIssueRuns._spec_entry` (1040) directly above it. Cost: the shape of a recorded spec audit is now spelled in five places; a schema field added to the row has to be found in all of them, and the helper module's docstring already flags that some of its builders have one caller - this diff had two chances to raise that count and took neither.

**6. `tests/helpers/journal.py:77` and `:104` - the bare `"spec_issues"` literal the new constant was created to remove**
The diff's own claim is that "there is one `SPEC_ISSUES_EVENT` now", and `TestSpecIssuesEventHasOneHome` enforces it - but only over `kstrl/`, so these two survive. The helper already imports `JOURNAL_REPAIR_EVENT` from `kstrl.evolution`, so importing `SPEC_ISSUES_EVENT` costs nothing. `audits_in()` (`:101-105`) additionally re-implements the filter that is now `EvolutionJournal.get_spec_audits()`. Cost: the guard's stated failure mode ("a reader added here reaches for the nearest spelling") is left standing in the file test authors read first.

**7. `kstrl/evolution.py:1497` (in `get_spec_audits`) - same one-liner as `get_repair_count` at 1467**
Both are `[e for e in self._read_all_entries() if e.get("event_type") == <CONST>]`. Low cost today (two sites), but it is the third such filter in the class counting the inline ones at 938/960/1331/1385, and a private `_entries_of_type(event_type)` would give the new `audits=`-style snapshot reuse one place to live.

## Not findings

`_blockers` (`tests/test_decompose.py:755`) has no existing equivalent - the inline `{**BLOCKER_ISSUE, ...}` at 1780 is a differently-shaped one-off. The `KSTRL_PACKAGE` constant duplication in isolation is a repo-wide convention (`test_atomicio.py:48`, `test_toml_readers.py:111`), so finding 2 is about `package_sources`/`parsed`, not the constant. No existing generic "constant, not literal" AST guard exists anywhere in `tests/`, so the guard's concept is genuinely new; only its plumbing is duplicated.

</details>

<details>
<summary><b>Simplification</b> (10 findings)</summary>

**1. `kstrl/evolution.py:1499-1537` - the `audits=` parameter's default path is production-dead, and it forces a redundant filter plus an extra contract to defend.**
`_journal_snapshot` is the only production caller of `get_spec_issue_runs`, and it *always* passes `audits=`. So `source = self.get_spec_audits() if audits is None else audits` exists for tests only, and on the live path the event-type filter runs twice over the same rows (once in `get_spec_audits`, again in the comprehension). The parameter also invents a public promise ("passing raw entries answers the same as passing audits") that nothing needs, and that promise is what pays for the 6-line docstring paragraph and `test_a_supplied_snapshot_answers_the_same_as_raw_entries` (test_evolution.py:1175-1187).

Simpler form that keeps both constraints (journal owns read + window, one parse per report):

```python
def window_spec_audits(audits: list[dict[str, Any]], project: str, last_n: int) -> list[dict[str, Any]]:
    mine = [e for e in audits if entry_str(e, "project") == project]
    return mine[-last_n:] if last_n > 0 else []
```

module-level in `evolution.py` beside `entry_str`; `get_spec_issue_runs(project, last_n=10)` becomes `return window_spec_audits(self.get_spec_audits(), project, last_n)` and `_journal_snapshot` calls `window_spec_audits(audits, project_name, lookback)`. Cost removed: one optional param, one redundant `event_type` comparison per row, ~6 docstring lines, one test.

**2. `kstrl/decompose.py:1350-1360, 1367-1369` - `_journal_snapshot`'s docstring now restates two comments sitting on its own call site.**
The paragraph "Runs are matched by project NAME, not by spec path or content hash..." duplicates the comment at `decompose.py:1985-1989` ("the trend is keyed on the project name... the spec is edited every round, so a content hash never matches, and #260's own loop renamed the file too"), and "MUST be read BEFORE this run's own audit is appended" duplicates `decompose.py:1977-1979`. Both paragraphs came from the deleted `_spec_convergence`, which was a *different* function; folding them into `_journal_snapshot` put them one line from their own restatement. Delete both from the docstring (the call-site comments are the ones a reader hits first), or delete the comments. ~12 lines.

**3. `kstrl/decompose.py:1347`, `kstrl/evolution.py:1484`, `tests/test_decompose.py:1640`, `tests/test_journal_one_writer.py:449` - the "if the journal ever compacts, rotates or gains a second segment" argument is spelled four times.**
Three of the four are added or rewritten by this diff. It belongs once, on `get_spec_audits` (evolution.py:1484), which is the thing that would change. The other three can cite it in a clause ("see `get_spec_audits`"). Same pattern for "one read so the two cannot disagree because the file moved between two reads": five copies now (decompose.py:1315, 1339, 1982; evolution.py:1519; test_evolution.py:1159), two of them new.

**4. `kstrl/decompose.py:1127` - `_build_convergence` should take the deleted wrapper's name.**
`_spec_convergence` existed only to window-then-build; with it gone, `_build_convergence` is the only convergence function and the `_build_` prefix distinguishes it from nothing. Rename to `_spec_convergence` (the name is free, and it matches the `SpecConvergence` it returns). 11 call sites, all in tests plus one in `_decompose_spec_impl`.

**5. `tests/test_evolution.py:1122-1128` and `tests/test_decompose.py:1590-1601` - two new private `_journal` / `_audit` helpers that already exist in `tests/helpers/journal.py`.**
`journal_at(tmp_path)` and `audit(project)` are there, and `test_decompose.py` already imports `tear` from that module. The new `TestSpecAudits._journal` is byte-identical to `TestOneJournalRead._journal`; the two `_audit` bodies differ only by a timestamp field. `test_evolution.py:1037` has a *third* `_journal` in the class immediately above. Simpler: add `spec_file: str | None = None` to `helpers.journal.audit` and import `audit, journal_at` in both files. ~20 lines of test scaffolding removed, and the three journal builders stop drifting.

**6. `tests/test_evolution.py:1308-1316` - `_constant_strings` is one call site and collapses into its caller.**
Five module-level AST helpers is one more than the walk needs. Inline it as the `any(...)` it feeds:

```python
if any(isinstance(c, ast.Constant) and c.value == SPEC_ISSUES_EVENT for c in ast.walk(operand))
```

That also drops the `isinstance(child.value, str)` guard, which exists only to satisfy the `set[str]` return type. The other four (`spec_issues_literals`, `_literal_event_writes`, `_literal_event_reads`, `_reads_event_type`) each carry distinct logic and should stay.

**7. `kstrl/decompose.py:1333-1377` - the 3-tuple wants to be a frozen dataclass.**
Four call sites now index it positionally (`_journal_snapshot(journal, project)[0]` at test_decompose.py:1207, 1260, 1579; `[1]` at 1678), so which of two same-typed `list[dict[str, Any]]` you get is decided by a bare integer. `@dataclass(frozen=True) class AuditSnapshot: audits, window, lookback` matches the project's own convention ("`@dataclass` for data containers, `frozen=True` when immutable") and makes `snapshot.window` read at the call site. Low-cost, and the production unpack at decompose.py:1984 is unaffected.

**8. `kstrl/decompose.py:1984` - the local `history` names the window, four lines above `_excluded_history` names something else.**
`audits, history, lookback = _journal_snapshot(...)`, then `_excluded_history(audits, ...)`. The docstring one function up calls the same value "the window". Rename the local to `window`.

**9. `tests/test_decompose.py:2079-2097` duplicates `tests/test_evolution.py:1189-1203`.**
Both pin the slice direction, and both docstrings make the same "measured with the rule inverted, the suite stayed green" claim. Since #314 the direction can only be wrong in one place (`get_spec_issue_runs`), and the unit test covers it in three lines; the decompose-level copy pays four full `_run` decompose invocations plus the `_blockers` helper for the same rule. It does add one thing the unit test does not: that the render prints the window in order. If that is the point, say so and keep it; otherwise the unit test is sufficient.

**10. `tests/test_decompose.py:1677` - `journal.config.lookback_runs = lookback` mutates the config in a loop.**
Works because `EvolutionConfig` is not frozen, but it means each iteration inherits the previous one's state on a shared object. `self._journal(tmp_path, entries)` is already a per-iteration builder; passing the lookback through it (or constructing `EvolutionConfig(journal_path=..., lookback_runs=lookback)`) is the same length and does not reach into another object's field.

## Not flagged

- `SPEC_ISSUES_EVENT` moving to `evolution.py` and the top-level `EvolutionConfig`/`EvolutionJournal` import replacing the function-local one are both net simplifications. No cycle (`evolution.py` does not import `decompose`), and `ruff check kstrl tests` is clean.
- The AST guard's scan root is `kstrl/` only, so three literal `"spec_issues"` spellings survive in tests (`tests/helpers/journal.py:80` and `:107`, `tests/test_evolution.py:1044`). Widening the walk is more work, not less; but `helpers/journal.py` already imports from `kstrl.evolution`, so those two could use the constant for free.
- Docstring length on `get_spec_audits` and `_excluded_history` is in keeping with the file. Only the four- and five-way repeats in finding 3 are genuine bloat.

</details>

<details>
<summary><b>Efficiency</b> (no findings)</summary>

## EFFICIENCY review: no findings worth acting on

I measured all four items you named. Every one lands below the noise floor of the thing it sits inside. Details and numbers below.

### 1. Module-level `from kstrl.evolution import (...)` - `kstrl/decompose.py:40-45`

**Not worth flagging.** Your measurement reproduces, and the production number is smaller than the one you took.

Method: `git archive main kstrl` and `git archive HEAD kstrl` into two isolated trees, timed `import kstrl.cli` in interleaved subprocesses (main, head, main, head, ...) to cancel machine drift, 30 pairs.

| condition | main | HEAD | delta |
|---|---|---|---|
| `PYTHONDONTWRITEBYTECODE=1`, caches cleared (n=30) | 202.6 ms median | 206.9 ms | **+4.3 ms (+2.1%)** |
| `__pycache__` warm (n=25) | 81.3 ms median | 82.1 ms | **+0.8 ms (+1.0%)** |

The warm row is the one that matters: the real checkout has `kstrl/__pycache__` populated, so an actual `ks` invocation pays +0.8 ms, not +4.3 ms. `PYTHONDONTWRITEBYTECODE=1` measures the compile, which production does once and then never again.

What is actually pulled in, by `sys.modules` diff: exactly three entries - `kstrl.evolution`, `csv`, `_csv`. Nothing else. Split by cold-import timing after `kstrl.cli` is already loaded: `csv` 0.22 ms, the `evolution` module body 3.58 ms. `kstrl.evolution`'s only imports are `kstrl.observability` and `kstrl.verify`, both of which `cli.py` already imports at module level, so there is no transitive fan-out. Its module-level statements are a logger, four constants, five `re.compile` calls and two dicts - no I/O, no env reads.

### 2. The re-applied event-type filter - `kstrl/evolution.py:1530-1537`

**Not worth flagging, and the diff is net negative on work anyway.**

Built a synthetic journal at your stated 6.9 KB per factory run, one spec audit per run:

| journal | `get_spec_audits()` (read + parse) | window as shipped | window without the redundant filter | redundant filter costs |
|---|---|---|---|---|
| 1011 KB, 2650 entries, 150 audits | 2934 us | 7.8 us | 5.8 us | **2.0 us (0.07%)** |
| 10108 KB, 26353 entries, 1500 audits | 34279 us | 86.5 us | 72.0 us | **14.5 us (0.04%)** |

Two microseconds behind a three-millisecond read. The filter is load-bearing for the documented contract (`audits=` accepts raw entries and still answers correctly, pinned by `test_a_supplied_snapshot_answers_the_same_as_raw_entries`), so removing it would trade a correctness guarantee for 0.07%.

Worth recording the other direction: **the diff removes more work than it adds.** `main`'s `decompose.py` called `_spec_audits(entries)` twice over the *whole* entry list - once at line 1198 (`_spec_convergence`) and once at line 1400 (`_excluded_history`). HEAD filters once inside `get_spec_audits` and shares the result. One full-entries pass at 2650 entries measures 30.8 us, so the net is about **-29 us per report**.

### 3. `_read_all_entries` - one read per report, confirmed

`_read_all_entries` has three call sites in `evolution.py` (`get_spec_audits`, the repair-row count, `_read_journal_entries`); only `get_spec_audits` is on this path, `_journal_snapshot` calls it once, and the window reuses that list via `audits=`. `tests/test_decompose.py::TestOneJournalRead::test_the_journal_is_parsed_once_per_report` counts calls at `_read_all_entries` itself and asserts `== 1`; the class passes (4 tests, 0.13 s). No caching layer exists or is needed.

### 4. `TestSpecIssuesEventHasOneHome` - `tests/test_evolution.py:1119` - **does not need the `parsed()` cache**

Measured: **0.21 s** for the class alone, **0.27 s** inside a full `test_evolution.py` run (of 0.45 s total). Breakdown over 127 modules / 65,797 lines / 237,085 nodes:

```
read only            1.7 ms
parse only         186.1 ms
walk pre-parsed     60.8 ms
guard as written   187.3 ms
```

The guard makes **exactly one** parse pass and **one** `ast.walk`, with both the `ast.Dict` and `ast.Compare` checks inside the same walk. `test_journal_one_writer`'s `_PARSED` cache exists because *three* walkers in that one file each parsed the package for themselves (0.20 s of its 0.71 s). A cache in front of a single-pass walker saves nothing - it is a dict lookup added to a path with no second lookup. The nested `_constant_strings` walk is gated behind `_reads_event_type`, so it only fires on comparisons that already mention `event_type`.

Scale: full suite is 294.33 s (4801 passed). This test is 0.09% of it, and it joins 13 other test files in `tests/` that already each make their own full-package `rglob("*.py")` pass. Sharing a parse across those files would save roughly 190 ms per file per suite run, but that is a cross-file coupling argument about an existing convention, not something this diff introduced.

### Also checked, nothing found

- No long-lived closures over an enclosing scope in production code. The two new closures (`counting`, `refuse`) are `monkeypatch.setattr` targets in tests and are reverted per-test.
- `entry_str` is imported by name into `decompose`, so the cross-module move costs no attribute lookup at call time.
- `_excluded_history` walks its `audits` list three times (two `sum()` genexps plus `_excluded_projects`) - about 4 passes over 150 items, single-digit microseconds, and identical in shape to `main`. The diff did not introduce it.

### One note on the target

`kstrl/decompose.py` has uncommitted working-tree changes from a sibling agent (`_journal_snapshot`'s 3-tuple has become a frozen `AuditSnapshot` dataclass). That is an altitude/clarity change; it does not alter any measurement above - same single read, same `audits=` reuse.

</details>

<details>
<summary><b>Altitude</b> (8 findings)</summary>

**1. `kstrl/evolution.py:46-52` - the constant landed in the right module; the same module's worst duplication of the same pattern is untouched, and the new guard cannot see it.**

The rationale written into the new constant is "a reader added HERE reaches for the nearest spelling, which was the literal." That argument is 9x truer for `component_result`, which is spelled as a bare literal nine times across five lines in this same file: `730` (write), `938`, `960`, `1331`, `1385` (each `entry.get("event_type", "component_result") == "component_result"`). The `.get` default there also encodes a schema rule ("a row with no `event_type` is a component result") in a string literal repeated four times.

Verdict on the brief's question: fixing one event type is the **right scope for issue 314** (which names one), and the other four are not the same defect - `role_usage` (evolution.py:613), `contract_result` (factory.py:3834), `autonomy_transition` (autonomy.py:787) and `findings_superseded` (pipeline.py:1256) are each written once with no production reader spelling the name back, so they never had the two-modules-two-spellings cost item 3 fixed. `component_result` is different: it is nine spellings in the module this PR is editing.

Cost: the PR establishes "journal event names are constants at the top of evolution.py" and leaves the file's own dominant counter-example in place, so the convention reads as an exception rather than a rule.
**Follow-up** for the `component_result` conversion (it touches five aggregates and their tests). But see 2 - the guard should be generalised in this PR so the follow-up is enforceable.

**2. `tests/test_evolution.py:1241-1345` - a 105-line AST detector hardcoded to one event name. Generalise it here; it is nearly free and green today.**

`spec_issues_literals`, `_literal_event_writes`, `_literal_event_reads` all close over `SPEC_ISSUES_EVENT` directly. The next event constant needs a copy of all of it, and the realistic outcome of "copy 105 lines" is "no guard."

The generalisation is one parameter: `spec_issues_literals(source, event_name)` driven from a set of enrolled constants. It passes today for `{SPEC_ISSUES_EVENT, JOURNAL_REPAIR_EVENT}` - I checked, `journal_repair` has zero literal spellings in `kstrl/` (only its definition at evolution.py:45 and the constant use at 1467). Enrolling the remaining four then becomes a one-line diff per event as each gets a constant, which is what makes finding 1's follow-up a real commitment rather than an intention.

One implementation note if you do it: for `entry.get("event_type", "component_result") == "component_result"`, `_constant_strings` walks into the getter call and will find the name in the `.get` default too, so the reported message names the whole call. Still a true positive, just an odd string; skip `ast.Call` defaults or accept the wording.
**Fix in this PR** (parameterise + enroll the two that pass). Conversions in the follow-up.

**3. `kstrl/evolution.py:1499-1541` - the `audits=` parameter is a special case layered onto the shared reader, exactly the shape this angle is looking for.**

`get_spec_issue_runs` now has two contracts: read the file, or window a list the caller already has. The optional-source branch forces a defensive re-application of the event-type filter (line 1533), and that in turn needed its own test to state the equivalence (`test_a_supplied_snapshot_answers_the_same_as_raw_entries`). The parameter is also misnamed relative to what it accepts - that test proves it takes raw entries, not audits.

The underlying mechanism generalises cleanly instead: a module-level pure function

```python
def spec_audit_window(audits: list[dict[str, Any]], project: str, last_n: int) -> list[dict[str, Any]]
```

called by `get_spec_issue_runs` (over `self.get_spec_audits()`) and by `decompose._journal_snapshot` (over the snapshot it already holds). One copy of the window rule - which is what #314 item 2 asked for - with no optional-source branch, no re-filter, and the equivalence test disappears because there is no second path to be equivalent to.

Cost of leaving it: a public method with a performance escape hatch that changes its data source, plus a test whose only job is to say the escape hatch behaves.
**Fix in this PR.**

**4. `kstrl/evolution.py:494-514` - `entry_str` is in the right module for the shared *rule* and the wrong module for the *function*.**

Nine of its ten call sites are in `decompose.py`; the single evolution-side use is the project match at line 1535. The docstring opens by defining it over "a JSON-decoded journal record" and then has to append a paragraph excusing the nested-issue case, which is the docstring paying for the placement.

Two separable things are fused here. The shared rule is "how a project is matched on an audit row" - that genuinely belongs on evolution, as `audit_project(entry) -> str`, and stating it that way removes the need for `decompose` to import a string accessor at all for that purpose. The generic accessor is a JSON util with no journal content; `kstrl/events.py` already has the identical idiom inlined four times (738, 739, 752, 767), which is where the real duplication is.

**Follow-up** for the shared util. The `audit_project` framing is small enough for this PR if you want the docstring exception gone.

**5. `kstrl/evolution.py:1499` - `get_spec_issue_runs` is named for its history; `get_spec_audits` is named for its layer.**

The sibling settles the vocabulary: the row is a spec audit. `get_spec_issue_runs` then spends two sentences of its own docstring correcting its name - "The last N recorded spec audits for `project`" and "`last_n` counts spec audits, not factory runs". A name that needs its docstring to disown it is the tell.

There is exactly one production caller (`decompose._journal_snapshot`) plus tests, so the rename is cheap now and gets more expensive with every future caller. `get_recent_spec_audits(project, last_n)` matches the sibling; or, under finding 3, the method collapses into `spec_audit_window` and the name question goes away.
**Fix in this PR.**

**6. `kstrl/decompose.py:1330, 1372-1376` - a bare 3-tuple whose first two elements are the same type.**

`_journal_snapshot` returns `tuple[list[dict[str, Any]], list[dict[str, Any]], int]`, unpacked positionally at the call site and indexed as `[0]` / `[1]` in four tests. Swapping the first two typechecks, and the trend/accounting split is exactly the pair #280 exists to keep straight. The project convention (CLAUDE.md: `@dataclass`, `frozen=True` when immutable) points at a three-field frozen container; it also gives the "all three or none" fact a name instead of a docstring paragraph.
**Fix in this PR** (small).

**7. `kstrl/evolution.py:1474` - the depth the issue was pointing at: the journal now owns *which rows are audits*, but not *what a row contains*.**

`get_spec_audits` returns `list[dict[str, Any]]`, and `decompose` still reads `"project"`, `"spec_file"`, `"timestamp"` and the nested `"severity"`/`"kind"`/`"summary"` by literal key across `_excluded_projects`, `_excluded_history`, `_stored_issues` and `_build_convergence`. That is the reason `entry_str` had to be exported to a second module at all (finding 4), and the reason the window rule needed a snapshot-passing escape hatch (finding 3).

A frozen `SpecAudit` returned by `get_spec_audits` would put the row's field names in the module that defines the schema, delete `entry_str` from `decompose` entirely, and make findings 4 and 5 moot. Today a journal schema change turns those fields into `""` silently in the convergence report.
**Follow-up** - it touches every test that builds raw audit dicts, and racing evolution.py again is what #314 exists to avoid. Worth filing with the reasoning above, because it is the fix the three smaller ones are approximating.

**8. `tests/test_journal_one_writer.py:445-451` - the inventory is the right mechanism. No positive rule is available. Minor nit on the tombstone.**

Dropping the `decompose.py` row is correct, and an inventory-of-expected-sites is the right shape here, not a bandaid: a positive rule "nothing outside evolution.py holds the journal path" cannot be written, because `cli.py:689` legitimately interpolates `journal.config.journal_path` into an operator message, and the `pipeline.py`/`workqueue.py` rows are a different file that the file's own docstring says cannot be told apart without the type resolution #324 is about. The inventory's stated design ("adding a row is not forbidden, it is the point") is the honest version of that.

Nit: the six-line tombstone comment describes an invariant the inventory does not enforce - holding the path is still allowed, as the `cli.py` row proves; what it enforces is that changing the set requires editing the dict and explaining why. And tombstones for removed rows accumulate. One line pointing at `get_spec_audits`, with the rationale living in that method's docstring (where it already is, verbatim), would carry the same information.
**Optional, this PR.**

</details>

### What was applied

| Finding | Action |
|---|---|
| Reuse 2, 3, 4 / Simplification 6 / Altitude 2 | Applied. The guard reuses `package_sources`, `parsed`, `label` and `folded_str` from `test_journal_one_writer`, is driven by an `ENROLLED_EVENT_CONSTANTS` table, and gained a positive control for the concatenated and f-string spellings plus one for the second enrolled name. `_constant_strings` is gone. |
| Reuse 5, 6 / Simplification 5 | Applied. `helpers.journal.audit` grows a `spec_file` argument and uses `SPEC_ISSUES_EVENT`; `audits_in` uses the constant. The new tests use `audit` and `journal_at`. |
| Simplification 7 / Altitude 6 | Applied. Frozen `AuditSnapshot`. |
| Simplification 8 | Applied, as `snapshot.window` at the call site. |
| Simplification 10 | Applied. A journal per lookback, and its own comment on why: appending to one file three times would compare a growing history against itself. |
| Simplification 2, 3 / Altitude 8 | Applied. The repeated arguments live on `get_spec_audits`; the docstring, the test and the tombstone cite it. |

### What was declined, and why

| Finding | Reason |
|---|---|
| **Simplification 1 / Altitude 3** - replace `audits=` with a module-level `window_spec_audits` | This is the one finding I am overruling on purpose. Both agents are right that the parameter is a second contract. But the shape they propose puts `get_spec_issue_runs` back to having **no production caller**, which is precisely the state issue #314 calls out ("a public method kept alive by its own tests, which is worse than either owning the rule or deleting it"). The issue's own second option is "make it the one place the window lives, called with pre-read entries", and `audits=` is literally that. The measured cost of the redundant filter is 2.0 us behind a 2934 us read (0.07%), and the filter is what makes the parameter safe against raw entries. |
| **Altitude 5** - rename `get_spec_issue_runs` | Declined for this PR. The issue and the decision to route it back both name this method; renaming it in the same commit that changes its signature makes the audit trail across #260, #280 and #314 harder to follow for a reader who has only the names. Worth a follow-up once it has settled. |
| **Simplification 4** - rename `_build_convergence` to `_spec_convergence` | Declined. Reusing a just-deleted function's name for a function with a different signature makes `git log -L` across #260 and #280 read as if one function changed shape. The clarity gain is small; the churn is 11 call sites. |
| **Reuse 1 / Altitude 4** - fold `entry_str` with `workqueue._as_str`, or split it into `audit_project` | Declined for this PR. The fold is a real duplication and belongs in `observability.py`, but it touches `workqueue.py`, `events.py`, `serve.py` and two TUI screens, well outside this seam. The `audit_project` split is worse than what shipped: `decompose` still needs the generic accessor for `spec_file`, `timestamp` and the issue fields, so it would get a private `_entry_str` back and the rule would live in two homes again. Follow-up. |
| **Altitude 1** - convert `component_result` and the other four event names | Declined for this PR, out of #314's scope. Enrolling `JOURNAL_REPAIR_EVENT` in the guard (finding 2, applied) is what makes the follow-up enforceable rather than intended: each conversion is now one line in `ENROLLED_EVENT_CONSTANTS`. |
| **Altitude 7** - a frozen `SpecAudit` row type | Declined for this PR. It touches every test that builds a raw audit dict, and racing `evolution.py` again is what #314 exists to avoid. This is the fix the smaller ones approximate, and it is worth filing. |
| **Reuse 7** - a private `_entries_of_type` on `EvolutionJournal` | Declined. Two sites, and the second is `get_repair_count`, which #327 landed this morning; the churn on another PR's fresh code costs more review than the duplication costs. |
| **Simplification 9** - drop the decompose-level direction test | Declined. It is not a duplicate: the unit test pins what the rule does, the end-to-end test pins that `decompose` **routes through** that rule and renders the window in order. Mutation 3 fails both, which is the point - if only the unit test existed, a decompose that quietly rewindowed for itself would still pass. |

## Known limits of the guards, stated rather than implied

- The inventory in `test_journal_one_writer.py` is what stops `decompose` reaching for the path again. It does not stop a NEW second copy of the window rule inside `_journal_snapshot` that happens to agree with the journal's; `test_the_window_is_the_journals_own_over_the_journals_own_lookback` is an agreement test and can only fail on a rule that differs. Code review is the mechanism for that one.
- The event-name walk covers `kstrl/` only. `tests/` is not walked; the two literals that were there are converted, but a new one in a test would not fail.
- Folding decides `"spec" + "_issues"` and the plain f-string. It does not decide `"".join(...)`, `%`-formatting, or a name resolved at run time, exactly as `folded_str` discloses.

https://claude.ai/code/session_01DHYpsJ8ecHLh5HdrswJtJK

### DIFF
--- a/kstrl/decompose.py
+++ b/kstrl/decompose.py
@@ -1330,34 +1330,39 @@ def _excluded_history(
     )
 
 
-def _journal_snapshot(
-    journal: EvolutionJournal | None,
-    project_name: str,
-) -> tuple[list[dict[str, Any]], list[dict[str, Any]], int]:
+@dataclass(frozen=True)
+class AuditSnapshot:
+    """One read of the recorded spec-audit history, in three views.
+
+    A container rather than a tuple because ``audits`` and ``window``
+    are both ``list[dict[str, Any]]``: swapping them typechecks, and
+    they are exactly the pair #280 exists to keep straight - the trend
+    reads the window, the accounting under it reads everything the
+    window left out.
+    """
+
+    #: Every spec audit the journal holds, across every project.
+    audits: list[dict[str, Any]]
+    #: This project's last ``lookback`` audits: what the trend scores.
+    window: list[dict[str, Any]]
+    #: How far back the window reaches, carried so the render can tell
+    #: a windowed-out audit from one the trend could not score.
+    lookback: int
+
+
+def _journal_snapshot(journal: EvolutionJournal | None, project_name: str) -> AuditSnapshot:
     """The recorded audit history behind one convergence report.
 
-    Three things from ONE read: every spec audit the journal holds,
-    this project's window of them, and how far back that window
-    reaches. Together because a journal that is absent has none of the
-    three, which keeps that single fact here instead of three
+    All three views from ONE read, because a journal that is absent has
+    none of the three: that keeps the fact here instead of three
     conditionals at the call site.
 
     Read through ``EvolutionJournal`` rather than through its storage
-    path (#314). The journal's own reader is the only one that would
-    survive the file compacting, rotating or gaining a second segment;
-    a caller holding the path would quietly return less than the
-    journal holds while the trend kept printing, and silent loss of
-    the excluded-history accounting is the defect #280 exists to fix.
-
-    The window comes from ``get_spec_issue_runs`` rather than from a
-    second copy of its rule here, so the trend and the accounting
-    printed under it cannot disagree about which audits belong to this
-    project. Runs are matched by project NAME, not by spec path or
-    content hash: the spec is edited between every round by
-    construction (so a content hash never matches), and the recorded
-    loop in #260 renamed the file mid-loop (so the path does not
-    either). A previous audit of a different file is still reported,
-    with the file names named.
+    path, and windowed by ``get_spec_issue_runs`` rather than by a
+    second copy of its rule here (#314). Why each matters is on those
+    two methods; the short version is that a caller holding the path
+    survives no change of storage, and a second copy of the window rule
+    can drift from the one the accounting uses.
 
     MUST be read BEFORE this run's own audit is appended to the
     journal, or the "previous run" the trend compares against is this
@@ -1369,11 +1374,14 @@ def _journal_snapshot(
     78 MB, against an architect call measured at 119 to 210 seconds.
     """
     if journal is None:
-        return [], [], 0
+        return AuditSnapshot(audits=[], window=[], lookback=0)
     audits = journal.get_spec_audits()
     lookback = journal.config.lookback_runs
-    window = journal.get_spec_issue_runs(project_name, lookback, audits=audits)
-    return audits, window, lookback
+    return AuditSnapshot(
+        audits=audits,
+        window=journal.get_spec_issue_runs(project_name, lookback, audits=audits),
+        lookback=lookback,
+    )
 
 
 def _surface_convergence(
@@ -1981,15 +1989,15 @@ def _decompose_spec_impl(
     # ONE read feeds both the trend and the accounting under it, so the
     # two are computed over the same snapshot and cannot disagree
     # because the file changed between two reads.
-    audits, history, lookback = _journal_snapshot(journal, project_name)
+    snapshot = _journal_snapshot(journal, project_name)
     # #280: the trend is keyed on the project name, so a rename starts a
     # fresh one and the runs before it drop out of view. Keying
     # differently would be worse (the spec is edited every round, so a
     # content hash never matches, and #260's own loop renamed the file
     # too), so what is fixed is the silence rather than the key.
     _surface_convergence(
-        _build_convergence(spec_issues, spec_path.name, history),
-        _excluded_history(audits, project_name, spec_path.name, lookback),
+        _build_convergence(spec_issues, spec_path.name, snapshot.window),
+        _excluded_history(snapshot.audits, project_name, spec_path.name, snapshot.lookback),
         project_name,
         ui,
     )
--- a/kstrl/decompose.py
+++ b/kstrl/decompose.py
@@ -1339,6 +1339,24 @@ class AuditSnapshot:
     they are exactly the pair #280 exists to keep straight - the trend
     reads the window, the accounting under it reads everything the
     window left out.
+
+    ``frozen=True`` for the rebinding guarantee, not to make this a
+    value that can key a dict. Because ``eq`` is on too, the decorator
+    would otherwise generate a ``__hash__`` that hashes the two lists
+    and raises ``TypeError: unhashable type: 'list'`` from inside a
+    method nobody wrote, so the class advertises a capability it does
+    not have. Tuples would not fix it: the elements are ``dict``, and
+    hashing a tuple hashes its elements, measured. ``__hash__ = None``
+    is the honest answer: set in the body, it is an explicit hash the
+    decorator leaves alone. It says unhashable, names THIS class when
+    somebody tries, and leaves ``==`` and the frozen fields untouched.
+
+    Not generalised, deliberately. Measured: nine other frozen
+    dataclasses in ``kstrl/`` declare a top-level ``list``/``dict``/``set``
+    field and still advertise a hash that raises. They are left alone
+    here because this PR introduced ``AuditSnapshot`` and that is the one
+    it owns; a sweep of the other nine wants its own change with a guard
+    behind it, rather than nine edits nothing keeps true.
     """
 
     #: Every spec audit the journal holds, across every project.
@@ -1348,6 +1366,8 @@ class AuditSnapshot:
     #: How far back the window reaches, carried so the render can tell
     #: a windowed-out audit from one the trend could not score.
     lookback: int
+
+    __hash__ = None  # type: ignore[assignment]
 
 
 def _journal_snapshot(journal: EvolutionJournal | None, project_name: str) -> AuditSnapshot:
--- a/kstrl/decompose.py
+++ b/kstrl/decompose.py
@@ -37,6 +37,12 @@ from kstrl.events import (
     RunStarted,
     SpecIssueRecorded,
 )
+from kstrl.evolution import (
+    SPEC_ISSUES_EVENT,
+    EvolutionConfig,
+    EvolutionJournal,
+    entry_str,
+)
 from kstrl.guards import ScopeHazard, scope_entry_hazard
 from kstrl.linear import (
     LinearClient,
@@ -50,7 +56,6 @@ from kstrl.manifest import (
     Manifest,
 )
 from kstrl.names import validate_branch_name, validate_component_id
-from kstrl.observability import read_progress_events
 from kstrl.prd import PRD
 
 logger = logging.getLogger(__name__)
@@ -62,7 +67,6 @@ SPEC_ISSUE_APPLIES_COMPONENT = "component"
 SPEC_ISSUE_APPLIES_SPEC = "spec"
 
 if TYPE_CHECKING:
-    from kstrl.evolution import EvolutionJournal
     from kstrl.ui.base import UI
 
 
@@ -796,20 +800,6 @@ def _surface_spec_issues(issues: list[SpecIssue], ui: UI) -> None:
 # next to manifest.json so one directory holds the decompose outputs.
 SPEC_ISSUES_REL_PATH = Path("scripts") / "kstrl" / "spec-issues.json"
 
-# The journal's discriminator for one recorded spec audit. Named because
-# this module both writes it and reads it back, and the two must agree.
-#
-# It is NOT the only copy: ``EvolutionJournal.get_spec_issue_runs``
-# hardcodes the same literal, and that file is under concurrent edit on
-# another branch, so it is not converted here. The constant belongs on
-# the storage layer rather than on this writer; filed as #314. What
-# keeps the two from drifting meanwhile is a mechanism rather than this
-# comment - ``test_the_journal_reader_agrees_with_the_event_name_written``
-# writes an entry through this module and reads it back through the
-# journal, so changing this name alone fails that test and fourteen
-# others.
-SPEC_ISSUES_EVENT = "spec_issues"
-
 
 def _issue_dict(issue: SpecIssue) -> dict[str, str]:
     """One issue as the five JSON keys every artifact writes it under."""
@@ -1010,8 +1000,6 @@ def _spec_audit_journal(root_dir: Path, ui: UI) -> EvolutionJournal | None:
     Which exceptions that covers is ``load_or_none``'s to know, not
     this call site's.
     """
-    from kstrl.evolution import EvolutionConfig, EvolutionJournal
-
     config = EvolutionConfig.load_or_none(root_dir, warn=ui.warn)
     if config is None or not config.enabled:
         return None
@@ -1032,7 +1020,7 @@ def _record_spec_issues_event(
     but the failure is surfaced as a warning rather than swallowed:
     the journal is an audit trail, so a silent skip would defeat it.
     No ``run_id`` field: decompose runs before a factory run id exists,
-    which is why ``EvolutionJournal.get_spec_issue_runs`` reads these
+    which is why ``EvolutionJournal.get_spec_audits`` reads these
     entries rather than the run-windowed reader.
     """
     if journal is None:
@@ -1124,9 +1112,9 @@ def _stored_issues(entry: dict[str, Any]) -> list[SpecIssue] | None:
         return None
     stored = [
         SpecIssue(
-            severity=_entry_str(i, "severity"),
-            kind=_entry_str(i, "kind"),
-            summary=_entry_str(i, "summary"),
+            severity=entry_str(i, "severity"),
+            kind=entry_str(i, "kind"),
+            summary=entry_str(i, "summary"),
         )
         for i in raw
         if isinstance(i, dict)
@@ -1151,7 +1139,7 @@ def _build_convergence(
     for entry in history:
         stored = _stored_issues(entry)
         if stored is not None:
-            audits.append((_entry_str(entry, "spec_file"), stored))
+            audits.append((entry_str(entry, "spec_file"), stored))
             trend.append(sum(1 for i in stored if i.severity == "blocker"))
     if not audits:
         return None
@@ -1166,36 +1154,6 @@ def _build_convergence(
         previous_spec_file=previous_spec_file,
         repeated=sum(1 for i in previous_issues if _issue_identity(i) in current_ids),
         blocker_trend=(*trend, current_counts["blocker"]),
-    )
-
-
-def _spec_convergence(
-    issues: list[SpecIssue],
-    entries: list[dict[str, Any]],
-    project_name: str,
-    spec_file: str,
-    lookback: int,
-) -> SpecConvergence | None:
-    """Compare this run against this project's recorded audit history.
-
-    Runs are matched by project name, not by spec path or content
-    hash: the spec is edited between every round by construction (so a
-    content hash never matches), and the recorded loop in #260 renamed
-    the file mid-loop (so the path does not either). A previous audit
-    of a different file is still reported, with the file names named.
-
-    ``lookback`` is how far back the trend reaches: the journal's own
-    knob rather than a number invented here, read as "the last N spec
-    audits" - decompose writes one audit per run, with or without a
-    factory run behind it.
-
-    MUST be called on entries read BEFORE this run's own is appended to
-    the journal, or the "previous run" it compares against is this one.
-    """
-    return _build_convergence(
-        issues,
-        spec_file,
-        _windowed_audits(_spec_audits(entries), project_name, lookback),
     )
 
 
@@ -1293,48 +1251,39 @@ class ExcludedHistory:
         return max(0, self.own_recorded - counted - self.unreadable(counted))
 
 
-def _entry_str(entry: dict[str, Any], key: str) -> str:
-    """A string field of a journal entry, or "" for anything else.
-
-    NOT ``str(entry.get(key, ""))``: that renders a JSON ``null`` as
-    the literal ``"None"``, which round 1 of review reproduced as a
-    phantom project named 'None' passing the emptiness guard below and
-    a spec file printed as ``None``. A null field is an absent field.
-    """
-    value = entry.get(key)
-    return value if isinstance(value, str) else ""
-
-
 def _excluded_projects(
-    entries: list[dict[str, Any]],
+    audits: list[dict[str, Any]],
     project_name: str,
     spec_file: str,
 ) -> tuple[ExcludedProject, ...]:
-    """Spec audits in ``entries`` recorded under some other project.
+    """Recorded spec audits under some project OTHER than this one.
+
+    Takes audits, not raw journal entries: which rows are spec audits
+    is ``EvolutionJournal.get_spec_audits``'s to decide (#314), and a
+    second copy of that rule here is how the accounting and the trend
+    could come to disagree about the same journal.
 
     Ordered so the display cap drops only the weakest evidence: a
     project that audited the file this run audited sorts first, then
     the rest by how much history they hold. The cap itself never drops
     a spec-file match; see ``_excluded_line``.
 
-    Entries with no project name are skipped rather than grouped under
+    Audits with no project name are skipped rather than grouped under
     "": an unnamed project is not somewhere the operator can go and
     look, so pointing at it is not evidence.
     """
     by_project: dict[str, list[str]] = {}
     last_seen: dict[str, str] = {}
-    for entry in entries:
-        if entry.get("event_type") != SPEC_ISSUES_EVENT:
-            continue
-        project = _entry_str(entry, "project")
+    for entry in audits:
+        project = entry_str(entry, "project")
         if not project or project == project_name:
             continue
-        by_project.setdefault(project, []).append(_entry_str(entry, "spec_file"))
+        by_project.setdefault(project, []).append(entry_str(entry, "spec_file"))
         # Only a timestamp that exists replaces one that exists. Round 2
         # of review: assigning unconditionally let one trailing entry
         # with no timestamp erase a good date every earlier entry for
         # that project carried, losing evidence to a single bad row.
-        if timestamp := _entry_str(entry, "timestamp"):
+        if timestamp := entry_str(entry, "timestamp"):
             last_seen[project] = timestamp
     excluded = [
         ExcludedProject(
@@ -1349,34 +1298,8 @@ def _excluded_projects(
     return tuple(sorted(excluded, key=lambda e: (not e.read_this_spec, -e.audits, e.project)))
 
 
-def _spec_audits(entries: list[dict[str, Any]]) -> list[dict[str, Any]]:
-    """Just the recorded spec audits, in file order."""
-    return [e for e in entries if e.get("event_type") == SPEC_ISSUES_EVENT]
-
-
-def _windowed_audits(
-    audits: list[dict[str, Any]],
-    project_name: str,
-    last_n: int,
-) -> list[dict[str, Any]]:
-    """The trend's history: the last ``last_n`` audits of one project.
-
-    The same rule as ``EvolutionJournal.get_spec_issue_runs``, applied
-    to entries already in memory so one read serves both the trend and
-    the accounting below it. Round 2 of review: the call site has both
-    results in scope, so the second parse of the same file bought
-    nothing. ``test_the_windowing_rule_matches_the_journals_own``
-    pins this against the journal's copy, because a duplicated rule
-    that drifts would make the trend and the accounting disagree.
-    Folding both into the journal is filed as #314, and doing so leaves
-    ``get_spec_issue_runs`` with no production caller today.
-    """
-    mine = [e for e in audits if _entry_str(e, "project") == project_name]
-    return mine[-last_n:] if last_n > 0 else []
-
-
 def _excluded_history(
-    entries: list[dict[str, Any]],
+    audits: list[dict[str, Any]],
     project_name: str,
     spec_file: str,
     lookback: int,
@@ -1386,10 +1309,12 @@ def _excluded_history(
     Empty for a first audit in a fresh repo, so the lines it feeds
     never fire on the common case.
 
-    Takes ``entries`` rather than reading, so the caller reads once and
-    the trend and the accounting are computed over the same snapshot.
-    That also removes any chance of the two disagreeing because the
-    file changed between two reads.
+    Takes the audits rather than reading them, so the caller reads once
+    and the trend and the accounting are computed over the same
+    snapshot. That also removes any chance of the two disagreeing
+    because the file changed between two reads. Selecting the audits
+    out of the journal's entries is the journal's own job (#314), not a
+    rule restated here.
 
     Deliberately NOT windowed by ``lookback_runs``; ``lookback`` is
     carried only so the render can tell a windowed-out audit from one
@@ -1397,40 +1322,46 @@ def _excluded_history(
     excludes that was itself windowed would omit history silently,
     which is the bug this exists to fix.
     """
-    audits = _spec_audits(entries)
     return ExcludedHistory(
-        own_recorded=sum(1 for e in audits if _entry_str(e, "project") == project_name),
+        own_recorded=sum(1 for e in audits if entry_str(e, "project") == project_name),
         projects=_excluded_projects(audits, project_name, spec_file),
-        unattributed=sum(1 for e in audits if not _entry_str(e, "project")),
+        unattributed=sum(1 for e in audits if not entry_str(e, "project")),
         lookback=lookback,
     )
 
 
-def _journal_snapshot(journal: EvolutionJournal | None) -> tuple[list[dict[str, Any]], int]:
-    """Every entry in the journal, and how far back the trend may reach.
+def _journal_snapshot(
+    journal: EvolutionJournal | None,
+    project_name: str,
+) -> tuple[list[dict[str, Any]], list[dict[str, Any]], int]:
+    """The recorded audit history behind one convergence report.
 
-    Both together because both callers need both, and because a
-    journal that is absent has no entries AND no window: returning
-    the pair keeps that single fact in one place instead of a
-    conditional at the call site.
+    Three things from ONE read: every spec audit the journal holds,
+    this project's window of them, and how far back that window
+    reaches. Together because a journal that is absent has none of the
+    three, which keeps that single fact here instead of three
+    conditionals at the call site.
 
-    Read through ``read_progress_events`` rather than through
-    ``EvolutionJournal``, because no public reader on it returns the
-    whole entry set: ``get_spec_issue_runs`` filters to one project
-    inside the reader by design.
+    Read through ``EvolutionJournal`` rather than through its storage
+    path (#314). The journal's own reader is the only one that would
+    survive the file compacting, rotating or gaining a second segment;
+    a caller holding the path would quietly return less than the
+    journal holds while the trend kept printing, and silent loss of
+    the excluded-history accounting is the defect #280 exists to fix.
 
-    That is a layering compromise and not a free one. It reaches past
-    ``EvolutionJournal`` to that journal's own storage path, so a
-    journal that ever compacts, rotates or gains a second segment would
-    leave this reading less than the journal holds while the journal's
-    own reader kept working. Silent loss of the accounting is precisely
-    what #280 exists to fix, so the risk is pinned rather than trusted:
-    ``test_the_windowing_rule_matches_the_journals_own`` fails if the
-    two readers ever see different entries. A reader on
-    ``EvolutionJournal`` is the right home and would delete this
-    function; it is not added here because that file is under
-    concurrent edit on another branch. Filed as #314 rather than
-    raced.
+    The window comes from ``get_spec_issue_runs`` rather than from a
+    second copy of its rule here, so the trend and the accounting
+    printed under it cannot disagree about which audits belong to this
+    project. Runs are matched by project NAME, not by spec path or
+    content hash: the spec is edited between every round by
+    construction (so a content hash never matches), and the recorded
+    loop in #260 renamed the file mid-loop (so the path does not
+    either). A previous audit of a different file is still reported,
+    with the file names named.
+
+    MUST be read BEFORE this run's own audit is appended to the
+    journal, or the "previous run" the trend compares against is this
+    one.
 
     One read per decompose, and the whole file: measured at 6.9 KB per
     factory run in this repo, so 1 MB is about 150 runs and 10 MB about
@@ -1438,8 +1369,11 @@ def _journal_snapshot(journal: EvolutionJournal | None) -> tuple[list[dict[str, 
     78 MB, against an architect call measured at 119 to 210 seconds.
     """
     if journal is None:
-        return [], 0
-    return read_progress_events(journal.config.journal_path), journal.config.lookback_runs
+        return [], [], 0
+    audits = journal.get_spec_audits()
+    lookback = journal.config.lookback_runs
+    window = journal.get_spec_issue_runs(project_name, lookback, audits=audits)
+    return audits, window, lookback
 
 
 def _surface_convergence(
@@ -2047,15 +1981,15 @@ def _decompose_spec_impl(
     # ONE read feeds both the trend and the accounting under it, so the
     # two are computed over the same snapshot and cannot disagree
     # because the file changed between two reads.
-    entries, lookback = _journal_snapshot(journal)
+    audits, history, lookback = _journal_snapshot(journal, project_name)
     # #280: the trend is keyed on the project name, so a rename starts a
     # fresh one and the runs before it drop out of view. Keying
     # differently would be worse (the spec is edited every round, so a
     # content hash never matches, and #260's own loop renamed the file
     # too), so what is fixed is the silence rather than the key.
     _surface_convergence(
-        _spec_convergence(spec_issues, entries, project_name, spec_path.name, lookback),
-        _excluded_history(entries, project_name, spec_path.name, lookback),
+        _build_convergence(spec_issues, spec_path.name, history),
+        _excluded_history(audits, project_name, spec_path.name, lookback),
         project_name,
         ui,
     )
--- a/kstrl/evolution.py
+++ b/kstrl/evolution.py
@@ -44,6 +44,15 @@ JOURNAL_SCHEMA_VERSION = 2
 # grepped: it is the only durable trace that a crash tore the file.
 JOURNAL_REPAIR_EVENT = "journal_repair"
 
+# #260: the event_type of one recorded spec audit. ``decompose`` writes
+# these rows and :meth:`EvolutionJournal.get_spec_audits` selects on
+# them, so the name belongs on the layer that defines the journal's
+# schema rather than on the writer (#314). It lived in ``decompose``
+# until then, with this module holding a second copy as a literal, and
+# the cost of that placement is the reason it moved: a reader added
+# HERE reaches for the nearest spelling, which was the literal.
+SPEC_ISSUES_EVENT = "spec_issues"
+
 
 # The header row record_run writes to experiments.tsv, at module scope so
 # that a test can assert against the columns the writer actually emits
@@ -480,6 +489,29 @@ def signatures_from_findings(phase: str, findings: Iterable[Finding]) -> list[st
 
 def _timestamp_now() -> str:
     return datetime.now(UTC).strftime("%Y-%m-%dT%H:%M:%SZ")
+
+
+def entry_str(entry: dict[str, Any], key: str) -> str:
+    """One string field of a JSON-decoded journal record, "" when absent.
+
+    A null or non-string field is an ABSENT field, not a value to be
+    stringified. ``str(None)`` renders the literal "None", which #280
+    round 1 reproduced as a phantom project named 'None' in the
+    convergence report and a spec file printed as ``None``. Nothing is
+    assumed about a record beyond it being a JSON object, so a journal
+    written by an older version, or edited by hand, still reads.
+
+    Lives here rather than in ``decompose`` (#314) because the window
+    in :meth:`EvolutionJournal.get_spec_issue_runs` matches a project
+    by this rule and the report's accounting matches by the same one.
+    Two copies of it would let the trend and the accounting disagree
+    about which audits belong to the project being reported on.
+
+    Applies to a record's nested objects too, which is what the stored
+    issue list is: same JSON, same rule.
+    """
+    value = entry.get(key)
+    return value if isinstance(value, str) else ""
 
 
 def _journal_line(entry: dict[str, Any]) -> str:
@@ -1436,31 +1468,71 @@ class EvolutionJournal:
         )
 
     # ------------------------------------------------------------------
-    # get_spec_issue_runs
+    # get_spec_audits / get_spec_issue_runs
     # ------------------------------------------------------------------
 
-    def get_spec_issue_runs(self, project: str, last_n: int = 10) -> list[dict[str, Any]]:
-        """The last N recorded spec audits for ``project``, oldest first (#260).
+    def get_spec_audits(self) -> list[dict[str, Any]]:
+        """Every recorded spec audit in the journal, oldest first (#314).
+
+        The whole set, across every project, because the caller that
+        accounts for the history a windowed trend leaves out needs
+        exactly what the window drops: filtering by project here would
+        hide the thing it is asking for.
+
+        This is the read surface a caller uses INSTEAD of opening
+        ``config.journal_path`` for itself. The difference is not
+        cosmetic: if the journal ever compacts, rotates or gains a
+        second segment, this method is what changes, while a caller
+        holding the path would quietly return less than the journal
+        holds - and silent loss of the excluded-history accounting is
+        the defect #280 exists to fix.
 
         Deliberately NOT routed through :meth:`_read_journal_entries`:
         that reader keeps only entries whose ``run_id`` is among the
-        last N distinct run ids, and a ``spec_issues`` entry carries no
-        ``run_id`` at all (decompose runs before a factory run id
-        exists), so every one of them is dropped there. Reading the raw
-        entries is what makes the architect's own history readable.
+        last N distinct run ids, and a spec audit carries no ``run_id``
+        at all (decompose runs before a factory run id exists), so
+        every one of them is dropped there. Reading the raw entries is
+        what makes the architect's own history readable.
+        """
+        return [e for e in self._read_all_entries() if e.get("event_type") == SPEC_ISSUES_EVENT]
+
+    def get_spec_issue_runs(
+        self,
+        project: str,
+        last_n: int = 10,
+        audits: list[dict[str, Any]] | None = None,
+    ) -> list[dict[str, Any]]:
+        """The last N recorded spec audits for ``project``, oldest first (#260).
+
+        The one place the window rule lives (#314). ``decompose`` had a
+        second copy of it, and two copies of one rule can drift: if
+        they had, the convergence trend and the accounting printed
+        under it would have disagreed about the same journal.
 
         ``last_n`` counts spec audits, not factory runs - a spec audit
         happens once per decompose, whether or not a factory run
         follows. Windowed here rather than by the caller, matching
         :meth:`get_experiment_trends`.
 
+        ``audits`` lets a caller that has already called
+        :meth:`get_spec_audits` window that snapshot instead of reading
+        the file a second time, so the window and any accounting over
+        the same entries cannot disagree because the file moved between
+        two reads. The event-type filter is applied either way, so
+        passing raw entries answers the same as passing audits.
+
         Nothing is assumed about an entry beyond it being a JSON
-        object, so journals written by older versions read cleanly.
+        object, so journals written by older versions read cleanly. A
+        project is matched by :func:`entry_str`, so a null or
+        non-string ``project`` field is an unattributed audit rather
+        than a project named "None".
         """
+        source = self.get_spec_audits() if audits is None else audits
         runs = [
             entry
-            for entry in self._read_all_entries()
-            if entry.get("event_type") == "spec_issues" and entry.get("project") == project
+            for entry in source
+            if entry.get("event_type") == SPEC_ISSUES_EVENT
+            and entry_str(entry, "project") == project
         ]
         return runs[-last_n:] if last_n > 0 else []
 
@@ -1606,8 +1678,8 @@ class EvolutionJournal:
         """Read JSONL journal and return entries from the last N distinct runs.
 
         Entries without a ``run_id`` are dropped, because the window is
-        defined in terms of runs. ``spec_issues`` entries are exactly
-        that case; :meth:`get_spec_issue_runs` reads those instead.
+        defined in terms of runs. Spec audits are exactly that case;
+        :meth:`get_spec_audits` reads those instead.
         """
         entries = self._read_all_entries()
         if not entries:
--- a/tests/helpers/journal.py
+++ b/tests/helpers/journal.py
@@ -26,7 +26,12 @@ from __future__ import annotations
 from pathlib import Path
 from typing import Any
 
-from kstrl.evolution import JOURNAL_REPAIR_EVENT, EvolutionConfig, EvolutionJournal
+from kstrl.evolution import (
+    JOURNAL_REPAIR_EVENT,
+    SPEC_ISSUES_EVENT,
+    EvolutionConfig,
+    EvolutionJournal,
+)
 from kstrl.observability import read_progress_events
 
 #: A partial JSONL line: valid JSON up to the point the process died,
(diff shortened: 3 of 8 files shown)


0xfauzi:
## P2-1 fixed: the guard now has a net under it

Pushed as `d32ef8c`.

The finding was right, and it was worse than six shapes. I planted 33 shapes plus 4
negative controls into `kstrl/evolution.py` one at a time and ran the guard against
each. Before: **6 caught, 27 missed**. After: **32 caught, 1 missed by design**, and
that one is disclosed in a docstring and pinned by a test that asserts the miss.

Every run used the pycache discipline, because CPython invalidates cached bytecode on
`(mtime_seconds, size)` and two same-size edits inside one second reuse stale bytecode:

```
find . -name __pycache__ -type d -exec rm -rf {} +
PYTHONDONTWRITEBYTECODE=1 uv run pytest ...
```

### What changed, and why it is not just a wider matcher

Widening the shape list was necessary and not sufficient. Enumerating node types does
not converge by inspection: round 1 found six misses, my widened version fixed those
and I then found seven more ordinary shapes it still could not see (`setattr`, a
parameter default, a tuple-unpacked assignment, a loop variable, a function returning
the bare name, `dict(zip(...))`, a three-element pair list). A third round would find
more. That is the same treadmill #324 records eleven guards on.

So the guard now has the two-layer shape `tests/test_journal_one_writer.py` already
documents, and I moved it into its own file, `tests/test_event_names_have_one_home.py`.

**Layer 1, `event_name_spellings`, is the net.** It counts every expression in
`kstrl/` whose folded value IS an enrolled event name, per module, pinned:

```python
EXPECTED_EVENT_NAME_SPELLINGS = {
    "decompose.py": 3,             # the architect's JSON key twice, plus the TUI artifact label
    "evolution.py": 2,             # the two declarations: this is the one home
    "tui/screens/decompose.py": 1, # the TUI reading that label back
}
```

Three rows, six occurrences. A row cannot be written or selected by a name the module
never spells, so a second spelling in any shape has to appear here first. It
enumerates no node types and resolves nothing, which is the point.

Note what the inventory itself found: only **one** of the six occurrences is the
journal's vocabulary. `spec_issues` is also the architect's JSON key AND the TUI's
artifact label (`ArtifactWritten(label="spec_issues")` at `decompose.py:1991`, read
back at `tui/screens/decompose.py:282`). Three vocabularies share the word. That is
why layer 2 draws its line at the `event_type` column and why layer 1 is a pinned
count rather than an assertion that the count is zero.

Equality of the folded value, not substring, which is the opposite of the choice
`folded_filename_sites` makes next door and for the opposite reason: that guard hunts
a filename and prose folds to a docstring CONTAINING it, while `DECOMPOSE_PROMPT`
spells `spec_issues` inside a prompt body thousands of characters long. Substring here
would put every prose mention in the inventory.

**Layer 2, `literal_event_names`, is the message.** Layer 1 can only say "this
module's spelling count moved", which is the wrong thing to say when the answer is
"you wrote a journal row with a bare literal, import the constant".

### The six named shapes, each planted alone, each now RED on both layers

```
===== B  dict() call writer =====
AssertionError: The set of places that spell an enrolled journal event name changed. [...]
AssertionError: A journal row is written or selected by a bare literal instead of the
constant evolution.py declares for it. Import the constant.
Sites: {'evolution.py': ["SPEC_ISSUES_EVENT: line 1710: writes 'spec_issues' as the event_type"]}
2 failed in 0.77s

===== C  item assignment after construction =====
Sites: {'evolution.py': ["SPEC_ISSUES_EVENT: line 1711: writes 'spec_issues' as the event_type"]}
2 failed in 0.74s

===== D  setdefault =====
Sites: {'evolution.py': ["SPEC_ISSUES_EVENT: line 1710: writes 'spec_issues' as the event_type"]}
2 failed in 0.87s

===== F  update() with kwargs =====
Sites: {'evolution.py': ["SPEC_ISSUES_EVENT: line 1710: writes 'spec_issues' as the event_type"]}
2 failed in 0.78s

===== G  match/case reader =====
Sites: {'evolution.py': ["SPEC_ISSUES_EVENT: line 1711: compares event_type against 'spec_issues'"]}
2 failed in 0.80s

===== J  walrus into a compare =====
Sites: {'evolution.py': ["SPEC_ISSUES_EVENT: line 1710: compares event_type against 'spec_issues'"]}
2 failed in 0.86s
```

Each names the right line and the right direction, so the message tells the author
what to do rather than only that something is wrong.

### The full shape table

Measured, not reasoned about. Every row was planted individually into
`kstrl/evolution.py` under the pycache discipline. GREEN means that layer reported
clean with the shape present.

| # | Shape | Before | L1 after | L2 after |
|---|---|---|---|---|
| A | `{"event_type": X}` dict literal | RED | RED | RED |
| B | `dict(event_type=X)` | GREEN | RED | **RED** |
| C | `row["event_type"] = X` after construction | GREEN | RED | **RED** |
| D | `row.setdefault("event_type", X)` | GREEN | RED | **RED** |
| E | `{**base, "event_type": X}` | RED | RED | RED |
| F | `row.update(event_type=X)` | GREEN | RED | **RED** |
| G | `match e.get("event_type"): case X:` | GREEN | RED | **RED** |
| H | alias on one line, compare on the next | GREEN | RED | **RED** |
| I | `e.get("event_type", X)` default | GREEN | RED | **RED** |
| J | walrus into a compare | GREEN | RED | **RED** |
| K | `TypedDict` / model constructor keyword | GREEN | RED | **RED** |
| L | `functools.partial(f, event_type=X)` | GREEN | RED | **RED** |
| M | dict comprehension building the row | GREEN | RED | **RED** |
| N | `row.pop("event_type", X)` default | GREEN | RED | **RED** |
| O | `in {"...", "..."}` set/tuple of names | RED | RED | RED |
| P | `.startswith(X)` on a read | GREEN | RED | **RED** |
| Q | `if`/`elif` chain of `==` | RED | RED | RED |
| R | dict-dispatch table keyed by the name | GREEN | **RED** | GREEN (deliberate) |
| S | `operator.itemgetter("event_type")` | GREEN | RED | **RED** |
| T | `"{}_issues".format("spec")` | GREEN | GREEN | GREEN (**disclosed + pinned**) |
| U | `case {"event_type": X}` mapping pattern | GREEN | RED | **RED** |
| V | `dict([("event_type", X)])` | GREEN | RED | **RED** |
| W | `case "role_usage" \| "spec_issues"` | GREEN | RED | **RED** |
| X | `"spec_issues" == e["event_type"]` | RED | RED | RED |
| Y | `"spec" + "_issues"` assembled | RED | RED | RED |
| Z | annotated alias, then compare | GREEN | RED | **RED** |
| AA | read behind a function boundary | GREEN | **RED** | GREEN (deliberate) |
| AB | `e.get("event_type") == names["spec"]` | GREEN | GREEN | GREEN (**same class as T**) |
| AC | `_SPEC = "spec_issues"`, then compare | GREEN | **RED** | GREEN (deliberate) |
| AD | `def emit(event_type="spec_issues")` | GREEN | **RED** | GREEN |
| AE | `setattr(row, "event_type", "spec_issues")` | GREEN | **RED** | GREEN |
| AF | `SPEC, OTHER = "spec_issues", "role_usage"` | GREEN | **RED** | GREEN |
| AG | name via a loop variable off a list literal | GREEN | **RED** | GREEN |
| AH | `def kind(): return "spec_issues"` | GREEN | **RED** | GREEN |
| AI | `dict(zip(["event_type"], ["spec_issues"]))` | GREEN | **RED** | GREEN |
| AJ | a three-element pair list | GREEN | **RED** | GREEN |

Shapes AD to AJ were **not on the review's list**. I found them by asking what layer 2
still could not see after the widening, which is how the net came to exist: every one
of them is caught by layer 1 and none by layer 2, and a third round of node-type
enumeration would have found an eighth.

Negative controls:

| # | Shape | L1 | L2 |
|---|---|---|---|
| N1 | `data.get("spec_issues", [])`, the architect's key | RED (a new row, with a reason) | GREEN |
| N2 | the name in prose in a docstring | GREEN | GREEN |
| N3 | another column compared against the name | RED (a new row, with a reason) | GREEN |
| N4 | an unrelated row and an unrelated comparison | GREEN | GREEN |

N1 and N3 are not false positives, they are the design: a NEW spelling of the word in
any vocabulary bumps a row, and the diff that adds it is where somebody says why.
Layer 2 stays silent on both, which is what keeps the message right. N2 stays green on
both because prose folds to a whole docstring, not to the bare name.

### The one disclosed miss, pinned

`tests/test_event_name_shapes.py::TestTheDisclosedMisses` asserts it, so the
disclosure fails if it stops being true:

**A name the interpreter has to build** (T, AB). `str.format`, `"".join`,
`%`-formatting, a run-time lookup. Constant folding answers `None`, so neither layer
can see it. The bound: a writer of such a row still has to reach the `event_type`
column, and a reader still has to obtain a value to compare, so this hides an
individual site rather than the practice. It is the residual `folded_str` already
discloses next door, reached from here.

R, AA and AC used to be on this list. They are not any more, because layer 1 catches
all three. `TestLayerOneCatchesWhatLayerTwoCannot` pins nine such cases, each asserting
BOTH halves (layer 2 silent, layer 1 counting), so the two-layer claim is a test and
not prose.

### The rule I removed, and why

My first version of the fix had `_assignment_hits` flag **any** assignment whose value
folds to an enrolled name, with the enrolled constants exempt. It caught AC. It also
broke the column discipline every other rule holds, and the `/simplify` altitude pass
found the cost, which I confirmed:

- it would fail `SPEC_ISSUES_KEY = "spec_issues"`, the obvious next cleanup of the
  architect's JSON vocabulary in `decompose.py`, with a message telling the author to
  import the JOURNAL's constant;
- it would fail `kstrl/events.py:343` and `:646`, which are the event STREAM's
  discriminator, the moment the enrolment follow-up this file commits to reaches
  `contract_result` and `autonomy_transition`. A guard that fails the follow-up it
  exists to enable gets silenced.

Narrowed to `x["event_type"] = X` and `self.event_type = X`. Layer 1 counts the
declaration instead, where a row with a reason is the normal outcome.

### Layer 2's visible inventory, before and after

Run the matcher over `kstrl/` for all seven real journal event names, not just the two
enrolled ones. Before: 9 rows. After: **15 rows**, and every one of the six is
accounted for.

| New row | Why it appeared |
|---|---|
| `evolution.py:938` writes `'component_result'` | the `.get` DEFAULT, a second spelling on a line that already had a compare row |
| `evolution.py:960` writes `'component_result'` | same |
| `evolution.py:1331` writes `'component_result'` | same |
| `evolution.py:1385` writes `'component_result'` | same |
| `events.py:343` writes `'contract_result'` | `type: ClassVar[str] = 'contract_result'` |
| `events.py:646` writes `'autonomy_transition'` | same |

No row was lost. None is for an ENROLLED name, so the pinned assertion stays green on
unchanged source. The last two are the ones that made me narrow `_assignment_hits`: an
earlier draft reported them and would have failed once those names are enrolled.

Side note worth having: `ENROLLED_EVENT_CONSTANTS`'s comment claims `component_result`
is a bare literal "nine times in `evolution.py` alone". The widened matcher now counts
exactly nine there. Before it could see five.

### No vacuous rules, measured

Each of the 23 matcher functions stubbed to a constant in turn, then the controls
re-run. **All 23 noticed**, none silently:

```
event_name_spellings 9    _hits_at 29             _keys_on_the_column 17
_reads_event_type 14      _is_event_type_read 14  _write_hit 14
_read_hit 15              _names_the_event 12     _literal_event_reads 11
_call_event_hits 8        _literal_event_writes 7 _pair_hits 5
event_type_aliases 5      _alias_sweep 5          _bound_names 5
_keyword_hits 4           _keyed_argument_hits 3  _pattern_literals 3
_assignment_hits 2        _names_the_column 2     _match_event_reads 2
_mapping_pattern_reads 1  _method_on_a_read_hits 1
```

### Gates

- Fast tier: **4843 passed**, 28 skipped, 6 xfailed. Collected 4837 -> 4877 (+40:
  +44 shape controls, +2 guard tests, -7 detector tests that moved out of
  `test_evolution.py`, +1 for `AuditSnapshot`).
- `uv run mypy kstrl/ --strict`: clean, 127 files. `uv run ruff check kstrl/ tests/`: clean.
- `pre-commit run --all-files`: all hooks pass, including complexipy, the cyclomatic
  ratchet, codespell, vulture, gitleaks, the 800-line ratchet and no-em-dash.
- complexipy on the new guard: highest is 7, gate is 15.
- Cost: the guard test's alias set is now hoisted off the name loop. Measured over 127
  modules: layer 2 **483 ms -> 382 ms**, layer 1 adds 84 ms. It was O(names x modules)
  on a computation that does not depend on the name, and this file commits to five
  more names.

### Untouched, as established by the review

- #327 one-writer guard: **5 rows / 15 occurrences**, `found == EXPECTED_JOURNAL_PATH_SITES`
  is True. Zero diff on `tests/test_journal_one_writer.py`.
- `tests/test_check_name_enrolment.py`: still **19 names**. Zero diff.
- H3 prompt snapshots: 8 prompts / 8 snapshots, zero diff on
  `tests/test_prompt_versions.py`. No enrolled prompt was edited, so no calibration
  re-run is owed.

---

## P3-1: `AuditSnapshot.__hash__`

`frozen=True` with `eq` on, so the decorator generated a `__hash__` over two `list`
fields that raised `TypeError: unhashable type: 'list'` from a method nobody wrote.

**Suppression, not tuples, and I measured why.** Tuple fields do not fix it, because
the elements are `dict[str, Any]` and hashing a tuple hashes its elements:

```
tuple fields: TypeError: unhashable type: 'dict'
suppressed:   TypeError: unhashable type: 'AuditSnapshot'
```

So `__hash__ = None` in the class body, which the decorator sees as an explicit hash
and leaves alone. `frozen=True` stays, because the rebinding guarantee is what it was
for; `eq` stays, because a test compares two snapshots. Pinned by
`test_the_snapshot_says_unhashable_instead_of_pretending`, which asserts both halves,
since the message alone would pass on a class that kept the broken generated hash.

**Not generalised, reported instead.** Measured precisely (top-level
`list`/`dict`/`set` annotation only, so a `Callable` whose signature mentions `dict[...]`
is not miscounted): **10** frozen+eq dataclasses in `kstrl/` have this shape. One is
now suppressed; the other **nine** still advertise a hash that raises:

```
Transition, _Buckets, SpecConvergence, PolicyConfig, PolicyEvaluation,
RetryPreview, ComponentScope, Outcome, ScrubbedProjectContext
```

`AuditSnapshot` is the one this PR introduced, so it is the one fixed here. A sweep of
the other nine wants its own change with a guard behind it, not nine edits nothing
keeps true.

---

## `/simplify` findings, verbatim, including the declined

Four agents (reuse, simplification, efficiency, altitude).

**Applied:**

1. *Altitude:* "The guard is fixed one level too shallow. The deeper invariant exists,
   is already implemented next door for the sibling problem in four lines, and I
   measured that it subsumes every shape this diff enumerates plus seven more the
   widened matcher still misses." -> **applied**, and it is the largest thing in this
   commit. Verified the claim myself before adopting: 3 rows, 6 occurrences, and it
   catches R, AA, AC and AD-AJ.
2. *Altitude:* "`_assignment_hits` is too broad, and it is the one rule that breaks the
   guard's own stated line." -> **applied**, narrowed to the column. Reasoning above.
3. *Altitude + Simplification:* "The seam is wrong, and the ratchet will not tell
   anyone" (the 800-line hook reports rather than fails once a file is over, so nothing
   would ever prompt the split). -> **applied**, guard moved to
   `tests/test_event_names_have_one_home.py`; `test_evolution.py` is 181 lines shorter.
4. *Efficiency:* "254 calls to `event_type_aliases`, 127 distinct trees [...] exactly
   127 calls, half, are pure duplicates" and "at the 7 names the diff plans, the wasted
   work alone (0.96 s) exceeds the whole file's current runtime". -> **applied**,
   `aliases` is now an optional hoisted argument. Measured 483 -> 382 ms; the controls
   need no edit because they use the default.
5. *Simplification:* "`_bound_names` exists for a branch nothing pins. Measured: delete
   the `ast.NamedExpr` branch and all 36 tests pass. `test_a_walrus_into_a_compare`
   does not reach it." -> **applied**, added `test_a_walrus_bound_on_an_earlier_line`.
6. *Simplification:* "the `ast.List` half of the bare-pair branch is unexercised." ->
   **applied**, added `test_a_bare_pair_in_a_list_counts_too`.
7. *Simplification:* "`_pair_hits`'s `and value is not None` can never be False [...]
   1626 dict literals across `kstrl/` and `tests/` yield zero `None` values." ->
   **applied**, condition and the `| None` on the value half both removed.
8. *Simplification:* "`_read_hit(node, operand)` has a parameter with no observable
   behaviour." -> **applied**, collapsed to one argument.
9. *Simplification:* "two spellings of 'spells the event name', no stated rule [...] the
   boundary is load-bearing in exactly one place, arbitrary in two, and written down
   nowhere." -> **applied** as documentation: `_keyed_argument_hits` and
   `_method_on_a_read_hits` now each say why they fold shallow or deep.
10. *Simplification:* "the 2-line field comment restates the docstring paragraph three
    lines above." -> **applied**, comment deleted.

**Declined, with reasons:**

11. *Reuse + Simplification, both flagged it:* extract the fixed-point loop shared by
    `event_type_aliases` and `journal_aliases` into one helper taking the sweep as a
    callable. **Declined.** The duplicated part is five lines of `while True`, and the
    two sweeps share no name resolution at all: one resolves journal paths through
    `mentions_journal` plus a lineno exemption, the other resolves event-type reads.
    #324's lesson is about duplicated RESOLUTION being independently holed, and the
    resolution is exactly the part that is not shared. Against that, the extraction
    puts a `Callable` indirection into a guard this PR was told not to perturb, on a
    branch that is otherwise approved. Worth doing in the sweep that also fixes the
    nine dataclasses, not here.
12. *Simplification:* "prose duplicated across two files [...] the matcher region is 141
    lines of code and 140 lines of prose." **Declined on the ratio, partly applied on
    the duplication.** I removed the repeated seam paragraph from the controls file.
    The remaining overlap is the module docstring stating the two-layer architecture and
    `literal_event_names` stating what round 1 missed, which are two different readers.
    The prose-to-code ratio is the house norm here (`test_journal_one_writer.py` is
    584 lines for a comparable amount of logic) and it is what makes the disclosures
    checkable.
13. *Altitude:* "`__hash__ = None` is a one-off on a 13-instance shape, with no
    mechanism." **Partly declined.** Real, and the count is nine not thirteen (their
    filter counted `Callable[..., dict[...]]` and `tuple[..., dict]` annotations). Kept
    the fix because this PR introduced `AuditSnapshot`, and reported the other nine
    above rather than sweeping them in a PR about an AST guard.

---

## Not fixed here, by instruction

- **P3-2**: four literal event-name sites remain in `tests/` (`test_decompose.py:805,
  1037, 1132`, `test_evolution.py:1055`), and `test_decompose.py:802-805` re-implements
  the selection rule and reads a kstrl-written file via `journal.read_text()` with no
  `encoding="utf-8"`. Pre-existing, filed separately.
- **P3-3**: `_excluded_history`'s one-bucket invariant is false at `project_name == ""`
  (measured identically on base and head: `sum=8` against `audits=5`). Pre-existing,
  filed separately.

PR #332, opened 2026-09-01
feat(decompose): four ways to close a question, and halt only on escalation (#260)

0xfauzi:
Step 3 of #260, the paid step. Gives the architect four ways to close a question and halts on one of them.

Steps 1 and 2 were free and shipped first: #322 routes the non-blocker findings into the component PRDs. This one edits `DECOMPOSE_PROMPT`, so it owes an H2 calibration run and an H3 version bump.

## Why this exists, in one paragraph

Five paid architect runs against one real spec. Every run halted, zero components were produced, and audit volume was flat at 21, 32, 20, 21, 23 findings with zero exact repeats across all 117, so there was nothing to converge on. Not one of the 26 blockers was a real escalation; the two genuine owner questions were labelled major and minor. Run 2's single blocker complains that the claude permission mapping cannot be derived, and its own `suggestion` quotes the six values it had just measured with `claude --help`: it solved its own blocker and halted on it. The full write-up, the replay evidence and the round-1 `/simplify` dispositions are in a PR comment, unedited.

## Where everything is

- **This body**: the H2 calibration for the shipped prompt, then what round 2 and round 3 changed and why, with the numbers.
- **PR comment, round-1 write-up**: the problem statement, the free replay validation, the first three paid captures, the prompt-inflation disclosure as it stood after round 1, and what round-1 `/simplify` applied and declined.
- **PR comment, round-1 `/simplify` findings**: four agent reports, verbatim.
- **PR comment, round-2 `/simplify` findings**: four agent reports, verbatim.

## H2 calibration: 3.0.0 measured, and which capture measured which prompt

The owner approved the paid re-run at about $1.25 on the grounds that 2.0.0 to 3.0.0 was a schema change, not a wording tweak. It is done: `baseline-20260901-203244.json`, architect fixtures only (`-k architect`), haiku, three runs per fixture, twelve agent calls, 39m 47s, **$1.3520**.

**3.0.0 is LEVEL, not better and not worse.** Fixture for fixture, it reproduces 2.0.0's third capture exactly.

```
fixture                    role                      1.4.2 pre-change     2.0.0 cap3           3.0.0 NEW
spec-01-no-error-handling  architect                 3/3 cons=1.00 err=0  3/3 cons=1.00 err=0  3/3 cons=1.00 err=0
spec-02-unspecified-auth   architect                 3/3 cons=1.00 err=0  2/3 cons=0.67 err=0  2/3 cons=0.67 err=0
[...]

Issue #312, opened 2026-08-31
append_entries eats the next journal entry after a torn write


0xfauzi:
Found by `/code-review xhigh` on PR #311 (issue #280). Pre-existing in
`EvolutionJournal.append_entries`, newly load-bearing because #280's
convergence note now counts journal entries and reports the total to the
operator as evidence.

## What happens

`append_entries` opens the journal in append mode and writes
`json.dumps(entry) + "\n"` per entry. It never checks whether the file it
is appending to ends in a newline. A crash mid-write leaves a partial line
with no trailing newline, so the NEXT entry is concatenated onto the
fragment and the two become one unparseable line. The tolerant reader skips
it, and the second entry is gone permanently.

`kstrl/evolution.py:1294` (`append_entries`):

```python
self.config.journal_path.parent.mkdir(parents=True, exist_ok=True)
with open(self.config.journal_path, "a") as f:
    for entry in entries:
        f.write(json.dumps(entry, separators=(",", ":")) + "\n")
```

## Reproduction

Append audit A, write the torn fragment `{"event_type": "spec_iss` (byte
for byte the fixture `tests/test_decompose.py` already uses for the
torn-tail case), then append audit B through the real writer:

```
after audit A: 1 line(s)
after the crash: file ends with newline? False

--- journal on disk ---
{"timestamp":"...","project":"alpha","event_type":"spec_issues",...}
{"event_type": "spec_iss{"timestamp":"...","project":"beta","event_type":"spec_issues",...}

--- what any reader sees ---
1 entry(s) parsed: ['alpha']

audits written: 2 (alpha, beta)
audits readable: 1
```

The journal now holds two audits and reports one. Nothing warns.

## Why it matters more now

`tests/test_decompose.py::TestExcludedHistory::test_a_torn_line_does_not_cost_the_note`
establishes the torn-tail crash as a scenario the system must survive, but
it only covers the READ side. The append path then destroys the next entry
written, so surviving the read is not enough: the audit trail loses a
record it was asked to keep, and #280's note prints a count that is one
short with no indication anything is missing.

This is the same class of defect as #280 itself: a number quietly computed
over less data than was written.

## Scope

Every writer of the journal is affected, not just spec audits.
`record_run` goes through the same method, so a crash mid-write also eats
the next `component_result` batch, which feeds
`get_cross_run_patterns`, `get_concern_hit_rate` and
`get_fact_utilization`.

## Suggested fix

[...]

Issue #294, opened 2026-08-31
A scope that could not be read is reported as diff_scope, which tells the engineer to fix something it cannot affect


0xfauzi:
Raised by a `/simplify` altitude pass during PR #293 (merged, closes #269), declined there with reasons, and recorded here so the argument is not lost.

## The observation

When a component's scope cannot be read at plan time, the failure is reported through `check_diff_scope` as `failed_check = diff_scope`, carrying `allowed_paths_error`. That has been true since R1.5.

But `diff_scope` means "the diff touched files outside the allowlist", and the retry context tells the engineer to narrow its diff. When the real problem is that the allowlist could not be read, the engineer is being asked to fix something it cannot affect. It will narrow the diff, fail identically, and burn the attempt.

The suggestion was to give the fail-closed scope signal its own ungated check, `check_scope_source`, so the failure names its own cause.

## Why PR #293 declined it, and why that reasoning holds

Three arguments, all sound:

1. `diff_scope` is an API. Three consumers key on that string, including the evolution journal's failure signatures. Renaming or splitting it is a wider change than a scope fix.
2. `diff_scope` has carried `allowed_paths_error` since R1.5, so this is long-standing behaviour rather than something PR #293 introduced.
3. **After that PR, the state is unreachable from `run_factory`.** `_preflight_component_scope` now refuses an unresolved component before the scheduler launches anything, so no engineer attempt is spent on it. The misleading failure can no longer cost money on the normal path.

Point 3 is what makes this low priority rather than urgent. It is a clarity problem now, not a spend problem.

## What is still worth doing, and when

The state is unreachable from `run_factory`, which is not the same as unreachable. `run_mechanical_verification` is called directly by `ks verify`, and any future caller that bypasses the plan-time preflight reopens it. A check whose name matches its cause is also simply better for the operator reading a failure summary.

So: worth doing when someone is next in `verify.py`'s check vocabulary for another reason, and worth doing carefully because of point 1. Not worth a PR of its own.

## Related

- #269 (closed by PR #293) established the plan-time snapshot that makes the state unreachable on the normal path.
- The same PR moved the PRD tamper refusal off `diff_scope` and onto `check_prd_stories`, on the argument that "defence in depth an unrelated toggle can disable is not defence in depth". This is the other half of that tidy-up: a check should be named for what it checks.

PR #339, opened 2026-09-01
fix(evolution): give the unenrolled gates a category, and an infrastructure one for the four with no home (#315)

### LINKED ISSUE #315: Three Phase 1 gates are miscategorised as engineer-loop failures in the evolution journal
Three Phase 1 gates are filed in the evolution journal under the engineer loop, because `evolution._CATEGORY_BY_CHECK` does not carry their names.

## The defect

`evolution.category_for_check` maps a check name onto a `FailurePattern.category` and falls through to `"iteration"` for anything the table does not carry. Measured by AST-walking `kstrl/` for every check name it constructs and mapping each through the function:

```
fixtures        -> iteration      (kstrl/fixtures.py)
policy_envelope -> iteration      (kstrl/verify.py)
test_adequacy   -> iteration      (kstrl/verify.py)
```

All three are mechanical verification gates. `signatures_from_verification` iterates every failed check with no filter, so each reaches `split_signature` -> `category_for_check` -> `FailurePattern.category` at `evolution.py` in both `extract_failure_patterns` and `get_cross_run_patterns`. A recurring approved-fixtures failure, policy envelope breach or test-adequacy block is therefore reported by `ks evolve` as an engineer-loop problem rather than a verification one.

Found while landing #294 (PR #310), which added a fourth name to the same table and needed to know whether the convention it was following actually held.

## Update: five more names, and they are not all the same defect

PR #310 round 2 widened the walk to cover the OTHER producer of the same prefix: `pipeline.fail(..., signatures=["<prefix>:<code>"])`, which builds a signature directly without ever constructing a `CheckResult`. The walk now sees 19 names rather than 13, and five more land on `"iteration"`:

```
aborted      -> iteration   pipeline.py, "aborted:shutdown"
token_budget -> iteration   pipeline.py, "token_budget:exceeded"
pr           -> iteration   pipeline.py, "pr:merge-conflict"
diff         -> iteration   pipeline.py, "diff:fetch-failed"
engineer     -> iteration   pipeline.py, "engineer:no-progress-stall"
```

Only one of those five is correct as it stands. `engineer:no-progress-stall` IS an engineer-loop failure, so `"iteration"` is the right answer and enrolling it would only make the table state explicitly what the fallback already gets right.

The other four are not engineer-loop failures and they are not verification failures either. A shutdown, a blown token ceiling, a merge conflict and a failed diff fetch are infrastructure and budget conditions, and the journal has no category for them: `_CATEGORY_BY_CHECK` offers `verification`, `review`, `security` and `contract`, and the fallback is `iteration`. So they cannot be enrolled correctly without first deciding what they should be called. `autonomy_replay.INFRA_FAILURE_PREFIXES` already treats `pr:`, `git:`, `infra:`, `timeout:` and now `scope_unreadable:` as infrastructure aborts for a different consumer, which is prior art for an `infrastructure` category and also evidence that the two consumers currently disagree about what `pr:merge-conflict` is.

This is why all eight sit in `UNENROLLED_AT_INTRODUCTION` rather than being fixed in passing: three are a straight misfiling, four need a new category decided first, and one is arguably already right.

## The mechanism now exists

PR #310 added `tests/test_check_name_enrolment.py`. It AST-walks `kstrl/` for every check name passed to `CheckResult` or `verify._failed_gate_result`, AND every `<prefix>:<code>` literal passed as a `signatures=` argument, resolving module-level string constants (`gateparse.GATE_TEST` and friends, `verify.SCOPE_UNREADABLE_CHECK`) as well as literals, and fails on any name absent from `_CATEGORY_BY_CHECK`.

The eight names above are grandfathered in `UNENROLLED_AT_INTRODUCTION` so the gate could land without also changing recorded categorisation. The set is guarded in both directions: a name that stops being emitted, or that gets enrolled and left in the set, fails the test. So the list cannot rot, and no NEW check can join it without a visible edit.

Stated limit: the walk cannot see `review`, `security` and `contract`, because `signatures_from_findings` composes those from a runtime `phase` argument and no literal exists to find. They are asserted by hand in `test_the_runtime_composed_phases_are_enrolled`.

## What this issue is for

1. Enrol `fixtures`, `policy_envelope` and `test_adequacy` as `verification`, and delete them from `UNENROLLED_AT_INTRODUCTION`.
2. Decide whether the journal gains an `infrastructure` category, and if so enrol `aborted`, `token_budget`, `pr` and `diff` into it. Reconcile with `autonomy_replay.INFRA_FAILURE_PREFIXES` so one answer serves both consumers.
3. Enrol `engineer` as `iteration`, which changes nothing but removes it from the grandfathered set.
4. Decide what happens to journal history already written under `"iteration"`.

That last part is why none of it was done in #310. The category is recorded per pattern at read time rather than stored, so enrolling any of these changes what `ks evolve` reports about runs that already happened. That is defensible, arguably a straight bug fix, but it is a behaviour change to the evolution journal with nothing to do with a scope-check rename, and it deserves its own justification rather than riding along.

## Verification

- `uv run pytest tests/test_check_name_enrolment.py -v`
- The walk currently sees 19 check names: `aborted`, `bad_patterns`, `dead_code`, `diff`, `diff_scope`, `engineer`, `fixtures`, `linter`, `mutation_testing`, `policy_envelope`, `pr`, `prd_stories`, `review`, `scope_unreadable`, `self_critique`, `test_adequacy`, `test_suite`, `token_budget`, `typecheck`.
- Eight of them are unenrolled and grandfathered: `aborted`, `diff`, `engineer`, `fixtures`, `policy_envelope`, `pr`, `test_adequacy`, `token_budget`. Each measured through `category_for_check` returns `iteration`.

0xfauzi:
Closes #315.

## What was wrong

`evolution.category_for_check` maps a check name onto a `FailurePattern.category` and falls through to `"iteration"` for anything `_CATEGORY_BY_CHECK` does not carry. Eight of the nineteen names the enrolment walk could see were unenrolled, so the journal filed them under the engineer loop: three Phase 1 gates (`fixtures`, `policy_envelope`, `test_adequacy`), four failures recorded outside a `CheckResult` (`aborted`, `token_budget`, `pr`, `diff`), and `engineer`, which the fallback got right by accident.

The four in the middle had no correct home. The table offered `verification`, `review`, `security` and `contract`, and the fallback was the engineer's loop. A shutdown, a blown token ceiling, a merge conflict and a diff that would not fetch are none of those.

While reviewing the fix, a third producer of check names turned up that the guard could not see at all. See "What the review found" below.

## What changed

1. `fixtures`, `policy_envelope`, `test_adequacy` are enrolled as `verification`.
2. A new `infrastructure` category holds `aborted`, `token_budget`, `pr`, `diff` and (round 2) `provisioning`.
3. `engineer` is enrolled as `iteration` explicitly, along with `unknown`. Neither changes an answer; both stop the table being readable as "most of the names".
4. `autonomy_replay.INFRA_FAILURE_PREFIXES` derives from `evolution.INFRASTRUCTURE_CHECKS` instead of retyping the same judgement. The two consumers disagreed about `pr:merge-conflict`: plumbing to the replay, an engineer-loop failure to the journal. Enrolling a check as infrastructure now reaches both in one edit.
5. `verify` and `provisioning` are enrolled, from the third producer the review found.
6. `_classify_check` no longer returns a category. All three callers discarded it, so it was a second, unreachable answer that could silently disagree with the live one.
7. `tests/test_check_name_enrolment.py` loses its grandfathered set entirely, gains the third producer, gains a blind-guard test, and gains a whole-table pin.

`UNENROLLED_AT_INTRODUCTION` is empty and deleted. The guard now admits no exceptions.

### The deliberate divergence, and why it is not "one answer serves both"

`scope_unreadable` stays `verification` in the journal and remains an infrastructure abort for the replay. The two consumers ask different questions. The journal asks which part of the factory produced a failure, and a scope refusal is a Phase 1 gate result: #294 gave it its own gate, its own table row and its own proposal arm on exactly that basis. The replay asks whether the run yielded a verdict about the factory's JUDGEMENT, and a gate's honest verdict can still answer no, because nothing was ever measured about the change.

So the relation is: infrastructure implies non-decisive, plus a documented remainder. The remainder is `autonomy_replay._REPLAY_ONLY_PREFIXES`, and `TestTheTwoInfrastructureConsumersAgree` pins both halves, through `RunRecord.infra_aborted` rather than against the constant.

`git:`, `infra:` and `timeout:` stay in that remainder. Measured with `git log -S`: all three strings entered the tree in 606cbb1, the commit that created the tuple, and no call site has ever produced them. They are kept because an experiments.tsv row outlives the code that wrote it and a prefix that matches nothing costs nothing. They are deliberately NOT enrolled in `_CATEGORY_BY_CHECK`, which describes names kstrl emits.

## Exactly what a user sees differently

Measured, not reasoned: the same seeded journal (two runs, signatures `pr:merge-conflict` and `fixtures:missing-approved-file`) read on `origin/main` and on this branch.

**`ks evolve` stdout: identical.** Byte for byte. The CLI prints `[check_name] description (seen in N components)` and never prints a category. `kstrl/cli.py` contains the string "category" zero times. The generated proposal files are also byte-identical (`diff -r` over the two `.kstrl/proposals` directories reports no difference), because `propose_improvements` branches on `pattern.check_name`, never on `pattern.category`.

**The evolve screen's patterns table changes.** That table (`ks tui`, evolve, patterns tab) has a `category` column, and it is the only surface in the product that displays this value.

Before:

```
  pr           merge-conflict                  2    1  iteration
  fixtures     missing-approved-file           2    1  iteration
```

After:

```
  pr           merge-conflict                  2    1  infrastructure
  fixtures     missing-approved-file           2    1  verification
```

This applies to runs that already happened. The category is computed at read time and never stored: measured by writing a real journal through `record_run` and listing the keys of the resulting entry, which are `schema_version, timestamp, run_id, project, component_id, event_type, status, retries, error, check_name, error_signature, failure_signatures, failed_phase, failed_check, duration_seconds, iteration_count, findings, findings_summary, usage, knowledge_utilization`. No `category` key, and no value equal to any category name. experiments.tsv stores the signature (`pr:merge-conflict`), not a category. Nothing on disk carries a stale category, so no migration exists to write.

**`ks autonomy replay` changes.** This is the behaviour change with teeth. `INFRA_FAILURE_PREFIXES` goes from 5 entries to 9:

```
before: pr:  git:  infra:  timeout:  scope_unreadable:
after:  aborted:  diff:  git:  infra:  pr:  provisioning:
        scope_unreadable:  timeout:  token_budget:
```

A run whose modal failure signature is `aborted:shutdown`, `token_budget:exceeded`, `diff:fetch-failed` or `provisioning:worktree-setup-failed` is now reported as infra-aborted and excluded from `decisive_runs`, instead of counting as evidence about the factory's judgement. It shows up in the report's `infra-aborted: N (excluded)` line rather than the `decisive: N` line, and it no longer feeds a simulated promotion or demotion. That is the module's own definition applied consistently: a run that died on a shutdown or a worktree it could not build says nothing about whether the factory's reviews are trustworthy.

Nothing else in the product reads `FailurePattern.category`. The full enumeration of consumers: `kstrl/tui/screens/evolve.py:205` (renders it), and nothing else. `kstrl/reducer.py`, `kstrl/pipeline.py`, `kstrl/pr.py` and `kstrl/findings.py` all use `Finding.category`, a different taxonomy with different values.

## Corrections to the issue text

Six claims checked, four confirmed, three corrected.

**Confirmed.**

- The walk really did see 19 names, and they are exactly the 19 the issue lists. Printed.
- All eight grandfathered names really did return `iteration` through `category_for_check`. Printed.
- The category really is read-time and nothing on disk carries it. Measured, above.
- `propose_improvements` really does branch on the check's name, not its category (`evolution.py:986` onward), so the triage note is right and proposals are unchanged by this fix.

**Corrected.**

1. **`ks evolve` was never the surface.** The issue says a recurring fixtures failure "is therefore reported by `ks evolve` as an engineer-loop problem". The `ks evolve` CLI never prints the category. The miscategorisation is real, but the surface is the evolve screen's patterns table, plus the autonomy replay through `INFRASTRUCTURE_CHECKS`.

2. **`_CATEGORY_BY_CHECK` was not the only place a category was decided.** `_classify_check` returned `(check_name, category)` and every one of its three call sites discarded the second element. Nothing got it right by accident because nothing read it, but a dead answer that can silently disagree with the live one is how the live one later gets fixed in the wrong file. It now returns the check name only.

3. **The walk had a third producer, not two.** Both #310 and the issue describe two producers (`CheckResult` names and `signatures=` literals). There is a third: `pipeline._record_failure_signatures` falls back to `signature_for_error(phase or "unknown", error)` when a `fail` / `retry_or_fail` call carries no `signatures=`, so the `phase=` argument becomes the check name. Two live sites, both unenrolled: `provisioning` (`factory.py:3414`) and `verify` (`pipeline.py:2146`).

Two smaller notes, neither a correction to the issue:

- The test module's own docstring claimed `review`, `security` and `contract` "appear in no literal anywhere in `kstrl/`". False for `review`: `pipeline` emits `review:divergence`, `review:setpoint-budget-exhausted` and `review:infrastructure` as literals, so the walk does see it. Fixed. The genuinely invisible enrolled names are `security`, `contract`, `verification` and `unknown`, and they are now enumerated with a reason each in `ENROLLED_BUT_INVISIBLE` and pinned.
- Two `signatures=` sites remain that the walk cannot read: `pipeline.py:2849` builds `f"{phase}:coverage-unverified:{reason}"` and `pipeline.py:3340` uses a conditional expression. Both resolve to prefixes that are already enrolled (`review`, `security`), so neither hides an unenrolled name today.

## What the review found

`/simplify` ran four agents against the first commit. Two findings were serious enough to change the shape of the fix, and both are the repo's own dominant defect class.

**A guard that was blind, not red.** The walk covered two of the three producers. Measured on the first commit: `provisioning:worktree-setup-failed`, a worktree that would not build, was filed under the engineer loop AND counted as decisive evidence for autonomy promotion, while `test_every_emitted_name_is_enrolled` was green next to a docstring saying the guard admitted no exceptions. The walk now resolves `phase=` on `fail` / `retry_or_fail` calls that carry no `signatures=`, and sees 21 names. It requires the `phase=` keyword rather than matching the callable name alone: `feature_cmd` has a local helper also called `fail`, and matching on the name credited six of its call sites with the name "unknown".

**Two guards that could not fail.** `INFRA_FAILURE_PREFIXES` is built as `derived | replay_only`, so `test_every_infrastructure_check_is_an_infra_abort_for_the_replay` restated the constructor, and `set(INFRA_FAILURE_PREFIXES) - derived == set(_REPLAY_ONLY_PREFIXES)` reduces to `replay_only - derived == replay_only`, which holds for any contents of `replay_only`. Both are gone. The class now pins the tuple's contents and runs every assertion through `RunRecord.infra_aborted`, the property that actually decides a run's fate.

## Mutations planted, and what happened

Every run under the `__pycache__` discipline: `find . -name __pycache__ -type d -exec rm -rf {} +` then `PYTHONDONTWRITEBYTECODE=1 uv run pytest ... -p no:cacheprovider`.

| Mutation | Expected | Result |
|---|---|---|
| M1: add `signatures=["planted_gate:unenrolled"]` to a pipeline failure | red | RED, `test_every_emitted_name_is_enrolled` |
| M2: rename a `CheckResult(name="fixtures")` to `"planted_check"` | red | RED, same test |
| M3: typo a category value (`"verifcation"`) | red | RED, whole-table pin |
| M4: delete the `"pr": "infrastructure"` row | red | RED, 3 enrolment tests plus 2 pre-existing `test_autonomy_ladder` tests, which is the derivation proving it is live |
| M5: move `policy_envelope` into a function-local so the walk cannot resolve it | red | RED, blind-guard test |
| M6: same for `dead_code`, an enrolled name no hand-written assertion lists | red here, green on main | RED on this branch, **GREEN on `origin/main`** |
| M7: change a `retry_or_fail(phase="verify")` to `phase="deployment"` | red | RED, 3 tests, including the anti-vacuity one |
| M8: add a bogus `engineer:` to `_REPLAY_ONLY_PREFIXES` | red | RED, prefix-contents test. Under the assertion it replaced, provably green |

M6 is the one worth reading. `dead_code` is enrolled, was never grandfathered, and is not one of the names the anti-vacuity test lists by hand. On `origin/main`, hiding it from the walk leaves all 12 tests passing: the guard could not fail on a name it could not see. On this branch it fails with "Newly invisible: ['dead_code']". That is the #324 defect class, closed for the whole table rather than for four hand-picked names.

## Verification

- `uv run pytest tests/ -v`: 4789 passed, 28 skipped, 6 xfailed.
- `uv run mypy kstrl/ --strict`: no issues in 127 source files.
- `uv run ruff check kstrl/ tests/` and `ruff format --check`: clean.
- Pre-commit hooks on both commits: all passed (complexipy, cyclomatic ratchet, codespell, vulture, gitleaks, shellcheck, file-length ratchet, em-dash, mock, agent-docs).
- `gitleaks dir . --redact` (the manual tree-scan hook) reports 1 finding, `aws-access-token` at `tests/test_policy_envelope.py:222`. It reports the same 1 finding on `origin/main`. Pre-existing test fixture, untouched here.
- Coverage: 90.11% against the 79.91% ratchet.

## Known divergence not closed here, and not created here

`factory._infra_casualty` (`factory.py:2701`) decides infrastructure-ness for the LIVE autonomy ladder from `Finding.is_infrastructure_error`, a different signal from the signature taxonomy this PR unifies. `token_budget` and `diff` failures attach an infrastructure finding; `pr:merge-conflict` and `aborted:shutdown` do not. So the live ladder can call a run judged while the replay calls it plumbing. That disagreement already existed for `pr:` before this PR, and this PR adds `aborted:` to it. Closing it means editing `kstrl/factory.py`, which has an open PR against it (#332), so it is left alone and stated. It deserves its own issue.

`docs/atlas/atlas.json` carries a stale `INFRA_FAILURE_PREFIXES` value: it still reads `('pr:', 'git:', 'infra:', 'timeout:')`, which has been wrong since #294 added `scope_unreadable:`. No generator for it exists in this repo and no CI job checks it, so it is left alone rather than hand-edited into a different kind of wrong.

## /simplify findings, verbatim

Four agents, one angle each, run against the first commit (94c5f6d). Reproduced in full, including the ones declined.

### Reuse

> **1. `kstrl/autonomy_replay.py:93-94` (and the predicate at `:115`) - the colon-suffixed prefix tuple re-implements `evolution.split_signature`.**
> `evolution.py:462` already owns the `"<check>:<code>"` format (`split_signature` returns `(check, code)`), and the enrolment walk's own docstring says signature literals "reach `category_for_check` via `split_signature`". The new code instead re-derives the format by appending `":"` to every check name and prefix-matching. The module now imports from `evolution` anyway, so `split_signature(self.common_failure)[0] in _INFRA_CHECKS` is available and gives the identical answer on every input (a signature with no colon fails both ways; `diff_scope:` fails both ways). Cost: the colon is hand-written in four places (`autonomy_replay.py:87`, `:94`; `tests/test_check_name_enrolment.py:365`, `:376`), and `test_the_prefixes_cannot_swallow_a_neighbouring_check` (`tests/test_check_name_enrolment.py:392-397`) exists only to guard a hazard the existing helper does not have. Two parsers of one string format, one of them untested against the format's owner.
>
> **2. `tests/test_check_name_enrolment.py:365, 389, 396-397` - the new tests re-implement `RunRecord.infra_aborted` instead of calling it.**
> `kstrl/autonomy_replay.py:113-115` is the consumer; `tests/test_autonomy_ladder.py:437` already exercises it as `runs[0].infra_aborted is True`. The class is named `TestTheTwoInfrastructureConsumersAgree` but asserts `"x:code".startswith(INFRA_FAILURE_PREFIXES)` inline, never touching the consumer. Cost: if `infra_aborted` stops reading `INFRA_FAILURE_PREFIXES` (which is what fix 1 would do), all four tests stay green while asserting nothing about the thing they are named after - the same blind-guard failure mode the file's own docstring cites as this repo's dominant defect.
>
> **3. `kstrl/evolution.py:410-413` - the four `"infrastructure"` rows are a third encoding of a fact the emitting call sites already carry twice.**
> At each of the four sites `pipeline` already passes `phase="aborted"/"token_budget"/"pr"/"diff"` next to the signature literal, and `kstrl/findings.py:82-87` (`Finding.is_infrastructure_error`) is the existing shared predicate for infrastructure-ness - `factory.py:2701` and `serve.py:902` both consume it, and `serve.py`'s docstring says explicitly that a second copy of the rule is how the taxonomies drift. The new table is a fourth list nobody derives from those. Measured cost, not hypothetical: `pipeline.py:1583` (token_budget) and `:2671` (diff) do build `Finding.infrastructure_error`, but `pipeline.py:1788` (`pr:merge-conflict`) and `:1378` (`aborted:shutdown`) build none. So after this change `evolution` calls those two runs infrastructure while `factory._infra_casualty` calls the same component a judged failure, and both feed a `decisive` definition (`factory.py:2707`, `autonomy_replay.py:124`). The change removes one two-way disagreement about `pr:merge-conflict` and opens another.
>
> **4. `kstrl/evolution.py:384-386` and `:388-422` - the key set now duplicates `context.PHASE_RANK` in full.**
> All seven `PHASE_RANK` keys (`kstrl/context.py:36-44`: engineer, verification, diff, review, security, contract, pr) are now keys of `_CATEGORY_BY_CHECK`, and four of the six `CATEGORIES` values are also `PHASE_RANK` keys. `context.py:229-230` already validates a phase against that dict and raises on an unknown one. Cost: a new phase needs a `PHASE_RANK` entry, a `_CATEGORY_BY_CHECK` row and possibly a `CATEGORIES` member, with no link between them - and the enrolment walk provably cannot see a runtime-composed phase (`ENROLLED_BUT_INVISIBLE`, `tests/test_check_name_enrolment.py:86-99`), so a new phase is precisely the case the new guard misses. Deriving the phase-named rows from `PHASE_RANK` would make that case red instead of invisible.
>
> **5. `tests/test_check_name_enrolment.py:86-99` vs `:307-319` - two hand-maintained lists of runtime-composed phases in one file, already disagreeing.**
> `ENROLLED_BUT_INVISIBLE` lists `security` and `contract` as "composed from a runtime phase argument"; `test_the_runtime_composed_phases_are_enrolled` iterates `("review", "security", "contract")` for the same reason, and the new docstring has to explain why `review` is in one and not the other. Cost: a new composed phase needs both edits, the divergence is carried as prose rather than as a derivation, and the second list is a literal tuple that nothing checks against the first.
>
> **6. `tests/test_check_name_enrolment.py:101-111` - `CATEGORY_DECIDED_IN_315` restates 8 of the 22 table rows.**
> Low value, and pinning data tables is a defensible pattern; flagging the shape only. Cost: a partial mirror of `_CATEGORY_BY_CHECK` keyed to an issue number rather than to a property, so 14 rows are unpinned and a future author has no rule telling them whether a new row belongs in it. `test_every_row_carries_a_category_that_exists` (`:321`) is the property-shaped guard and covers the whole table.
>
> Nothing in the diff duplicates `review.VALID_CONCERN_CATEGORIES` or `security.VALID_CATEGORIES`; `CATEGORIES` follows that existing convention rather than colliding with it, and `INFRASTRUCTURE_CHECKS` being derived from the table (rather than typed out) is the right call - findings 1 and 3 are about which taxonomy it derives from, not about the derivation itself.

Disposition: **2 applied** (every assertion in that class now goes through `RunRecord.infra_aborted`, and the helper says why). **6 applied**, harder than proposed: the partial mirror is now a whole-table pin, which subsumed the property-shaped guard, so that test is gone too. **1 declined**: `split_signature` and prefix matching are the same answer on every input the module can see, so this is a style preference, and the prefix tuple is a documented public constant with an existing rationale ("a new plumbing family is one entry, not a new code path"). Rewriting it widens the diff into behaviour-neutral churn in a module #315 only needed to reconcile. The colon test survives as `test_a_judgement_failure_still_counts_as_evidence`, which now earns its place as the other direction of the class. **3 declined as out of scope, and stated in the PR instead**: the `factory._infra_casualty` divergence is real and pre-existing for `pr:`, this PR adds `aborted:` to it, and closing it means editing `kstrl/factory.py`, which has an open PR (#332). It deserves its own issue. **4 declined**: deriving category rows from `context.PHASE_RANK` couples two taxonomies that agree by coincidence, not by construction (`PHASE_RANK` orders retry context; `_CATEGORY_BY_CHECK` classifies failures), and the round-2 fix addresses the underlying complaint directly by making the walk SEE phase-derived names. **5 applied**: the two lists no longer disagree, and the docstring says why the second still exists (a pin is satisfied by editing the expectation; the named test says why the row cannot go).

### Simplification

> **1. `tests/test_check_name_enrolment.py:307` - `test_the_runtime_composed_phases_are_enrolled` is now dead weight, and its docstring is now false.**
> The docstring still says "removing `security` from the table does not fail any other test here", and the diff appends a sentence about `review` instead of deleting that claim. Measured by popping each row from `_CATEGORY_BY_CHECK` and running the siblings: `security` and `contract` now redden `test_the_walk_covers_every_enrolled_name_but_these` (they are in `ENROLLED_BUT_INVISIBLE`), and `review` reddens `test_every_emitted_name_is_enrolled`. All three are held twice. Cost: a stale justification a future reader will trust, plus a third guard nobody can retire without re-deriving what I just derived. Simpler form: delete the test; `ENROLLED_BUT_INVISIBLE` already carries the same two names with the same reason string.
>
> **2. `tests/test_check_name_enrolment.py:364` and `:371` - the two `TestTheTwoInfrastructureConsumersAgree` agreement tests restate the constructor they are testing.**
> `INFRA_FAILURE_PREFIXES` is built as `{f"{c}:" for c in INFRASTRUCTURE_CHECKS} | set(_REPLAY_ONLY_PREFIXES)`, so `test_every_infrastructure_check_is_an_infra_abort_for_the_replay` cannot fail: `"X:any-code".startswith("X:")` is true for every derived member by construction. And the second test's LHS reduces to `set(_REPLAY_ONLY_PREFIXES) - derived`, so it can only fail on an overlap between the two halves. It cannot see a bogus prefix added to `_REPLAY_ONLY_PREFIXES`, which is precisely what its docstring promises to catch ("Anything beyond this list is an undocumented divergence"). Cost: two tests and a class docstring that read as policy coverage while pinning nothing, so the real content of the tuple is unpinned. Simpler and strictly stronger, one test replacing both: `assert set(INFRA_FAILURE_PREFIXES) == {f"{c}:" for c in INFRASTRUCTURE_CHECKS} | {"scope_unreadable:", "git:", "infra:", "timeout:"}` with those four spelled out in the test file.
>
> **3. `kstrl/autonomy_replay.py:50` - "the first four come from `evolution.INFRASTRUCTURE_CHECKS`" is false of the constant it annotates.**
> The tuple is `sorted(...)`, so it is `('aborted:', 'diff:', 'git:', 'infra:', 'pr:', 'scope_unreadable:', 'timeout:', 'token_budget:')` and the first four include two replay-only prefixes (`git:`, `infra:`) while excluding two derived ones (`pr:`, `token_budget:`). Cost: a reader who indexes the tuple to find the derived half gets the wrong half; the comment describes the pre-sort mental model. Fix: "four of them", or name them.
>
> **4. `tests/test_check_name_enrolment.py:343` - `assert name in _check_names()` duplicates the blind-guard test, eight times over.**
> Measured by hiding `pr` from the walk: `test_the_walk_covers_every_enrolled_name_but_these` already goes red, with the better diagnostic ("Newly invisible: ..., a refactor probably moved a name out of the shape the walk matches"). Cost: one cause reddens nine tests instead of one, and the parametrized message ("`pr` is no longer emitted by kstrl/") is the misleading one, since the usual reason is a walk that went blind, not a deleted call site. Simpler: keep only the `category_for_check(name) == category` line, which is the one thing the test's name promises.
>
> **5. `kstrl/evolution.py:384` - `CATEGORIES` is a new public constant that no code in `kstrl/` reads.**
> Sole consumer is `test_every_row_carries_a_category_that_exists`. Nothing validates `FailurePattern.category` against it at construction, so the new field comment "One of CATEGORIES" is a claim only a test upholds. It is also exactly `set(_CATEGORY_BY_CHECK.values())` today, so it earns its keep only as a hand-written list the guard compares against - and that list belongs beside the guard. Simpler: move the frozenset into the test file next to the assertion (identical guard strength), leaving `evolution.py` fourteen lines and one public name lighter. Keep the "not `Finding.category`'s `infrastructure_error`" paragraph, but attach it to `_CATEGORY_BY_CHECK`, where the spelling decision actually lives.
>
> **6. `kstrl/evolution.py:237` and `:487` - the same non-obvious fact is now written twice in near-identical words.**
> "decided by `category_for_check` ... never written to the journal, so correcting the table also corrects what is reported about runs already recorded" (field comment) and "The answer is computed here, at read time, and never stored in the journal, so a correction to the table also corrects what `ks evolve` reports about runs that already happened" (function docstring). Cost: two copies to keep in step, and the field comment is the copy nobody editing the function will see. Simpler: keep it on `category_for_check`, where the behaviour is, and reduce the field comment to "One of `CATEGORIES`; see `category_for_check`".
>
> Nothing here is a correctness bug: the 18 tests in the file pass, and `INFRA_FAILURE_PREFIXES` derives correctly.

Disposition: **2, 3, 4, 5 and 6 applied**, 4 by deleting the parametrized test altogether once the whole-table pin subsumed it, and 5 by deleting `CATEGORIES` and moving its prose onto the table. **1 declined in part**: the stale sentence is gone, but the test stays. A whole-table pin is satisfied by editing the expectation, so the named test is what tells the author WHY `security` and `contract` cannot go, and its cost is one assertion loop. The rewritten docstring says exactly that.

### Efficiency

> **No efficiency defects worth fixing.** Both of the specific concerns you flagged were measured and came back clean. Numbers below so the conclusion is auditable.
>
> **1. `kstrl/autonomy_replay.py:41` - the new `from kstrl.evolution import INFRASTRUCTURE_CHECKS` does not meaningfully cost anything.** Not a finding.
>
> - Raw A/B, 5 runs each, warm: `import kstrl.autonomy, kstrl.verify` (origin/main's exact dep set for this module) = **26.7-27.5 ms**; `import kstrl.autonomy_replay` at HEAD = **29.9-30.8 ms**. Delta approximately **3.1 ms**, split by `-X importtime` into `kstrl.observability` 759 us + `kstrl.evolution` own body 875 us (the 42.3 ms cumulative you see under `kstrl.evolution` is almost entirely `kstrl.verify`, which `autonomy_replay` already imported before this change).
> - **No CLI path imports `autonomy_replay` eagerly.** The only import site in `kstrl/` is `kstrl/cli.py:4464`, inside the `autonomy_replay_cmd` body. Repo-wide, the only other importers are two test files.
> - On that one real path the marginal cost is smaller still, because `kstrl.cli` already loads `observability` and `verify` (verified: after `import kstrl.cli`, `'kstrl.observability' in sys.modules` is `True`, `'kstrl.evolution'` is `False`). Measured `import kstrl.evolution` after `import kstrl.cli`: **1.02, 1.03, 1.04, 1.07 ms** (first run 4.38 ms, cold-cache outlier). Against `ks autonomy replay --help` wall time of **0.10-0.11 s**, that is approximately 1% of the single command that reaches it and 0% of every other command.
>
> **2. `tests/test_check_name_enrolment.py:290` - `test_the_walk_covers_every_enrolled_name_but_these` reuses the cached walk. No second parse.** Not a finding.
>
> - `uv run pytest tests/test_check_name_enrolment.py --durations=5`: **18 passed in 0.57s** (0.32 s on a warm re-run), with `test_the_walk_sees_the_gates_it_claims_to_guard` holding **0.24-0.33 s** of it and everything else under the 5 ms floor. At `--durations=0 -vv`, the new test and all five other new items report **0.00s**.
> - Direct proof: `ast.parse` monkeypatched around `pytest.main` counts **187** calls for the whole file against **127** `.py` files in `kstrl/`; `_parsed_modules` is `@lru_cache(maxsize=1)` and is the file's only `ast.parse` site (line 188). The package is parsed exactly once. Item count went 12 to 18; the 6 new items cost nothing measurable.
>
> Sub-noise observations, listed only for completeness
>
> 3. `kstrl/autonomy_replay.py:115` - the derived prefix tuple grew from 5 entries to 8, so `common_failure.startswith(INFRA_FAILURE_PREFIXES)` does more work. Measured worst case (no match, every prefix tried), 2M iterations: **39.6 ns to 44.6 ns, +5.0 ns per call**. `infra_aborted` is a non-cached property read approximately 4x per row in `replay`; a 1000-row `experiments.tsv` pays approximately 20 us total.
>
> 4. `kstrl/evolution.py:384` - `CATEGORIES` has no production consumer: `grep` finds it referenced only by a docstring comment and by `tests/test_check_name_enrolment.py:328`. It is a 6-element frozenset built at every `import kstrl.evolution`. Construction of both new module-level constants together measures **0.76 us per import**. Correct placement (module level, not per call); flagging only that it is a test-only export living in a production module.
>
> 5. `kstrl/evolution.py:309` - the `_classify_check` narrowing is a small net **win**, not a cost: dropping the discarded tuple element saves **2.4 ns per call** (37.8 to 35.4 ns, 2M iterations) on the legacy-fallback path.
>
> Nothing here is worth changing. The derived-constant work is all at module scope where it belongs, the AST walk is properly cached, and the one import the diff adds lands behind an already-lazy CLI import.

Disposition: nothing to apply. Observation 4 was independently raised by two other agents and is applied there (`CATEGORIES` is gone). Observation 3 is 5 ns and the tuple is now 9 entries rather than 8, which is 1 ns more of nothing.

### Altitude

> The change is mostly at the right depth: replacing a convention with a table plus a walk is the correct level, and killing `_classify_check`'s dead second return value is a genuine depth fix (one decision, one place). Five things sit at the wrong altitude.
>
> **1. The guard is written at the call-site-shape level, not at the choke point, and it already misses a third producer.**
> `tests/test_check_name_enrolment.py:81,159,244` / `kstrl/evolution.py:456`
>
> Three things in `kstrl/` produce a check prefix that reaches `category_for_check`. The walk covers two (`CheckResult`/`_failed_gate_result` name args, `signatures=[...]` literals). The third is `pipeline._record_failure_signatures` (`kstrl/pipeline.py:1291`), which falls back to `signature_for_error(phase or "unknown", error)` whenever `signatures=` is absent, so the `phase=` argument becomes the check name. Measured on this tree:
>
> ```
> provisioning   sig=provisioning:worktree-setup-failed   category=iteration  infra_abort=False
> verify         sig=verify:mechanical-verification-...   category=iteration  infra_abort=False
> ```
>
> Two live call sites hit it: `kstrl/factory.py:3413` (`pipeline.fail(..., phase="provisioning", check="worktree_setup")`, no `signatures=`) and `kstrl/pipeline.py:2150` (`retry_or_fail(..., phase="verify", check="diff_scope")`, no `signatures=`). Cost: a worktree-provisioning failure - the purest infrastructure failure in the system - is still filed under the engineer loop by `ks evolve` and still counts as a **decisive** run for autonomy promotion. That is exactly the defect #315 claims to close, and `test_every_emitted_name_is_enrolled` is green (18 passed) while it happens, next to a docstring saying "the guard now admits no exceptions at all". The altitude fix is to guard where the prefix is produced (`signature_for_error` / `_record_failure_signatures`) rather than enumerating call shapes, or at minimum to resolve `phase=` in `_check_names`.
>
> **2. Deriving the replay's predicate from the journal's category turns a taxonomy edit into an autonomy-policy edit - and it already fired, unannounced.**
> `kstrl/autonomy_replay.py:93`
>
> `INFRA_FAILURE_PREFIXES` went from 5 prefixes to 8: `aborted:`, `diff:` and `token_budget:` are new. That feeds `RunRecord.infra_aborted` to `decisive` to `decisive_runs` to `MIN_DECISIVE_RUNS`, i.e. the promotion gate. The docstrings argue only about `pr:merge-conflict`; nothing states that the replay's decisive-run definition just widened, and no test in `tests/test_autonomy_ladder.py` covers the three new prefixes. The pin that would catch a future recategorisation, `CATEGORY_DECIDED_IN_315` (`tests/test_check_name_enrolment.py:101,336`), is a bare `assert category_for_check(name) == category` whose docstring mentions only `ks evolve`; someone fixing a journal categorisation will satisfy it by editing the expectation and never learn they moved an autonomy threshold.
>
> **3. The taxonomy is in the journal module, and a peer consumer now imports it - the repo's own habit is a third, small module.**
> `kstrl/autonomy_replay.py:41`, `kstrl/evolution.py:429`
>
> `kstrl/` is full of tiny shared-vocabulary modules that exist precisely so two consumers do not import each other: `delimiters.py` (38 lines), `runid.py` (43), `gateparse.py` (119), `procgroup.py`, `atomicio.py`, `statedir.py`. Here a 300-line replay module takes a module-level dependency on a 1678-line journal/proposals module to get one frozenset; `kstrl/autonomy.py:781` keeps its own `from kstrl.evolution import ...` inside a function body. Concrete cost: `_CATEGORY_BY_CHECK` stays private while `INFRASTRUCTURE_CHECKS` is a public projection shaped for exactly one consumer, so the next consumer that needs "which checks are verification" gets a second bespoke export plus a second bespoke test class. `checks_in_category(category)`, or a `kstrl/failure_taxonomy.py` that both import, costs the same to write today.
>
> **4. `CATEGORIES` as a bare `frozenset[str]` is against the house pattern, and it exists only to feed a test.**
> `kstrl/evolution.py:384`
>
> Measured: 14 `StrEnum` classes in `kstrl/`. The two closest analogues both use *StrEnum as the authority plus a frozenset derived from it* - `kstrl/divergence.py:155` (`_VALID_MODES = frozenset(m.value for m in DivergenceMode)`) and `kstrl/review.py:65` (`VALID_CRITERION_VERDICTS = frozenset({ReviewVerdict.PASS.value, ...})`). Bare `frozenset[str]` is reserved for vocabularies kstrl *validates but does not author*: LLM output (`review.VALID_CONCERN_CATEGORIES`, `security.VALID_CATEGORIES`, `knowledge.VALID_SCOPES`) and config strings (`linear._VALID_AUTH_MODES`). These six categories are authored in-house, as literals, in the same file. Cost: `CATEGORIES` has exactly one consumer, the new `test_every_row_carries_a_category_that_exists` (`tests/test_check_name_enrolment.py:321`), which exists only because the table's values are unchecked `str`. Typing the table `dict[str, Category]` makes a typo a `mypy --strict` error at the edit site and deletes both the constant and the test.
>
> **5. The `scope_unreadable` divergence is the model missing an attribute, not a special case worth three docstrings.**
> `kstrl/autonomy_replay.py:86`, `tests/test_check_name_enrolment.py:364,371`
>
> A check has two independent properties: which part of the factory produced the failure (category), and whether the run said anything about the factory's judgement (decisive). Expressing the second as "category == infrastructure, plus a hand-kept remainder tuple" is precisely why `scope_unreadable` needs an essay to defend. One table with two columns per check states it once and deletes `_REPLAY_ONLY_PREFIXES`, the derivation, and the divergence tests. Supporting evidence that the current shape proves little: `test_every_infrastructure_check_is_an_infra_abort_for_the_replay` (line 364) cannot fail while `INFRA_FAILURE_PREFIXES` is built from `INFRASTRUCTURE_CHECKS` on line 93, and `test_the_replays_extra_prefixes_are_exactly_the_documented_ones` (line 371) is `(A | B) - A == B`. Both pin the derivation, not the taxonomy. `test_the_prefixes_cannot_swallow_a_neighbouring_check` is the one in that class that earns its place.

Disposition: **1 applied**, and it is the most valuable finding of the four reviews. Verified independently before acting: the phase fallback is real, the two sites are real, and both names were unenrolled. The walk now resolves `phase=`, and `provisioning` and `verify` are enrolled. I took the narrower of the two suggested fixes (resolve `phase=` in the walk, not a guard inside `signature_for_error`) because the walk is where the other two producers are already covered and a runtime guard cannot fail at review time. **2 applied**: the widening is now stated in the PR body in the "what a user sees differently" section with the before and after tuple, and the pin's own comment says an added row can move a run out of the replay's evidence. **5 applied in part** (the two tautologies are gone) and **declined in part**: a second column per check is a redesign of a taxonomy this issue was asked to correct, not replace, and it would move `scope_unreadable` again, which #294 settled deliberately. The divergence is stated once, in the constant that holds it, and tested through the consumer. **3 declined**: a new module for one frozenset adds a mechanism to remove an import, and the import is measured at 1 ms behind an already-lazy CLI import. If a second consumer appears, that is the moment the module earns its keep. **4 declined**: the whole-table pin makes the runtime check unnecessary rather than making it stronger, so the frozenset and its test are deleted rather than replaced by an enum. A `Category` StrEnum would type-ripple into `FailurePattern.category`, which is written to no schema and read by exactly one Rich `Text()` call. The house-pattern argument is fair and this is a judgement call.

---

## Round 3: the altitude pass, and where this stops

Six findings from an altitude review, all measured. Four were do-now. The round was then scoped down by the owner mid-flight, for a process reason worth recording: every PR in this batch hand-rolls its own AST matcher and hand-rolls the same hole. Instance 8 was #336, instance 9 was round 1 here, instance 10 is the producer below. That is not converging, so **#324 (one shared `tests/helpers/astwalk.py`) is being pulled forward to close the class in one place** and this PR stops widening its own matcher.

### A1. A third caller of the funnel, and a live check name nothing guarded

`pipeline._record_failure_signatures` files a failure under its PHASE whenever no truthy `signatures` reaches it. Round 2 keyed the guard on `fail` and `retry_or_fail`, which are two of that funnel's THREE callers. The third is `_fail_pr_flow`:

```python
self._record_failure_signatures(comp, "pr", comp.error, None)
```

Positional, so no `phase=` keyword. Not a `fail`/`retry_or_fail` call. Not an assignment into a `*failure_signatures` mapping. And `"pr"` has no colon, so the spellings net could not see it either. `_producer_sites` returned nothing for that node: neither resolved, nor recorded as a blind site. Mutating `"pr"` to `"bogus_flow"` left the suite green.

The producer is now keyed on the funnel itself, the way `test_journal_one_writer.py` keys on `append_entries` being the only writer, and on `PhaseFailure`, whose `phase` reaches the same fallback through `_route_failure`.

Two things had to move with it.

**The signatures test ignored the value.** `"signatures" in keywords` was true for a literal `None`, which silenced the phase, and true for `signatures=failure.signatures` at `_route_failure`, where the value is decided at run time and the fallback really can fire. It is now `_signatures_taken`, three-valued: `True` means the recorder provably takes them and the phase is unreachable, `False` means provably nothing is there, `None` means neither producer may skip.

**The positional index is derived, not written down.** The first version of this fix hardcoded four integers. The altitude pass measured what they bought: six of eight off-by-one mutations to them left every guard green, because 25 of the 27 call sites in `kstrl/` pass `phase` by keyword and the two that do not resolve to nothing. An index is a fact about the definition, so `astfold.parameter_index` reads it off the `def` (or the dataclass body). It also keeps `feature_cmd`'s one-argument local `fail` out of the census **by construction**, deleting the special case the integers needed a paragraph to defend.

Census 21 to 23 names. `BLIND_SITES` gains five pass-through rows, each with a stated reason.

### The distinction this round is actually about

`BLIND_SITES` looked like `EXPECTED_JOURNAL_PATH_SITES` and does not behave like it, and that resemblance is what let `_fail_pr_flow` survive two review rounds.

- `EXPECTED_JOURNAL_PATH_SITES` inventories **every place the resource is obtained**. A new way of obtaining it lands in the list whatever it looks like. Closed by construction.
- `BLIND_SITES` inventories **the places the walk gave up**. It is closed only over the producer shapes the walk already enumerates. A producer written in a shape neither matcher matches is not resolved AND not recorded: it is silence, and the ledger staying the same length is not evidence of anything.

That is now written in the `BLIND_SITES` docstring itself, naming #324 as the tracking item, so the next reader hits it where they would otherwise inherit the wrong prior.

It is also pinned rather than merely disclosed. `tests/test_signature_chokepoint.py` is the one guard here with the precedent's guarantee: a check name cannot reach `category_for_check` without touching a `*failure_signatures` name, and there are five such sites in `kstrl/`. Measured both ways:

- The day `_fail_pr_flow` was written, `self._record_failure_signatures` went from 2 occurrences to 3 and this pin would have gone red, with no matcher to widen and no shape to recognise.
- A new writer reaching the mapping through a **local alias** with a **runtime-built** signature is invisible to the census, to `BLIND_SITES`, to the shapes file and to the spellings net. It fails this pin and nothing else. Measured by planting it.

### A4. The two sibling branches that still had P2-1's bug

All four `phase="pr"` HITL-adjacent outcomes, measured before this round:

| outcome | `infra_aborted` | `decisive` | verdict |
|---|---|---|---|
| `pr:rejected-at-hitl-checkpoint` | True | False | fixed by P2-1 |
| `pr:retry-requested-at-hitl-checkpoint` | True | False | NOT fixed |
| `pr:parked-awaiting-merge-approval-...` | True | False | NOT fixed |
| `pr:pr-merge-failed-conflict-with-base` | True | False | genuinely plumbing |

A human asking for changes is a verdict on the change by this PR's own argument, and it was being discarded 46 lines below the fix. It now carries `signatures=["review:hitl-changes-requested"]`. PARKED deliberately keeps the `pr:` fallback with a comment saying why: nobody answered the gate, so there is no verdict to record.

The rule is stated **once**, at the first of the three branches, and the other two point at it. Three copies of one rule in prose is the signature of a missing mechanism, and the mechanism is keying the fallback on `check` rather than `phase` (the branches already carry `hitl_reject` / `merge_gate` / `hitl_retry`, which is exactly the vocabulary that distinguishes them, while all three carry `phase="pr"`). That is blocked on `factory.py`.

One claim was also corrected rather than defended: "the phase stays `pr` because that is where it happened" is not true by the module's own accounting. `_phase_started(comp, "pr")` fires below all three checkpoint branches.

### A5. Three of the four replay-only prefixes can never match

Only `scope_unreadable:` is live. `git:`, `infra:` and `timeout:` were re-measured four ways: none appears in `kstrl/` as a signature prefix, none is in `_CATEGORY_BY_CHECK`, none is in `_classify_check`'s range, and none is producible by any enumerated producer. The comment now says which one is live at the **top** of the block.

They are not deleted. The earlier draft defended keeping them by saying deletion would change how historical rows are read, and that cannot be true alongside the measurement above it: if no version ever emitted the strings, no `experiments.tsv` row contains them. The false sentence is gone. They stay because a prefix that matches nothing costs nothing, and because removing them is a decision about what the constant is for, worth making on its own rather than inside a categorisation fix.

### A3 and A6: settled, with the reasons recorded

**A3** (a named ledger type for `component_failure_signatures`) is blocked on `factory.py`, which #332 holds. The pipeline-only half would leave two writers outside the ledger, strictly worse than the current suffix match. Worth recording: what holds that producer is the invisibility pin, not the suffix matcher, verified by mutating the container name to an alias and watching two tests go red.

**A6** (`_from_the_callers`) stays. It is a general rule that fails safe into `BLIND_SITES` and was measured red. Its docstring now says it assumes callers call by BARE NAME, so routing through a dispatch table or a `functools.partial` would quietly return the row to `BLIND_SITES`.

## The producer table, with the mutation that proves each row

Every mutation ran after `find . -name __pycache__ -type d -exec rm -rf {} +` with `PYTHONDONTWRITEBYTECODE=1`, because CPython invalidates bytecode on `(mtime_seconds, size)` and two same-size edits inside one second reuse stale bytecode. Every file was restored to its pre-mutation SHA-256, checked.

| what was mutated | result |
|---|---|
| `_fail_pr_flow`'s `phase="pr"` to `"bogus_flow"` | RED: `test_every_emitted_name_is_enrolled` |
| `hitl_retry`'s `signatures=` deleted | RED: `test_checkpoint_retry_consumes_a_retry` |
| `hitl_reject`'s `signatures=` deleted | RED: `test_checkpoint_reject_fails_component` |
| `_record_failure_signatures` dropped from `PHASE_FALLBACK_CALLS` | RED, 5 failed |
| `PhaseFailure` dropped from `PHASE_FALLBACK_CALLS` | RED, 2 failed |
| the positional fallback dropped (`index = None`) | RED, 4 failed |
| `_signatures_taken` always True | RED, 8 failed |
| `_failed_gate_result` dropped from `CHECK_NAME_CALLS` | RED, 1 failed |
| the `ast.Assign` producer branch deleted | RED, 7 failed |
| `_from_the_callers` returns nothing | RED, 1 failed |
| `parameter_index` off by one | RED, 5 failed |
| the receiver no longer excluded from the index | RED, 5 failed |
| dataclass fields ignored by the index | RED, 1 failed |
| a fourth caller of the funnel planted, literal phase | RED, 2 failed |
| a new caller planted whose phase is computed | RED, 2 failed (chokepoint + ledger) |
| a new writer planted via a local alias, runtime-built signature | RED, 1 failed: **the chokepoint pin alone** |

## P2-5: `ks autonomy replay` can change its exit status on unchanged data

Disclosure, not a code change. `kstrl/cli.py` ends that command with `sys.exit(0 if report.sufficient_data else 2)`, and this PR moves runs out of `decisive_runs`, which is what `sufficient_data` counts. A scripted caller treating non-zero as failure can start failing on an `experiments.tsv` that has not changed.

Measured, the same file replayed on `origin/main` and on this branch:

```
10 rows, all dominated by provisioning:worktree-setup-failed
  origin/main  decisive=10  infra_aborted=0   Final L2  sufficient_data=True   exit=0
  this branch  decisive=0   infra_aborted=10  Final L1  sufficient_data=False  exit=2
```

Being precise about what that file is: it is constructed, not a captured run. On the only two real `experiments.tsv` files available, this branch changes **nothing**. The live 5-row file reports `decisive=2 infra_aborted=3 exit=2` on both sides; the 856-row archive reports `decisive=800 infra_aborted=0 merged=737 L2 exit=0` on both sides, with **zero rows** moving verdict. Those rows predate the four newly-derived prefixes: their `common_failure` values are legacy slugs with no colon (`test-failure`, `mechanical-verification-failed`, `did-not-complete`), so no prefix matches on either side. The exit-status change is real, but it needs a row dominated by `aborted:`, `diff:`, `provisioning:` or `token_budget:` to fire.

## P2-2 and P2-4, both landed

**P2-2**: a Ctrl-C that discarded a run which had merged 24 components still printed "Components merged: 24" two lines under "infra-aborted: 1 (excluded)". Two populations under one label, and the bigger one counted for nothing. The report now says `Components merged: N (in decisive runs)` and adds `in excluded runs: N (not counted)` when there is something to disclose, so the total stays recoverable.

**P2-4**: the comment at `evolution.py` claimed one live/replay disagreement. There are six, in both directions, and they are now counted rather than exemplified.

## Deliberately left for #324

Named so nobody re-finds them and thinks they are new:

1. **`BLIND_SITES` is not closed over unenumerated producer shapes.** Disclosed in its docstring, backed by the chokepoint pin, not otherwise fixed here.
2. **`tests/test_check_name_shapes.py` is a fifth bespoke positive-control file.** It was already written when the scope call came, so it stays and was not extended. It pins matcher branches usefully, but it enumerates shapes its author already thought of, so it cannot close the unknown-unknown gap it was written for. #324 replaces the need for it.
3. **Two AST string folders in `tests/`.** `astfold.fold` subsumes `test_journal_one_writer.folded_str`: measured, they agree on all 22,738 name-free expressions in `kstrl/`, and all 471 disagreements across 119,605 expressions are the deliberate name-resolution branch. The strict answer is a two-line wrapper over the new one. Not merged here because that file guards #312 and changing what it folds is that guard's own measurement to make. The `astfold` docstring now says so rather than implying the copies must stay.
4. **`called_name` is the fifth copy of a three-line helper** (`test_procgroup.py`, `test_atomicio.py`, `test_process_scoping.py`, `test_timeout_enforcement.py` have the other four). Exporting one and deleting four touches files a parallel subprocess sweep holds.
5. **Two independent parse caches over the same 124 files**, one in `astfold`, one in `test_journal_one_writer`, each with its own measurement paragraph. Roughly 162 ms of avoidable work per session by the existing helper's own number, plus two vocabularies for the same file (`kstrl/pipeline.py` vs `pipeline.py`) across guards that cite each other.
6. **A third qualified-scope walk.** `astfold.scoped_nodes` is the better implementation (measured: it covers 18,066 nodes `test_tui_config_walk.own_nodes` never yields), but retiring the two older copies onto it changes what those guards report and needs their own measurement.

## Known coverage gap, pre-existing and unchanged (P3-3)

Five of the eight `return` statements in `evolution._classify_check` are never executed by the suite: rewriting all five to `return "linter"` left it green. Identical on `origin/main`, 38 missed statements on both sides, so this PR neither creates nor widens it. Recorded here rather than filed, per the move to class sweeps.

### DIFF
--- a/kstrl/autonomy_replay.py
+++ b/kstrl/autonomy_replay.py
@@ -43,23 +43,13 @@ from kstrl.verify import SCOPE_UNREADABLE_CHECK
 
 DEFAULT_EXPERIMENTS_PATH = Path(".kstrl/experiments.tsv")
 
-#: `common_failure` prefixes that mark an infrastructure casualty rather
-#: than a judgement. Kept as a prefix tuple so a new plumbing failure
-#: family is one entry, not a new code path.
+#: The prefixes this module adds to the ones the journal's own
+#: taxonomy supplies. It asks a WIDER question than that taxonomy does:
+#: the journal asks which part of the factory produced a failure, this
+#: asks whether the run yielded a verdict about the factory's JUDGEMENT
+#: at all, and a failure can be a gate's honest verdict and still answer
+#: that with no.
 #:
-#: #315: the first four come from `evolution.INFRASTRUCTURE_CHECKS`
-#: rather than being retyped here. The two consumers disagreed about
-#: `pr:merge-conflict` - plumbing to this module, an engineer-loop
-#: failure to the journal - and a taxonomy that answers a question twice
-#: will eventually answer it two ways. Deriving means enrolling a check
-#: as infrastructure in `_CATEGORY_BY_CHECK` reaches both in one edit,
-#: and the enrolment guard forces every emitted name through that table.
-#:
-#: _REPLAY_ONLY_PREFIXES is the deliberate remainder: this module asks a
-#: WIDER question than the journal's category does. The journal asks
-#: which part of the factory produced a failure; this asks whether the
-#: run yielded a verdict about the factory's JUDGEMENT at all, and a
-#: failure can be a gate's honest verdict and still answer that with no.
 #: `scope_unreadable:` (#294) is the harness failing to establish its own
 #: input: the component's scope could not be read from the pre-run tree,
 #: so nothing was ever measured about the change. The same reasoning that
@@ -90,6 +80,24 @@ _REPLAY_ONLY_PREFIXES: tuple[str, ...] = (
     "timeout:",
 )
 
+#: `common_failure` prefixes that mark an infrastructure casualty rather
+#: than a judgement. Kept as a prefix tuple so a new plumbing failure
+#: family is one entry, not a new code path.
+#:
+#: #315: every check the journal files as `infrastructure` is in here by
+#: construction, rather than being retyped. The two consumers disagreed
+#: about `pr:merge-conflict` - plumbing to this module, an engineer-loop
+#: failure to the journal - and a taxonomy that answers a question twice
+#: will eventually answer it two ways. Deriving means enrolling a check
+#: as infrastructure in `_CATEGORY_BY_CHECK` reaches both in one edit,
+#: and the enrolment guard forces every emitted name through that table.
+#: `_REPLAY_ONLY_PREFIXES` above is the deliberate remainder. The sorted
+#: tuple interleaves the two halves, so read the two sources, not the
+#: order.
+#:
+#: `tests/test_check_name_enrolment.py` pins the contents through
+#: `RunRecord.infra_aborted` rather than against this tuple, so a
+#: property that stopped consulting it would be caught.
 INFRA_FAILURE_PREFIXES: tuple[str, ...] = tuple(
     sorted({f"{check}:" for check in INFRASTRUCTURE_CHECKS} | set(_REPLAY_ONLY_PREFIXES))
 )
--- a/kstrl/autonomy_replay.py
+++ b/kstrl/autonomy_replay.py
@@ -38,15 +38,28 @@ from kstrl.autonomy import (
     AutonomyState,
     DemotionTrigger,
 )
+from kstrl.evolution import INFRASTRUCTURE_CHECKS
 from kstrl.verify import SCOPE_UNREADABLE_CHECK
 
 DEFAULT_EXPERIMENTS_PATH = Path(".kstrl/experiments.tsv")
 
 #: `common_failure` prefixes that mark an infrastructure casualty rather
-#: than a judgement. Derived from the recorded failure taxonomy: `pr:` is
-#: push/create/merge plumbing. Kept as a prefix tuple so a new plumbing
-#: failure family is one entry, not a new code path.
+#: than a judgement. Kept as a prefix tuple so a new plumbing failure
+#: family is one entry, not a new code path.
 #:
+#: #315: the first four come from `evolution.INFRASTRUCTURE_CHECKS`
+#: rather than being retyped here. The two consumers disagreed about
+#: `pr:merge-conflict` - plumbing to this module, an engineer-loop
+#: failure to the journal - and a taxonomy that answers a question twice
+#: will eventually answer it two ways. Deriving means enrolling a check
+#: as infrastructure in `_CATEGORY_BY_CHECK` reaches both in one edit,
+#: and the enrolment guard forces every emitted name through that table.
+#:
+#: _REPLAY_ONLY_PREFIXES is the deliberate remainder: this module asks a
+#: WIDER question than the journal's category does. The journal asks
+#: which part of the factory produced a failure; this asks whether the
+#: run yielded a verdict about the factory's JUDGEMENT at all, and a
+#: failure can be a gate's honest verdict and still answer that with no.
 #: `scope_unreadable:` (#294) is the harness failing to establish its own
 #: input: the component's scope could not be read from the pre-run tree,
 #: so nothing was ever measured about the change. The same reasoning that
@@ -56,13 +69,29 @@ DEFAULT_EXPERIMENTS_PATH = Path(".kstrl/experiments.tsv")
 #: evidence about the factory's judgement. #294 makes that more likely
 #: rather than less, because refusing the component before its engineer
 #: runs makes this the modal signature of the run instead of one of
-#: three retries.
-INFRA_FAILURE_PREFIXES: tuple[str, ...] = (
-    "pr:",
+#: three retries. The journal keeps it under `verification` all the same:
+#: it IS a Phase 1 gate result, and #294 gave it its own gate, its own
+#: table row and its own proposal arm on that basis. That divergence is
+#: the point of splitting this constant in two, and
+#: `tests/test_check_name_enrolment.py` pins both halves of it.
+#:
+#: `git:`, `infra:` and `timeout:` name families no version of kstrl has
+#: ever emitted: measured with `git log -S`, all three strings entered
+#: the tree in 606cbb1, the commit that created this tuple, and no call
+#: site produces them. Kept because an experiments.tsv row outlives the
+#: code that wrote it and the cost of a prefix that matches nothing is
+#: zero. They are NOT enrolled in `_CATEGORY_BY_CHECK`: that table
+#: describes names kstrl emits, and the enrolment guard would make any
+#: real `git:` gate enrol itself on the day it lands.
+_REPLAY_ONLY_PREFIXES: tuple[str, ...] = (
+    f"{SCOPE_UNREADABLE_CHECK}:",
     "git:",
     "infra:",
     "timeout:",
-    f"{SCOPE_UNREADABLE_CHECK}:",
+)
+
+INFRA_FAILURE_PREFIXES: tuple[str, ...] = tuple(
+    sorted({f"{check}:" for check in INFRASTRUCTURE_CHECKS} | set(_REPLAY_ONLY_PREFIXES))
 )
 
 
--- a/kstrl/autonomy_replay.py
+++ b/kstrl/autonomy_replay.py
@@ -50,6 +50,13 @@ DEFAULT_EXPERIMENTS_PATH = Path(".kstrl/experiments.tsv")
 #: at all, and a failure can be a gate's honest verdict and still answer
 #: that with no.
 #:
+#: ONE OF THE FOUR IS LIVE. `scope_unreadable:` is emitted by kstrl and
+#: is the reason this constant exists; `git:`, `infra:` and `timeout:`
+#: match nothing this code can produce, and the paragraph below says how
+#: that was measured. Stated here as well as there because a reader who
+#: takes the list at face value inherits the wrong prior about how much
+#: of the ladder's evidence these four remove.
+#:
 #: `scope_unreadable:` (#294) is the harness failing to establish its own
 #: input: the component's scope could not be read from the pre-run tree,
 #: so nothing was ever measured about the change. The same reasoning that
@@ -63,14 +70,28 @@ DEFAULT_EXPERIMENTS_PATH = Path(".kstrl/experiments.tsv")
 #: it IS a Phase 1 gate result, and #294 gave it its own gate, its own
 #: table row and its own proposal arm on that basis. That divergence is
 #: the point of splitting this constant in two, and
-#: `tests/test_check_name_enrolment.py` pins both halves of it.
+#: `tests/test_infrastructure_category_consumers.py` pins both halves of
+#: it, through the property rather than against this tuple.
 #:
 #: `git:`, `infra:` and `timeout:` name families no version of kstrl has
 #: ever emitted: measured with `git log -S`, all three strings entered
 #: the tree in 606cbb1, the commit that created this tuple, and no call
-#: site produces them. Kept because an experiments.tsv row outlives the
-#: code that wrote it and the cost of a prefix that matches nothing is
-#: zero. They are NOT enrolled in `_CATEGORY_BY_CHECK`: that table
+#: site produces them. #339 re-measured it four ways and got the same
+#: answer: none of the three appears in `kstrl/` as a signature prefix,
+#: none is in `_CATEGORY_BY_CHECK`, none is in `_classify_check`'s range,
+#: and none is producible by any of the four check-name producers
+#: `tests/test_check_name_enrolment.py` enumerates.
+#:
+#: Kept, and #339 review is why that needs saying carefully: an earlier
+#: draft of this comment defended them as protecting historical rows,
+#: which cannot be true alongside the sentence above it. If no version
+#: ever emitted the strings then no experiments.tsv row contains them,
+#: and deleting them would change nothing that exists. They stay because
+#: a prefix that matches nothing costs nothing to carry and because
+#: removing them is a decision about what this constant is FOR - whether
+#: it guards against families kstrl has never written - which is worth
+#: making on its own rather than as a tidy-up inside a categorisation
+#: fix. They are NOT enrolled in `_CATEGORY_BY_CHECK`: that table
 #: describes names kstrl emits, and the enrolment guard would make any
 #: real `git:` gate enrol itself on the day it lands.
 _REPLAY_ONLY_PREFIXES: tuple[str, ...] = (
@@ -95,8 +116,8 @@ _REPLAY_ONLY_PREFIXES: tuple[str, ...] = (
 #: tuple interleaves the two halves, so read the two sources, not the
 #: order.
 #:
-#: `tests/test_check_name_enrolment.py` pins the contents through
-#: `RunRecord.infra_aborted` rather than against this tuple, so a
+#: `tests/test_infrastructure_category_consumers.py` pins the contents
+#: through `RunRecord.infra_aborted` rather than against this tuple, so a
 #: property that stopped consulting it would be caught.
 INFRA_FAILURE_PREFIXES: tuple[str, ...] = tuple(
     sorted({f"{check}:" for check in INFRASTRUCTURE_CHECKS} | set(_REPLAY_ONLY_PREFIXES))
@@ -181,7 +202,22 @@ class ReplayReport:
     decisive_runs: int
     infra_aborted_runs: int
     projects: list[str]
+    #: Merges in DECISIVE runs only, which is the population the ladder
+    #: promotes on. #339: this used to sum every row, so a run Ctrl-C'd
+    #: after merging 24 components printed "Components merged: 24" two
+    #: lines under "infra-aborted: 1 (excluded)" while contributing
+    #: nothing to a promotion. Two populations, one label.
     components_merged: int
+    #: The merges the line above deliberately leaves out, so the total is
+    #: still recoverable and the exclusion is visible rather than a
+    #: number that quietly got smaller. Rendered only when non-zero,
+    #: unlike the infra-aborted count above it, because a report with
+    #: nothing excluded has nothing to disclose.
+    #:
+    #: No default: a hand-built report would otherwise claim silently
+    #: that nothing was excluded, which is the exact under-report this
+    #: field exists to stop.
+    merged_in_excluded_runs: int
     would_promote: list[str] = field(default_factory=list)
     would_demote: list[str] = field(default_factory=list)
     final_level: int = int(AutonomyLevel.L1_SUPERVISED)
@@ -200,7 +236,12 @@ class ReplayReport:
             f"Runs recorded:        {self.total_runs}",
             f"  decisive:           {self.decisive_runs}",
             f"  infra-aborted:      {self.infra_aborted_runs} (excluded)",
-            f"Components merged:    {self.components_merged}",
+            f"Components merged:    {self.components_merged} (in decisive runs)",
+            *(
+                [f"  in excluded runs:   {self.merged_in_excluded_runs} (not counted)"]
+                if self.merged_in_excluded_runs
+                else []
+            ),
             f"Projects:             {', '.join(self.projects) or '(none)'}",
             "",
             "Thresholds replayed (ALL UNMEASURED PLACEHOLDERS):",
@@ -254,7 +295,8 @@ def replay(runs: list[RunRecord]) -> ReplayReport:
         decisive_runs=sum(1 for r in runs if r.decisive),
         infra_aborted_runs=sum(1 for r in runs if r.infra_aborted),
         projects=sorted({r.project for r in runs if r.project}),
-        components_merged=sum(r.completed for r in runs),
+        components_merged=sum(r.completed for r in runs if r.decisive),
+        merged_in_excluded_runs=sum(r.completed for r in runs if not r.decisive),
     )
     for run in runs:
         if not run.decisive:
--- a/kstrl/evolution.py
+++ b/kstrl/evolution.py
@@ -234,9 +234,7 @@ class FailurePattern:
     # "scope_creep" for a review concern) - the part after the colon in
     # the full "<check>:<code>" signature
     error_signature: str
-    # One of CATEGORIES, decided by category_for_check when the pattern
-    # is built and never written to the journal, so correcting the table
-    # also corrects what is reported about runs already recorded.
+    # A _CATEGORY_BY_CHECK value; see category_for_check.
     category: str
 
 
@@ -367,23 +365,29 @@ def _classify_check(error: str) -> str:
 # difference is the number.
 _DIGIT_RUN_RE = re.compile(r"\d+")
 
-#: The categories a FailurePattern can carry. ``"infrastructure"``
-#: (#315) is for a failure that is neither a gate's verdict on the
-#: change nor the engineer's loop: the run was shut down, hit the token
-#: ceiling, could not merge, could not fetch its own diff. Before it
-#: existed those fell through to ``"iteration"`` and ``ks evolve``
-#: reported them as engineer-loop problems, which is advice aimed at an
-#: agent that could not have prevented any of them.
-#:
-#: Not ``Finding.category``'s ``"infrastructure_error"``
-#: (:mod:`kstrl.findings`), which marks one ROLE RUN that failed to
-#: execute inside an otherwise healthy component. Different taxonomy,
-#: different consumer, deliberately different spelling so a grep for
-#: either does not silently return the other.
-CATEGORIES: frozenset[str] = frozenset(
-    {"verification", "review", "security", "contract", "infrastructure", "iteration"}
-)
-
+# The categories are verification, review, security, contract,
+# iteration and, since #315, infrastructure. That last one is for a
+# failure that is neither a gate's verdict on the change nor the
+# engineer's loop: the run was shut down, hit the token ceiling, could
+# not merge, could not fetch its own diff, could not provision a
+# worktree. Before it existed those fell through to "iteration", so the
+# journal filed them as engineer-loop problems and the evolve screen's
+# category column said so: a verdict about an agent that could not have
+# prevented any of them.
+#
+# NOT Finding.category's "infrastructure_error" (kstrl/findings.py),
+# which marks one ROLE RUN that failed to execute inside an otherwise
+# healthy component. Different taxonomy, different consumer,
+# deliberately different spelling so a grep for either does not silently
+# return the other. They also disagree, on purpose and measurably:
+# factory's live autonomy accounting asks the FINDING question
+# (`_infra_casualty`) while the replay asks the SIGNATURE question, and
+# `pr:merge-conflict` attaches no infrastructure finding. Reconciling
+# those two is not this table's job.
+#
+# tests/test_check_name_enrolment.py pins the whole table row by row, so
+# a new row, a dropped row or a typo in a category is a red test with
+# the diff as its audit trail.
 _CATEGORY_BY_CHECK = {
     "linter": "verification",
     "typecheck": "verification",
@@ -404,17 +408,30 @@ _CATEGORY_BY_CHECK = {
     "fixtures": "verification",
     "policy_envelope": "verification",
     "test_adequacy": "verification",
+    # #315 round 2: a failure recorded with no signatures= is filed
+    # under its PHASE (pipeline._record_failure_signatures), so these
+    # two are check names as much as any gate is. "verify" is the phase
+    # spelling of the mechanical gates above; "provisioning" is a
+    # worktree that would not build, which no agent can write code
+    # against.
+    "verify": "verification",
+    "provisioning": "infrastructure",
     # #315: recorded outside a CheckResult, by pipeline.fail(signatures=
     # ["<prefix>:<code>"]). None of these is a verdict on the change.
     "aborted": "infrastructure",
     "token_budget": "infrastructure",
     "pr": "infrastructure",
     "diff": "infrastructure",
-    # #315: the fallback already answers "iteration" for this one. The
-    # row is here so the table states every name kstrl emits rather than
-    # most of them, and so that a reader cannot tell an unenrolled name
-    # from a deliberate one by its absence.
+    # #315: the fallback already answers "iteration" for these two. The
+    # rows are here so the table states every name kstrl emits rather
+    # than most of them, and so that a reader cannot tell an unenrolled
+    # name from a deliberate one by its absence. "unknown" is what
+    # _classify_check returns when it cannot recognise a legacy error
+    # string: not the engineer's fault so much as nobody's, and inventing
+    # a category for "we could not tell" would be a worse answer than
+    # the one this table has always given.
     "engineer": "iteration",
+    "unknown": "iteration",
     "review": "review",
     "security": "security",
     "contract": "contract",
@@ -484,8 +501,11 @@ def category_for_check(check_name: str) -> str:
     uncategorised.
 
     The answer is computed here, at read time, and never stored in the
-    journal, so a correction to the table also corrects what
-    ``ks evolve`` reports about runs that already happened.
+    journal - measured: a ``record_run`` entry has no ``category`` key
+    and experiments.tsv records the signature only. So a correction to
+    the table also corrects what is reported about runs that already
+    happened. The one surface that displays it is the evolve screen's
+    patterns table; the ``ks evolve`` CLI prints the check name.
     """
     return _CATEGORY_BY_CHECK.get(check_name, "iteration")
 
--- a/kstrl/evolution.py
+++ b/kstrl/evolution.py
@@ -379,11 +379,26 @@ _DIGIT_RUN_RE = re.compile(r"\d+")
 # which marks one ROLE RUN that failed to execute inside an otherwise
 # healthy component. Different taxonomy, different consumer,
 # deliberately different spelling so a grep for either does not silently
-# return the other. They also disagree, on purpose and measurably:
-# factory's live autonomy accounting asks the FINDING question
-# (`_infra_casualty`) while the replay asks the SIGNATURE question, and
-# `pr:merge-conflict` attaches no infrastructure finding. Reconciling
-# those two is not this table's job.
+# return the other.
+#
+# They also disagree, on purpose and measurably. factory's live autonomy
+# accounting asks the FINDING question (`_infra_casualty`) while the
+# replay asks the SIGNATURE question, and #339 review counted the
+# divergence rather than leaving it at the one example this comment used
+# to give. Six signatures, in both directions:
+#
+#   - the whole `pr:` family, `provisioning:` and `aborted:shutdown` are
+#     infrastructure to the replay and attach no finding at all, so the
+#     live side calls them judgement;
+#   - `review:infrastructure` and `security:infrastructure` are the
+#     reverse: a finding is attached, but `review`/`security` are not
+#     infrastructure prefixes, so the replay counts them as evidence;
+#   - `scope_unreadable:` depends on which producer fired - the Phase 1
+#     gate attaches a finding, the pre-launch refusal in factory does
+#     not.
+#
+# Reconciling those is not this table's job (#332 holds factory.py), but
+# an undercount was, because it read as a single known exception.
 #
 # tests/test_check_name_enrolment.py pins the whole table row by row, so
 # a new row, a dropped row or a typo in a category is a red test with
@@ -494,11 +509,31 @@ def category_for_check(check_name: str) -> str:
     four of them into the ``"infrastructure"`` category that had to be
     invented to hold them, and nothing is grandfathered any more.
 
-    ``tests/test_check_name_enrolment.py`` is the mechanism: it
-    AST-walks ``kstrl/`` for every constructed check name and every
-    ``signatures=["<prefix>:<code>"]`` literal, and fails on one this
-    table does not carry. A new gate cannot reach the journal
-    uncategorised.
+    ``tests/test_check_name_enrolment.py`` is the mechanism, and #339
+    review corrected what it establishes. It AST-walks ``kstrl/`` for
+    the four places a component's failure signatures are written - a
+    ``CheckResult`` name, a ``signatures=`` argument, the ``phase=`` of
+    a failure recorded with neither, and a direct
+    ``component_failure_signatures[...] = [...]`` assignment - and fails
+    on a name this table does not carry. What it does NOT establish is
+    that no gate can reach the journal uncategorised, which is what this
+    docstring claimed while the fourth producer was invisible to it and
+    eight of nine shape mutations survived.
+
+    What it establishes now, in three files and stated as three separate
+    claims because they hold three separate things:
+
+    - a producer site the walk RECOGNISES and cannot read is enumerated
+      in its ``BLIND_SITES`` ledger rather than dropped, which is what
+      the old walk did;
+    - ``tests/test_signature_spellings.py`` pins every signature-shaped
+      string in the package whatever container it sits in, so a producer
+      in a shape nobody has thought of still has to appear somewhere;
+    - ``tests/test_check_name_shapes.py`` pins what the walk SAYS about
+      each shape, including the shapes it says nothing about, because a
+      shape the walk does not recognise leaves no trace in either of the
+      two above. That is a limit, not a proof, and it is the limit
+      ``pipeline._fail_pr_flow`` lived in until #339 review.
 
     The answer is computed here, at read time, and never stored in the
     journal - measured: a ``record_run`` entry has no ``category`` key
--- a/kstrl/evolution.py
+++ b/kstrl/evolution.py
@@ -234,7 +234,10 @@ class FailurePattern:
     # "scope_creep" for a review concern) - the part after the colon in
     # the full "<check>:<code>" signature
     error_signature: str
-    category: str  # "verification", "review", "security", "contract", "iteration"
+    # One of CATEGORIES, decided by category_for_check when the pattern
+    # is built and never written to the journal, so correcting the table
+    # also corrects what is reported about runs already recorded.
+    category: str
 
 
 @dataclass
@@ -303,8 +306,8 @@ def _normalize_error(error: str) -> str:
     return slug[:80] if slug else "unknown"
 
 
-def _classify_check(error: str) -> tuple[str, str]:
-    """Return (check_name, category) inferred from the error text.
+def _classify_check(error: str) -> str:
+    """Return the check name inferred from the error text.
 
     The legacy fallback, used wherever a component has no in-memory
     ``failure_signatures`` entry - which is EVERY ``ks evolve`` or
@@ -320,25 +323,32 @@ def _classify_check(error: str) -> tuple[str, str]:
     state no agent can influence, which is the thing the arm exists to
     prevent. Matched FIRST because the text also contains words the
     later rules claim.
+
+    #315: this returned ``(check_name, category)`` until every caller
+    was measured to discard the category, making it a SECOND place a
+    category was decided that no consumer could reach. A dead answer
+    that can silently disagree with the live one is how the live one
+    later gets "fixed" in the wrong file, so the category is gone from
+    here and ``category_for_check`` is the only decision.
     """
     lower = error.lower()
 
     if SCOPE_UNREADABLE_ERROR_PREFIX.lower() in lower:
-        return SCOPE_UNREADABLE_CHECK, "verification"
+        return SCOPE_UNREADABLE_CHECK
     if any(kw in lower for kw in ("ruff", "flake8", "pylint", "lint")):
-        return "linter", "verification"
+        return "linter"
     if any(kw in lower for kw in ("mypy", "pyright", "typecheck", "type error")):
-        return "typecheck", "verification"
+        return "typecheck"
     if any(kw in lower for kw in ("pytest", "test", "assert", "unittest")):
-        return "test_suite", "verification"
+        return "test_suite"
     if any(kw in lower for kw in ("review", "finding", "reviewer")):
-        return "review", "review"
+        return "review"
     if any(kw in lower for kw in ("contract", "integration")):
-        return "contract", "contract"
+        return "contract"
     if any(kw in lower for kw in ("mechanical verification failed",)):
-        return "verification", "verification"
+        return "verification"
 
-    return "unknown", "iteration"
+    return "unknown"
 
 
 # ---------------------------------------------------------------------------
@@ -357,6 +367,23 @@ def _classify_check(error: str) -> tuple[str, str]:
 # difference is the number.
 _DIGIT_RUN_RE = re.compile(r"\d+")
 
+#: The categories a FailurePattern can carry. ``"infrastructure"``
+#: (#315) is for a failure that is neither a gate's verdict on the
+#: change nor the engineer's loop: the run was shut down, hit the token
+#: ceiling, could not merge, could not fetch its own diff. Before it
+#: existed those fell through to ``"iteration"`` and ``ks evolve``
+#: reported them as engineer-loop problems, which is advice aimed at an
+#: agent that could not have prevented any of them.
+#:
+#: Not ``Finding.category``'s ``"infrastructure_error"``
+#: (:mod:`kstrl.findings`), which marks one ROLE RUN that failed to
+#: execute inside an otherwise healthy component. Different taxonomy,
+#: different consumer, deliberately different spelling so a grep for
+#: either does not silently return the other.
+CATEGORIES: frozenset[str] = frozenset(
+    {"verification", "review", "security", "contract", "infrastructure", "iteration"}
+)
+
 _CATEGORY_BY_CHECK = {
     "linter": "verification",
     "typecheck": "verification",
@@ -371,10 +398,36 @@ _CATEGORY_BY_CHECK = {
     "mutation_testing": "verification",
     "prd_stories": "verification",
     "verification": "verification",
+    # #315: mechanical gates that the table did not carry, so every
+    # approved-fixtures failure, policy-envelope breach and
+    # test-adequacy block was filed under the engineer loop.
+    "fixtures": "verification",
+    "policy_envelope": "verification",
+    "test_adequacy": "verification",
+    # #315: recorded outside a CheckResult, by pipeline.fail(signatures=
+    # ["<prefix>:<code>"]). None of these is a verdict on the change.
+    "aborted": "infrastructure",
+    "token_budget": "infrastructure",
+    "pr": "infrastructure",
+    "diff": "infrastructure",
+    # #315: the fallback already answers "iteration" for this one. The
+    # row is here so the table states every name kstrl emits rather than
+    # most of them, and so that a reader cannot tell an unenrolled name
+    # from a deliberate one by its absence.
+    "engineer": "iteration",
     "review": "review",
     "security": "security",
     "contract": "contract",
 }
+
+#: Checks whose failures are infrastructure, derived from the table so
+#: there is ONE answer rather than two lists to keep in step:
+#: :data:`kstrl.autonomy_replay.INFRA_FAILURE_PREFIXES` builds its
+#: prefixes from this, so enrolling a check as infrastructure reaches
+#: both consumers in one edit (#315).
+INFRASTRUCTURE_CHECKS: frozenset[str] = frozenset(
+    name for name, category in _CATEGORY_BY_CHECK.items() if category == "infrastructure"
+)
 
 # Cap on distinct per-check signatures so one catastrophic run (e.g. 40
 # distinct ruff rules) cannot flood the journal entry.
@@ -416,21 +469,23 @@ def split_signature(signature: str) -> tuple[str, str]:
 def category_for_check(check_name: str) -> str:
     """Map a check/gate name to a FailurePattern category.
 
-    An unlisted name falls through to "iteration", which files a Phase 1
-    gate under the engineer loop. Enrolling a new check in
-    ``_CATEGORY_BY_CHECK`` is a convention with no mechanism, and
-    measured on this tree the convention does not hold: AST-walking
-    ``kstrl/`` for ``CheckResult`` names finds ``fixtures``,
-    ``policy_envelope`` and ``test_adequacy`` absent from the table, so
-    three verification gates are already miscategorised here (#315).
-    That predates #294 and correcting it changes recorded
-    categorisation, so it is not that change's to make.
+    An unlisted name falls through to "iteration", which files a gate
+    under the engineer loop. Enrolling a new check in
+    ``_CATEGORY_BY_CHECK`` was a convention with no mechanism, and
+    measured during #294 the convention did not hold for eight of the
+    nineteen names ``kstrl/`` emits. All eight are enrolled as of #315,
+    four of them into the ``"infrastructure"`` category that had to be
+    invented to hold them, and nothing is grandfathered any more.
 
     ``tests/test_check_name_enrolment.py`` is the mechanism: it
-    AST-walks ``kstrl/`` for every constructed check name and fails on
-    one this table does not carry, with those three grandfathered by
-    name so the gate could land without the behaviour change. A new
-    check cannot join them quietly.
+    AST-walks ``kstrl/`` for every constructed check name and every
+    ``signatures=["<prefix>:<code>"]`` literal, and fails on one this
+    table does not carry. A new gate cannot reach the journal
+    uncategorised.
+
+    The answer is computed here, at read time, and never stored in the
+    journal, so a correction to the table also corrects what
+    ``ks evolve`` reports about runs that already happened.
     """
     return _CATEGORY_BY_CHECK.get(check_name, "iteration")
 
@@ -676,7 +731,7 @@ class EvolutionJournal:
                 comp_signatures = list(failure_signatures.get(comp.id) or [])
                 if not comp_signatures:
                     # Legacy fallback: classify the flattened string.
-                    legacy_check, _ = _classify_check(comp.error)
+                    legacy_check = _classify_check(comp.error)
                     legacy_sig = _normalize_error(comp.error)
                     if legacy_sig:
                         comp_signatures = [f"{legacy_check}:{legacy_sig}"]
@@ -759,12 +814,7 @@ class EvolutionJournal:
                 continue
             sigs = list(failure_signatures.get(comp.id) or [])
             if not sigs:
-                sigs = [
-                    signature_for_error(
-                        _classify_check(comp.error)[0],
-                        comp.error,
-                    )
-                ]
+                sigs = [signature_for_error(_classify_check(comp.error), comp.error)]
             for sig in sigs:
                 failure_sigs[sig] = failure_sigs.get(sig, 0) + 1
         common_failure = max(failure_sigs, key=failure_sigs.get, default="") if failure_sigs else ""  # type: ignore[arg-type]
@@ -847,7 +897,7 @@ class EvolutionJournal:
                 continue
             sigs = list(signatures_by_component.get(comp.id) or [])
             if not sigs:
-                legacy_check, _ = _classify_check(comp.error)
+                legacy_check = _classify_check(comp.error)
                 legacy_sig = _normalize_error(comp.error)
                 if not legacy_sig:
                     continue
--- a/kstrl/pipeline.py
+++ b/kstrl/pipeline.py
@@ -1797,7 +1797,20 @@ class ComponentPipeline:
         comp.completed_at = _iso_now()
         comp.failed_phase = "pr"
         comp.failed_check = "pr_flow"
-        self._record_failure_signatures(comp, "pr", comp.error, None)
+        # Spelled out with keywords, and read that way. #339: this
+        # is the third caller of the funnel and the only one that is
+        # not `fail` / `retry_or_fail`, so no `signatures=` is ever
+        # handed to it and the phase becomes the check name. It was
+        # invisible to every layer of
+        # `tests/test_check_name_enrolment.py`: not a fail call, no
+        # keyword, no colon in "pr" for the spellings net to see.
+        # Mutating "pr" to "bogus_flow" left 4737 tests green.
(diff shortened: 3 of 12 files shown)


0xfauzi:
# `/simplify` on this branch: four parallel reviewers, findings verbatim

Reuse, simplification, efficiency and altitude, run concurrently against `git diff origin/main` in the working tree. Everything below is quoted as returned. Disposition follows each block. Line numbers are from the tree as each agent read it and have since moved; the chokepoint pin was split into `tests/test_signature_chokepoint.py` after the 800-line ratchet fired.

Two of the four flagged that the worktree changed under them mid-review. That is on me: a mutation experiment and a sibling fix were in flight. Where a number could have been contaminated I re-measured it myself before acting, and say so.

---

## REUSE

> ### F1. `tests/test_autonomy_ladder.py:960-972` -- a hand-rolled `RunRecord` the PR's own new helper already builds, byte for byte
>
> The PR moved five builders into `tests/helpers/replay.py` and imported two of them into this file, but left the third site behind. [...] Measured, not argued: constructing both and comparing gives `True` on every field. Call it `failing_run("review:prd_criterion")`.
>
> Cost: 12 lines, and the `RunRecord` import at `tests/test_autonomy_ladder.py:31` exists for this one site (grep: only two references in the file). The helper's own docstring at `tests/helpers/replay.py:5-7` says it exists because "#339 review measured five hand-rolled builders across two files"; this is the fifth, still hand-rolled two lines under `_clean = staticmethod(clean_run)`.

**APPLIED.** Verified the import count independently (two references, one of them the definition line). Now `failing_run("review:prd_criterion")`, and the `RunRecord` import is gone.

> ### F2. `tests/helpers/astfold.py:155-228` -- a second copy of the constant-folding rules, in a file whose stated purpose is that there is only one
>
> Measured over `kstrl/`: on the 22,738 expressions containing no `ast.Name`, `folded_str(node)` and the strict reading of `fold(node, {})` (`None` when the result carries a `HOLE`) give the **identical** answer, 0 divergences. Over all 119,605 expressions, only 471 differ, and all 471 are the deliberate `_fold_name` branch. So `folded_str` is today derivable in three lines from `fold` once name resolution is a parameter [...] Cost: `tests/test_journal_one_writer.py:116-182` (67 lines, 4 functions) is deletable against one extra parameter on `fold`.

**DECLINED as a change, APPLIED as a correction.** The measurement is right and the altitude reviewer reached the same conclusion independently. But `test_journal_one_writer.py` guards #312, and changing what it folds is that guard's own measurement to make, not a side effect of a categorisation fix. What I did fix is the `astfold` docstring, which argued the two were "deliberately NOT" the same thing: it now carries both numbers and says plainly that the strict answer is a two-line wrapper and the copies should converge. The wrong prior was the actual defect.

> ### F3. `tests/helpers/astfold.py:70-84` -- `_string_bindings` re-implements `assignment_parts`, and re-learns its lesson
>
> Same destructuring, same AnnAssign lesson written twice in two docstrings. The replacement is four lines and drops both `# type: ignore` [...] Correct direction is to move `assignment_parts` into `tests/helpers/astfold.py` and have `test_journal_one_writer.py` [...] import it from there, rather than a helper importing from a test module.

**DECLINED.** The agent's own preferred fix requires editing `test_journal_one_writer.py`, same reason as F2. The minimal alternative it offers has a helper module importing from a test module, which the agent itself calls the wrong direction. Left for #324, which moves all of this into one place anyway. Note the simplification reviewer found something worse in the same docstring, and that part is fixed: see S3.

> ### F4. `tests/helpers/astfold.py:87-99` -- `called_name` is the fifth copy of a one-liner already in four test files
>
> Same result including the `""` sentinel. Cost is small in lines (4 x 1), but this PR created the first shared AST helper module in `tests/helpers/` and wrote a fifth copy instead of exporting one and deleting four.

**DECLINED, scope.** Two of the four (`test_procgroup.py`, `test_timeout_enforcement.py`) are held by a parallel subprocess sweep. Recorded in the PR body as left for #324.

> ### F5. `tests/helpers/astfold.py:31, 102-114` -- `parsed_modules` duplicates `package_sources` + `parsed` + `label`, and parses `kstrl/` a second time per session
>
> Two independent caches over the same 124 files, each with its own measurement paragraph justifying itself. Concrete cost: with both guards in a run, `kstrl/` is read and parsed twice, roughly 162 ms of avoidable work per session by the existing helper's own number. Second cost is vocabulary: astfold keys files as `"kstrl/pipeline.py"` while `label()` keys them as `"pipeline.py"`, so `BLIND_SITES` / `EXPECTED_SPELLINGS` and `EXPECTED_EVENT_NAME_SPELLINGS` name the same file two ways across guards that cite each other in their docstrings.

**DECLINED, scope.** Same file, same reason. The vocabulary half is the more annoying of the two and is recorded in the PR body.

> ### F6. `tests/helpers/astfold.py:263-288` -- `scoped_nodes` is a third qualified-scope walk (weaker finding, consolidation is not free)
>
> I measured this one before reporting it, and it does not reduce to "call the old one": over `kstrl/`, `scoped_nodes` and `scopes()+own_nodes()` agree on 140,633 node/qualname pairs, disagree on 854, and `scoped_nodes` covers 18,066 nodes the other never yields [...] So the new one is the better implementation, in the right place, and the reuse debt is that the two older copies were not retired onto it.

**DECLINED, and the agent agrees it is not a swap.** Retiring the older copies changes what those guards report. Recorded for #324.

> ### F7. `tests/test_infrastructure_category_consumers.py:30-38` -- `_run_dominated_by` duplicates `failing_run`, both added in this PR
>
> Two names for "a run that failed on this signature", one file apart, in the same change. Cost: 9 lines plus the reader's question of which one to reach for.

**APPLIED.** `_run_dominated_by` is now an alias for `failing_run`, with the reason kept as a comment because the "assert through the property, not the constant" point is worth keeping.

> ### F8. `tests/test_signature_spellings.py:113-127` and `tests/test_check_name_enrolment.py:521-537` -- the same fold-and-take-the-head walk, twice, both new
>
> `_names_in` is `signature_heads` plus the caller rescue. Small (7 lines), listed because it is the exact "two copies that can be widened apart" risk the PR cites as its reason for creating `astfold.py`, and both copies arrive in this diff.

**APPLIED.** `astfold.folded_nodes` is now the single walk; the net is a set comprehension over it and `_names_in` layers the rescue on top.

---

## SIMPLIFICATION

> ### 1. An anti-vacuity assertion that cannot fail when the thing it names is deleted
>
> **Mutation that leaves it green:** replace the body of `_from_the_callers` with `return frozenset()`, deleting the entire interprocedural rescue. Measured: `"security" in check_names()` stays `True` (it is produced at another `kstrl/pipeline.py` site, not by the rescue) [...] Concrete fix: assert `("kstrl/pipeline.py", "ComponentPipeline._coverage_failure", "f'{phase}:coverage-unverified:{reason}'") not in blind_sites()`, which is the property that actually moves.

**APPLIED, and it is the best finding of the four reviews.** A test labelled as the guard on 70 lines of machinery was inert. Re-measured after the fix: under the same mutation the class goes from 1 failure to 2, and the new assertion is one of them.

> ### 2. Dead branch: `signature_head`'s hole check is unreachable
>
> Verified two ways: exhaustive search over every string up to length 6 from `{a,b,0,_,:,.,-,NUL,A}` found no value that both fullmatches the shape and holds a `HOLE` before its first colon; and the mutation `return head` leaves 215/215 green with the census, blind ledger and spelling inventory byte-identical. [...] Cost: a reader takes the line as the guard against an invented check name; the regex is.

**APPLIED.** Deleted.

> ### 3. A docstring justification that the package contradicts
>
> Measured: `kstrl/gateparse.py:50-52` declares `GATE_TEST`, `GATE_TYPECHECK` and `GATE_LINT` as plain `ast.Assign`. `GATE_TOOLS` at line 67 is annotated but binds a dict, which `_string_bindings` skips regardless. The only three module-level annotated string constants in the whole package are `kstrl/agents/base.py:439 ARCHITECT_ROLE`, `kstrl/names.py:58 ROLE_KEY_PREFIX` and `kstrl/serve.py:784 SPAWNED_RUN_KIND`, none of them a check name.

**APPLIED.** The claim was false and this PR had promoted it into a new shared helper, so it read as a measured fact about a module two files depend on. The branch stays as future-proofing against a logged defect class; it is no longer sold as a live requirement.

> ### 4. A test duplicated across the two new files
>
> Identical fixture string. The shapes version asserts strictly more (it pins the expression text as well as the empty name list) in three lines; the enrolment version hand-rolls the `scoped_nodes` + `_signature_sites` + `_names_in` composition in twelve.

**APPLIED.** The twelve-line copy is gone, with a comment pointing at the survivor.

> ### 5. `BLIND_SITES`: three pairs of rows that differ only in prose
>
> Simpler form: carry the count in the row [...] and compare a `collections.Counter` of the triples, giving eleven rows and one reason per site. The tradeoff is a less readable failure message, so this is a judgement call.

**DECLINED.** The agent names the tradeoff correctly and I come down the other way: the failure message is what a future reader acts on, and a `Counter` diff reads worse than three redundant rows. The row text is what makes the ledger readable rather than regenerable.

> ### 6. `test_every_blind_site_says_why` reads only its own literal
>
> No mutation of `kstrl/`, `astfold.py` or either matcher can turn it red [...] It is not a strict tautology (an author could type `""` fourteen lines up), but it is the one test in the PR with no reachable failure mode from the code under guard.

**DECLINED.** Accurate, and it is the point: it guards the ledger's own discipline, which is data an author edits, not code. A row with no reason is a silencer, and that is the failure it catches.

> ### 7. The helper extraction re-duplicates itself in the same PR
>
> `tests/helpers/replay.py` exists because "#339 review measured five hand-rolled builders across two files". This comprehension is a sixth: it is exactly `clean_run(i)` with one field overridden.

**APPLIED.** Now `replace(clean_run(i), common_failure=...)`.

> ### 8. Assertions the builder already guarantees
>
> `_run_dominated_by` fixes `completed=0, failed=1`, so `(completed + failed) > 0` is constant-True for every record it builds and `RunRecord.decisive` reduces to `not infra_aborted`. Each of these five `.decisive` assertions is therefore the negation of the `infra_aborted` assertion sitting beside it. [...] Each pair also rebuilds the record and repeats the signature literal, so a typo in one copy silently changes what the second assertion tests.
>
> **Line 102** (`assert "scope_unreadable" not in INFRASTRUCTURE_CHECKS`) restates line 101 through the one-line comprehension at `kstrl/evolution.py`.

**APPLIED.** One `decisive` case survives, in the parametrized test, binding the record once; the class docstring says why the others went. The `INFRASTRUCTURE_CHECKS` line is gone with a comment saying what entails it. This is the same defect class that cost this PR two tests in round 2, so it is worth naming that the extraction of a shared builder is what made it easy to write four more.

> ### 9. Conditions the type system or a sibling clause already gives
>
> - `signatures is not None and` is unreachable as a *condition* [...] It survives only to narrow `ast.expr | None` for mypy, which the comment above it does not say.
> - `ast.Tuple | ast.Set` in `_signatures_taken`. No call site passes either.
> - `isinstance(name, str)` is always true.

**PARTIALLY APPLIED.** The first is a real documentation defect and now says it is a type narrowing rather than a condition. The other two are declined: `_signatures_taken` answering three shapes rather than one is what makes it safe to widen the producer set later, and the `isinstance` is a two-word cost against a `getattr` whose type nobody should have to re-derive.

> ### 10. `@cache` with zero hits
>
> I instrumented `_keyword_values` and counted the calls the whole census makes to it: exactly one [...] `_census` is itself `@lru_cache(maxsize=1)`, so the decorator never serves a second lookup.

**DECLINED, with the measurement recorded.** True and harmless. The docstring now names the crossover at which indexing beats the per-key walk (two keys), which is the fact a future author needs. Removing a correct one-line cache to save nothing is not an improvement.

> ### 11. A private name imported across test modules
>
> `tests/test_check_name_shapes.py:40` imports `_producer_sites` from `tests.test_check_name_enrolment`; line 41 imports the public `signature_heads`. Same role, two conventions.

**DECLINED, scope.** Fair, and renaming ripples through a file the owner has asked me to stop expanding. Left for #324.

> ### 12. A production rewrite the guard does not need
>
> The call was expanded from one positional line to a six-line keyword call under an eight-line comment. [...] Not wrong, but the diff reads as if the spelling were required.

**DECLINED, but the comment was rewritten anyway** for a different reason (the altitude pass found the same three-branch prose repeated three times). The keyword spelling stays because four positional arguments ending in a bare `None` is worse to read than the line it replaced, and the test that pins the positional spelling says in its own docstring that the fix does not depend on it.

> ### 13. Three more astfold guards no test distinguishes (lower confidence)
>
> Green mutation, identical census, and no fixture anywhere: `_fold_concat`'s non-`Add` rejection, `_fold_concat`'s one-side-must-fold rule, and `_fold_placeholder`'s `!r`/format-spec rejection. All three are defensible future-proofing, but `astfold.py:192` sets the standard itself ("a branch nothing exercises is a branch nobody knows is broken").

**DECLINED, scope.** The agent is right that this file sets that standard and then breaks it three times. Adding three fixtures means extending `tests/test_check_name_shapes.py`, which is exactly what the owner has stopped. Recorded.

> ### 14. `Folded.first_hole` (judgement call)
>
> Measured: it is load-bearing for exactly one thing, keeping the `_coverage_failure` row out of `BLIND_SITES`. [...] A `_leading_placeholder(expr)` helper at the single consumer would let `fold` return `str | None`, delete the `Folded` dataclass and `_fold_part` [...] Roughly line-neutral, but it turns a field carried by every fold into a lookup at the one site that needs it.

**DECLINED.** The right call at a different time. It is a redesign of the shared helper, line-neutral by the agent's own estimate, and #324 is about to rewrite that layer.

> ## One thing outside the brief, for the caller
>
> Adding one row to `_CATEGORY_BY_CHECK` is now a four-place edit [...] I measured this by adding `"deploy": "infrastructure"`: three tests across two files went red. Each pin is separately argued in the PR, so this is a cost to weigh, not a defect.

**Noted, not changed.** Real and worth the reviewer's attention. Each of the four pins has a distinct reason in the PR body and I would not drop any of them individually, but four places for one row is a number somebody should push back on if they disagree.

---

## EFFICIENCY

> **1. `parameter_index` is a per-key cache over a full-package walk -- 6 misses per run.** Measured with `cache_info()` across the four-file run: **6 misses, 13 hits**. Each miss re-walks all 127 modules to find one `def`. Six misses cost **580.6 ms** loaded (**241.5 ms** quiet) [...] Two of the six sit inside `_census()`, which I timed at **180 ms** before `parameter_index` landed and **304 ms** after. [...] Cheaper: one `@lru_cache(maxsize=1)` pass building `{name: [positional-param lists]}`. I measured it: **58.5 ms quiet / 76.2 ms loaded, 1,775 names**, and it is flat in key count. 4.1x quiet, 7.6x loaded.

**APPLIED.** This was my own regression, introduced two hours earlier while fixing the altitude finding. Now one cached pass. Re-measured after: the four-file set went from 1.28-1.43 s to **0.90 s**, and the two 0.13 s shape tests dropped off the slowest list.

> **2. `_keyword_values` has the identical shape and could share that same walk.** One miss per run: **59.7 ms quiet / 85.6 ms loaded**. It is linear in keys -- I measured k=1..4 at 62 / 100 / 144 / 209 ms. Its docstring rejects an index citing "1,742 entries and 60 ms". I reproduced that number exactly (**1,745 entries, 63-65 ms**) -- but the index is *flat* where the per-key walk is linear, so the crossover is at **k=2**, not never.

**DECLINED as a change, APPLIED as a correction.** At one key a run the two are a wash, and fusing two unrelated index builds into one walk to save 45 ms is coupling I would rather not add on a landing PR. But the docstring said "indexed is not worth it" as if that were a property, and it is a property of k=1. It now carries the linear-versus-flat numbers and names the crossover.

> **3. `signature_spellings()` is uncached and called by two tests.** **93 ms quiet / 102 ms loaded** per call; the second is pure duplicate.

**APPLIED.** `@lru_cache(maxsize=1)`, as `parsed_modules` and `_census` already use.

> **4. The hole-marker test takes a traversal of its own.** **63 ms quiet / 71 ms loaded**. Its incremental cost if it rode the spellings walk is **13.5 ms**. Saves ~50 ms.

**DECLINED.** It walks for a different predicate (raw `ast.Constant` holding the marker, not folded values), so fusing means a combined walker returning two unrelated things. 50 ms is not worth that coupling in a file whose whole point is being readable.

> ## Measured and *not* findings (so nobody re-checks)
>
> `parsed_modules()` cache works: 1 miss / 12 hits. `pool()`: 108,560 hits / 1 miss. `scoped_nodes` is not quadratic: 92.9 ms for 237,162 pairs. `_census()` cache works. `signature_container_sites()` is uncached but has one caller: 75 ms, correct as written. Production `replay()`: the new second `sum()` pass is **0.078 ms per 10,000 rows**. Production import edge: `autonomy_replay` now imports `evolution` (**+18.9 ms**) [...] Non-issue. None of these files appears in the full suite's slowest-20.

**Noted.** The production numbers are the ones that mattered and both are non-issues.

---

## ALTITUDE

The altitude pass is written up in full in the PR body, because four of its six findings are the substance of this round rather than cleanups. In short:

- **A1, the derived index: APPLIED.** Four hand-written integers next to a walk that parses the tree stating them, and *six of eight off-by-one mutations to them left every guard green*. Now read off the `def`. It also deletes the special case that kept `feature_cmd`'s local `fail` out of the census, because a definition with no `phase` parameter never enters the table by construction.
- **A2, the chokepoint inventory: APPLIED**, and it is the durable finding. `BLIND_SITES` resembles `EXPECTED_JOURNAL_PATH_SITES` and does not behave like it: the precedent inventories every place a resource is *obtained* and is closed by construction, while `BLIND_SITES` inventories the places the walk *gave up* and is closed only over shapes the walk already enumerates. `tests/test_signature_chokepoint.py` has the precedent's guarantee. Measured: a new writer reaching the mapping through a local alias with a runtime-built signature is invisible to the census, the ledger, the shapes file and the net, and fails that pin alone.
- **A3, the HITL signature literals: NO FINDING**, and the agent checked properly (11 pre-existing sites of the same idiom). Its residual point, that 27 lines of comment restate one rule three times because the rule has no place to live, is applied: stated once, pointed at twice.
- **A4, `_REPLAY_ONLY_PREFIXES`: APPLIED.** It caught me defending three dead prefixes with a historical-rows argument that contradicts the measurement one paragraph above it. If no version ever emitted them, no row contains them. The false sentence is gone; the prefixes stay for a reason that survives being checked.
- **A5, two folders in `tests/`: docstring corrected**, same as reuse F2.

---

## What I would look at first as a reviewer

The three findings that were *false claims in the tree*, not style: the inert anti-vacuity assertion (S1), the `gateparse` docstring (S3), and the historical-rows defence (A4). All three read as measured facts and none of them was one. That is the pattern this PR keeps generating, and it is the reason the owner has pulled #324 forward rather than accepting a fourth round here.

PR #328, opened 2026-09-01
refactor(procgroup): one copy of the safe-pgid guard, and a test for verify's (#308)

0xfauzi:
Closes #308.

Tidying, not a bug fix, exactly as the issue's own correction comment says. The
three-way diff below agrees with it: the copies had not drifted on any decision
any of them makes, so there is no bug narrative here and none has been
manufactured.

## 1. The three-way diff, measured before anything moved

The three copies were driven over one matrix of pid / `getpgid` / `killpg`
outcomes using the real production functions, not by reading them. `serve`
returns a pgid and the other two call `killpg`, so the comparable observable is
"which pgid, if any, was returned or reached `killpg`", plus whether the
direct-child fallback (`fb`) fired.

```
case                         | serve            | verify           | fb   | agents.proc      | fb
-------------------------------------------------------------------------------------------------
magicmock pid (coerces to 1) | pgid None        | pgid None        | term | pgid None        | term
pid None                     | pgid None        | pgid None        | term | pgid None        | term
pid 0                        | pgid None        | pgid None        | term | pgid None        | term
pid 1                        | pgid None        | pgid None        | term | pgid None        | term
pid -1                       | pgid None        | pgid None        | term | pgid None        | term
pid 2, pgid 99999            | pgid 99999       | pgid 99999       | -    | pgid 99999       | -
pid 2, pgid 1 (broadcast)    | pgid None        | pgid None        | term | pgid None        | term
pid 2, pgid 0                | pgid None        | pgid None        | term | pgid None        | term
pid 2, pgid -1               | pgid None        | pgid None        | term | pgid None        | term
pid 2, pgid == ours          | pgid None        | pgid None        | term | pgid None        | term
pid 2, getpgid ESRCH         | RAISED ProcessLookupError | pgid None | term | pgid None | term  <== serve/verify DIVERGE
pid 2, getpgid EPERM         | RAISED PermissionError    | pgid None | term | pgid None | term  <== serve/verify DIVERGE
pid 2, getpgid OSError       | RAISED OSError   | pgid None        | term | pgid None        | term  <== serve/verify DIVERGE
pid True (bool is an int)    | pgid None        | pgid None        | term | pgid None        | term
pid 2, killpg EPERM          | pgid 99999       | pgid 99999       | term | pgid 99999       | term

divergences: 3
```

Two findings, and one correction to the correction.

**A. `verify._signal_process_group` and `agents.proc._signal_group` were
behaviourally identical, whole function.** Zero divergences across all 15
[...]

Issue #309, opened 2026-08-31
PS_TIMEOUT_SECONDS does not bound an unkillable ps, so the reap check can hang the daemon


0xfauzi:
Split out of #298 (PR #301) round-2 review, finding 4. Referenced from `kstrl/procgroup.py`.

## The claim that was wrong

`kstrl/procgroup.read_group_liveness` passes `timeout=PS_TIMEOUT_SECONDS` to `subprocess.run`, and the docstring claimed this bounds how long a wedged `ps` can stall the caller. It does not, for the exact case it named.

## Why

CPython `subprocess.run` handles its own timeout as:

```
except TimeoutExpired as exc:
    process.kill()
    ...
    process.wait()      # no timeout
    raise
```

and the enclosing `with Popen(...)` calls `__exit__`, which waits again, also without a timeout. A `ps` wedged in an uninterruptible sleep (D state) cannot be killed, so both waits block indefinitely.

The docstring named exactly that case: "a ps stalled on a D-state read (NFS, a hung container runtime)".

## Consequence

`read_group_liveness` is called from `serve.terminate_process_group` on the timed-out-run path, which runs while the daemon holds the serve lock. A D-state `ps` therefore hangs the daemon mid-reap, holding the lock, with no timeout anywhere above it.

Rare, but the whole point of the timeout was this case.

## Options

1. Replace `subprocess.run` with `Popen` plus `communicate(timeout=...)`, then `kill()`, then a second bounded `communicate`. If the second also times out, ABANDON the child and return "cannot see". This gives a real bound and leaks one unreapable `ps` per occurrence, which becomes a zombie of the daemon. Requires adding `kstrl/procgroup.py` to `POPEN_ALLOWLIST` in `tests/test_timeout_enforcement.py`.
2. Run the probe off the daemon critical path (a thread with a deadline, abandoning the worker) so a hang cannot hold the serve lock.
3. Accept the hang and document it. This is the current state, now correctly documented.

Option 1 is a real bound at the cost of a leaked process. Option 2 keeps the lock free but leaves the fork outstanding either way. The trade needs a decision, not a default.

## What has been done

`kstrl/procgroup.py`s docstring now says what the timeout does and does not cover and points here, so it no longer claims a bound it lacks. `tests/test_procgroup.py::TestThePsCallIsBounded` pins that the kwarg is passed, which is what it can check; it explicitly cannot detect this.

## Related

[...]

Issue #298, opened 2026-08-31
process_group_alive treats a zombie as running, so a reaped-but-unwaited group reads as live


0xfauzi:
Found while fixing #292 (PR #297), by measurement rather than by reading, and left out of that PR as out of scope. PR #297 pins the current behaviour with a test that will fail when someone fixes this, so the two changes cannot silently disagree.

## The defect

`serve.process_group_alive` maps `PermissionError` to `True`. The reasoning is presumably that a signal refused for permission reasons proves something is there to refuse it, which is sound as far as it goes.

But a **zombie stays signalable**. A process that has been SIGKILLed and not yet reaped by its parent still accepts a signal-zero probe, so a group holding nothing but a corpse reads as live.

`terminate_process_group` uses that answer to decide whether a run was reaped. So it can conclude a run is still running when every process in it is already dead and merely unwaited-for.

## How it was found

While fixing #292, the obvious implementation of "check the specific process group" was `os.killpg(pgid, 0)`. Measured on a real process tree, with the parent alive and not yet reaping: a SIGKILLed group reported as **present for the entire six-second window and never converged**.

That measurement is what stopped PR #297 swapping a machine-state flake for a reap-timing flake. Its test now uses `ps` group membership with zombies excluded. The same measurement is what exposes this production path, which uses the signal-probe shape.

## Why it matters

The window is small but not empty: it is exactly the interval between a child dying and its parent reaping it, which is precisely when a shutdown or timeout path is asking whether the group is gone. A wrong "still alive" there means the caller waits, or escalates, or reports a run as in flight when it is not.

## Suggested fix

Distinguish "a process exists in this group" from "a signal was refused". Options:

1. Use group membership from `ps` with zombies excluded, which is what PR #297's test now does and is the exact question being asked.
2. Keep the signal probe but treat `PermissionError` as unknown rather than alive, and resolve it another way.
3. On Linux, read `/proc/<pid>/stat` and check the state field for `Z`. Not portable to macOS, which this project targets first.

(1) matches the question. The caller wants to know whether anything is still running, and a zombie is not running.

## Test note

PR #297 added a test that pins the current, wrong behaviour deliberately, so that whoever fixes this sees it fail rather than discovering the coupling later. Update it as part of the fix rather than working around it.

PR #340, opened 2026-09-02
fix(procgroup): a child is signalled, reaped and let go of in one place (#326, #329)

### LINKED ISSUE #326: verify.run_scrubbed abandons a killed child without registering it, and never closes its pipe ends
Named as an explicit follow-up by PR #321 (issue #309), which fixed the same defect at one site and says so in its own module docstring under "THIS FIXES ONE SITE, NOT THE CLASS".

## The defect

`verify.run_scrubbed` kills a child on timeout and gives up on it without registering what it dropped. Two consequences:

1. **The abandoned child can be a permanent zombie.** `Popen.__del__` calls `_warn(..., ResourceWarning)` BEFORE `_active.append(self)` (CPython 3.12.8 `subprocess.py:1139` and `:1145`). Under `PYTHONWARNINGS=error` the warn raises, `__del__` aborts, and the registration never happens. Reproduced under `warnings.simplefilter("error")` as a real `95800 Z <defunct>`.

   `kstrl/serve.py:1634` already records this codebase crashing a daemon under that exact setting and calls it "a common CI setting".

2. **It never closes its two pipe ends.** This is the same leak PR #321 fixed in `procgroup._kill_or_abandon` via a `finally` and did not fix here.

## Why this one is wider than #309 was

`run_scrubbed` is the worst of the three siblings by a long way:

- It needs **no D-state child at all**. It only needs something outside the process group holding the pipe write end, which a forked grandchild does routinely.
- It runs **once per verification command per iteration**, not once per timed-out run. `run_scrubbed` has roughly 15 call sites.

## The other two siblings, same class, narrower windows

- `serve._wait_out_grace`
- `agents.proc.DeadlineStreamer.kill`

## Suggested shape

PR #321 landed `_register_abandoned` and `_release_pipes` in `kstrl/procgroup.py`, which is a leaf module and importable without a cycle. A shared `dispose(process, grace)` there, called from all three sites, is the natural fix.

Note one measured subtlety from #321 before copying its approach: registering ONLY on a kstrl-owned list is better under `-W error` and WORSE under default warnings, because CPython sweeps `_active` on every `Popen.__init__` and kstrl has around 70 spawn sites, so the stock path reaps an abandoned child at the next git call. #321 registers in both places for that reason. A shared helper should keep that.

Also from #321: guard the `_active` append with `hasattr(process, "_internal_poll")` rather than `isinstance(process, subprocess.Popen)`. A test double in `_active` crashes the next real spawn anywhere in the process, and the isinstance form is unusable because the test seam replaces that name with a function.

## Separately, pre-existing, different gate

`tests/test_procgroup.py`'s `ps`-owner net has the same import-alias hole that #321 fixed in the timeout audit: measured, it returns `[]` for `from subprocess import run as _run`. Related to #324.

### LINKED ISSUE #329: serve.terminate_process_group signals a caller-supplied pgid with no safety guard
Flagged during PR #328 (issue #308), and deliberately declined there because fixing it is a behaviour change rather than a cleanup.

## The gap

PR #328 unified the safe-pgid guard into `procgroup.safe_pgid(process)`, which refuses to signal a pgid that would broadcast (0), and refuses the caller's own group. Three call sites now route through it.

`serve.terminate_process_group(process, pgid=...)` does not. It signals a caller-supplied pgid with **no guard at all**. It is the one bare-pgid path left outside `procgroup`, and it is protected by provenance only: it is safe today because of where its callers happen to get the value, not because anything checks.

## Why it matters

`ks serve` holds the daemon singleton lock for its entire process lifetime. A signal to pgid 0 broadcasts to the caller's whole process group, and a signal to the daemon's own group is a total loss of the daemon. PR #328's mutation sweep demonstrated the shape of this concretely: removing the own-group check does not fail a test, it **kills the test runner with signal 15**, because a pre-existing serve test sets `fake.pid = os.getpid()` and does not patch `killpg`.

That is a test harness. The same mistake reached from a real caller takes down the daemon.

## What has to be decided

The reason #328 declined it: `safe_pgid` answers a question about a `Popen`, and this function receives a bare integer. Guarding it means deciding what `terminate_process_group` RETURNS when handed an unsafe pgid. Options:

1. Refuse and report, which changes the return contract and every caller that ignores it.
2. Refuse and raise, which is fail-closed but reaches callers that currently cannot fail here.
3. Keep provenance as the contract and make it explicit: type the parameter so a bare int cannot be passed, or take the `Popen` and derive the pgid inside.

Option 3 is the one that removes a mechanism rather than adding a check, which is the repo's stated preference. It also matches how the other three sites now work.

## First step

Trace every caller and record where each pgid comes from, before choosing. If provenance is genuinely sufficient at every site today, this is a latent hazard rather than a live bug, and option 3 closes it permanently at low cost.

0xfauzi:
Closes #326. Closes #329.

A class sweep, not two site fixes. Five defects of this class had already been
found one at a time, each one's investigation turning up the next, and
`kstrl/procgroup.py` carried a comment saying `THIS FIXES ONE SITE, NOT THE
CLASS` with the other three named underneath it. They were still there four PRs
later. This closes the class and puts a guard on it.

## What a user sees differently

- **A timed-out `ks verify` no longer leaks a child.** `verify.run_scrubbed`
  killed the command on expiry and abandoned it: no reap, no register, both pipe
  ends still held. Under `PYTHONWARNINGS=error` that child is a permanent
  zombie, because CPython's `Popen.__del__` calls `_warn(..., ResourceWarning)`
  BEFORE `_active.append(self)` (3.12.8 `subprocess.py:1139` then `:1145`), so
  the warn raises, `__del__` aborts, and nothing is left holding the pid.
- **An aborted agent run stops spending.** Every adapter's `run` is a generator,
  and `decompose.collect_agent_output` raises `AgentOutputTooLarge` from inside
  `for line in agent.run(...)` today. Nothing then called `finish`, so the agent
  CLI kept running to its own completion, and kept billing, after the caller had
  decided to throw the answer away.
- **`ks serve` cannot signal itself off the machine.** `terminate_process_group`
  took a caller-supplied pgid straight to `os.killpg` with no check of any kind.

## The two issues

**#326 - a killed child is abandoned.** Three sites killed a child and let go of
it with no reap, no register and no pipe close: `verify.run_scrubbed`,
`serve.run_supervised` and `agents.proc.DeadlineStreamer.kill`.

**#329 - a pgid is signalled with no guard.** `serve.terminate_process_group`
was safe by PROVENANCE: every caller happened to source its pgid from
`safe_pgid`. Provenance is not a mechanism. #328 declined this as a behaviour
change rather than a cleanup, and its own mutation sweep is the evidence for why
it matters: dropping the own-group check does not fail a test, it kills the test
RUNNER with signal 15, because a serve test sets `fake.pid = os.getpid()`. `ks
serve` holds the daemon singleton lock for its whole process lifetime, so the
same mistake from a real caller takes the daemon down.

## Review round: the guard went blind, and that was the important finding

`kstrl/procdispose.py` was split out of `procgroup` and landed on neither
`POPEN_ALLOWLIST` nor any gate scope, so `test_no_allowlisted_module_waits_without_a_deadline`
kept scanning the same four files it always had. Measured with the gate's own
`_wait_calls` matcher:

| revision | gate-visible waits | agents/proc | procgroup | serve | verify | procdispose |
|---|---|---|---|---|---|---|
| `origin/main` | 11 | 3 | 2 | 3 | 3 | absent |
| this branch, before the fix | 7 | 2 | 1 | 2 | 2 | 2, unscanned |
| this branch, after the fix | 9 | 2 | 1 | 2 | 2 | 2, scanned |

`origin/main:kstrl/procgroup.py:379` is verbatim the code now at
`procdispose.py:148`. It was guarded and stopped being guarded, and **nothing
went red** - the gate's anti-vacuity check is per file, and all four remaining
files still showed a wait.

**The consequence was a surviving mutant.** Deleting the deadline from
`reap_or_abandon`'s `process.wait(timeout=grace)` passed the entire suite, 4977
passed and zero failures, on the line whose own docstring says it is why the
function exists. `DeadlineStreamer.kill` calls it as its last leg, so a
grandchild in uninterruptible IO outside the killed group would have blocked the
harness thread forever. That is #309 on the agent path.

**This is instance eleven of the #324 class, and the variant that matters most:
a guard that goes BLIND rather than RED because a refactor moved code out from
under it.** The other ten are guards that pass a mutation they can see. This one
stopped looking. The fix is two named scopes rather than one:
`POPEN_ALLOWLIST` says who may construct a child and `CHILD_WAIT_SCOPE` says who
may wait on one, and `test_the_spawner_table_names_the_same_modules_the_gate_scopes_do`
now asserts both ties exactly - every row of `EXPECTED_SPAWNERS` is in the wait
scope, and the spawning rows are exactly the allowlist. Removing `procdispose`
from either goes red.

### Four more, all from the same review

1. **The streamer rule matched a method NAME, never an object.** Measured:
   replacing `finally: streamer.close()` with `finally: _tmp.close()` left the
   guard green on a confirmed live leak, and `agents/codex.py` already carries a
   second `try`/`finally` in the same scope for its temp file, so one edit of
   that `unlink()` to a `close()` would have switched the rule off for that
   adapter. `_finally_disposes` now checks the receiver and requires the `try`
   not to start before the construction.
2. **`STREAMER_DISPOSALS` accepted `finish` and `kill`.** The PR body claimed
   this was narrowed and the commit did not do it - a body claim reading as
   measured fact that was not. It is narrowed now, and pinned:
   `finish()` on a live child takes 10.0028s, which on the abandonment path is
   the billed spend the rule exists to stop, and `kill()` never calls `_settle`,
   so it leaves `_disposed` False and the streamer in `_ACTIVE`.
3. **The fifth disposal did not obey the rule this PR mechanises.**
   `DeadlineStreamer.kill`'s SIGTERM grace was guarded by
   `except subprocess.TimeoutExpired` alone. Reproduced with SIGINT delivered
   1.0s into a 5s grace on a `trap '' TERM` child: the group stayed alive,
   `_disposed` stayed False, the streamer stayed in `_ACTIVE`, and the
   `KeyboardInterrupt` displaced the caller's own exception. It now carries
   `except BaseException`, the escalation, and a bare `raise`, and all three
   disposals put `_settle()` in a `finally`. `agents/proc.py` moves from 0 to 1
   in the `base_disposal_guards` column, so the rule is mechanised for all four
   modules rather than three plus a comment explaining the fourth.
4. **Three vocabulary gaps.** `from os import wait` then `wait()` gave 40 passed
   on the guard file and 4977 on the full suite, because `wait` cannot go in the
   vocabulary (the bare word costs 9 false rows) and had no receiver for layer 2
   to resolve. Fixed by resolving the `ImportFrom`, which is the one place the
   ambiguity is removed. `loop.subprocess_exec`, `loop.subprocess_shell` and
   `webbrowser` are enrolled, measured at zero rows changed. The three receiver
   forms that need real dataflow are disclosed as row 51 with a test.
5. **`getoutput` and `getstatusoutput` were ungated.** The deleted rule banned
   them; the surviving `SPAWN_FUNCS` omitted them. Neither can take a `timeout`
   (`inspect.signature`: `(cmd, *, encoding=None, errors=None)`) and both run
   `sh -c cmd` and wait forever. Both are enrolled now. Not used in `kstrl/`
   today, so this is prevention.

### Two corrections to what this body previously claimed

- **The 600-second sentinel hang was not inherited.** `origin/main`'s reader
  `finally` contained no close at all. This PR introduced that bug during its own
  development and then fixed it. Saying otherwise credited the branch with
  repairing something it broke.
- **Both withdrawn mutations reached the right verdict for the wrong reason,
  and one of them was hiding a real defect.** `suppress(Exception)` does not
  cover `BaseException`, so with the sentinel written after the close a
  `KeyboardInterrupt`, `SystemExit`, `GeneratorExit` or `MemoryError` in the
  close still skipped the `put`. The withdrawn "sentinel first" mutation was
  strictly safer than what shipped, and the inline comment argued for the
  opposite of what the code did. The sentinel is posted first now, with a test
  that goes red on the old order. Breadth and order are two halves of one rule,
  not two arguments for one property.
- **The `spelled_tokens` measurement was wrong a second time.** It said the
  dotted split changes the inventory in "exactly ONE place" and it is two:
  `procgroup.py` as well as `procdispose.py`, from docstrings writing
  `agents.proc.DeadlineStreamer`. Both rows were already pinned, so only the
  prose was wrong - which is the rot that paragraph itself warns about, one round
  later. The docstring now says the number has been wrong twice and that the
  pinned rows, not the sentence, are the mechanism.

## Four more instances the sweep found, that neither issue names

1. **The abandoned generator, at five sites.** The five modules that own a child
   through `DeadlineStreamer` disposed of it on the straight-line path only.
   Four of them are generators. `finally: streamer.close()` is what reaches the
   abandonment path: CPython throws `GeneratorExit` at the suspended `yield`
   when a generator is closed or collected, and a `finally` that does not itself
   yield runs there.
2. **`except TimeoutExpired` alone, at two sites.** `verify` and `serve` guarded
   their collection with the timeout type only, where `procgroup._read_ps` has
   caught `BaseException` since #309. A `KeyboardInterrupt` or a `MemoryError`
   in the grace leaves the same child behind.
3. **A frozen constant.** `PS_KILL_GRACE_SECONDS` had become a default argument,
   evaluated once at import, so an operator tuning it changed nothing. Measured:
   two bound tests set it to 0.02 and got 1.0. `grace` is now a required
   argument on both disposals, which is a shape that cannot freeze.
4. **A close that skipped the queue sentinel.** The reader thread's pipe close
   sat above `self._queue.put(None)` under `suppress(OSError, ValueError)`. A
   `MagicMock` stdout raises `AttributeError` on close, which escaped, the
   thread died before posting the sentinel, and `lines()` waited on an unbounded
   `queue.get()`. Measured at 600 seconds on a real test run before it was
   found. The sentinel is now the last statement and the suppression is
   `Exception`, argued on `procdispose.close_quietly`.

## The shape of the fix

Six names cover the class, in two files that are one home.

`kstrl/procgroup.py` decides and sends: `safe_pgid` answers the question about a
`Popen`, `_refuse_group` is the one spelling of the two-part rule underneath it,
`signal_group` answers the same question about the bare integer that ends up in
`killpg`, and `signal_process_tree` sends the signal and degrades to the direct
child. `os.killpg` went from three modules to one.

`kstrl/procdispose.py` lets go: `drain_or_abandon` when the caller owns the
pipes, `reap_or_abandon` when other threads do. The split is measured, not
stylistic: with a reader thread blocked on stdout, a `close` from a second
thread had not returned after 3.005s, in both directions, because a
`BufferedReader` holds its io lock through a blocking raw read.

It is two files because one was 871 lines and the 800-line ratchet is a real
bound, and because reading a group and letting go of a child are two jobs.

## The site inventory

Every site in `kstrl/` that creates, signals, waits on, reaps or abandons a
child or a process group. The five columns are the five questions the sweep asked.

### The four that construct a `Popen`

| Site | pgid guard | kills: reaps + registers? | closes pipes on every path? | every wait bounded? | zombie (#298) |
|---|---|---|---|---|---|
| `procgroup._read_ps` | n/a, no signal | yes, `drain_or_abandon` | yes, `finally` under `except BaseException` | yes, `PS_TIMEOUT_SECONDS` then `PS_KILL_GRACE_SECONDS` | reads via `ps`, excludes state `Z` |
| `agents.proc.DeadlineStreamer` | `signal_process_tree` | yes, `reap_or_abandon` | yes, by the threads that own each end | yes, `term_grace`, then two 1.0s joins | n/a |
| `serve.run_supervised` | `signal_process_tree` via `kill` | yes, `drain_or_abandon` | yes, timeout path and `except BaseException` | yes, `timeout` then `ABANDON_GRACE_SECONDS` | n/a |
| `verify.run_scrubbed` | `signal_process_tree` | yes, `drain_or_abandon` | yes, timeout path and `except BaseException` | yes, `timeout` then `term_grace` | n/a |
| `procdispose.drain_or_abandon` / `reap_or_abandon` | n/a, the caller signalled | this IS the reap and the register | drain closes, reap deliberately does not | yes, `grace`, a REQUIRED argument | n/a |

`procdispose` is not a `Popen` construction site and is on the wait scope for
that reason: `POPEN_ALLOWLIST` says who may make a child, `CHILD_WAIT_SCOPE`
says who may wait on one, and they are two lists because conflating them is
how the gate went blind.

Before this PR: rows 2, 3 and 4 all answered **no** to "reaps and registers",
rows 3 and 4 answered **no** to "closes pipes on every path" (they had `except
TimeoutExpired` only), and rows 2, 3 and 4 each carried their own copy of the
`killpg` routine.

### The sites that signal without constructing

| Site | Before | After |
|---|---|---|
| `serve.terminate_process_group` | raw `os.killpg(pgid, sig)`, no guard (#329) | `signal_group(pgid, sig)`, refusal reported into the `GroupTermination` record |
| `serve._pid_alive` | its own inline `os.kill(pid, 0)` branch table, a fourth copy | `procgroup.pid_is_alive`; a zombie reads alive, which is correct for a LEASE (the pid is still allocated) and deliberately opposite to `read_group_liveness` |
| `verify._signal_process_group` | hand-rolled `getpgid` + `killpg` + fallback | one-line forward to `signal_process_tree` |
| `agents.proc.DeadlineStreamer._signal_group` | the same routine, written out again | one-line forward to `signal_process_tree` |
| `procgroup._kernel_says_group_is_empty` | own ESRCH branch table | `_probe_says_gone`, shared with `pid_is_alive` |
| `procgroup.signal_probe_alive` | already delegating (#309) | unchanged |

### The five that own a child through `DeadlineStreamer`

`agents/claude_code.py`, `agents/claude_sdk.py`, `agents/codex.py`,
`agents/custom.py`, `agents/liveness.py`. All five disposed on the straight-line
path only; all five now wrap the streaming region in `try: ... finally:
streamer.close()`. Four are generators, which is the shape that was open.

### Everything else, checked and left alone

- **13 modules call `subprocess.run` behind a `timeout=`** and nothing else:
  `breaker`, `contract`, `fixtures`, `git`, `intake_github`, `licensing`,
  `observability`, `pr`, `retry_plan`, `statedir`, `timeout`,
  `tui/screens/home`, `tui/screens/retry`. `subprocess.run` reaps its own child
  and closes its own pipes; the #309 bound on it is enforced by
  `tests/test_timeout_enforcement.py`.
- **`kstrl/factory.py`** spells `terminate` and `kill`, but they are
  `multiprocessing.Process` methods on `ProcessPoolExecutor` workers the
  executor owns, not a `Popen` this module spawned. **In class but NOT TOUCHED:
  the file is held by open PR #332.** The verdict is recorded here so the next
  sweep does not have to re-derive it.
- **`autonomy.py` and `inbox.py`** spell `system`. It is the word, not the call:
  `actor: str = "system"`. Pinned as rows rather than dropped from the net,
  because a net with a hand-tuned exception list is a net with a hand-tuned
  hole.

## The guard: what it sees, and what it does not

`tests/test_process_lifecycle.py` is what stops a sixth site landing outside the
six names. It is two layers, and the distinction between them is the whole
lesson of #324 - ten logged instances of a guard that passes the mutation it
exists to catch.

**Layer 1 is closed by construction, and that is the only claim here that does
not rest on a matcher being right.** `process_primitive_spellings` enumerates no
node types at all. It reads every string in every field of every node via
`ast.iter_fields`, adds each dotted component, and intersects with a fixed
vocabulary. There is no resolution in it to get wrong, so widening it is not a
thing a future author has to remember to do: a module that can reach a process
primitive has to SPELL it somewhere, and `EXPECTED_PROCESS_MODULES` pins who
spells what. That is the same property `EXPECTED_JOURNAL_PATH_SITES` has in
`tests/test_journal_one_writer.py`: it inventories every place the resource is
obtained, rather than every place somebody gave up.

**Layer 2 is a good error message, and it is a hand-rolled matcher.** Saying
which line, and catching the things layer 1 is structurally blind to - a SECOND
`Popen` inside a module that already spawns, a deleted disposal, a `finally`
that stopped disposing - needs resolution, and resolution is what #324 is about.
Four of the functions in `tests/helpers/proclifecycle.py` are domain-neutral AST
resolution and every one is a copy of something this suite already has:
`own_nodes` is the third copy of the innermost-scope walk and the three already
disagree; `os_module_names` reimplements the import half of
`test_tui_config_walk.collect_bindings` and none of the fixed-point rebind half
of `test_toml_readers._module_aliases`; `_receiver_is_os` needed
`test_tui_config_walk.dotted_name`; `callee_names` is the sixth spelling of
"reduce a `Call` to its callee name" in `tests/`. **That is what is deliberately
left for #324**, named in the module docstring so the issue can find it. The
process rules - the vocabulary, the disposal matchers, the spawner table - are
about processes rather than about ASTs and will be rewritten on top of
`astwalk.py`, not deleted by it.

**Five shapes are DISCLOSED rather than covered**, and each one is asserted in
`TestWhatTheGuardCannotSee` so the disclosure fails if it stops being true:

| Shape | Why |
|---|---|
| `__import__("".join(...))` with no constant fold | The module name does not exist until run time |
| `getattr(os, name)` with a computed attribute | Same, for the attribute |
| a spawner reached through a PARAMETER | Needs dataflow this walk does not do |
| a `shutil` helper that shells out internally | `shutil.which` spells nothing, measured at 0 `subprocess` references in `inspect.getsource(shutil)` |
| `o.wait()` through a subscript, an index or a parameter | The one name where layer 2 IS the coverage, see below |

**One claim in the guard's docstring had a silent exception and now states it.**
"Layer 2's resolvers are message quality, not coverage" is true for every name
but `wait`. The bare word cannot go in the vocabulary - measured, enrolling it
puts 9 modules in the inventory for `threading.Event.wait`, `Queue.wait` and
`Popen.wait`, and a net that reports 9 false rows is a net that gets silenced -
so for `wait` alone the receiver resolution is the coverage, and row 51
discloses the three forms that need real dataflow.

**Two rules of this class are deliberately NOT in this file.** "Every spawn that
waits carries a deadline" and "every wait inside a spawner carries one" are
#309's rules, and `tests/test_timeout_enforcement.py` has held both since #309
round 2. This PR first carried a second, weaker copy of each - the copies read
only `not any(kw.arg == "timeout")`, so `wait(timeout=None)` and `with
Popen(...)` walked through, which are the two forms #309 round 1 recorded as
having escaped its own first version, and they matched the LOCAL callee name, so
`from subprocess import run as _run` walked through the C2 alias fix too. Two
nets on one rule, the newer one weaker, is how a rule ends up enforced by
whichever copy is looked at. **They were deleted rather than widened**, and the
class docstring names the two tests that carry them.

`tests/test_process_lifecycle_shapes.py` is the positive control: 51 evasion
shapes, each with a verdict and a named test, and the table is ASSERTED rather
than decorative - `test_every_named_test_in_the_evasion_table_exists` parses the
rows and fails on a name that no longer resolves. The review found three ways
that assertion was weaker than it read, all now fixed: it matched with
`startswith` in BOTH directions, so rows 3, 4 and 5 were mutually satisfiable
(exact match now, which immediately caught four rows whose names had already
drifted); the vacuity control checked only the row count and the verdict
alphabet, never that the numbers are 1..51 and distinct; and
`test_every_bare_syscall_name_is_matched` iterated the same frozenset it
intersected with, so deleting `killpg` from the vocabulary made the loop run one
fewer time and pass.
`tests/test_procgroup_disposal.py` tests the six names on behaviour rather than
on placement, including an end-to-end test that abandons a real `CustomAgent`
generator mid-yield and asserts the child is gone.

## `/simplify`

Four review agents ran over the diff: reuse, simplification, efficiency,
altitude. Their reports are reproduced verbatim below. Verdicts first.

### Applied (26)

| # | Finding | What changed |
|---|---|---|
| A2 | `STREAMER_DISPOSALS` accepted `finish` and `kill` | Narrowed to `{"close"}`. `finish` is the ORDERLY disposal and gives the child ten seconds to leave; on the abandonment path that is ten seconds of spend on a discarded answer. The guard would have permitted exactly the shape `close`'s docstring argues against. |
| A4, S2 | `_finalise(process, *, close_pipes: bool)` | Replaced by `_register_if_unreaped`. The flag meant "skip the first statement"; `drain_or_abandon`'s `finally` is now two lines and `reap_or_abandon`'s is one. |
| A5, S8 | Four parallel dicts over the same modules | One `SpawnerRules` frozen dataclass, one row per module, four columns. Zeros are written out, and the measurement now runs over EVERY module rather than only the pinned ones, so a sixth spawner mismatches an implied zero rather than vanishing from a filtered dict. |
| A6 | `THIS FILE IS WHAT #324 SHOULD REPLACE` overstated by half | Corrected to name the four domain-neutral functions, with the prior art for each. An overstated scope is how a cleanup gets abandoned when it turns out bigger than advertised. |
| A7, S4a | One six-line rationale comment, copied verbatim four times | One line each, pointing at `DeadlineStreamer.close`. |
| S1 | `_grace()` and the `float \| None` it forced | Deleted; `grace` is a required argument on both disposals. A required argument cannot freeze at import the way the default did. |
| S3 | The three-line disposal tail written out three times | `_settle()`. `_breach` now guards on `_disposed` like the other two, with `timed_out` set above the guard because a deadline that fires after `finish` still has to record it. |
| S6 | `ABANDON_GRACE_SECONDS` defined 23 lines after its first use | Moved above `run_supervised`. The second bare `10.0` in the escalation table is now `GROUP_KILL_GRACE_SECONDS`, with a comment saying why it is a different quantity rather than a shared constant. |
| S10 | The 44-row evasion table was decoration with no mechanism | `test_every_named_test_in_the_evasion_table_exists` parses the rows and resolves every name, plus a vacuity control asserting all 44 rows still parse. Verified RED against two planted renames. |
| S11 | `function_scopes` public, one caller, no importer | `_function_scopes`. |
| S12 | `_FakeStreamer.closed` write-only | Deleted. |
| S13 | A comment contradicting the row directly under it | Corrected: `killpg` and `kill` are gone from `serve` and `verify`; `agents/proc.py`'s `kill` is `DeadlineStreamer.kill` and stays. |
| E1, E2 | `process_primitive_spellings` computed twice over 127 modules (160ms, 14% of the file's wall clock); `_module_trees` re-globbing and re-reading on all 9 call sites | Both cached. |
| E4 | The `spelled_tokens` docstring claimed the dotted split "changes the inventory in exactly zero places" | Re-measured: it is ONE place. `procdispose.py` gains `DeadlineStreamer` from two docstrings that write `agents.proc.DeadlineStreamer`. Docstring corrected, row pinned with the reason on it. A measured claim in prose does not re-measure itself, which is the argument for the pinned row. |
| E5 | `procs.read_pid` slept a flat 0.05s between polls: 481ms of the new file's 830ms | 0.005s. |
| E6 | `finish(timeout=0.2)` was the slowest test in the file | 0.02s; the assertion is a kill count, not a duration. |
| E7 | `close()` blocks for up to ~12s on the abandon path and the caller does not write the moment it runs | Stated in the docstring with the measurements (1.6us after an orderly `finish`, 1.3ms on a well-behaved child, 5.01s on one that traps SIGTERM), including that a generator's `finally` can be run by the collector at an arbitrary allocation point. |
| E10 | The diff tripled `_ABANDONED`'s producers and left its only sweep attached to `_read_ps`, which a `ks verify` process never reaches | `reap_abandoned()` now runs at the top of both disposals, so the producers sweep themselves for one `WNOHANG` poll per prior entry. |
| R7 | `_receiver_is_os` handled only a bare `ast.Name`, so `ns.os.kill(...)` was invisible | Flattens the attribute chain, which `test_tui_config_walk.dotted_name` had been doing all along. #324's case in one line. |
| R8 | `_is_broad` saw two of the four spellings of a broad handler | Four: bare, the name, `builtins.BaseException`, and a one-element tuple. The direction was fail-safe, and "it fails in the right direction" is how a matcher stays wrong. |
| R10 | `pid_is_alive` was a THIRD copy of the ESRCH branch table, in the file whose own docstring says "writing the branches out twice is how they drift apart again" | `_probe_says_gone(send)`. Each caller keeps its own eligibility guard, because a group must be one we may signal at all and a pid must be positive. |
| R11, E9 | `safe_pgid` and `signal_group` each spelled the two-part refusal rule for themselves | `_refuse_group(pgid) -> str`, returning the evidence because `serve` writes it into a `GroupTermination` record. `_may_signal_group` stays separate underneath: the signal-0 probes legitimately ask about our own group. |
| R13 | Two "close a pipe and swallow" helpers with two different policies, the narrow one on the disposal path | One `procdispose.close_quietly` with `suppress(Exception)`, carrying the 600-second measurement that justifies the width. |
| R1, R2 | Two rules re-implemented weaker than the existing versions next door | Deleted, with the class docstring naming `test_timeout_enforcement.py`'s two tests. |
| R3 | The `Popen` allowlist written out a second time | `test_the_spawner_table_names_the_same_modules_the_popen_allowlist_does` asserts the two agree, so a fifth spawner argued for in one place cannot be missing from the other. |
| R16, R17 | `_wait_gone` and the `SLEEPER` command string, the latter re-spelled inline five times in one file | `procs.wait_for_pid_to_die` and `procs.SLEEPER`, next to the group twin they mirror. |

### Declined (12), with reasons

| # | Finding | Why not |
|---|---|---|
| A1a | Make `lines()` dispose of itself in a `finally` | **Measured wrong.** CPython runs the inner generator's `finally` on `break` and on normal exhaustion too, not only on abandonment. `claude_code.py` breaks on the result event and then calls `finish(timeout=10)`; the other three fall out of the loop and call `finish()`. A self-disposing `lines()` would fire `close` first on the SUCCESS path, set `_disposed`, SIGTERM the child, and turn every `finish` in the tree into a no-op. There is no `except GeneratorExit` variant that rescues it, because `break` throws `GeneratorExit` too. |
| A1b, S4b, S5 | `__enter__`/`__exit__` so each site becomes `with streamer:` | Right shape, wrong PR. It replaces the five wrappers rather than removing them, and it needs `_disposes_in_finally` widened to accept an `ast.With`, a control for it, and `__enter__`/`__exit__` on the one streamer fake. Filed for #324's brief along with the four holes the altitude agent measured in the current matcher, which are real and are disclosed rather than hidden. |
| A3 | A `procgroup.owned()` context manager | Would ADD a mechanism. Two of the three sites need the drained output on the timeout branch, which a context manager cannot hand back. |
| S7 | Delete `verify._signal_process_group` and `DeadlineStreamer._signal_group`, both now one-line forwards | Real, and it is test churn across `test_hitl_env_scrub.py` and `test_timeout_enforcement.py`, neither otherwise touched here. The two wrappers cost four lines of code; retiring the tests that pin behaviour through them costs review attention this PR has already spent. |
| S9 | Parametrize the two pairs of near-identical test methods | Two of the four were deleted outright under R1 and R2, which is the stronger version of the same finding. The remaining pair now share `_pinned`. |
| R4 | `getpgid` is guarded by both `BARE_SYSCALLS` here and `test_safe_pgid.py`'s 14-row net | Kept. Unlike R1 and R2 this is not one rule with two nets: layer 1 here is a SPELLING census over the whole vocabulary, and dropping one token from a vocabulary to avoid an overlap is how a net starts having a hand-tuned hole. Both run; the stronger one still fires. |
| R5, R6, R9 | `own_nodes`, `os_module_names`, `callee_names` duplicate existing helpers | This is #324's deliverable by name, and it is disclosed in the module docstring with the prior art for each. Fixing three of the twelve copies in this PR leaves the same problem with a smaller number on it. |
| R12 | `drain_or_abandon` / `reap_or_abandon` share ~6 lines of skeleton | The reviewer's own verdict: no change worth making. The two-name public API is load-bearing for callers. |
| R18 | A shared `unreapable_process()` factory beside `procs.unkillable_ps` | Left. The existing fake is bound to `PS_ARGV` through `_patch_ps_popen` and is not directly reusable; extracting a second one is a helper-file change for one test class. |
| E3 | Three AST walks per module where one would do (221.8ms) | Left. Merging them means one walk with a dispatch inside it, which is a mechanism added to save 0.2s in a test file that now runs in 1.1s. |
| E8 | `_grace()` is clean | Moot: it was deleted under S1. |
| E11 | `test_journal_one_writer._PARSED` retains 65.7MiB for the session | Pre-existing, and out of this diff. Recorded so the third consumer knows the session-scoped-fixture shape is the answer. |

## Files this PR does not touch, and why

Per the file locks on the open PRs: `kstrl/factory.py` (#332) is in class and is
left alone, with its verdict recorded above. `kstrl/cli.py`,
`kstrl/decompose.py`, `kstrl/decisions.py`, `kstrl/tui/session.py`,
`kstrl/evolution.py`, `kstrl/autonomy_replay.py` and `kstrl/pipeline.py` are not
in class: none of them spells a process primitive.
`kstrl/decompose.collect_agent_output` is named repeatedly above as the caller
that abandons a generator, and it is NOT changed here - the fix is on the
callee, which is the right level for it and keeps this diff out of #332's file.

## Mutation evidence

**Round 1: 67 planted, 63 caught, 4 disclosed and pinned.**
**Round 2, after the review: 38 more planted, 38 caught.** Every one under the
`__pycache__` discipline: CPython invalidates bytecode on `(mtime_seconds,
size)`, so two same-size edits inside one second reuse stale bytecode and give a
false pass. Every run is `find . -name __pycache__ -type d -exec rm -rf {} +`
then `PYTHONDONTWRITEBYTECODE=1 uv run pytest ...`. A run not obtained that way
is not evidence.

Round 1 was 42 evasion shapes (38 RED, 4 disclosed with an anti-vacuity test
each) and 25 site reversions (all RED), including `S02` for #326's abandoned
child and `S04` for #329's unguarded signal.

Round 2 is every finding the review raised, planted as its own mutation:

| | mutation | result |
|---|---|---|
| R1, R2 | `from os import wait` / `... as _w`, then a bare call | RED |
| R3 | `loop.subprocess_exec` | RED |
| R4 | `webbrowser.open` | RED |
| R5, R5b | `subprocess.getoutput` / `getstatusoutput` | RED |
| R6 | codex's `finally` disposes a DIFFERENT object | RED |
| R7 | custom's `finally` uses `finish()` | RED |
| R8 | claude_code's `finally` uses `kill()` | RED |
| R9 | `procdispose` leaves `CHILD_WAIT_SCOPE` | RED |
| R10 | `agents/proc` narrows its broad clause | RED |
| R11 | **`reap_or_abandon` loses its deadline** (the surviving mutant) | RED |
| R12 | `drain_or_abandon` loses its deadline | RED |
| S23 | the sentinel stops being unconditional (`finally` to `else`) | RED |
| S23b | the sentinel is dropped from the `finally` | RED |
| S22 | `close_quietly` narrowed back to `(OSError, ValueError)` | RED |

**Two mutations changed verdict between rounds, and both changes are real
findings rather than bookkeeping.**

`S23` survived round 1. Nothing made the READ fail - the only regression test
raised from `close` - so a pipe dying mid-stream would have ended the reader
thread with no sentinel and left `lines()` on an unbounded `queue.get()`.
`test_a_read_that_raises_still_posts_the_sentinel` closes it.

`S22` survived a re-run in round 2, and that exposed the ordering defect above:
once the sentinel is posted BEFORE the close, narrowing the suppression no
longer produces a hang in the reader. Breadth is still load-bearing, on the
WRITER side, where there is no sentinel to reorder and a stdin end left open
blocks a child reading to EOF forever. `TestCloseQuietlySwallowsWhatever` pins
the breadth directly, and `S22` is RED again. The two halves of the rule now
have one test each instead of one test claiming both.

## Gates

| Gate | Result |
|---|---|
| `uv run pytest tests/ -q` | 5103 passed, 28 skipped, 6 xfailed, on the tree merged with `origin/main` |
| `uv run mypy kstrl/ --strict` | Success, 129 source files |
| `uv run pre-commit run` on the staged set | all hooks pass |
| `no file grows past 800 lines` | Passed. `procgroup.py` crossed it at 871 during the sweep, which is what forced the `procdispose.py` split; it is 616 now and `procdispose.py` is 339. |
| `cyclomatic complexity does not grow` | Passed. `undisposed_streamer_sites` hit 16 against the max of 15 while the receiver check was being added, and `_is_let_go_of` was extracted rather than the limit raised. |
| `forbid em dashes` | Passed |
| `Detect hardcoded secrets` | Passed |
| coverage ratchet, baseline 79.91 | 90.28% TOTAL, `coverage report --fail-under=79.91` exit 0. `procdispose.py` 100.00%, `procgroup.py` 98.44%, `agents/proc.py` 90.60%. |

**A note on `pre-commit run --all-files`**: it reports many complexipy failures
across `docs/` and `scripts/`, and every one is an artefact of the invocation
rather than a finding. The hook runs with `--staged`, so with the whole tree
staged it grades the whole tree against a limit that only new code is held to.
On a pristine `origin/main` worktree with nothing staged it reports Passed, and
on this branch's actual staged set it reports Passed. The gate that matters is
`git commit`, and it is green.

## What was declined this round, and why

- **Test-module double collection was fixed, not just noted.**
  `from tests.test_timeout_enforcement import TestSubprocessTimeoutAudit` made
  pytest collect 26 tests a second time under this file's node IDs, and the same
  applied to `TestWhatTheGuardCannotSee` in the shapes file. Both are now module
  imports, which is why the suite total moves down by 32 and then up by the new
  tests.
- **`docs/lessons/verify/pr-221/admission_gates.py` is left alone.** It is
  identical to `origin/main`, this PR touches no file under `docs/`, and
  rewriting a lesson artefact to satisfy a hook that does not gate it would be
  scope creep in a PR that is already large.
- **The `with DeadlineStreamer(...)` conversion still belongs to #324**, now
  with the receiver-binding and enclosure checks in place as the interim. The
  four holes the altitude review measured in the old matcher are two fewer:
  the wrong-object hole and the already-run-block hole are closed here, and the
  `with`-statement and cross-function ones are what the conversion is for.

---

The four `/simplify` reports are reproduced verbatim in a comment on this PR;
they did not fit alongside the review round inside GitHub's 65,536-byte body
cap. Verdicts for all 38 findings are in the two tables above.


0xfauzi:
## `/simplify`: the four reports, verbatim

Overflow from the PR body, which hit GitHub's 65,536-byte cap once the
review round was added. Verdicts for all 38 findings are in the body's
applied and declined tables. Reproduced unedited except that em dashes are
replaced with a hyphen, per the repo style rule.

## REUSE angle

Findings written to `/private/tmp/claude-501/-Users-wumpinihussein-Documents-code-ralph/85ffe45c-d098-4763-b6c4-141ae2357c6f/scratchpad/lane-flake-main/simplify/reuse.md`. Nothing in the worktree was touched, no tests were run. Every "what exists already" claim was checked by reading the named file, including the Textual 8.2.8 source under the venv.

## 1. HIGH - `parsed` / `label` / `test_sources` are a copy, not an import
`tests/test_settle_discipline.py:88-118` re-declares `tests/test_journal_one_writer.py:25-62`. The two `parsed` bodies are byte-identical and the docstring justifying the text-keyed cache is restated in both. The repo has a written house rule for this: `tests/test_journal_guard_detects.py:16-19` states it ("Importing the walkers across test modules is the house pattern here"), and `tests/test_event_names_have_one_home.py:56-62` does it (`from tests.test_journal_one_writer import assignment_parts, folded_str, label, package_sources, parsed`). The new guard is the one AST module that copies instead. Change: import `parsed` (its body touches no package anchor), delete lines 105-118; `label`/`test_sources` need a `root` parameter first.

## 2. HIGH - a third "nodes owned by this scope", in a new and costlier spelling
`tests/test_settle_discipline.py:390` (`parent_map`) + `:395` (`own_awaits`) + `:408` (`enclosing_function`) build a whole-module parent map and walk upward. `tests/test_toml_readers.py:300` (`_own_nodes`) and `tests/test_tui_config_walk.py:201` (`own_nodes`) are byte-identical to each other and answer it by not descending. The drift has already started: both existing copies treat `ast.Lambda` as a scope boundary, `enclosing_function` does not. Change: `own_awaits` becomes a filter over `own_nodes(fn)`, `enclosing_function` goes. Keep `parent_map` for the four functions that genuinely walk upward.

## 3. HIGH - a third copy of the fixed-point taint sweep
`app_derived_names`/`_binding_sweep` (`:320`/`:366`) is the same loop as `journal_aliases`/`alias_sweep` (`test_journal_one_writer.py:320`/`:341`) and `event_type_aliases`/`_alias_sweep` (`test_event_names_have_one_home.py:445`/`:476`), whose docstring already says it copied the first. On the binding side, `bindings` (`:252`) re-derives `assignment_parts` (`test_journal_one_writer.py:284`) and `_bound_names` (`test_event_names_have_one_home.py:486`) and extends them. The branch's version is the best of the three, so promote it and have the others delegate.

## 4. HIGH - `test_evolve_screen.py::_open` is a second copy of the shared helper, deepened by this branch
`tests/test_evolve_screen.py:79-106` now performs the identical three waits in the identical order as `tests/helpers/tui_screens.py:46-69`, same `what=` string included, and `:109-110` `_home_app` is character-for-character `home_app` (`:40-42`). On main these were only loosely related. The branch also restated the same three-condition rationale twice, at `tui_screens.py:8-22` and `test_evolve_screen.py:80-93`, which is what that module's own docstring says it exists to prevent. Change: extract `evolve_on(app, pilot)` in the helper, import it and `home_app`.

## 5. HIGH - `lambda: app.screen` written 80 times because `mounted` has no default
80 of 106 `mounted` sites spell the second argument that way and nothing else. `settle.py:155-161` gives the right reason the parameter is a callable, but that argues for a default, not 80 restatements. Checked and rejected: passing `app` directly, because `App._get_dom_base` (`textual/app.py:941-943`) returns `default_screen`, not `screen`. Change: default `node` to `lambda: pilot.app.screen`.

## 6. MEDIUM-HIGH - one site inlines `mounted`'s body to dodge an exception this Textual does not raise
`tests/test_tui_embed.py:360-367` says "`mounted` cannot serve: query_one on 'Button' would raise TooManyMatches". In 8.2.8 `query_one` (`textual/dom.py:1462-1533`) returns the first breadth-first match and raises only `NoMatches`/`WrongType`/`InvalidQueryFormat`; `TooManyMatches` belongs to `query_exactly_one` (`:1585`). Use `mounted` for the wait, keep the separate `query("Button")` for the labels.

## 7. MEDIUM-HIGH - `drained` reimplements two things Textual ships, more weakly
`settle.py:173-199` is a one-pump subset of `Pilot._wait_for_screen` (`textual/pilot.py:473-533`), which is the first statement of `Pilot.pause`. The public `MessagePump.wait_for_refresh` (`textual/message_pump.py:469-484`) is stronger: via `InvokeLater` -> `Screen._invoke_later` (`screen.py:1274`), `Screen._on_idle` (`screen.py:1175-1186`) refuses to run those callbacks while layout or repaint is pending, so it fires after the layout, which is exactly the condition the banner flake needed. The docstring derives everything from first principles and mentions neither.

## 8. MEDIUM - 23 of 25 `push_screen` calls are unawaited then polled, 2 await the awaitable
`App.push_screen` returns `AwaitMount` (`textual/app.py:2895`, documented at `:2916` as awaiting "the screen and its children"). Two sites await it, 23 do not. Inherited from main, but every one of these lines was touched. Note `AwaitMount` genuinely would not cover `EvolveScreen`'s lazily mounted `TabbedContent` panes, so the fix is "await plus a documented poll where a later frame is involved", confirmed with a run.

## 9. MEDIUM - `mounted` drops `query_one`'s `expect_type`
It mirrors two of the three overloads and not `query_one(selector, expect_type)` (`textual/dom.py:1457-1460`), so 9 sites hand-roll `cast(DataTable, ...)` and others carry `# type: ignore[attr-defined]`. Add `expect=` and forward it; a wrong type then becomes Textual's `WrongType` instead of a cast that lies.

## 10. MEDIUM - 41 byte-identical settle blocks, and only one file hoisted its repeat
Predicate and `what=` both identical: 5x `app._safe_mode_reasons`, 5x `... is not None`, 4x `table.rows`, 4x `preview_stage.display`, 3x `form.region.height`, 3x `app.store.state.components`, 3x `"comp-a" in ...`, plus seven pairs. `tests/test_config_screen.py:32` hoists its repeated message to `_LOADED` and uses it five times; that instinct is applied in one of the sixteen rewritten files. Around 150 lines.

## 11. MEDIUM - "the screen changed" spelled two ways, 49 times, never named
38 `isinstance(app.screen, X)` sites and 11 identity sites, both spellings within ten lines of each other at `tests/test_tui_detail.py:95` and `:103`. `settle.py` names three conditions and the second-commonest wait in the branch is not one of them. Add `screen_is` / `screen_left`. `App.pop_screen` does return `AwaitComplete` (`app.py:3096`), but nearly all these transitions go through `pilot.press`, so the framework API does not apply.

## 12. LOW - `mounted` has one shape; three sites need the other two
`tests/test_tui_detail.py:251` and `tests/test_decompose_screens.py:334` both inline `lambda: not screen.query(RunHeader)` with the same `what=`. A three-line `unmounted` sibling covers them; finding 6 covers the third site.

## Angles that turned up nothing
pytest and its plugins (`pytest-asyncio`, `pytest-cov`, `pytest-textual-snapshot`) ship no wait-for-condition. No existing `eventually`/`wait_until`/`poll_until` in the repo, so `settled` is genuinely new; the sync relatives in `tests/helpers/procs.py:47` and `:351` drive a different resource. `Pilot.wait_for_animation` / `wait_for_scheduled_animations` wait on the animator, not layout. `WorkerManager.wait_for_complete` is weaker than the four worker waits need and unbounded with no argument. The `hits`/`awaits` snippet helpers in `test_settle_shapes.py:36-48` are deliberate pattern reuse of `test_journal_guard_detects.py`, cited in its own docstring, not duplication.

---

## SIMPLIFICATION angle

Review complete. One important caveat first, then the findings.

**Worktree drift**: `/private/tmp/.../wt-subproc` is being edited live by another agent while I read it. `git status` now shows `MM` on `procgroup.py`, `proc.py`, `serve.py`, `verify.py` and an untracked `kstrl/procdispose.py` (304 lines) that splits the disposal half out of `procgroup.py`. Everything below is against the **staged** snapshot (`git show :<path>`), which is what I was asked to review. Finding 13 may already be in hand.

## Findings

**1. `kstrl/procgroup.py:638` - `_grace` and the `float | None` it forces exist for exactly one call site.**
Every caller passes a grace except `_read_ps` (`procgroup.py:529`, `drain_or_abandon(process)`). Simpler form: write `drain_or_abandon(process, PS_KILL_GRACE_SECONDS)` there and make `grace: float` required on both disposals. The late-binding the 11-line docstring argues for is satisfied identically - the module global is read inside `_read_ps`'s body at call time, so `monkeypatch.setattr("kstrl.procgroup.PS_KILL_GRACE_SECONDS", 0.02)` in `tests/test_procgroup.py:300,328,396` still works. Cost as written: a 14-line function (12 of them docstring) and an optional parameter on two public signatures, to avoid one argument at one site.

**2. `kstrl/procgroup.py:652` - `_finalise(process, *, close_pipes: bool)` is a two-statement body behind a flag that means "skip the first statement".**
Called twice, `True` at :587 and `False` at :634. Simpler form: a `_register_if_unreaped(process)` helper (the `returncode is None` check, which is the genuinely shared part), then `drain_or_abandon`'s finally is two lines and `reap_or_abandon`'s is one. Removes the flag, removes a name that says nothing, and puts `_release_pipes` visibly in the function whose docstring is about closing pipes.

**3. `kstrl/agents/proc.py:174-176, 211-213, 256-258 - the same three-line disposal tail written out three times.**
`self._reader.join(timeout=1.0)` / `self._writer.join(timeout=1.0)` / `_ACTIVE.discard(self)` appears verbatim in `finish`, `close` and `_breach`. The diff added the third copy. A private `_settle()` collapses it. Separately: `close()` (:178) is `finish(timeout=0)` plus an unconditional `kill()` - `Popen.wait(timeout=0)` raises `TimeoutExpired` immediately for a live child and escalates to the same `self.kill()`. The two methods differ by one line and 30 lines of docstring.
On `_disposed` specifically: **keep it.** It is not derivable. `returncode is not None` is true for a child that exited on its own, which still needs the joins and the deregistration; `self in _ACTIVE` would work only if the `discard` moved to the top of each method, which changes what `kill_active_process_groups` can still reach during the joins.

**4. `kstrl/agents/{claude_code.py:147, codex.py:146, claude_sdk.py:172, custom.py:64}` - one 6-line comment, copy-pasted verbatim four times.**
It also duplicates `DeadlineStreamer.close`'s own docstring (`proc.py:178-206`), which says the same thing better. Delete all four; the docstring is one hop away.
The bigger cost is structural: five hand-written `try:` / `finally: streamer.close()` blocks (`claude_code:186, claude_sdk:199, custom:88, codex:188, liveness:210`) plus the reindentation of every adapter body. `DeadlineStreamer.__enter__`/`__exit__` would make all five `with DeadlineStreamer(...) as streamer:` and delete the boilerplate. **But the new guard forecloses that**: `proclifecycle._disposes_in_finally` (:441) matches only `ast.Try` with a `finalbody`, and the failure message at `test_process_lifecycle.py:392` literally instructs "Wrap the streaming in `try: ... finally: streamer.close()`". A `with` form would be reported as undisposed. If the context-manager form is wanted, the matcher has to accept `ast.With` on a `DeadlineStreamer` too.

**5. `kstrl/agents/codex.py` - nesting reaches five levels.**
The new inner `try` sits inside the pre-existing outer `try` (last-message tempfile cleanup), so the loop body runs at method → try → try → for → if → if. This is the one adapter where the context-manager form in finding 4 buys a real indentation level, not just two lines.

**6. `kstrl/serve.py:1471` - `ABANDON_GRACE_SECONDS` is defined 23 lines after its first use, and its own comment points at the wrong neighbour.**
First read at :1448, second at :1463, defined at :1471. Legal, but every reader hits the name before the value. Its comment says it is named so "an operator tuning `GROUP_TERM_GRACE_SECONDS` needs to see it" - yet the bare `10.0` sitting directly under `GROUP_TERM_GRACE_SECONDS` in the escalation table (`:1562`) was left unnamed. The two `10.0`s are not the same quantity (drain grace vs post-SIGKILL wait), so this is not duplication; it is half the job done. Move the constant above `run_supervised` and name the second one, or name neither.

**7. `kstrl/verify.py:106` and `kstrl/agents/proc.py:260` - two one-line forwarders carrying 12-line docstrings each.**
Both bodies are now `signal_process_tree(x, sig)`. Both docstrings justify the wrapper by "tests pin the guard through it" - and those tests (`test_hitl_env_scrub.py`, `test_timeout_enforcement.py:379-460`) now duplicate `test_procgroup_disposal.py::TestSignalProcessTreeDegrades`. The honest simplification is to delete both wrappers, call `signal_process_tree` at the four call sites, and retire the pinning tests to the new behaviour file. Neither module's row in `EXPECTED_PROCESS_MODULES` changes. This is the largest single deletion available; it costs test churn, so it is a judgement call, not a defect.

**8. `tests/test_process_lifecycle.py:199, 211, 229, 250 - four parallel dicts keyed on the same four modules.**
`EXPECTED_POPEN_SITES`, `EXPECTED_COMMUNICATE_CALLS`, `EXPECTED_BASE_DISPOSAL_GUARDS`, `EXPECTED_DISPOSAL_CALLS` are all `dict[str, int]` over `{procgroup, agents/proc, serve, verify}`. Two of them carry a comment saying they are "KEYED ON `EXPECTED_POPEN_SITES` rather than on its own list" - which is prose doing what one shared row would do structurally. Simpler form: one table, one row per module, four columns (`None` for the communicate column at the home, where the rule is "outside the home"). A fifth spawner then cannot be added without filling in every column, which is exactly what the comments are asking for.
Also: the *key sets* of `EXPECTED_POPEN_SITES` and `EXPECTED_COMMUNICATE_CALLS` are already derivable from `EXPECTED_PROCESS_MODULES` (:138) - the rows containing `"Popen"` and `"communicate"` respectively, verified exact. Only the counts are new information.

**9. `tests/test_process_lifecycle.py:336 and :365, and :401 and :418 - two pairs of near-identical methods.**
`test_every_spawn_that_waits_carries_a_deadline` and `test_every_wait_on_a_spawned_child_carries_a_deadline` are the same comprehension over `untimed_calls(tree, <SET>)` with a different scope predicate and a different frozenset; one `@pytest.mark.parametrize` covers both. `test_every_popen_module_says_how_it_guards_a_broad_exit` and `test_every_popen_module_still_calls_a_disposal` both build a `dict[str, int]` over `EXPECTED_POPEN_SITES` and assert equality; folding finding 8's table makes them one assertion. Around 40 lines.

**10. `tests/test_process_lifecycle_shapes.py:35-82 - the 44-row table is unenforced and already incomplete.**
I checked it mechanically: all 44 "Test" cells do resolve to a real test name today, so it is not stale - but 27 of the file's 66 tests are absent from it, and nothing asserts the mapping. In a codebase whose rule is "if there is no mechanism, there is no plan", a hand-maintained index of test names sitting above the test names is decoration that has to be re-synced on every rename. Two options that both carry their weight: keep only the 4 `DISCLOSED` rows (those carry information the test names do not) and delete the 40 `COVERED` ones, or add a test that parses the docstring and asserts every named test exists. As written it is 48 lines of module docstring with no guard.

**11. `tests/helpers/proclifecycle.py:433 - `function_scopes` is public with one caller and no importer.**
Three lines of `ast.walk` + `isinstance`, used only by `undisposed_streamer_sites` at :453, imported by neither test file. Inline it or make it `_function_scopes`. To answer the brief directly: `own_nodes` and `function_scopes` are **not** redundant and neither is derivable from the other - one enumerates scopes, the other enumerates nodes no nested scope owns. `own_nodes` is the substantive one and is directly tested (`test_own_nodes_stops_at_a_nested_scope`).

**12. `tests/test_claude_sdk_agent.py:131, 149 - `self.closed` is write-only.**
Set in `__init__` and in `close()`, never asserted anywhere. The fake only needs the method to exist so the new `finally` does not raise `AttributeError`. Delete both lines and leave `close()` as `pass` (which is what `finish()` on the line above already does), or assert it.

**13. `tests/test_process_lifecycle.py:153-156 - a comment that contradicts the line under it.**
"`killpg` and `kill` are gone from all three non-home ones as of this PR; a token coming back is a hand-rolled copy of a `procgroup` routine" sits immediately above `"agents/proc.py": ("DeadlineStreamer", "Popen", "kill", "subprocess")`. That `kill` is `DeadlineStreamer.kill` and `streamer.kill()` - legitimate and permanent. A reader following the comment would try to remove a token that must stay. One-word fix: "gone from serve and verify".

**14. `kstrl/procgroup.py` (staged) is 871 lines, 68% docstring.**
`_grace` is 1 line of code under 11 of docstring; `signal_group` is 44 lines of which 26 are docstring. Two disposal functions whose bodies are 8 and 7 lines carry 100 lines of prose between them. The untracked `kstrl/procdispose.py` in the worktree suggests this is already being addressed.

## Clean

- **`GroupSignal`'s four fields are all genuinely read.** `serve.py:1565-1575` reads `vanished`, `sent`, `denied` and `refused`, and maps the last two onto different `GroupTermination` columns for a documented reason; `signal_process_tree:423` reads `sent`. `tests/test_procgroup_disposal.py:137-151` pins `denied` and `refused` apart. Not over-built. (`vanished` is technically derivable as `not sent and not denied and not refused`, but the named field is worth more than the field it saves - leave it.)
- **`_close_quietly` (`proc.py:68`)** is two uses of a genuinely shared body with a real justification for `suppress(Exception)`. Fine.
- **No dead code left behind.** `ruff check kstrl/ tests/` passes, and an AST sweep for module-level names in `procgroup.py`, `agents/proc.py` and `tests/helpers/proclifecycle.py` with zero references anywhere in `kstrl/` or `tests/` found none. The only dead *state* is finding 12.
- One near-duplicate I checked and am **not** flagging: `_kernel_says_group_is_empty` (`procgroup.py:809`) keeps its own `os.killpg` + guard + branch table rather than delegating to the new `signal_group`. It is not redundant - `signal_group` refuses the caller's own group, and `_may_signal_group`'s docstring records that probing our own group with signal 0 is deliberately allowed. The module docstring's "THE ONE PLACE A REAL SIGNAL REACHES A GROUP" is accurate; "real" is load-bearing.

---

## EFFICIENCY angle

Findings, all measured in the staged worktree (note: a sibling agent began editing `kstrl/procgroup.py` mid-review; every number below is against the staged diff, taken before that).

## Test-guard cost

**1. `tests/test_process_lifecycle.py:352` - the second full `process_primitive_spellings` pass is the single biggest waste in the diff.**
`test_no_new_module_reaches_for_a_process_primitive` (line 274) and `test_every_spawn_that_waits_carries_a_deadline` (line 352) each walk all 127 modules with it. Instrumented over a real run of the file: 263 calls, 332 ms total; one full pass is **160 ms**, i.e. ~14% of the file's 1.13 s wall, spent computing an identical answer. Memoising the frozenset per tree removes it.

**2. `tests/test_process_lifecycle.py:256` - `_module_trees()` re-globs and re-reads all 127 files on every one of its 9 call sites.**
The brief's premise is wrong in the good direction: `parsed()` in `tests/test_journal_one_writer.py:55` **is** caching, keyed on source text, so the package is parsed **once** (140 ms cold), not seven times. What repeats is the I/O around it. Measured: cold `_module_trees()` 140.5 ms, warm 4.9 ms (rglob+sort 0.6 ms, 127 × `read_text` 1.8 ms, 127 dict lookups over 2.5 MiB of string keys 0.14 ms). Instrumented in-run: 9 calls, 197 ms. A module-level `_module_trees` cache saves the **8 warm repeats = 39 ms** and 1016 redundant `read_text` calls. Smaller than finding 1, and it is the cheaper of the two to fix.

**3. `tests/test_process_lifecycle.py:290` - three AST walks per module where one would do.**
`bare_syscall_calls(tree) + os_syscall_calls(tree, os_module_names(tree))` is three independent `ast.walk` sweeps over the same 237,523 nodes. Measured **221.8 ms** over the package, the most expensive single assertion after finding 1. For context, the file makes ~9 full-package sweeps in total and they account for essentially all of its 1.13 s: spellings 332 ms, syscalls 222 ms, `undisposed_streamer_sites` 182 ms, `calls_named("Popen")` 66 ms, `calls_named("communicate")` 68 ms.

**4. `tests/helpers/proclifecycle.py:168` - `spelled_tokens` costs 11%, and its "zero places" measurement is wrong.**
The dot-split over all 237k nodes: **171.0 ms** vs **153.7 ms** for the `spelled_strings`-only version, so **+17 ms (+11%)**. Cheap. But the docstring at line 181 says "over the 127 modules of `kstrl/` the split changes the inventory in exactly zero places" - I re-ran it: it changes **one**. `procgroup.py` gains the token `DeadlineStreamer` purely from prose (`agents.proc.DeadlineStreamer` inside two docstrings, split on `.`), which is precisely the prose false positive the next sentence claims prose is safe from. The row is pinned at `tests/test_process_lifecycle.py:143` with a comment that reads as if it were a real spelling. Either the docstring claim needs correcting or the row needs re-justifying.

## `tests/test_procgroup_disposal.py`

**5. Lines 73 / 292 / 391 / 420 / 477 - the real processes are not the cost; `procs.read_pid`'s poll is.**
Measured directly: `subprocess.Popen` of the `sh -c` sleeper is **1.8 ms**, `drain_or_abandon` on it is **0.3 ms**, and `procs.read_pid` is **54.5 ms** - because `tests/helpers/procs.py:62` sleeps a flat 0.05 s between polls and the shell has never written by the first check. Instrumented over the whole file: **10 calls, 481 ms of its 830 ms (58%)**. So no test here should be swapped for a fake - the real child is 0.2% of the file's runtime and every one of these assertions is about real kernel or real pipe behaviour. Dropping `read_pid`'s poll interval from 0.05 s to 0.005 s would take the file from ~0.83 s to ~0.39 s. That is a one-line change in a pre-existing helper, not in this diff, but this diff is now its heaviest user.

**6. `tests/test_procgroup_disposal.py:429` - `streamer.finish(timeout=0.2)` is the slowest test in the file at 0.26 s.**
The 0.2 s is a deliberate wait on a child that will not exit, and the docstring argues why the child must not exit. Nothing in the assertion (`kills == [1]`) needs 200 ms rather than 20 ms; `timeout=0.02` would give back ~0.18 s. Marginal.

`--durations=10` for the file, as requested: 0.26 / 0.09 / 0.08 / 0.08 / 0.06 / 0.06 / 0.06 / 0.06 / 0.05 / 0.01 s, 22 passed in 0.86 s. All three new files together: 2.25 s (lifecycle 1.21, shapes 0.07, disposal 0.86).

## Production code

**7. `kstrl/agents/proc.py:178` `close()` - free on the hot path, seconds on the abandon path, and that is the intended trade.**
Measured: orderly run then `finish()` then the new `finally: streamer.close()` costs **1.6 microseconds** (the `_disposed` bool short-circuits). On the abandon path with a child that honours SIGTERM: **1.3 ms**. With a child that traps SIGTERM and `term_grace=5.0`: **5.01 s**, and up to ~12 s worst case (5 s SIGTERM grace + 5 s `reap_or_abandon` grace + two 1 s thread joins). So it does add blocking work where there was none, but the thing it replaces is an agent CLI running to completion and spending tokens, so it is a large net win. The one thing worth stating in the docstring: a generator's `finally` can be run by the GC at an arbitrary allocation point, so that multi-second block can land somewhere the caller did not write it.

**8. `kstrl/procgroup.py:638` `_grace()` - clean, as you suspected.** One function call and one `is None` test per disposal, which is once per killed child. Not measurable. The default-argument argument in its docstring is correct and the cost is nil.

**9. `kstrl/procgroup.py:406` `signal_process_tree` - the pgid guard now runs twice per signal.**
`safe_pgid` already ends with `if not _may_signal_group(pgid) or pgid == os.getpgrp(): return None` (line 329), and `signal_group` (line 362) opens by re-running both checks. Cost is negligible - `os.getpgrp()` is **111 ns** - but it is two copies of the same invariant on the same call path, in a PR whose whole subject is removing duplicated copies of that invariant. Worth a comment saying the second check is for the bare-integer entry point and is redundant when reached through `signal_process_tree`.

**10. `kstrl/procgroup.py:250` `_ABANDONED` - the diff triples the producers and leaves one consumer.**
`_reap_abandoned()` is called from exactly one place, `_read_ps()` (line 508), reachable only through `read_group_liveness`, whose only production caller is `serve.py:1645` on a timed-out supervised run. This diff adds three new producers: `verify.run_scrubbed` (the highest-frequency spawn in the factory, once per verification command per iteration), `serve.run_supervised`, and `DeadlineStreamer.kill`. In a `ks verify` or `ks run` process that never calls `read_group_liveness`, `_ABANDONED` is append-only for the process lifetime, and `reap_or_abandon` entries additionally pin two **open** file descriptors (`close_pipes=False`). Only unreapable children register, so the rate is low and the fd limit here is 1,048,576, but the sweep is now attached to the wrong site. A `_reap_abandoned()` at the top of `drain_or_abandon`/`reap_or_abandon` would make the producers sweep themselves for the cost of one `WNOHANG` poll per prior entry.

**11. `tests/test_journal_one_writer.py:52` `_PARSED` - 65.7 MiB held for the whole session, pre-existing, now with a second consumer.**
Measured with tracemalloc: caching 127 ASTs plus their source text retains **65.7 MiB** (peak 67.5 MiB) for the life of the pytest process, never cleared. No xdist in `pyproject.toml`, so it is one copy, not one per worker. Not introduced here and not worth changing on its own, but the new guard file is now the second thing keeping it alive; if a third arrives, a session-scoped fixture with teardown is the shape.

## Clean

`kstrl/agents/proc.py:68` `_close_quietly` (two calls per streamer lifetime, one `suppress` context each), the `_release_pipes` widening from two pipes to three (`drain_or_abandon` end to end measures 0.3 ms), the four adapter `try/finally` wrappers (no work added to the streaming loop itself), and `import signal` in `procgroup.py` (a builtin C module already loaded) all cost nothing measurable. `tests/test_process_lifecycle_shapes.py` runs in 0.07 s and does not touch `kstrl/` - it is the cheap half of the guard and is built right.

---

## ALTITUDE angle

Everything I need is measured. Here is the review.

---

## ALTITUDE REVIEW - process lifecycle PR

Base is green (`81 passed` on the two static guard files). All claims below are measured, with the probe scripts at `/private/tmp/.../scratchpad/simpAlt_gen.py` and `simpAlt_guard.py`.

---

### 1. Should `DeadlineStreamer` own the disposal? A self-disposing `lines()` is measurably wrong; `__enter__`/`__exit__` is right but not for this PR.

**Self-disposing `lines()` - DECLINE, and it is not a taste call.** The suggestion is plausible on the abandonment path and fails on the other two. Measured with `simpAlt_gen.py`:

| consumer does | inner generator's `finally` runs? |
|---|---|
| abandons the outer generator (`close()` or GC) | yes |
| **`break`s out of the `for` loop** | **yes, before the code after the break** |
| **loop exhausts normally** | **yes, before the code after the loop** |

Rows 2 and 3 are fatal. `kstrl/agents/claude_code.py:163` `break`s on the result event and then calls `streamer.finish(timeout=10)`; `codex.py`, `custom.py` and `claude_sdk.py` fall out of the loop and call `finish()`. If `lines()` disposed in a `finally`, `close()` would fire first, set `_disposed = True`, SIGTERM the child, and turn every `finish()` in the tree into a no-op - replacing the orderly 10-second flush-and-exit with an immediate kill on the *success* path. And there is no `except GeneratorExit` variant that rescues it, because CPython throws `GeneratorExit` into the inner generator on `break` too (row 2). `lines()` cannot distinguish "consumer broke out and will finish properly" from "consumer walked away", so it must not decide.

**`__enter__`/`__exit__` - the right shape, DECLINE for this PR, file as follow-up.** It works mechanically: a `with` inside an abandoned outer generator does run `__exit__` (`simpAlt_gen.py` case d: `['enter', 'inner-finally:d', 'exit']`). Concretely it is `def __enter__(self): return self` / `def __exit__(self, *exc): self.close()` (~6 lines on `kstrl/agents/proc.py`), and each site becomes `with streamer:` replacing `try:` … `finally: streamer.close()` - so it **replaces** the five wrappers rather than removing them. One line saved per site is not the argument.

The argument is that it closes four holes I measured in the guard that enforces the wrapper (`simpAlt_guard.py`, all four **PASS** the guard today):

- **A** - `finally: logfile.close()` in the same scope excuses an undisposed streamer. `_disposes_in_finally` never checks the disposal targets the streamer.
- **B** - two streamers in one scope, one disposed: passes.
- **C** - `finally: streamer.finish(timeout=10)` passes (see finding 2).
- **D** - the `try/finally` sits in an unrelated `if` branch; the streaming loop is outside it.

`with DeadlineStreamer(...) as s:` makes acquisition and release structural, so all four close by construction - which is exactly the layer-1-vs-layer-2 distinction `tests/test_process_lifecycle.py`'s own docstring draws. **But** the conversion touches 5 adapters + the class + `tests/helpers/proclifecycle.py` + both new guard test files, and three of the four holes are shapes nobody has written. In a PR this size, take finding 2 instead and let #324 carry this, since #324 is already reworking the matcher.

### 2. `STREAMER_DISPOSALS` accepts the exact shape this PR exists to remove. **APPLY.**

`tests/helpers/proclifecycle.py:1601` - `STREAMER_DISPOSALS = frozenset({"close", "finish", "kill"})`. So `finally: streamer.finish(timeout=10)` satisfies the guard (measured, case C). `DeadlineStreamer.close`'s own docstring says why that is wrong: *"KILL RATHER THAN WAIT … giving it ten seconds to finish work whose result is already discarded is ten seconds of spend for nothing."* The guard permits the ten seconds of spend. Narrow it to `frozenset({"close"})`; every real site already calls `close`, so this costs nothing and the message ("Wrap the streaming in `try: ... finally: streamer.close()`") becomes true. One-word change, closes the only hole of the four that is adjacent to real code.

### 3. `with procgroup.owned(process)` for the three `except BaseException:` blocks. **DECLINE.**

The differing grace does not break it (it is a constructor argument). What breaks it is that two of the three sites need the *return value* of the disposal on a path the context manager cannot serve: `verify.py:180` and `serve.py:1447` both do `stdout, stderr = drain_or_abandon(...)` inside their `except TimeoutExpired` branch, and that branch is not an error exit - it builds the result. So `owned()` would have to coexist with an explicit `drain_or_abandon` at the same site, giving two disposal mechanisms per function and requiring `owned()` to track whether a disposal already happened - which is re-deriving `DeadlineStreamer._disposed` in a second place. Only `procgroup._read_ps` (`kstrl/procgroup.py:521`) would shrink cleanly, and it is the one site that already had the clause. Three copies of a 3-line block that each re-raise is within the budget for repetition; the pinned `EXPECTED_BASE_DISPOSAL_GUARDS` is what stops a fourth site omitting it.

Worth noting though: `tests/test_process_lifecycle.py`'s failure message prescribes the *shape* ("The clause is `except BaseException:`, a call to … and a bare `raise`") rather than the outcome, so the guard now cements this copy. That is acceptable for a pinned inventory, but say it in the PR body rather than letting a future author read it as forbidding `owned()`.

### 4. `_finalise(process, close_pipes: bool)` is a flag argument with two constant call sites. **APPLY.**

`kstrl/procgroup.py:652`, called at `:588` with `close_pipes=True` and `:635` with `close_pipes=False`. The caller has *already* expressed pipe ownership by choosing `drain_or_abandon` vs `reap_or_abandon`; the boolean is the same decision encoded a second time, one indirection lower. Delete `_finalise` and write:

```python
def _register_if_unreaped(process: subprocess.Popen[str]) -> None:
    if process.returncode is None:
        _register_abandoned(process)
```

then `finally: _release_pipes(process); _register_if_unreaped(process)` in `drain_or_abandon` and `finally: _register_if_unreaped(process)` in `reap_or_abandon`. The difference between the two disposals is then visible at the two disposals, which is where the module docstring argues it belongs.

Do **not** take the tempting further step of moving the `returncode is None` check into `_register_abandoned` itself: `tests/test_procgroup_disposal.py:3154` monkeypatches `_register_abandoned` with `registered.append` and asserts `registered == []` for a reaped child. That seam only works while the guard is outside it.

### 5. Five pinned inventories → one table with columns for the four spawner modules. **APPLY.**

`tests/test_process_lifecycle.py` has five tables. One (`EXPECTED_PROCESS_MODULES`, 25 rows) is layer 1 and must stay separate. The other four are all about the same four modules and should be one table:

- `EXPECTED_POPEN_SITES` - `{procgroup, agents/proc, serve, verify}`
- `EXPECTED_BASE_DISPOSAL_GUARDS` - same four, with an **explicit** `"agents/proc.py": 0`
- `EXPECTED_DISPOSAL_CALLS` - same four
- `EXPECTED_COMMUNICATE_CALLS` - **only two rows**, because its test filters on `if (count := ...)` so `agents/proc.py` drops out **implicitly** at zero, and `procgroup.py` is excluded by name

So two of the four express "zero" one way and the other two the other way, and a fifth spawner added to `EXPECTED_POPEN_SITES` forces a row in tables 2 and 3 (both keyed on it) but silently skips table 4. Replace with one frozen dataclass keyed on the module (`popen_sites`, `communicate_calls`, `base_disposal_guards`, `disposal_calls`), keep the four separate assertions by projecting a column (`{m: getattr(e, col) for m, e in TABLE.items()}`) so the four distinct failure messages survive. That makes the docstring's stated intent - *"a fifth spawner cannot be added without a row here saying which of the two it is"* - structurally true instead of true for two of four columns. ~30 lines, one file, no production change.

### 6. `tests/helpers/proclifecycle.py` vs #324. Clean seam is enough - but the PR deepens the wrong dependency. **APPLY the small half.**

Don't build anything to anticipate `astwalk.py`; the section markers are a good enough seam and pre-generalising a matcher nobody has designed yet is the opposite of this angle. Two corrections though:

- The module docstring says **"THIS FILE IS WHAT #324 SHOULD REPLACE"**, and that overstates by about half. `spelled_strings`, `spelled_tokens`, `callee_names`, `calls_named`, `own_nodes`, `function_scopes`, `_is_broad`, `_hit` are domain-neutral and belong in `astwalk.py`; `PROCESS_VOCABULARY`, `BARE_SYSCALLS`, `undisposed_streamer_sites`, `base_disposal_guards` and the rest are process rules that will be *rewritten on top of* it, not deleted. Saying "replace" will mis-scope #324 into an archaeology exercise. Change it to name the eight functions.
- More concretely: `tests/helpers/proclifecycle.py:1217` does `from tests.test_journal_one_writer import folded_str`, and `tests/test_process_lifecycle.py:1873` does `from tests.test_journal_one_writer import label, package_sources, parsed`. A helper importing a *test module* is the wrong direction, and this PR takes that module's importer count from 2 to 4. Those four functions are generic, ~45 lines, and are precisely the seam #324 needs. Moving them to `tests/helpers/astwalk.py` now - leaving `test_journal_one_writer.py` importing from there - is mechanical, shrinks #324's blast radius, and is the one piece of anticipation that pays for itself. If that feels like scope creep in a PR this size, it is a legitimate DECLINE; but then do not let the docstring claim the seam is already drawn.

### 7. Four verbatim copies of a six-line rationale comment. **APPLY.**

`grep -c` confirms the identical block in `claude_code.py`, `claude_sdk.py`, `codex.py`, `custom.py` (a fifth variant in `liveness.py`). Twenty-four lines of duplicated prose whose full argument already lives, better, on `DeadlineStreamer.close`. Four copies is four things to drift. Cut each site to one line - `# The consumer can walk away mid-yield; see DeadlineStreamer.close (#326).` - and keep the argument in one place. Same reasoning the PR applies to the code applies to the comment about the code.

---

**Summary of verdicts:** APPLY 2 (one word, real), 4 (flag argument), 5 (five tables → two), 7 (comment duplication), and the docstring half of 6. DECLINE 1's self-disposing `lines()` (measured wrong), 1's `with` conversion (right shape, wrong PR - hand it to #324 with the four measured holes as its brief), and 3's `procgroup.owned()` (the timeout branches need the drained output, so it would add a mechanism rather than remove one).

PR #341, opened 2026-09-02
fix(tests): the broadcast was the flake, and every TUI settle read now waits on a condition

0xfauzi:
## Summary

A CI flake in one TUI test. The brief said two `await pilot.pause()` calls were
too few for the LAYOUT. **That diagnosis was wrong: the flake was a BROADCAST.**
`KstrlTuiApp` checks safe mode on a thread and broadcasts `[]` to every live
screen, setting `display = False`, and a hidden widget answers with the zero
region. That is the `[0, 0, 1]` CI saw, character for character; a layout that
merely had not happened gives `[0, 0, 0]`. Waiting harder on the layout made the
test WORSE, 14 failures in 20 loaded runs against 0 for `main`, which is how the
wrong diagnosis was caught.

The class is real, so the sweep stands: 235 fixed `pilot.pause` calls across 17
files became 308 waits on a condition, through one helper with a wall-clock
deadline that RAISES rather than returning unsettled. Two AST guard layers keep
new instances out, a third guards the helper. Test-only; no production file is
touched.

| the four numbers | result | how it was produced |
|---|---|---|
| repeat runs | **0 failures in 40 runs**, 14,080 test executions | commit `1b5010f`: `uv run pytest` over the 23 converted-plus-guard files, 20x idle then 20x under 16 shell busy-loops on 10 cores (loadavg 4.07 to 28.70). Appendix N lists the 23 files and the per-run times |
| planted production defects | **19 planted, 19 caught**, 16 at the test's own assertion, 3 at a named settle condition | commit `1b5010f`: one one-line edit at a time to a `kstrl/tui/` file, then the TUI tier, `__pycache__` cleared and `PYTHONDONTWRITEBYTECODE=1` per run. Appendix G lists the 19 edits and where each landed |
| tests whose wait is load-bearing | **61** after the sweep, against **52** before it | AFTER, on `1b5010f`: `settled` given an early `return`, `uv run pytest tests/`, count failures outside the 5 guard files. BEFORE, on `583acd0`: every fixed non-loop `pilot.pause` replaced with `pass`, run one file at a time under a 420s bound, with the one file that hangs re-run per test at 60s. Raw pytest output for all 33 runs is kept. Appendix E |
| planted helper defects | **12 planted, 12 caught.** Ten passed silently before this round | commits `7c5afe3` and `1b5010f`: one one-line edit at a time to `tests/helpers/settle.py`, then the 5 settle files. Appendix K tables each with the test that names it |

[...]

PR #343, opened 2026-09-02
docs: the guard-design rules and the class-sweep convention the batch paid for

0xfauzi:
Documentation only. No behaviour change, no code touched.

Three things this batch cost real rework to learn, written into `CLAUDE.md` where the next author reads them before repeating them. `AGENTS.md` is a symlink, so it moves with it.

## The guard-design rules

The dominant defect class in this repo, and the one being re-fought inside every PR: **a guard that passes the mutation it exists to catch**. Eleven logged instances, and every single miss in the **skip direction**, the guard going blind rather than red. Four rules, all checkable:

1. **Closed by construction beats a ledger of give-ups.** Inventory every place the resource is OBTAINED and count what you saw, so a new shape shows up as an unexplained census delta. A ledger of the places the walk gave up is closed only over shapes someone already enumerated. `EXPECTED_JOURNAL_PATH_SITES` is the named precedent; `tests/test_event_names_have_one_home.py` is the current best, its outer layer enumerating no node types at all.
2. **`assert hits(...) == []` is not a control.** It passes when the walk is correctly narrow AND when the walk has been switched off, which is the same shape as the defect it is meant to exclude. A genuine disclosed miss belongs in a blind-spot record plus a strict xfail, so a later widening XPASSes and fails loudly.
3. **Know which direction your guard is wrong in.** A guard that FLAGS may over-match; the cost is a false positive someone reads. A guard that CLEARS must be narrow, because over-matching converts a resolution into a clearing and deletes the mechanism. #342 measured exactly that in three of sixteen migrated guards, one of them letting a config load escape onto the Textual event loop.
4. **A guard you did not mutate is a guard you did not test**, per layer, not per file.

## The `__pycache__` trap

A mutation result that did not delete `__pycache__` first is not a result. CPython invalidates bytecode on `(mtime_seconds, size)`, so two edits of the same size inside one second reuse stale bytecode and report a **false pass**. This has produced a wrong answer here more than once, so it is now written down with the exact command.

## Two conventions

- **One defect class, one PR.** The issue list grew by thirteen in three days under per-site fixes, and the growth was almost entirely the same five findings re-filed under new numbers.
- **Merging into a moving main is where the defects are.** Four defects in one batch survived a textually clean merge and were caught only by running the full suite on the merged tree. Hence: re-run every static guard's site census before and after the merge and diff the two, because a code move in another file changes what a walk sees without touching yours.

## Verification

Every claim was checked against the tree rather than taken from memory: `tests/test_event_names_have_one_home.py` exists, `EXPECTED_JOURNAL_PATH_SITES` is at `tests/test_journal_one_writer.py:307`, `tests/helpers/astwalk/` has its six modules, and the "thirteen flag, three clear" split is `tests/helpers/astwalk/__init__.py:43-44` verbatim.

`uv run pre-commit run --all-files` passes, including the em-dash hook and the AGENTS.md drift check.


0xfauzi:
## Fact-check round, and what it changed

A reviewer checked every factual claim in the added text against `origin/main` at `0f5bf45`, with instructions to ignore style entirely and verify claims only. Nothing came back FALSE, but **four claims came back UNSUPPORTED**, and chasing them found that one was a misstatement of the measurement behind it. All are now fixed in `f97504e`.

This matters more than a normal docs correction, because the entries being added are about evidence discipline. An unsourceable number in the document that tells you to source your numbers is the same defect it warns about.

### What could not be sourced, and what replaced it

| claim | verdict | now |
|---|---|---|
| "the issue list grew by thirteen in three days ... the same five findings re-filed" | UNSUPPORTED, and a misstatement of the original finding | the label query, which anyone can run: six `absorbed` and four `class-sweep` issues, all filed on one day |
| "three of twenty helpers were drop-in, fifteen had no equivalent, five needing a parent map" | UNSUPPORTED, came from a lane report never committed | numbers dropped, rule kept, with the checkable part cited: `grep -rn parent tests/helpers/astwalk/` finds only `Path.parent` |
| "four defects survived a textually clean merge", enumerating four | one confirmed verbatim, three unsourceable | cites the two that are recorded: `SPAWN_FUNCS` with zero readers (PR #342), and 750 + 800 merging to 827 against an 800 ratchet (PR #341) |
| the census pin's "13 + 4 where the pin said 17" | UNSUPPORTED from main, since PR #341 is unmerged | attributed to PR #341 and stated without the unverifiable figures |

The reviewer did real work to reach those verdicts, not a shrug: it pulled all 94 issues with creation and close dates and computed daily counts, and ran `git log --all -S` across 94 branches including unpushed ones.

### What was verified TRUE

- `EXPECTED_JOURNAL_PATH_SITES` at `tests/test_journal_one_writer.py:307`.
- `tests/test_event_names_have_one_home.py`'s outer layer really does enumerate no node types: the predicate is `lambda node: folded_str(node) in names` over `ast.walk`, with zero `isinstance` checks in the walk.
- `Bindings.attributes` marks an inferred origin `guessed` (`resolve.py:162-163`), and `resolve.py:500` uses it to gate the clear direction.
- "eleven instances", now attributed: #324 records ten, and `tests/test_timeout_enforcement.py:1337` names its own as instance eleven.
- **The CPython claim, by direct experiment rather than recollection.** Confirmed the default pyc header is timestamp-based, then imported a module with `X = 1`, rewrote the source to `X = 2` at the same byte length, forced the mtime into the same second with `os.utime`, and imported again in a fresh process. It returned the stale `X = 1`.

Two claims remain sourced from docstrings rather than from re-execution (the thirteen-flag three-clear split, and the instance count). Both are attributed to the file that states them, so a reader who doubts one knows where to go.

PR #347, opened 2026-09-05
fix(decompose): one classifier for the audit buckets, and no blank project name (#338)

### LINKED ISSUE #338: _excluded_history's one-bucket-per-audit invariant is false at project_name == ""
Found while reviewing PR #336 (issue #314). Pre-existing: measured identically on that PR's base and head, so it was not introduced there and was left out of its scope.

## The defect

`_excluded_history` in `kstrl/decompose.py` documents a partition property: every audit lands in exactly one of "own project", "other project" or "unattributed", so the three buckets sum to the audit count. At `project_name == ""` they do not.

Measured on one fixture, both sides of PR #336:

```
base  project='': audits=5 own=3 other=2 unattributed=3 sum=8 invariant_holds=False
head  project='': audits=5 own=3 other=2 unattributed=3 sum=8 invariant_holds=False
```

Five audits, eight bucket placements. Three audits are counted twice.

The cause is that the buckets are computed by two different rules. One is raw equality (`entry.get("project") == project`), the other treats a null, a non-string or an absent value as unattributed. For every non-empty project name the two rules agree and the partition holds, which is why nothing has caught this. At `""` they diverge: an audit with no project satisfies both "matches my project" and "has no project".

## Why it is reachable

PR #336's review established this independently: `ks decompose --project-name` is `required=True type=STRING` with `callback: None`, and Click accepts `""` (exit 0). The full `decompose_spec` path was driven at `project_name=""` without raising. So an empty project name is not a hypothetical, and #336 pinned the raw-equality rule as the production answer precisely because it was reachable.

## What needs deciding

Not just which rule wins, but whether `""` should be a legal project name at all. Two candidate fixes, and they are not equivalent:

1. Reject `""` at the CLI boundary. Then the partition holds by construction and no downstream code needs to know. This is a behaviour change to a currently accepted input.
2. Make the two rules agree. Then `""` stays legal and the partition holds because both sides treat an absent project the same way. Note PR #336 deliberately kept them divergent so that nothing already recorded changed, so this reverses a decision made with evidence, and owes its own.

Either way the invariant should be pinned by a test that fails on the current code.

## Verification

Construct a journal with three audits carrying no `project` key plus two carrying a different one, then compute the three buckets at `project_name=""` and assert they sum to the audit count. It currently reports 8 against 5.

0xfauzi:
## Summary

`ExcludedHistory` documents a partition: every recorded spec audit lands in exactly one of this project's, another project's, or no project recorded, so the three buckets sum to the audits on disk. The first and third counts were separate predicates over the same list, `entry_str(e, "project") == project_name` and `not entry_str(e, "project")`. Those ask the same question when `project_name` is `""`, so an audit whose project field is absent, null or not a string satisfied both and was counted twice. Measured on one fixture: five audits on disk, eight bucket placements.

`_audit_bucket` in `kstrl/decompose.py` is now the only place the rule lives, and one call per audit places it once, so the sum is an identity rather than two predicates happening to agree. The order is own before unattributed, which keeps `own_recorded` equal to what `EvolutionJournal.get_spec_issue_runs` counts at `""`: that method windows by the identical expression, so calling those audits unattributed here would report 0 under a trend showing 3, and would make the rendered note saying neither counts them false about audits the trend just counted. For every non-empty name the order is unobservable.

`--project-name` is also refused at the boundary. It is an identity: it keys the journal audits, the decision register and, under `--single-pr`, the branch. Click accepted `""` on `ks decompose`, and `ks factory`'s body check rejected `""` while passing `"   "`. `_reject_blank_project_name` is a Click callback on all three `--project-name` options in the CLI tree, exiting 2 and naming the option before the architect runs, which is measured at 119 to 210 seconds against a frontier model. It never strips: `" x "` passes through verbatim, because substituting `"x"` would write a manifest, a branch and a journal audit under a name the operator never typed.

The third option, `ks queue add --project-name`, is the round-1 review finding. Its `""` default is the sentinel `serve._derive_project_name` turns into `queue-<id>`, and Click runs a callback over an option's own default, so the refusal is gated on `ctx.get_parameter_source(param.name) is not ParameterSource.DEFAULT`. A `""` the operator never typed is returned untouched; an explicit `""` or `"   "` is refused. `decompose` and `factory` have no default, so an absent optional still arrives as `None` and is skipped by the first clause. A source the callback cannot resolve is refused, because a check that cannot prove the value came from the default must flag rather than clear.

`tests/test_project_name_boundary.py` is the guard for instance N+1, closed by construction rather than by a list of enrolled options: it walks the Click tree, counts every parameter it obtains that spells the `--project-name` flag or binds the `project_name` destination, and pins both the count at three and the three command paths, so a fourth option is an unexplained census delta and a walk that has gone blind fails on a missing path instead of agreeing with an empty result. Round 2 of review planted `@click.option("--project-name", "project")`, an idiom `cli.py` already uses eight times, and the census keyed on the destination alone cleared it; the key now takes either spelling, a test plants that shape on a throwaway group, and the layer that clears asserts the walk saw something before it clears.

Closes #338

## Behaviour changes

- `ks decompose --project-name ""` used to produce a manifest with `projectName: ""`, a register keyed `""` and a journal audit with `"project": ""`. It now exits 2 before the architect runs.
- `ks factory --spec ... --project-name "   "` used to be accepted. It now exits 2, and the message for `""` changes from `--project-name is required with --spec` to Click's.
- `ks factory --manifest m.json --project-name ""` used to be accepted with the name ignored, because the manifest carries its own. It now exits 2. Raised by review as an undisclosed change; the manifest branch never reads `project_name`, so nothing downstream changes.
- `ks queue add --project-name ""` and `--project-name "   "` now exit 2. `ks queue add` with the flag absent is unchanged: the item still stores `""` and `serve` still derives `queue-<id>` from it. The option's help text now says so.
- A queue item already on disk with a whitespace-only `project_name`, added before this change, used to run. It is now poisoned on its next attempt: `serve` passes the stored name to the child `ks factory`, the callback exits 2, and `classify_run` records that as `UNCLASSIFIABLE` with `may_retry=False`, so the item gets `needs_human` and an inbox item on attempt 1. There is no queue migration. Deriving `queue-<id>` from a name the operator typed was considered and rejected as the silent substitution the callback itself refuses; if the owner wants this closed in code, the consistent version is for `serve` to refuse the item with a reason that names the cause.

Tested, by running it:

- Full suite, bytecode deleted first and `PYTHONDONTWRITEBYTECODE=1`: **5583 passed, 37 skipped, 38 xfailed** in 285s. Branch baseline before any of this was 5564 passed, 37 skipped, 38 xfailed; 19 new cases.
- `uv run mypy kstrl/ --strict` clean, `uv run ruff check kstrl/ tests/` clean, `uv run python scripts/gen_docs.py --check` clean. `file_length_ratchet.py` and `cyclomatic_ratchet.py` exit 0, and the full `pre-commit` hook set ran on each commit.
- The three static guards that AST-walk `decompose.py` and `cli.py`, run before and after the round-2 changes: `tests/test_journal_one_writer.py`, `tests/test_event_names_have_one_home.py`, `tests/test_config_preflight.py`, 117 passed and 1 xfailed both times. `origin/main` has not moved from 568bca4, which is the merge base, so there is nothing to merge.
- Red first: the new parametrized `test_the_partition_holds_for_every_project_name` reports eleven placements for seven audits at `""` on the previous production code, and passes at the other four names. Five of the six `TestBlankProjectName` cases are red on the base commit; the sixth, `""` on `factory`, was already refused by that command's own body check, so it was green before and after. It is `"   "` and `"\t"` that command let through.
- Mutations, each planted alone with `__pycache__` deleted first, every one confirmed to apply before the run. All caught, none hung, none failed to plant. M1, restoring the two separate sums: **caught**, at `""` only, 11 placements against 7 audits. M2, swapping the two branches of `_audit_bucket`: **caught**, on the window-agreement assertion, which is the one that fixes which bucket takes the audit rather than only that the sum is right. M3, dropping each `decompose` and `factory` callback: **caught**. M6, removing the parameter-source gate: **caught**, the `queue add` default path goes red. M7, removing the `queue add` callback: **caught**, by the callback census and the three explicit-blank cases. M8, stopping the Click-tree walk from recursing into groups: **caught**, by the path census. M9, adding a fourth unguarded `project_name` option to `ks queue ls`: **caught**, by both census layers.
- The serve-side consequence of a whitespace name, measured by the round-1 reviewer against the real `classify_run` rather than reasoned from the spawn site: a child `ks factory` exiting 2 with Click's message is `UNCLASSIFIABLE` with `may_retry=False`, so the item is poisoned on attempt 1 with `needs_human` and an inbox item. Loud, but the refusal landed in the daemon rather than at the command that took the input, and the cause was only in `evidence["output_tail"]`. That path is now unreachable from `ks queue add` for items queued after this lands, which refuses the name at parse time and queues nothing; an item queued before it still takes that path, as the behaviour change above says.
- Round-2 mutations, same discipline: reverting the census key to the destination name alone: **caught**, by the planted renamed-destination test. Widening the callback to clear on every source but `COMMANDLINE`: **caught**, by the environment-source test and the unresolvable-name test. Flipping the unresolvable-name fallback to `DEFAULT`: **caught**, by the unresolvable-name test. Full suite after the round-2 changes: **5588 passed, 37 skipped, 38 xfailed** in 339s, bytecode deleted first and PYTHONDONTWRITEBYTECODE=1; the first run of this suite was killed part way by another lane and re-run from scratch.

Assumed, not measured: nothing outstanding. The one assumption in the round-1 body, the serve-side behaviour above, has been measured.

## Checklist

## Provenance

None. No `*_PROMPT` constant and no text reaching an LLM role's prompt is touched, so H2 and H3 do not apply.

## Quality pass

A four-angle cleanup review over the diff applied two changes, neither of which moves behaviour: `_excluded_projects` now binds `project` below the classifier guard instead of above it, so the value is not recomputed for skipped entries; and the two `_invoke` methods in `tests/test_cli.py`, the same eleven-flag argv with the project name varying, collapse to one module-level `_invoke_spec_command`, taking the file from 737 to 726 lines. One finding was skipped: `entry_str(entry, "project") == project_name` still appears in `kstrl/evolution.py::get_spec_issue_runs` as well as in `_audit_bucket`, and a shared predicate would make the agreement hold by construction. That is a change to a shared journal API outside this diff, and the agreement is currently pinned by an assertion in the new test, so it is left as a named follow-up rather than folded in here.

`_audit_bucket` is called twice per audit, once by `_excluded_projects` and once by `_excluded_history`. Raised in review and left: the list is a journal's spec audits, tens of entries, and `_excluded_projects` has eleven direct test callers whose signature is not worth changing for an unmeasurable cost.

## Handoffs from review, not fixed here

- A branch-unsafe name still pays for the architect: `--single-pr --project-name "my project"` runs the architect for 119 to 210 seconds and then `validate_branch_name` exits 1. Pre-existing, recorded in this lane's plan, and a different class (expensive-failure ordering); hoisting the check touches `decompose_spec` and the TUI launch path.
- The TUI strips the project name (`tui/screens/launch.py` reads `.value.strip()`) while the CLI keeps `" x "` verbatim, so one identity has two answers at two entrances. Pre-existing and not an instance of this class, since the TUI already refuses blanks. Settling it means deciding whether verbatim or stripped is the one answer.

### DIFF
--- a/kstrl/cli.py
+++ b/kstrl/cli.py
@@ -186,13 +186,22 @@ def _reject_blank_project_name(
     substituting "x" would write a manifest, a branch and a journal
     audit under something else.
 
-    Not on ``ks queue add``: its ``--project-name`` defaults to "" as
-    the sentinel ``serve`` derives ``queue-<id>`` from, and Click runs
-    a callback over an option's own default. A queued item with a
-    whitespace name still reaches this check, because ``serve`` spawns
-    a child ``ks factory --project-name``, which refuses it with exit 2.
+    On all three ``--project-name`` options, ``ks queue add``
+    included, whose "" default is the sentinel ``serve`` derives
+    ``queue-<id>`` from. Click runs a callback over an option's own
+    default, so the refusal is gated on the parameter SOURCE rather
+    than left off the option: a "" the operator never typed is
+    returned untouched, an explicit "" or "   " is refused. A source
+    this cannot resolve is refused too, because a check that cannot
+    prove the value came from the default must flag, not clear.
+
+    ``tests/test_project_name_boundary.py`` walks the Click tree and
+    fails if a fourth ``project_name`` parameter appears without this
+    callback, so the boundary is closed by census rather than by a
+    list of the options someone remembered.
     """
-    if value is not None and not value.strip():
+    source = ctx.get_parameter_source(param.name) if param.name else None
+    if value is not None and not value.strip() and source is not ParameterSource.DEFAULT:
         raise click.BadParameter("must not be empty or whitespace-only", ctx=ctx, param=param)
     return value
 
@@ -4810,7 +4819,12 @@ def _resolve_queue_item(queue: Any, item_id: str, ui_impl: UI) -> Any:
 @click.argument("spec", type=click.Path(exists=True, path_type=Path))
 @click.option("--priority", type=int, default=0, help="Higher runs first")
 @click.option("--title", default="", help="Human label (default: spec filename)")
-@click.option("--project-name", default="", help="Factory project name")
+@click.option(
+    "--project-name",
+    default="",
+    callback=_reject_blank_project_name,
+    help="Factory project name",
+)
 @click.option(
     "--auto-merge/--stop-at-pr",
     "auto_merge",
--- a/kstrl/cli.py
+++ b/kstrl/cli.py
@@ -196,9 +196,10 @@ def _reject_blank_project_name(
     prove the value came from the default must flag, not clear.
 
     ``tests/test_project_name_boundary.py`` walks the Click tree and
-    fails if a fourth ``project_name`` parameter appears without this
-    callback, so the boundary is closed by census rather than by a
-    list of the options someone remembered.
+    fails if a fourth ``--project-name`` option, or a fourth parameter
+    bound to ``project_name``, appears without this callback, so the
+    boundary is closed by census rather than by a list of the options
+    someone remembered.
     """
     source = ctx.get_parameter_source(param.name) if param.name else None
     if value is not None and not value.strip() and source is not ParameterSource.DEFAULT:
@@ -4823,7 +4824,7 @@ def _resolve_queue_item(queue: Any, item_id: str, ui_impl: UI) -> Any:
     "--project-name",
     default="",
     callback=_reject_blank_project_name,
-    help="Factory project name",
+    help="Factory project name (omit it and serve derives queue-<id>)",
 )
 @click.option(
     "--auto-merge/--stop-at-pr",
--- a/kstrl/cli.py
+++ b/kstrl/cli.py
@@ -165,6 +165,38 @@ def _use_cli_value(ctx: click.Context, name: str) -> bool:
     return ctx.get_parameter_source(name) == ParameterSource.COMMANDLINE
 
 
+def _reject_blank_project_name(
+    ctx: click.Context,
+    param: click.Parameter,
+    value: str | None,
+) -> str | None:
+    """Refuse an empty or whitespace-only project name at the boundary (#338).
+
+    The project name is an identity: it keys the journal audits, the
+    decision register and, under ``--single-pr``, the branch. "" was
+    accepted by Click and is the one value at which the convergence
+    accounting counted an audit with no project as BOTH this project's
+    and unattributed, reporting five audits as eight. Refusing it here
+    means no downstream reader has to carry a special case for it, and
+    the refusal lands before the architect is invoked, measured at 119
+    to 210 seconds against a frontier model on a real spec.
+
+    Rejects or returns the value VERBATIM; it never strips. " x " is a
+    strange name but it is the name the operator typed, and quietly
+    substituting "x" would write a manifest, a branch and a journal
+    audit under something else.
+
+    Not on ``ks queue add``: its ``--project-name`` defaults to "" as
+    the sentinel ``serve`` derives ``queue-<id>`` from, and Click runs
+    a callback over an option's own default. A queued item with a
+    whitespace name still reaches this check, because ``serve`` spawns
+    a child ``ks factory --project-name``, which refuses it with exit 2.
+    """
+    if value is not None and not value.strip():
+        raise click.BadParameter("must not be empty or whitespace-only", ctx=ctx, param=param)
+    return value
+
+
 # Accepted spellings for the agent type across the config surface.
 # kstrl.toml documents "claude" | "codex" | "custom"; the --agent-type
 # flags and KSTRL_AGENT_TYPE historically use "claude-code" | "codex" |
@@ -2006,6 +2038,7 @@ def feature(
 @click.option(
     "--project-name",
     required=True,
+    callback=_reject_blank_project_name,
     help="Name for this factory project",
 )
 @click.option(
@@ -2216,6 +2249,7 @@ def decompose(
 )
 @click.option(
     "--project-name",
+    callback=_reject_blank_project_name,
     help="Name for this factory project (required with --spec)",
 )
 @click.option(
--- a/kstrl/decompose.py
+++ b/kstrl/decompose.py
@@ -1588,7 +1588,7 @@ class ExcludedHistory:
       wording claimed the whole journal.
     - ``projects`` covers every audit under some other project name.
     - ``unattributed`` covers audits whose ``project`` field is absent,
-      null or not a string, FOR A NON-EMPTY project name; at "" those
+      null, empty or not a string, FOR A NON-EMPTY project name; at "" those
       audits are this project's own, which is what :func:`_audit_bucket`
       decides. Round 2 of review found these counted by neither of the
       other two: three audits on disk reported as one.
@@ -1644,7 +1644,7 @@ def _audit_bucket(
     it cost. ``own_recorded`` was ``entry_str(e, "project") ==
     project_name`` and ``unattributed`` was ``not entry_str(e,
     "project")``. Those are two spellings of the same question at
-    ``project_name == ""``, so an audit with an absent, null or
+    ``project_name == ""``, so an audit with an absent, null, empty or
     non-string project satisfied both and was counted twice: five
     audits on disk, eight bucket placements. One call per audit places
     it once, so the three counts sum to the audits by construction
--- a/kstrl/decompose.py
+++ b/kstrl/decompose.py
@@ -11,7 +11,7 @@ from collections import Counter
 from collections.abc import Callable, Sequence
 from dataclasses import dataclass
 from pathlib import Path
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, Literal
 
 from kstrl.agents.base import (
     ARCHITECT_COMPONENT,
@@ -1575,8 +1575,11 @@ class ExcludedHistory:
     Three buckets, because there are three ways for the report to see
     less than the journal holds, and #280 is that any of them going
     unsaid is the failure the report cannot afford. Each round of
-    review found another one being missed, so they are enumerated here
-    and every ``spec_issues`` entry falls into exactly one:
+    review found another one being missed, so they are enumerated here.
+    Every ``spec_issues`` entry falls into exactly one, and
+    :func:`_audit_bucket` is the only place that decides which: #338 is
+    what happened when two of the three were computed by separate
+    predicates that agreed at every project name except "".
 
     - ``own_recorded`` counts THIS project's audits on disk. The trend
       may count fewer, because ``lookback_runs`` windows it and because
@@ -1629,6 +1632,42 @@ class ExcludedHistory:
         return max(0, self.own_recorded - counted - self.unreadable(counted))
 
 
+def _audit_bucket(
+    entry: dict[str, Any],
+    project_name: str,
+) -> Literal["own", "other", "unattributed"]:
+    """Which of ``ExcludedHistory``'s three buckets one spec audit is in.
+
+    The ONLY place the rule lives, because #338 is what two copies of
+    it cost. ``own_recorded`` was ``entry_str(e, "project") ==
+    project_name`` and ``unattributed`` was ``not entry_str(e,
+    "project")``. Those are two spellings of the same question at
+    ``project_name == ""``, so an audit with an absent, null or
+    non-string project satisfied both and was counted twice: five
+    audits on disk, eight bucket placements. One call per audit places
+    it once, so the three counts sum to the audits by construction
+    rather than by the two predicates happening to disagree.
+
+    The ORDER is the decision, and it is own before unattributed.
+    :meth:`EvolutionJournal.get_spec_issue_runs` windows the trend by
+    the identical ``entry_str(entry, "project") == project`` expression,
+    so at "" the trend counts those audits. Calling them unattributed
+    here would make ``own_recorded`` 0 for a trend showing 3, and would
+    make the rendered note claiming neither the trend nor the
+    cross-project line counts them false about audits the trend just
+    counted. Agreeing with the window keeps one answer for one journal.
+
+    For every non-empty ``project_name`` the order is unobservable:
+    ``project == project_name`` and ``not project`` cannot both hold.
+    """
+    project = entry_str(entry, "project")
+    if project == project_name:
+        return "own"
+    if not project:
+        return "unattributed"
+    return "other"
+
+
 def _excluded_projects(
     audits: list[dict[str, Any]],
     project_name: str,
@@ -1654,7 +1693,7 @@ def _excluded_projects(
     last_seen: dict[str, str] = {}
     for entry in audits:
         project = entry_str(entry, "project")
-        if not project or project == project_name:
+        if _audit_bucket(entry, project_name) != "other":
             continue
         by_project.setdefault(project, []).append(entry_str(entry, "spec_file"))
         # Only a timestamp that exists replaces one that exists. Round 2
@@ -1700,10 +1739,11 @@ def _excluded_history(
     excludes that was itself windowed would omit history silently,
     which is the bug this exists to fix.
     """
+    buckets = Counter(_audit_bucket(e, project_name) for e in audits)
     return ExcludedHistory(
-        own_recorded=sum(1 for e in audits if entry_str(e, "project") == project_name),
+        own_recorded=buckets["own"],
         projects=_excluded_projects(audits, project_name, spec_file),
-        unattributed=sum(1 for e in audits if not entry_str(e, "project")),
+        unattributed=buckets["unattributed"],
         lookback=lookback,
     )
 
--- a/kstrl/decompose.py
+++ b/kstrl/decompose.py
@@ -1692,9 +1692,9 @@ def _excluded_projects(
     by_project: dict[str, list[str]] = {}
     last_seen: dict[str, str] = {}
     for entry in audits:
-        project = entry_str(entry, "project")
         if _audit_bucket(entry, project_name) != "other":
             continue
+        project = entry_str(entry, "project")
         by_project.setdefault(project, []).append(entry_str(entry, "spec_file"))
         # Only a timestamp that exists replaces one that exists. Round 2
         # of review: assigning unconditionally let one trailing entry
--- a/kstrl/decompose.py
+++ b/kstrl/decompose.py
@@ -1588,8 +1588,10 @@ class ExcludedHistory:
       wording claimed the whole journal.
     - ``projects`` covers every audit under some other project name.
     - ``unattributed`` covers audits whose ``project`` field is absent,
-      null or not a string. Round 2 of review found these counted by
-      neither of the other two: three audits on disk reported as one.
+      null or not a string, FOR A NON-EMPTY project name; at "" those
+      audits are this project's own, which is what :func:`_audit_bucket`
+      decides. Round 2 of review found these counted by neither of the
+      other two: three audits on disk reported as one.
 
     ``lookback`` is carried so the render can separate the two reasons
     the trend counts fewer than ``own_recorded``. An audit outside the
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -575,6 +575,46 @@ class TestResolveBaseBranch:
         assert resolve_base_branch(absent, root) == "master"
 
 
+def _halting_decompose(monkeypatch: pytest.MonkeyPatch) -> dict[str, object]:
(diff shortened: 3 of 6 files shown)


tests/test_project_name_boundary.py
```diff
@@ -0,0 +1,148 @@
...
+        for name in command.list_commands(ctx):
+            sub = command.get_command(ctx, name)
+            assert sub is not None, f"{where} lists {name!r} but cannot resolve it"
+            yield from _every_parameter(sub, here)
+
+
+def _project_name_sites() -> dict[str, click.Parameter]:
+    sites: dict[str, click.Parameter] = {}
+    for where, param in _every_parameter(cli):
+        if param.name == "project_name":
```
0xfauzi: **The census is keyed on the Python destination name, not on the `--project-name` flag, so it is blind to the option shape it exists to catch.**

`param.name == "project_name"` only holds when Click derives the variable name from the flag. `@click.option("--project-name", "project")` declares the same operator-facing boundary and binds it to `project`, and both layers then go quiet.

Measured against the real tree at this commit (added the command to `cli` at runtime, then ran `_project_name_sites()`):

```
baseline sites: ['cli decompose', 'cli factory', 'cli queue add']
after --project-name bound to 'project': ['cli decompose', 'cli factory', 'cli queue add']
  sneaky params opts: [['--project-name']]
```

The count stays at three, the path pin stays satisfied, and the clearing layer clears. Adding the same option with the default destination name IS caught (`uncovered: ['cli sneaky2']`), so the walk works; it is the key that is wrong.

This fails in the skip direction, which CLAUDE.md records as the repo's most-repeated defect (eleven logged instances, every miss in the skip direction), and the module docstring's claim that the boundary is "closed by census rather than by a list of the options someone remembered" is not true of a name-keyed walk. Key on the resource instead:

```suggestion
        if param.name == "project_name" or "--project-name" in param.opts:
```

(and let the count/path pin absorb the delta if the two ever disagree).

kstrl/cli.py
```diff
@@ -4776,7 +4819,12 @@ def _resolve_queue_item(queue: Any, item_id: str, ui_impl: UI) -> Any:
 @click.argument("spec", type=click.Path(exists=True, path_type=Path))
 @click.option("--priority", type=int, default=0, help="Higher runs first")
 @click.option("--title", default="", help="Human label (default: spec filename)")
-@click.option("--project-name", default="", help="Factory project name")
+@click.option(
+    "--project-name",
+    default="",
+    callback=_reject_blank_project_name,
```
0xfauzi, line 4826: **`serve` still hands an unvalidated on-disk `project_name` to a child `ks factory`, which this PR just taught to hard-refuse it.**

`kstrl/serve.py:2457` is `project_name = running.project_name or _derive_project_name(running)`, and `subprocess_factory_runner` puts that string straight into `--project-name` argv. `"   "` is truthy, so the `or` never fires; the child now exits 2 at parse time. Per this PR's own round-1 measurement that is `UNCLASSIFIABLE` with `may_retry=False`, so the item is poisoned on attempt 1 with `needs_human`.

The PR body says "That path is now unreachable from `ks queue add`" — true for items queued *after* this lands. An item already sitting in `.kstrl/queue/queued/` with `"project_name": "   "` in its `meta.json` is still reachable, `QueueItem.from_dict` applies no validation, and the daemon has no migration. Before this change that item ran; after it, it is poisoned. That is not in the Behaviour changes list or the CHANGELOG entry.

The deeper fix is to make the sentinel test match the refusal, at the one place serve derives:

```python
project_name = running.project_name.strip() and running.project_name or _derive_project_name(running)
```

so a blank or whitespace name on disk derives `queue-<id>` rather than spawning a child that cannot parse it.

kstrl/cli.py
```diff
@@ -2006,6 +2047,7 @@ def _target(embed_ctx: EmbeddedContext) -> int:
 @click.option(
     "--project-name",
     required=True,
+    callback=_reject_blank_project_name,
```
0xfauzi, line 2051: **The boundary refuses blanks before the architect but still lets `--single-pr` pay for the architect and then fail on a name git cannot make a ref from.**

The callback's stated purpose is that "the refusal lands before the architect is invoked, measured at 119 to 210 seconds", and it deliberately passes `" x "` through verbatim (`tests/test_cli.py::test_a_name_with_surrounding_space_reaches_decompose_verbatim` pins that).

Under `--single-pr`, `_component_branch` returns `f"kstrl/factory/{project_name}"` and `kstrl/decompose.py:2586` runs `validate_branch_name(branch)`. `_BRANCH_CHARSET_RE` is `^[A-Za-z0-9._/-]+$`, so `" x "`, `"my project"`, `"feat:auth"` and any unicode dash all fail — and that check sits *after* the architect call and the journal audit, raising `ValueError` which `ks factory` turns into exit 1. The comment on that block already says it "can only fire for a project_name (user input) that is not branch-safe".

So the expensive-failure class the callback exists to close is only half closed: the blank case is refused at parse, the branch-unsafe case still costs a full architect call. `validate_branch_name` is a pure function in `kstrl/names.py` and `decompose_spec` has both `project_name` and `single_pr` at entry, so the general fix is to hoist that single-PR branch pre-validation to the top of `decompose_spec` rather than adding a blank-only special case at the CLI. That covers the TUI entrance too.

kstrl/decompose.py
```diff
@@ -1629,6 +1634,42 @@ def windowed_out(self, counted: int) -> int:
         return max(0, self.own_recorded - counted - self.unreadable(counted))
 
 
+def _audit_bucket(
+    entry: dict[str, Any],
+    project_name: str,
+) -> Literal["own", "other", "unattributed"]:
+    """Which of ``ExcludedHistory``'s three buckets one spec audit is in.
+
+    The ONLY place the rule lives, because #338 is what two copies of
```
0xfauzi, line 1643: **"The ONLY place the rule lives" is not accurate, and the ORDER decision below depends on the copy that is still there.**

`kstrl/evolution.py:1640` still spells the own-bucket predicate independently:

```python
and entry_str(entry, "project") == project
```

The docstring three paragraphs down argues that own-before-unattributed is correct *because* `get_spec_issue_runs` "windows the trend by the identical `entry_str(entry, "project") == project` expression". So the second copy is not incidental — it is the definition this function's ordering is derived from, and the two are only kept in step by one assertion in one parametrized test over five names on one fixture.

Add a `.strip()`, a `casefold()`, or a `!= ""` short-circuit to `get_spec_issue_runs` (any of which reads as a reasonable local improvement) and `own_recorded` silently stops matching the trend printed directly above it — which is #338's failure mode restored, with only `test_the_partition_holds_for_every_project_name` standing between. That is two definitions of one rule, and CLAUDE.md's own note on #260 is that "the weaker one is the one the gate consults".

Since `_audit_bucket` is already exported-adjacent in this module, having the window call it (`_audit_bucket(entry, project) == "own"`) — or moving the classifier into `kstrl/evolution.py` beside `entry_str` and importing it here — makes the agreement hold by construction instead of by a test fixture.

kstrl/cli.py
```diff
@@ -4776,7 +4819,12 @@ def _resolve_queue_item(queue: Any, item_id: str, ui_impl: UI) -> Any:
 @click.argument("spec", type=click.Path(exists=True, path_type=Path))
 @click.option("--priority", type=int, default=0, help="Higher runs first")
 @click.option("--title", default="", help="Human label (default: spec filename)")
-@click.option("--project-name", default="", help="Factory project name")
+@click.option(
+    "--project-name",
+    default="",
+    callback=_reject_blank_project_name,
+    help="Factory project name",
```
0xfauzi: **The `""` sentinel and the `""` refusal now contradict each other, and the help text documents neither.**

The callback's own docstring calls this default "the sentinel `serve` derives `queue-<id>` from", and this change then refuses the operator from spelling that sentinel. The refusal keys on *how the value arrived*, not on what it means, so:

```sh
ks queue add spec.md --project-name "$PROJECT"   # PROJECT unset
```

now exits 2 and queues nothing, where before it queued an item that `serve` named `queue-<id>`. An unset shell variable expanding to `""` is the single most common way an empty string reaches a command line, and it is indistinguishable to the operator from the supported "omit the flag" spelling.

Two things would make this defensible rather than surprising, and neither is here: the help string still reads `"Factory project name"` with no mention of the derivation *or* the refusal, and the CHANGELOG entry frames it purely as a fix. If the intent is that only a *typed* blank is wrong, the error message should say so — `must not be empty or whitespace-only; omit --project-name to have serve name this item queue-<id>` — so the operator learns the escape hatch from the failure.

kstrl/cli.py
```diff
@@ -165,6 +165,47 @@ def _use_cli_value(ctx: click.Context, name: str) -> bool:
...
+    The project name is an identity: it keys the journal audits, the
+    decision register and, under ``--single-pr``, the branch. "" was
+    accepted by Click and is the one value at which the convergence
+    accounting counted an audit with no project as BOTH this project's
+    and unattributed, reporting five audits as eight. Refusing it here
+    means no downstream reader has to carry a special case for it, and
+    the refusal lands before the architect is invoked, measured at 119
+    to 210 seconds against a frontier model on a real spec.
+
+    Rejects or returns the value VERBATIM; it never strips. " x " is a
```
0xfauzi, line 184: **The TUI decompose boundary does exactly the silent substitution this paragraph forbids, and the census cannot see it.**

`kstrl/tui/screens/launch.py:169`:

```python
project = self.query_one("#decompose-project", Input).value.strip()
```

That `project` goes into `DecomposeLaunch(project_name=project)` and reaches the same `decompose_spec` as the CLI value. So typing `" x "` into the launch form writes a manifest, a branch and a journal audit under `"x"` — "under something else", in this docstring's words — while typing it on the CLI writes them under `" x "`. One identity, two boundaries, two answers, and nothing anywhere records that the two disagree.

The census in `tests/test_project_name_boundary.py` walks the Click tree only, so it is structurally incapable of covering this fourth entrance; the module docstring scopes the class as "a blank project name accepted at a **CLI** boundary", but the defect class as the code actually experiences it is "a project name normalised differently at different boundaries".

CLAUDE.md: "One defect class, one PR. Fix every instance plus the guard that catches instance N+1, rather than one site per PR with sibling issues filed behind it." Dropping the `.strip()` on the TUI value (keeping the existing `if not project` check against the stripped form) makes the two entrances agree, and the guard for N+1 would be a census over the sites that construct `DecomposeLaunch` rather than a Click-only walk.

tests/test_project_name_boundary.py
```diff
@@ -0,0 +1,148 @@
...
+    def test_the_tree_has_exactly_the_three_known_project_name_options(self) -> None:
+        assert set(_project_name_sites()) == EXPECTED_PROJECT_NAME_SITES
+
+    def test_every_one_of_them_carries_the_rejecting_callback(self) -> None:
+        uncovered = {
+            where
+            for where, param in _project_name_sites().items()
+            if param.callback is not _reject_blank_project_name
+        }
+        assert uncovered == set()
```
0xfauzi, line 119: **`assert uncovered == set()` is green both when every site carries the callback and when the walk returned nothing.**

CLAUDE.md names this shape directly: "`assert hits(...) == []` is not a control. It passes when the walk is correctly narrow AND when the walk has been switched off, which is the same shape as the defect."

The module docstring answers this — "it pins both the count and the command paths, so a walk that has gone blind fails red instead of agreeing with an empty result" — but that pin lives in the *sibling* test. This test, which is the clearing layer, is load-bearing on a neighbour staying present and correct. Delete or `xfail` `test_the_tree_has_exactly_the_three_known_project_name_options` and this one keeps passing over an empty walk.

One line makes the clearing layer self-sufficient:

```suggestion
        sites = _project_name_sites()
        assert sites, "the Click-tree walk found no project_name parameter at all"
        uncovered = {
            where for where, param in sites.items() if param.callback is not _reject_blank_project_name
        }
        assert uncovered == set()
```

kstrl/decompose.py
```diff
@@ -1700,10 +1741,11 @@ def _excluded_history(
     excludes that was itself windowed would omit history silently,
     which is the bug this exists to fix.
     """
+    buckets = Counter(_audit_bucket(e, project_name) for e in audits)
```
0xfauzi, line 1744: **The partition is still assembled by two independent traversals; the sum is an identity only because both happen to call the same classifier.**

This line buckets every audit, and then `_excluded_projects(audits, ...)` on the next line walks the same list a second time and re-derives the "other" bucket for itself (and re-calls `entry_str(entry, "project")` on top of `_audit_bucket`'s call). `other_audits` is `sum(p.audits for p in self.projects)` — a count produced by a different pass than the two counts beside it.

That is structurally the same shape as the defect being fixed: three numbers claimed to sum to the input, computed by separate walks that agree today. `_audit_bucket` being shared is what makes them agree, which is a real improvement, but it leaves the *identity* resting on "nobody changes how `_excluded_projects` selects" rather than on construction. One pass closes it:

```python
by_bucket: dict[str, list[dict[str, Any]]] = {"own": [], "other": [], "unattributed": []}
for entry in audits:
    by_bucket[_audit_bucket(entry, project_name)].append(entry)
return ExcludedHistory(
    own_recorded=len(by_bucket["own"]),
    projects=_group_other_projects(by_bucket["other"], spec_file),
    unattributed=len(by_bucket["unattributed"]),
    lookback=lookback,
)
```

This is not the cost point the PR body already answers ("tens of entries... an unmeasurable cost") — it is that `assert own + other + unattributed == len(audits)` would then be true by construction rather than by a test fixture. The eleven direct callers of `_excluded_projects` are a real obstacle; splitting the grouping out from the selection is what removes it.

tests/test_cli.py
```diff
@@ -575,85 +575,157 @@ def test_absent_value_detects(self, tmp_path: Path, absent: str | None) -> None:
...
     ) -> None:
-        seen = self._capture(monkeypatch)
+        seen = _halting_decompose(monkeypatch)
         root = _repo_on(tmp_path, "master")
-        assert self._invoke("factory", root, "--base-branch", "trunk").exit_code == 2
+        assert _invoke_spec_command("factory", root, "--base-branch", "trunk").exit_code == 2
         assert seen["base_branch"] == "trunk"
+
+
+class TestBlankProjectName:
```
0xfauzi, line 683: **The one behaviour change this PR calls out as previously undisclosed is the one with no test.**

From the PR body: "`ks factory --manifest m.json --project-name ""` used to be accepted with the name ignored, because the manifest carries its own. It now exits 2. Raised by review as an undisclosed change."

Every case in this class goes through `_invoke_spec_command`, which always passes `--spec`. Nothing exercises the `--manifest` branch with a blank name. A later author who decides the manifest path should keep accepting the flag (a defensible reading — `project_name` is genuinely unread there) moves the check from the option into the `--spec` branch's body and restores the old behaviour with the whole suite green.

One case pins it:

```python
def test_a_blank_name_is_refused_even_with_a_manifest(self, tmp_path: Path) -> None:
    """The manifest carries its own project name, so the flag is unread
    on that branch - but it is still refused, because the boundary is
    the option and not the branch that happens to consume it."""
    root = _repo_on(tmp_path, "master")
    manifest = root / "m.json"
    manifest.write_text("{}\n")
    result = CliRunner().invoke(
        cli, ["factory", "--manifest", str(manifest), "--project-name", "",
              "--root", str(root), "--ui", "plain", "--no-color"]
    )
    assert result.exit_code == 2
    assert "--project-name" in result.output
```

kstrl/decompose.py
```diff
@@ -1585,8 +1588,10 @@ class ExcludedHistory:
       wording claimed the whole journal.
     - ``projects`` covers every audit under some other project name.
     - ``unattributed`` covers audits whose ``project`` field is absent,
-      null or not a string. Round 2 of review found these counted by
-      neither of the other two: three audits on disk reported as one.
+      null or not a string, FOR A NON-EMPTY project name; at "" those
```
0xfauzi: **This amended bullet enumerates three shapes; `_audit_bucket` puts a fourth in the same bucket.**

At a non-empty `project_name`, an entry carrying `"project": ""` — present, a string — is classified `unattributed` by the code below (`project == project_name` is false, `not project` is true). The bullet says the bucket covers fields that are "absent, null or not a string", which excludes it.

The new test fixture contains exactly that row (`audit("", "empty.md")`), so the code is doing the right thing and only the prose is short. Since the whole point of the edit on this line is that the bucket's membership rule changed, it is worth spelling the rule the same way `_audit_bucket` does:

```suggestion
    - ``unattributed`` covers audits whose ``project`` field is absent,
      null, empty or not a string - anything :func:`entry_str` reads as
      "" - FOR A NON-EMPTY project name; at "" those
```

Same observation applies to `_audit_bucket`'s own docstring ("an audit with an absent, null or non-string project").

Issue #318, opened 2026-08-31
config._load_toml: a non-UTF-8 kstrl.toml crashes instead of raising ConfigError


0xfauzi:
`config._load_toml` catches only `tomllib.TOMLDecodeError`, so a config file with one non-UTF-8 byte crashes the CLI instead of raising `ConfigError`.

## The mechanism

```python
try:
    with open(path, "rb") as f:
        data: dict[str, Any] = tomllib.load(f)
except tomllib.TOMLDecodeError as exc:
    raise ConfigError(f"Invalid TOML in {path}: {exc}") from exc
```

`tomllib.load` decodes the stream as UTF-8 itself, so a `kstrl.toml` carrying a latin-1 name or a stray `0x80` raises `UnicodeDecodeError`. That is a `ValueError`, not a `TOMLDecodeError`, so it walks past this handler and out of the config loader as an unhandled traceback rather than the `ConfigError` every caller is written against, including the CLI's config preflight seam.

This is the rule CLAUDE.md already states from #291: a reader of any file must catch `ValueError` alongside `OSError`, because `UnicodeDecodeError` is a `ValueError` and escapes a fail-closed `except OSError`.

## Where it was found

The identical hole at `verify.py:806` (`_default_typecheck_command`) was found by the round-2 `/code-review` of #300 and fixed there, because it reached an ADVISORY verification report and took `ks feature` down before the agent had run. `config.py:520` is the same class at a different site.

`feedforward.py:540` and `:586` already use a bare `except Exception: return`, so they are unaffected.

## Why it is not in #300

#300 is a `ks feature` change at its last review round. Widening a config loader's failure contract is a different blast radius: `ConfigError` is caught in several places and turning a crash into one changes what those paths do. Filed rather than folded in.

## Suggested fix

`except (ValueError, OSError) as exc:` with the message naming both causes, plus a test that writes a `kstrl.toml` with one invalid byte and asserts `ConfigError`.

Issue #289, opened 2026-08-30
The TUI evolve screen loads config unguarded, so a malformed kstrl.toml crashes the home shell where the CLI now reports it cleanly


0xfauzi:
Found while implementing #272 (PR #287) and deliberately left out of it, because it sits on a path that bypasses the seam that PR adds rather than being fixed by it.

## The gap

`kstrl/tui/screens/evolve.py:178` calls `EvolutionConfig.load` with no guard. That load raises `ValueError` on malformed TOML, on a bad `lookback_runs`, or on a bad `KSTRL_EVOLUTION_LOOKBACK_RUNS`.

The screen is reachable from the home shell, which never runs the command preflight PR #287 installs, because the shell is not a click command invocation. So a malformed `kstrl.toml` takes down the TUI rather than producing the clean, named error the CLI now gives.

## Why it is not covered by #272

PR #287 resolves configuration once at command entry, on `_KstrlCommand.invoke`. That covers every `ks <command>` path. It does not cover screens entered from the home shell after startup, which construct their own config objects on demand.

So the two surfaces now behave differently for the same broken file: the CLI names the section, the coercion failure and the environment variable that set it, and exits cleanly, while the TUI raises.

## Related, already closed the same way elsewhere

Four `EvolutionConfig.load` sites were guarded during this batch, using `EvolutionJournal.open` or `load_or_none`, on the reasoning that the journal is an optional audit trail and degrading to "no journal" is honest. This is a fifth site of the same class, and the same reasoning applies: an evolve screen that cannot read the journal config should say so and show nothing, not crash the shell.

One caveat worth stating: for a screen whose entire purpose is the evolution journal, degrading silently to an empty view is arguably worse than an error. PR #287 already faced this and answered it with `_PREFLIGHT_REQUIRED`, which promotes a section to fatal for a command that is *about* it, so `ks evolve` stops rather than warning and then failing. The screen should do the equivalent: show a clear "configuration unreadable" state naming the problem, not an empty list and not a traceback.

## Worth checking at the same time

Whether other home-shell screens construct config objects on demand and have the same exposure. The fix is probably one shared pattern rather than a guard per screen.

## Status

Pre-existing, unchanged by PR #287, and independent of it.

PR #348, opened 2026-09-06
docs(config): state whose fault a RecursionError is, and pin both directions (#323)

### LINKED ISSUE #323: RecursionError has no provenance: config and config_preflight disagree about whose fault it is
Surfaced by the round-3 /simplify pass on PR #319 (issue #318). Not a regression; the ambiguity is newly visible because #319 widened the parse guard to `except Exception`, which makes `RecursionError` reachable as a reported cause for the first time.

## The problem

Two call sites classify the same exception in opposite directions and neither can tell them apart:

- `config_preflight.raise_if_defect` treats a `RecursionError` as kstrl's own defect, and keeps its traceback.
- `config.load_toml_document` treats it as the operator's bad file, and reports a `ConfigError` naming the path.

Both are right some of the time. A 600-deep nested array in `kstrl.toml` really is the operator's file. A runaway recursion inside kstrl really is kstrl's defect. The exception object carries nothing that separates them.

## Why it matters

The distinction is the one `raise_if_defect` exists to draw: report to the operator what the operator can act on, and keep a traceback for what only a developer can. With no provenance, one of the two sites is always wrong.

## Suggested shape

A predicate below both modules that answers "did this recursion happen inside the parse, or inside kstrl". Inspecting the traceback for frames belonging to the parser is one candidate; entering the parse with a recursion limit scoped to the call is another and is probably more honest, since it converts the question into a bound rather than a guess. Measure before choosing.

0xfauzi:
## Summary

Closes #323

#323 asked for a predicate separating "the operator nested a TOML array 600 deep" from "kstrl recursed forever", on the grounds that the exception object carries nothing that tells them apart. Measured on 3.12.8 and on 3.13.2, the object really does carry nothing: both directions give `builtins.RecursionError`, `str(exc)` "maximum recursion depth exceeded", the same single-string `args`, and no attribute a freshly constructed instance lacks.

A predicate is still not needed, because the two sites are in series rather than in competition. Every tomllib parse in `kstrl/` ends on a bare `except Exception`, so no document's `RecursionError` escapes the parse that raised it: `config.load_toml_document` converts it to a `ConfigError` naming the path, and the pyproject.toml and ruff.toml readers in `verify` and `feedforward` swallow it and fall back to a default or to no conventions at all. The closure is over the PARSES, not over the call graph, and that distinction is load-bearing: `init_wizard._detected_text` is a guarded block that reaches tomllib WITHOUT going through `load_toml_document`, via `resolve_verify_commands` to `verify._default_typecheck_command` on the project's pyproject.toml. The conclusion survives because that parse ends on `except Exception: return DEFAULT_TYPECHECK_COMMAND`, so a `RecursionError` there is swallowed and never reaches `raise_if_defect`. The two-layer census in `tests/test_toml_readers.py` is what keeps the parse set closed; layer 1, `EXPECTED_TOMLLIB_SPELLINGS`, resolves nothing and so cannot miss a reader that spells `tomllib` at all.

**No production behaviour changes.** What changes is that the layering is stated where each half of it lives, with its residual, and pinned by tests.

- `load_toml_document`'s docstring stops naming one depth. Measured at the default limit of 1000 on CPython 3.12.8, with one harness and one caller convention (a chain of exactly N frames above the module frame): nested arrays first raise at 497 levels for N=1, 472 for N=51 and 397 for N=201; inline tables at 331 / 314 / 264, being about three interpreter frames per level against an array's two. The threshold is the caller's remaining stack, not a property of the file, so no constant belongs there.
- The same docstring states the RESIDUAL rather than asserting it away. The catch-all converts any `Exception` out of the parse, so kstrl's own runaway that happens to bottom out inside a parse is reported to the operator as their broken file. Measured with a valid two-line kstrl.toml and a caller recursed to a known depth, stated as headroom because a loop parameter is not a frame count (round 2 found the first version three frames out): with 7 or more of the 1000 frames left the parse succeeds, with 5 or 6 left a valid file is reported unparseable with a `RecursionError` cause, and with 4 or fewer the error is raised before the guard and escapes it. A caller that deep is itself the runaway, so the residual is one runaway reported as a file rather than as a traceback, in a two-frame band. It is stated, not mechanised: a frame-counting predicate needs a threshold nobody has measured, and the measured headroom on the paths that exist is large. The parse runs 9 to 16 frames deep under `ks status`, 14 to 15 under `ks config show`, and 13 on the Textual event loop's own callbacks; one TUI parse runs on its caller's stack and inherits it, 22 from a bare script and 52 under pytest, so the 52 is a property of the test harness, not of the event loop.
- `raise_if_defect`'s docstring says why "ours" holds for `RecursionError`, and why inspecting the exception could not have decided it.
- `tests/test_recursion_provenance.py` pins both directions at `preflight_config`, and direction (b) again at `config_problem_lines` and at the `ks status` seam. Direction (b) uses a real self-calling loader, so a stack is actually exhausted and the frames are the runaway's; direction (a) uses real bytes. Both assert that `raise_if_defect` is in the traceback, so the tests can tell a deliberate re-raise from nothing having caught it at all.
- `tests/helpers/bad_toml.py`'s CLI depth comment is corrected. It said 495 through the CLI and blamed one frame. Binary search through the real CLI (`CliRunner` on `ks status`, first depth whose output carries `UNPARSEABLE_TOML_MESSAGE`): 489 parses, 490 is reported. The gap from 497 is 7 array levels, which at about two interpreter frames per level is the 9-to-16-frame depth of the CLI parse.

Rejected, with the measurement: traceback inspection discriminates cleanly (998 of 1000 traceback frames inside tomllib for a deep file, 0 of 1000 for a runaway) but is unnecessary for anything that reaches `raise_if_defect`; and a scoped `sys.setrecursionlimit` is process-wide, so it can raise a spurious `RecursionError` on another thread while moving the threshold rather than settling provenance.

**Tested.**

- Full suite: 5564 passed / 37 skipped / 38 xfailed on the merge base 568bca4, 5567 / 37 / 38 at the first review head, 5568 / 37 / 38 here. The delta is +4 and it is the four cases in the new file. `mypy kstrl/ --strict` clean over 129 source files, `ruff check kstrl/ tests/` clean, `gen_docs.py --check` clean, `ruff format --check` and `codespell@2.4.3` clean on all four changed files. The full pre-commit hook set ran on both commits of the fix round and every hook passed or skipped for want of matching files.
- `scripts/precommit/file_length_ratchet.py` passes with `kstrl/config.py` at 800 lines against its 800 ceiling. That is zero headroom: the next paragraph in that file is a split, not an edit. It is also what held the provenance paragraph to five lines, which is why the fuller statement of the residual lives in the test file's module docstring.
- Four mutations, each asserting an anchor count of exactly 1 before writing so a silent no-op plant reports COULD-NOT-PLANT rather than passing, each with `__pycache__` purged and `PYTHONDONTWRITEBYTECODE=1`, each bounded at 180s in its own process group with `start_new_session=True` and `os.killpg` on timeout so a hang is reported as a hang. None hung.
  - Narrowing `load_toml_document`'s broad clause to `ValueError`: CAUGHT, 1 of 4 red, with a raw `RecursionError` out of the parser.
  - Deleting `raise_if_defect` from `collect_config_problems`: CAUGHT, 3 of 4 red. Its first anchor reported COULD-NOT-PLANT (the comment block between the `except` and the call), which is recorded as its own outcome and retargeted rather than folded into a pass.
  - Narrowing `REJECTIONS` to `(ValueError, TypeError)`, which makes `raise_if_defect` unreachable: CAUGHT, 3 of 4 red. This was STILL GREEN in this file before this round's fix, measured at the first review head (the repo as a whole catches it in `tests/test_tui_config_guard.py` and `tests/test_serve_cli.py`), and it is the whole reason the `raise_if_defect` frame assertions exist.
  - Deleting `raise_if_defect` from `config_problem_lines` only: CAUGHT, 1 of 4 red. Also STILL GREEN in this file before this round; at repo level `tests/test_config_guard_survey.py` already catches it, so this kill is a duplicate there and new only for a real stack exhaustion. That surface has a `raise_if_defect` of its own which the traversal's call does not stand in for.
- The seam test covers only direction (b), because the direction (a) half restated the `[status-deep_nest]` cell of `TOML_PARSE_FAULTS` crossed with `SEAM_COMMANDS`. Verified rather than assumed: under the `ValueError` mutation, 13 of that matrix's 52 cells go red and `[status-deep_nest]` is one of them, out of 16 failures in the file.
- The in-process direction (b) test asserts `warn` is never called. `[evolution]` is the one degrading section, and a warning line there would hide a cycle rather than surface it. This is the only assertion in the suite about a DEFECT not being degraded into a warning; the `warnings == []` in `test_config_preflight.py` is about a clean config, which is a different claim.
- The Textual event-loop stack depth was listed as assumed in the first round and is now measured: four parses at 13, 13, 13 and 52, mounting the evolve screen under `app.run_test()` with a `tomllib.loads` spy.

**Assumed.**

- The `init_wizard._detected_text` route to `verify._default_typecheck_command` is read from the call chain and from `SURFACE_REJECTIONS`'s own docstring, not re-measured with a spy in this round. Nothing here depends on how many parses that route does, only on the fact that it does not go through `load_toml_document`.
- The 3.13.2 half of the "the object carries nothing" measurement was taken on a second interpreter under `uvx --python 3.13`, not in CI; CI runs 3.12.

After the round-2 text corrections and a merge of origin/main at 9890a33 (PR #347), the full suite on the merged tree is 5592 passed, 37 skipped, 38 xfailed in 323s, bytecode deleted first and `PYTHONDONTWRITEBYTECODE=1`; mypy strict, ruff, gen_docs and the four precommit ratchets are clean, and `kstrl/config.py` is at 800 of 800 lines.

## Measured handoff, not fixed here

`kstrl/feedforward.py:540` and `:586` are both `tomllib.loads(path.read_text(encoding="utf-8"))` INSIDE the `try`, so a read `OSError` is swallowed into the same silent `return` as a parse fault. That is rule (3) of the four states CLAUDE.md records from #318, "ALL the I/O outside the guarded block", and `verify.py:934` and `config.py:635` hoist the read correctly. `tests/test_toml_readers.py`'s failure message states the rule, but `handler_verdict` judges four things - unreadable, too wide, too narrow, out of order - and I/O position is not one of them, so neither offender is visible to it. It does not touch this PR's claim, since an `OSError` is not a `RecursionError`, and it is a different defect class, so per CLAUDE.md's "one defect class, one PR" it belongs here with the numbers attached rather than as a new issue or as scope creep in this branch. Two sites, one guard function to widen.

Seven of the ten `raise_if_defect` call sites are unpinned. Measured by round 2 of review: deleting `raise_if_defect(exc)` from `kstrl/config_report.py:435`, `kstrl/cli.py:3082`, `kstrl/tui/home.py:38`, `kstrl/tui/session.py:208` and `:260`, `kstrl/tui/screens/config.py:248` and `kstrl/tui/screens/init_wizard.py:89` at once leaves the full suite at 5568 passed, 37 skipped, 38 xfailed, byte-identical to the unmutated baseline. The three sites inside `kstrl/config_preflight.py` are the pinned ones. `tests/test_tui_config_walk.py::guarding_handler` clears a `try` whose except clause names `SURFACE_REJECTIONS` without reading the handler body, so `except SURFACE_REJECTIONS: cfg = None` is cleared too. That is a different defect class from #323 (call-site coverage of a re-raise rule, not exception provenance) and is left here as a handoff: the natural home for the census is `tests/test_tui_config_walk.py`, which already walks every TUI config load and has the bindings table to resolve the handler body.

## Checklist

No roadmap item: #323 came out of the round-3 simplify pass on #319, and no tracker doc names it. No CHANGELOG line either, for the same reason the #318 rounds added none: nothing an operator can observe changed.

## Provenance

None. No adversarial prompt body changed, so neither box applies.

### DIFF
--- a/kstrl/config.py
+++ b/kstrl/config.py
@@ -567,17 +567,17 @@ def load_toml_document(path: Path) -> dict[str, Any]:
     render the message.
 
     Provenance follows (#323): the parse runs 9 to 16 frames deep under
-    ``ks status``, 13 to 52 on the Textual event loop, so a
-    ``RecursionError`` here is the document's. Not by construction:
-    measured, a CALLER at 991 of 1000 frames gets a valid file called
-    unparseable, and IS the runaway. ``raise_if_defect`` has the rest.
+    ``ks status``, 13 on the Textual event loop's own callbacks (one TUI
+    parse inherits its caller's stack), so a ``RecursionError`` here is
+    the document's. Not by construction: measured, a CALLER with 5 or 6
+    of 1000 frames left gets a valid file called unparseable, and IS the
+    runaway. ``raise_if_defect`` has the rest.
 
     ``OSError`` is not in the catch-all's reach at all, which is the
-    stronger form of the guarantee two callers depend on. An earlier
-    draft re-raised it from inside the guard; that clause was correct
-    and unpinnable, because with the I/O already hoisted out there was
-    no way to make a test reach it. A special case no test can enter is
-    a special case that has not been deleted yet.
+    stronger form of the guarantee two callers depend on. Re-raising it
+    from inside the guard was correct and unpinnable: with the I/O
+    hoisted out, no test could reach that clause, and a special case no
+    test can enter is a special case that has not been deleted yet.
 
     The rule, after being wrong three times: a parser's error taxonomy
     belongs to the parser, and a reader naming any class narrower than
--- a/kstrl/config.py
+++ b/kstrl/config.py
@@ -567,9 +567,8 @@ def load_toml_document(path: Path) -> dict[str, Any]:
     render the message.
 
     Provenance follows (#323): the parse runs at stack depth 9 to 16
-    under ``ks status``, so a ``RecursionError`` here is the document's,
-    and every block ``config_preflight`` guards reaches tomllib only
-    through this function, so one above is ours - see ``raise_if_defect``.
+    under ``ks status``, so a ``RecursionError`` here is the document's.
+    One above this call is ours; ``raise_if_defect`` carries that half.
 
     ``OSError`` is not in the catch-all's reach at all, which is the
     stronger form of the guarantee two callers depend on. An earlier
--- a/kstrl/config.py
+++ b/kstrl/config.py
@@ -545,8 +545,10 @@ def load_toml_document(path: Path) -> dict[str, Any]:
       then refuses to build: ``max_iterations = <4301 digits>`` raises
       "Exceeds the limit (4300 digits) for integer string conversion"
       from ``sys.get_int_max_str_digits``. Walked past round 1.
-    - ``RecursionError``, at roughly 496 nested arrays or inline tables,
-      from tomllib's recursive-descent parser. It derives from
+    - ``RecursionError``, from tomllib's recursive-descent parser, at no
+      one depth: at the default limit nested arrays first fail at 497
+      levels from a one-frame caller and 396 under 200 more, inline
+      tables at 331 and 264; the caller's own stack sets it. It derives from
       ``RuntimeError``, NOT ``ValueError``, so it walked past round 2 -
       whose docstring, whose AST guard and whose CLAUDE.md line all
       asserted that ``ValueError`` WAS the whole class. Round 2 stated
@@ -563,6 +565,11 @@ def load_toml_document(path: Path) -> dict[str, Any]:
     covered, since it derives from ``Exception``, with the honest caveat
     that no handler can promise the interpreter has the headroom left to
     render the message.
+
+    Provenance follows (#323): the parse runs at stack depth 9 to 16
+    under ``ks status``, so a ``RecursionError`` here is the document's,
+    and every block ``config_preflight`` guards reaches tomllib only
+    through this function, so one above is ours - see ``raise_if_defect``.
 
     ``OSError`` is not in the catch-all's reach at all, which is the
     stronger form of the guarantee two callers depend on. An earlier
--- a/kstrl/config.py
+++ b/kstrl/config.py
@@ -547,7 +547,7 @@ def load_toml_document(path: Path) -> dict[str, Any]:
       from ``sys.get_int_max_str_digits``. Walked past round 1.
     - ``RecursionError``, from tomllib's recursive-descent parser, at no
       one depth: at the default limit nested arrays first fail at 497
-      levels from a one-frame caller and 396 under 200 more, inline
+      levels from a one-frame caller and 397 under 200 more, inline
       tables at 331 and 264; the caller's own stack sets it. It derives from
       ``RuntimeError``, NOT ``ValueError``, so it walked past round 2 -
       whose docstring, whose AST guard and whose CLAUDE.md line all
@@ -566,9 +566,11 @@ def load_toml_document(path: Path) -> dict[str, Any]:
     that no handler can promise the interpreter has the headroom left to
     render the message.
 
-    Provenance follows (#323): the parse runs at stack depth 9 to 16
-    under ``ks status``, so a ``RecursionError`` here is the document's.
-    One above this call is ours; ``raise_if_defect`` carries that half.
+    Provenance follows (#323): the parse runs 9 to 16 frames deep under
+    ``ks status``, 13 to 52 on the Textual event loop, so a
+    ``RecursionError`` here is the document's. Not by construction:
+    measured, a CALLER at 991 of 1000 frames gets a valid file called
+    unparseable, and IS the runaway. ``raise_if_defect`` has the rest.
 
     ``OSError`` is not in the catch-all's reach at all, which is the
     stronger form of the guarantee two callers depend on. An earlier
--- a/kstrl/config_preflight.py
+++ b/kstrl/config_preflight.py
@@ -132,26 +132,34 @@ def raise_if_defect(exc: BaseException) -> None:
 
     ``RecursionError`` is the one that needs an argument rather than a
     rule, and the argument is layering rather than inspection (#323). A
-    kstrl.toml with 600 nested arrays exhausts the stack inside
-    tomllib's recursive descent, and that is the operator's file, not a
-    defect of ours. It never arrives here. Every tomllib parse in
-    ``kstrl/`` ends on a bare ``except Exception``, and
-    ``EXPECTED_TOML_PARSES`` in ``tests/test_toml_readers.py`` is what
-    makes that a closed set rather than a count restated here:
-    ``config.load_toml_document`` re-raises ``ConfigError`` naming the
-    path, and the pyproject.toml and ruff.toml readers in ``verify`` and
-    ``feedforward`` fall back to a default. So a ``RecursionError`` that
+    kstrl.toml with 600 nested arrays exhausts the stack at the default
+    recursion limit inside tomllib's recursive descent, and that is the
+    operator's file, not a defect of ours. It never arrives here. Every
+    tomllib parse in ``kstrl/`` ends on a bare ``except Exception``, and
+    the two-layer census in ``tests/test_toml_readers.py`` is what makes
+    that a closed set rather than a count restated here - layer 1,
+    ``EXPECTED_TOMLLIB_SPELLINGS``, resolves nothing and so cannot miss
+    a reader that spells ``tomllib`` at all. ``load_toml_document``
+    re-raises ``ConfigError`` naming the path, and the pyproject.toml
+    and ruff.toml readers in ``verify`` and ``feedforward`` fall back to
+    a default or to no conventions at all. So a ``RecursionError`` that
     does reach this function is a cycle in kstrl's own code, and the
-    traceback this re-raise keeps is what locates it.
+    traceback this re-raise keeps is what locates it. The closure is
+    over the PARSES, not over the call graph: not every guarded block
+    goes through ``load_toml_document``, and ``init_wizard._detected_text``
+    is the one that does not, reaching
+    ``verify._default_typecheck_command`` on the project's
+    pyproject.toml instead.
 
     Inspecting the exception could not have settled it anyway. Measured
     on 3.12.8 and 3.13.2, both directions give
     ``builtins.RecursionError`` with ``str(exc)`` "maximum recursion
     depth exceeded" and no attribute of its own; only the frames differ,
-    and by the time they could be read the boundary has already
-    answered. ``tests/test_recursion_provenance.py`` pins both
-    directions, at :func:`preflight_config` and at the ``ks status``
-    seam.
+    and by the time they could be read ``load_toml_document`` has
+    already turned the document's into a ``ConfigError``.
+    ``tests/test_recursion_provenance.py`` pins both directions at
+    :func:`preflight_config`, and direction (b) again at
+    :func:`config_problem_lines` and at the ``ks status`` seam.
     """
     if isinstance(exc, RuntimeError) and type(exc).__module__.split(".")[0] != "kstrl":
         raise exc
@@ -554,6 +562,12 @@ def _blamed_toml_value(
     whose value reprs into that message is the one to look at. Reported
     only when exactly one key matches, and phrased as what the file
     says rather than as a diagnosis.
+
+    The ``suppress`` below would swallow a ``RuntimeError`` kstrl did
+    not define, which :func:`raise_if_defect` exists to re-raise. It
+    cannot reach one: this runs only AFTER that function has passed on
+    the exception being reported, and everything ``load_toml_section``
+    raises here has already been converted by ``load_toml_document``.
     """
     hits = []
     for name in sections:
--- a/kstrl/config_preflight.py
+++ b/kstrl/config_preflight.py
@@ -134,9 +134,10 @@ def raise_if_defect(exc: BaseException) -> None:
     rule, and the argument is layering rather than inspection (#323). A
     kstrl.toml with 600 nested arrays exhausts the stack inside
     tomllib's recursive descent, and that is the operator's file, not a
-    defect of ours. It never arrives here. There are four tomllib parses
-    in ``kstrl/`` - ``tests/test_toml_readers.py`` pins the census - and
-    every one of them ends on a bare ``except Exception``:
+    defect of ours. It never arrives here. Every tomllib parse in
+    ``kstrl/`` ends on a bare ``except Exception``, and
+    ``EXPECTED_TOML_PARSES`` in ``tests/test_toml_readers.py`` is what
+    makes that a closed set rather than a count restated here:
     ``config.load_toml_document`` re-raises ``ConfigError`` naming the
     path, and the pyproject.toml and ruff.toml readers in ``verify`` and
     ``feedforward`` fall back to a default. So a ``RecursionError`` that
--- a/kstrl/config_preflight.py
+++ b/kstrl/config_preflight.py
@@ -564,8 +564,8 @@ def _blamed_toml_value(
     says rather than as a diagnosis.
 
     The ``suppress`` below would swallow a ``RuntimeError`` kstrl did
-    not define, which :func:`raise_if_defect` exists to re-raise. It
-    cannot reach one: this runs only AFTER that function has passed on
+    not define, which :func:`raise_if_defect` exists to re-raise. None
+    has been reproduced: this runs only AFTER that function has passed on
     the exception being reported, and everything ``load_toml_section``
     raises here has already been converted by ``load_toml_document``.
     """
--- a/kstrl/config_preflight.py
+++ b/kstrl/config_preflight.py
@@ -129,6 +129,28 @@ def raise_if_defect(exc: BaseException) -> None:
     ``builtins`` and any dependency's ``RuntimeError`` is not something
     we modelled and so not something we can honestly blame a file for.
     ``tests/test_tui_config_guard.py`` pins both halves.
+
+    ``RecursionError`` is the one that needs an argument rather than a
+    rule, and the argument is layering rather than inspection (#323). A
+    kstrl.toml with 600 nested arrays exhausts the stack inside
+    tomllib's recursive descent, and that is the operator's file, not a
+    defect of ours. It never arrives here. There are four tomllib parses
+    in ``kstrl/`` - ``tests/test_toml_readers.py`` pins the census - and
+    every one of them ends on a bare ``except Exception``:
+    ``config.load_toml_document`` re-raises ``ConfigError`` naming the
+    path, and the pyproject.toml and ruff.toml readers in ``verify`` and
+    ``feedforward`` fall back to a default. So a ``RecursionError`` that
+    does reach this function is a cycle in kstrl's own code, and the
+    traceback this re-raise keeps is what locates it.
+
+    Inspecting the exception could not have settled it anyway. Measured
+    on 3.12.8 and 3.13.2, both directions give
+    ``builtins.RecursionError`` with ``str(exc)`` "maximum recursion
+    depth exceeded" and no attribute of its own; only the frames differ,
+    and by the time they could be read the boundary has already
+    answered. ``tests/test_recursion_provenance.py`` pins both
+    directions, at :func:`preflight_config` and at the ``ks status``
+    seam.
     """
     if isinstance(exc, RuntimeError) and type(exc).__module__.split(".")[0] != "kstrl":
         raise exc
--- a/tests/helpers/bad_toml.py
+++ b/tests/helpers/bad_toml.py
@@ -73,11 +73,15 @@ INT_LIMIT_TOML = b"[run]\nmax_iterations = " + b"9" * _INT_DIGITS + b"\n"
 #: ``ValueError``, and so walked past everything round 2 shipped. That is
 #: #318 round 3.
 #:
-#: Measured by binary search on this machine: 496 standalone, 495 through
-#: the CLI (the seam's own frames cost one level). Scaled off the live
-#: recursion limit rather than fixed at 600, so a run with a raised limit
-#: still nests past it: any depth above the limit exhausts the stack
-#: whatever the limit is, which is also why there is no ``max(600, ...)``
+#: Measured by binary search on this machine at the default limit: the
+#: first depth that raises is 497 from a one-frame caller and 490
+#: through ``ks status``. There is no one threshold. The caller's own
+#: stack sets it, and the CLI parse runs 9 to 16 frames deep, which at
+#: roughly two interpreter frames per array level is the 7 levels the
+#: two numbers differ by. Scaled off the live recursion limit rather
+#: than fixed at 600, so a run with a raised limit still nests past it:
+#: any depth above the limit exhausts the stack whatever the limit is,
+#: which is also why there is no ``max(600, ...)``
 #: floor - it could never bind. Inline tables recurse identically;
 #: arrays are used because they are terser per level. Measured cost of
 #: building it: 0.21 us, 2205 bytes.
--- a/tests/helpers/bad_toml.py
+++ b/tests/helpers/bad_toml.py
@@ -76,9 +76,9 @@ INT_LIMIT_TOML = b"[run]\nmax_iterations = " + b"9" * _INT_DIGITS + b"\n"
 #: Measured by binary search on this machine at the default limit: the
 #: first depth that raises is 497 from a one-frame caller and 490
 #: through ``ks status``. There is no one threshold. The caller's own
-#: stack sets it, and the CLI parse runs 9 to 16 frames deep, which at
-#: roughly two interpreter frames per array level is the 7 levels the
-#: two numbers differ by. Scaled off the live recursion limit rather
+#: stack sets it, and the CLI parse that raises runs 16 frames deep,
+#: which at roughly two interpreter frames per array level is the 7
+#: levels the two numbers differ by. Scaled off the live recursion limit rather
 #: than fixed at 600, so a run with a raised limit still nests past it:
 #: any depth above the limit exhausts the stack whatever the limit is,
 #: which is also why there is no ``max(600, ...)``
--- a/tests/test_recursion_provenance.py
+++ b/tests/test_recursion_provenance.py
@@ -9,36 +9,71 @@ series rather than in competition:
 - ``config.load_toml_document`` catches ``Exception`` around the parse
   and re-raises ``ConfigError`` naming the file, so the document's
   ``RecursionError`` never travels further than that call.
-- Every block whose exception reaches ``config_preflight.raise_if_defect``
-  reaches tomllib only through that function. So a ``RecursionError``
-  that arrives there is kstrl's own by construction, and re-raising it
-  with its traceback intact is right.
+- Every tomllib parse in ``kstrl/`` ends on a bare ``except
+  Exception``, so no document's ``RecursionError`` escapes the parse
+  that raised it: ``load_toml_document`` converts it to
+  ``ConfigError``, and the pyproject.toml and ruff.toml readers in
+  ``verify`` and ``feedforward`` swallow it. Not every guarded block
+  goes through ``load_toml_document`` - ``init_wizard._detected_text``
+  reaches ``verify._default_typecheck_command`` - which is why the
+  closure is over the PARSES and not over the call graph.
 
 These tests pin both directions in-process through
-``preflight_config``, and then direction (b) again at the CLI seam,
-where an operator sees the difference between a reported line and a
-traceback. Direction (a) uses real bytes, and direction (b) a real
-self-calling function rather than a constructed ``RecursionError``,
-because a stub would pass in every broken state.
+``preflight_config``, then direction (b) again at
+:func:`config_problem_lines` and at the CLI seam, where an operator
+sees the difference between a reported line and a traceback. Direction
+(a) uses real bytes, and direction (b) a real self-calling function
+rather than a constructed ``RecursionError``, because a stub would
+pass in every broken state.
 
-What was already pinned, and why it is not enough. Direction (a) lives
-at the loader in ``test_config_toml.py``
+THE RESIDUAL, WHICH IS STATED RATHER THAN TESTED
+------------------------------------------------
+``load_toml_document``'s catch-all converts any ``Exception`` out of
+the parse, so kstrl's OWN runaway that happens to bottom out inside a
+parse is reported to the operator as their broken file. Measured on
+CPython 3.12.8 at the default limit with a valid two-line kstrl.toml:
+a caller holding 991 or 992 of the 1000 frames gets that relabelling,
+and at 993 the ``RecursionError`` is raised before the guard and
+escapes it. A caller that deep is itself the runaway, so the residual
+is one runaway reported as a file rather than as a traceback, in a
+two-frame band. No assertion here pins it: a test would have to key on
+the interpreter's recursion limit and on this one function's frame
+cost, and would move whenever either did.
+
+WHAT IS ALREADY PINNED, AND WHAT IS NEW
+---------------------------------------
+``test_config_guard_survey.py`` is the closest prior art at the shared
+traversal. Its
+``test_the_entry_check_does_not_list_our_own_defect_as_a_config_problem``
+monkeypatches ``EvolutionConfig.load`` to raise ``NotImplementedError``
+- the other ``builtins`` ``RuntimeError`` subclass
+``raise_if_defect``'s docstring names beside ``RecursionError`` - and
+asserts it propagates through BOTH ``collect_config_problems`` and
+:func:`config_problem_lines`.
+Direction (a) lives at the loader in ``test_config_toml.py``
 (``test_a_recursion_error_is_reported_not_raised``), one call below
 :func:`preflight_config`, and at the seam as the ``deep_nest`` row of
 ``TOML_PARSE_FAULTS`` crossed with ``["status"]`` in
 ``test_config_preflight.py`` - which is why the seam test below covers
-only direction (b) rather than restating that cell. Direction (b) lives
-at :func:`load_or_report` in ``test_tui_config_guard.py``
+only direction (b) rather than restating that cell. Direction (b)
+lives at :func:`load_or_report` in ``test_tui_config_guard.py``
 (``test_every_runtimeerror_kstrl_did_not_define_is_re_raised``) and
-raises a CONSTRUCTED ``RecursionError``, so no stack is ever exhausted.
-Neither goes through ``collect_config_problems``, which is the
-traversal all three reporting surfaces share, and neither is a pair:
-the two halves are what make this a provenance claim rather than two
-unrelated handler checks.
+raises a CONSTRUCTED ``RecursionError``, so no stack is ever
+exhausted.
+
+Three things here are not in any of those: a REAL stack exhaustion, so
+the frames are the runaway's and the re-raise has something to keep;
+``warned == []``, which is the only assertion anywhere that the one
+DEGRADING section must not turn a defect of ours into a warning line
+(``test_config_preflight.py:test_a_clean_config_raises_nothing`` makes
+the same assertion about a CLEAN config, which is a different claim);
+and the CLI seam, where the operator gets the traceback rather than a
+reported line.
 """
 
 from __future__ import annotations
 
+import tomllib
 import traceback
 from pathlib import Path
 
@@ -47,16 +82,25 @@ from click.testing import CliRunner
 
 from kstrl.cli import cli
 from kstrl.config import ConfigError
-from kstrl.config_preflight import preflight_config
+from kstrl.config_preflight import config_problem_lines, preflight_config
 from kstrl.evolution import EvolutionConfig
 from tests.helpers.bad_toml import BROAD_FRAGMENT, DEEP_NEST_TOML
 
-VALID_TOML = "[agent]\nname = 'x'\n"
+VALID_TOML = "[agent]\ntype = 'claude-code'\n"
+
+#: Where the standard library's TOML parser lives, resolved rather than
+#: spelled: ``_parser.py`` is CPython-private, so a rename or a split
+#: would make a filename check go quiet instead of red.
+_TOMLLIB_DIR = Path(tomllib.__file__).parent
 
 
 def _runaway(root_dir: Path | None = None) -> None:
     """A loader that recurses forever. kstrl's defect, not the file's."""
     return _runaway(root_dir)
+
+
+def _frame_names(exc: BaseException) -> list[str]:
+    return [f.name for f in traceback.extract_tb(exc.__traceback__)]
 
 
 class TestRecursionErrorProvenance:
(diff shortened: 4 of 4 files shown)


kstrl/config.py
```diff
@@ -564,6 +566,10 @@ def load_toml_document(path: Path) -> dict[str, Any]:
     that no handler can promise the interpreter has the headroom left to
     render the message.
 
+    Provenance follows (#323): the parse runs at stack depth 9 to 16
+    under ``ks status``, so a ``RecursionError`` here is the document's.
+    One above this call is ours; ``raise_if_defect`` carries that half.
```
0xfauzi: **The provenance claim only holds in one direction, and the other direction is the one #323 asked about.**

`load_toml_document`'s catch-all converts *any* `Exception` out of `tomllib.loads` into `ConfigError` naming the file. That includes a `RecursionError` that kstrl's own cycle caused and that merely happened to bottom out inside the parse. Measured on this branch (3.12.8, `__pycache__` purged, `PYTHONDONTWRITEBYTECODE=1`), with a perfectly valid two-line `kstrl.toml` and a caller recursed to depth 992:

```
MISLABELLED at n=992 -> ConfigError: /.../kstrl.toml could not be parsed as TOML: maximum recursion depth exceeded
```

So "a `RecursionError` here is the document's" is not true by construction; it is true only when the caller's stack is shallow, which is the very thing the number beside it (9 to 16 under `ks status`) is measuring and which the PR body lists as *unmeasured* for the TUI surface. For reference, I measured `load_or_report` -> `load_toml_document` at depth 52 under a Textual screen (`ConfigProblemBanner` on `InboxScreen`), against 9-16 for `ks status`.

This is the mislabelling #323 asked for a predicate to prevent, in the direction the PR does not cover: kstrl's defect reported as the operator's broken file, exit 1, no traceback. Nothing in this PR pins it, and the docstring asserts it away. If the position is that this direction is acceptable, say so here rather than stating the boundary settles both.

tests/test_recursion_provenance.py
```diff
@@ -0,0 +1,100 @@
...
+at the loader in ``test_config_toml.py``
+(``test_a_recursion_error_is_reported_not_raised``), one call below
+:func:`preflight_config`, and at the seam as the ``deep_nest`` row of
+``TOML_PARSE_FAULTS`` crossed with ``["status"]`` in
+``test_config_preflight.py`` - which is why the seam test below covers
+only direction (b) rather than restating that cell. Direction (b) lives
+at :func:`load_or_report` in ``test_tui_config_guard.py``
+(``test_every_runtimeerror_kstrl_did_not_define_is_re_raised``) and
+raises a CONSTRUCTED ``RecursionError``, so no stack is ever exhausted.
+Neither goes through ``collect_config_problems``, which is the
```
0xfauzi: **"Neither goes through `collect_config_problems`" is false for both of the coverages this paragraph enumerates, and it misses the closest prior art entirely.**

1. `tests/test_config_guard_survey.py::test_the_entry_check_does_not_list_our_own_defect_as_a_config_problem` already monkeypatches `EvolutionConfig.load` with a function that raises `NotImplementedError` (a `builtins` `RuntimeError` subclass, one of the two classes `raise_if_defect`'s own docstring names beside `RecursionError`) and asserts it propagates through **both** `collect_config_problems` **and** `config_problem_lines`. Its docstring literally reads "The seam all three reporting surfaces route through". That is direction (b) at the traversal this file claims is uncovered.

2. The `deep_nest` x `["status"]` matrix cell cited two sentences earlier invokes `ks status`, which reaches `_KstrlCommand.invoke` -> `preflight_config` -> `collect_config_problems`. So direction (a) goes through it too.

The new file is also *narrower* on the axis it claims: the existing survey test covers `config_problem_lines`, the third reporting surface, and nothing here does.

tests/test_recursion_provenance.py
```diff
@@ -0,0 +1,100 @@
...
+    def test_a_kstrl_recursion_is_ours_and_keeps_its_traceback(
+        self, tmp_path: Path, monkeypatch: pytest.MonkeyPatch
+    ) -> None:
+        (tmp_path / "kstrl.toml").write_text(VALID_TOML, encoding="utf-8")
+        monkeypatch.setattr(EvolutionConfig, "load", staticmethod(_runaway))
+        warned: list[str] = []
+        with pytest.raises(RecursionError) as caught:
+            preflight_config(tmp_path, warn=warned.append)
+        frames = traceback.extract_tb(caught.value.__traceback__)
+        assert "_runaway" in [f.name for f in frames]
```
0xfauzi, line 127: **Neither direction-(b) test can tell "`raise_if_defect` re-raised it" from "nothing caught it at all", which is the whole claim.**

Mutation run on this branch, `__pycache__` purged, `PYTHONDONTWRITEBYTECODE=1`: narrow `REJECTIONS` to `(ValueError, TypeError)` in `kstrl/config_preflight.py:90`. That deletes `RuntimeError` from the seam's catch, so the runaway's `RecursionError` never reaches `raise_if_defect` and every domain `RuntimeError` (`ServeError`, `QueueError`, `InboxError`, `IntakeError`) starts escaping as a raw traceback. Result:

```
3 passed in 0.84s
```

The observable outcome is identical whether the seam re-raised deliberately or never caught it, so this is the repo's own dominant defect class: a guard that clears in the skip direction. `frames` is already in hand one line above; the discriminating assertion costs nothing.

```suggestion
        assert "_runaway" in [f.name for f in frames]
        # Re-RAISED by the seam, not merely never caught: narrowing
        # REJECTIONS to (ValueError, TypeError) makes raise_if_defect
        # unreachable and every assertion below still passes.
        assert "raise_if_defect" in [f.name for f in frames]
```

kstrl/config_preflight.py
```diff
@@ -129,6 +129,29 @@ def raise_if_defect(exc: BaseException) -> None:
...
+    traceback this re-raise keeps is what locates it.
+
+    Inspecting the exception could not have settled it anyway. Measured
+    on 3.12.8 and 3.13.2, both directions give
+    ``builtins.RecursionError`` with ``str(exc)`` "maximum recursion
+    depth exceeded" and no attribute of its own; only the frames differ,
+    and by the time they could be read the boundary has already
+    answered. ``tests/test_recursion_provenance.py`` pins both
+    directions, at :func:`preflight_config` and at the ``ks status``
+    seam.
```
0xfauzi: **Stale after commit 2 (`6af50be`).** This says `tests/test_recursion_provenance.py` "pins both directions, at `preflight_config` **and at the `ks status` seam**". The simplify commit deleted the direction-(a) half of the seam test and renamed it `test_the_seam_re_raises_ours_instead_of_reporting_a_file`; the seam now pins direction (b) only, with (a) living in `test_config_preflight.py`'s matrix. A reader following this sentence goes looking for a direction-(a) seam assertion in that file and does not find one.

This is exactly the drift the same commit's message invokes as its reason for editing the `config.py` half ("Two copies of a load-bearing claim are two copies that can drift") - it just moved the second copy instead of removing it.

tests/test_recursion_provenance.py
```diff
@@ -0,0 +1,100 @@
...
+    def test_a_parse_recursion_is_the_files_fault_and_names_the_file(self, tmp_path: Path) -> None:
+        toml = tmp_path / "kstrl.toml"
+        toml.write_bytes(DEEP_NEST_TOML)
+        with pytest.raises(ConfigError) as caught:
+            preflight_config(tmp_path, warn=lambda m: None)
+        assert str(toml) in str(caught.value)
+        assert BROAD_FRAGMENT in str(caught.value)
+        # The cause is kept, so the boundary is visible in the report
+        # rather than inferred from the message alone.
+        assert isinstance(caught.value.__cause__, RecursionError)
```
0xfauzi, line 116: **This test is a near-verbatim restatement of `tests/test_config_toml.py::test_a_recursion_error_is_reported_not_raised`.** That test asserts the same two facts on the same fixture:

```python
    assert isinstance(caught.value.__cause__, RecursionError)
    assert BROAD_FRAGMENT in str(caught.value)
```

The only addition here is `str(toml) in str(caught.value)`, and the `[status-deep_nest]` matrix cell this file cites already asserts `"kstrl.toml" in result.output` through `collect_config_problems`. So the union of the two coverages the module docstring calls insufficient already covers every assertion in this method.

The reasoning commit 2 used to delete the seam's direction-(a) half - "exactly the `[status-deep_nest]` cell ... that cell asserts strictly more" - applies to this method with one fewer step.

kstrl/config_preflight.py
```diff
@@ -129,6 +129,29 @@ def raise_if_defect(exc: BaseException) -> None:
...
     ``tests/test_tui_config_guard.py`` pins both halves.
+
+    ``RecursionError`` is the one that needs an argument rather than a
+    rule, and the argument is layering rather than inspection (#323). A
+    kstrl.toml with 600 nested arrays exhausts the stack inside
+    tomllib's recursive descent, and that is the operator's file, not a
+    defect of ours. It never arrives here. Every tomllib parse in
+    ``kstrl/`` ends on a bare ``except Exception``, and
+    ``EXPECTED_TOML_PARSES`` in ``tests/test_toml_readers.py`` is what
+    makes that a closed set rather than a count restated here:
```
0xfauzi: **Two of the four parses this sentence leans on put the I/O inside the guard, and nothing checks that.**

`kstrl/feedforward.py:540` and `:586` are both:

```python
    try:
        data = tomllib.loads(path.read_text(encoding="utf-8"))
    except Exception:
        return
```

so the `read_text` `OSError` is swallowed into the same silent default as a parse fault - rule (3) of the four CLAUDE.md states from #318 ("ALL the I/O outside the guarded block"). `verify.py:934` and `config.py:534` hoist the read correctly; these two do not. `test_toml_readers.py`'s failure message *states* the rule ("keep all I/O outside the guard") but the walk only judges handler class and clause order, so the two offenders are invisible - the guard-that-clears-in-the-skip-direction shape again.

It does not break the `RecursionError` claim (an `OSError` is not a `RecursionError`), but the sentence cites these sites as the evidence, so it is worth either fixing them in this PR or not resting the argument on them.

tests/test_recursion_provenance.py
```diff
@@ -0,0 +1,100 @@
...
+        self, tmp_path: Path, monkeypatch: pytest.MonkeyPatch
+    ) -> None:
+        (tmp_path / "kstrl.toml").write_text(VALID_TOML, encoding="utf-8")
+        monkeypatch.setattr(EvolutionConfig, "load", staticmethod(_runaway))
+        warned: list[str] = []
+        with pytest.raises(RecursionError) as caught:
+            preflight_config(tmp_path, warn=warned.append)
+        frames = traceback.extract_tb(caught.value.__traceback__)
+        assert "_runaway" in [f.name for f in frames]
+        assert "_parser.py" not in [Path(f.filename).name for f in frames]
```
0xfauzi: **An absence assertion with no control, spelled as a stdlib implementation filename.**

In direction (b) `tomllib` is never entered, so `"_parser.py"` cannot appear in these frames for reasons that have nothing to do with `raise_if_defect`. And the string is CPython's private module filename: if tomllib's parser module is renamed, split, or replaced by a C accelerator, this line goes quiet rather than red. Per CLAUDE.md rule (2), an assertion that passes both when the code is right and when the mechanism is gone is not a control. Either drop it or key it on `tomllib.__file__` / the `tomllib` package directory so a rename fails loudly.

tests/test_recursion_provenance.py
```diff
@@ -0,0 +1,100 @@
...
+        # [evolution] is the one degrading section. Its degrade path must
+        # NOT swallow this: a warning line here would hide the cycle.
+        assert warned == []
+
+    def test_the_seam_re_raises_ours_instead_of_reporting_a_file(
+        self, tmp_path: Path, monkeypatch: pytest.MonkeyPatch
+    ) -> None:
+        monkeypatch.chdir(tmp_path)
```
0xfauzi, line 166: Redundant: `tests/conftest.py::isolate_kstrl_state` is `autouse=True`, requests the same `tmp_path`, and its first statement is `monkeypatch.chdir(tmp_path)`. This line re-does it. (The sibling seam tests in `test_config_preflight.py` use `runner.isolated_filesystem()` via `_invoke` instead, which is a third convention for the same job in the same area.)

tests/test_recursion_provenance.py
```diff
@@ -0,0 +1,100 @@
...
+        (tmp_path / "kstrl.toml").write_text(VALID_TOML, encoding="utf-8")
+        monkeypatch.setattr(EvolutionConfig, "load", staticmethod(_runaway))
+        result = CliRunner().invoke(cli, ["status"], env={"KSTRL_NO_TUI": "1"})
```
0xfauzi: `env={"KSTRL_NO_TUI": "1"}` is dead weight here. `ks status` only takes the TUI branch when `sys.stdout.isatty() and sys.stdin.isatty()` (`cli.py:3454-3456`), and `CliRunner`'s streams are never ttys, so `use_tui` is already False. Carrying it suggests the test depends on it; it does not, and a reader cannot tell that from here.

kstrl/config.py
```diff
@@ -564,6 +566,10 @@ def load_toml_document(path: Path) -> dict[str, Any]:
     that no handler can promise the interpreter has the headroom left to
     render the message.
 
+    Provenance follows (#323): the parse runs at stack depth 9 to 16
```
0xfauzi: **The depth model added here contradicts a measured comment one file over that the PR does not update.** `tests/helpers/bad_toml.py:76` still reads "496 standalone, 495 through the CLI (the seam's own frames cost one level)".

Measured on this branch (3.12.8, limit 1000, binary search on first failing nest depth):

```
caller extra frames   0 -> 496
caller extra frames  15 -> 489
caller extra frames  50 -> 471
caller extra frames 200 -> 396
```

The 471/396 figures reproduce this docstring exactly, so the new model is right. But 15 extra caller frames cost 7 levels, not one, and I measured the `ks status` parse at stack depth 10-17 (`tomllib.loads` spy through `CliRunner`), so the CLI figure should be around 489, not 495. One of the two numbers is wrong and they now sit in two files. Since this PR's whole second commit is about not leaving two copies of a load-bearing measurement, `bad_toml.py`'s belongs in the same edit.

kstrl/config_preflight.py
```diff
@@ -129,6 +129,37 @@ def raise_if_defect(exc: BaseException) -> None:
...
+    operator's file, not a defect of ours. It never arrives here. Every
+    tomllib parse in ``kstrl/`` ends on a bare ``except Exception``, and
+    the two-layer census in ``tests/test_toml_readers.py`` is what makes
+    that a closed set rather than a count restated here - layer 1,
+    ``EXPECTED_TOMLLIB_SPELLINGS``, resolves nothing and so cannot miss
+    a reader that spells ``tomllib`` at all. ``load_toml_document``
+    re-raises ``ConfigError`` naming the path, and the pyproject.toml
+    and ruff.toml readers in ``verify`` and ``feedforward`` fall back to
+    a default or to no conventions at all. So a ``RecursionError`` that
+    does reach this function is a cycle in kstrl's own code, and the
```
0xfauzi, line 146: **Seven of the ten `raise_if_defect` call sites are unpinned; deleting all seven leaves the suite green.**

Measured on this head, `__pycache__` purged, `PYTHONDONTWRITEBYTECODE=1`, full suite bounded: deleting `raise_if_defect(exc)` from all of

- `kstrl/config_report.py:435` (`build_config_report._resolve`)
- `kstrl/cli.py:3082` (`ks config show`)
- `kstrl/tui/home.py:38`
- `kstrl/tui/session.py:208` and `:260`
- `kstrl/tui/screens/config.py:248`
- `kstrl/tui/screens/init_wizard.py:89`

in one tree gives **5568 passed, 37 skipped, 38 xfailed** - byte-identical to the baseline this PR reports. `config_report.py:435` alone was also measured on its own: same numbers, green.

So the mechanism this docstring says "the traceback this re-raise keeps is what locates it" depends on is enforced at exactly 3 of 10 sites (the three in this file), and two of those three were pinned by this PR. That is a ledger of the sites somebody remembered, not a closure. CLAUDE.md: *"One defect class, one PR. Fix every instance plus the guard that catches instance N+1"* and *"Prefer closed by construction over a ledger of give-ups: inventory every place the resource is OBTAINED and count what you saw."*

The inventory is cheap and already has a home: every `except REJECTIONS` / `except SURFACE_REJECTIONS` handler in `kstrl/`, counted, each one required to call `raise_if_defect` in its body (the one exemption, `_blamed_toml_value`'s `suppress`, is the row that carries the reason this PR already wrote). `tests/test_config_guard_survey.py` is where its behavioural half already lives, and `tests/helpers/astwalk` already resolves `SURFACE_REJECTIONS` by origin for `tests/test_tui_config_walk.py`.

tests/test_recursion_provenance.py
```diff
@@ -0,0 +1,174 @@
...
+    def test_the_third_reporting_surface_re_raises_ours_too(
+        self, tmp_path: Path, monkeypatch: pytest.MonkeyPatch
+    ) -> None:
+        """``config_problem_lines`` has a ``raise_if_defect`` of its own.
+
+        It catches SURFACE_REJECTIONS around the traversal and folds an
+        unparseable document back in as one line. That catch sees this
+        ``RecursionError`` a second time, so deleting only ITS
+        ``raise_if_defect`` turns the runaway into a returned string
+        while the test above stays green.
```
0xfauzi, line 154: **This docstring's premise does not reproduce: the mutation it exists for is already caught.**

The claim is that deleting only `config_problem_lines`' own `raise_if_defect` "turns the runaway into a returned string while the test above stays green", and the PR body records it as *"Also STILL GREEN before this round."*

Measured on this head with that exact deletion planted (anchor count asserted at 1, `__pycache__` purged, `PYTHONDONTWRITEBYTECODE=1`), running only the pre-existing files:

```
FAILED tests/test_config_guard_survey.py::test_the_entry_check_does_not_list_our_own_defect_as_a_config_problem
E       Failed: DID NOT RAISE NotImplementedError
tests/test_config_guard_survey.py:267
1 failed, 36 passed, 1 xfailed
```

`test_the_entry_check_does_not_list_our_own_defect_as_a_config_problem` already asserts `pytest.raises(NotImplementedError)` around **both** `collect_config_problems` **and** `config_problem_lines` - the module docstring above cites it as "the closest prior art" and then reads it as covering only the first. `NotImplementedError` is in `SURFACE_REJECTIONS`, so without the re-raise it comes back as `[str(exc)]` and that test fails.

So this test is a duplicate of existing coverage and the docstring's justification for it is wrong. Either delete it, or keep it and say what it adds that `test_config_guard_survey.py:247` does not (a real stack exhaustion rather than a raised instance is a defensible answer; "the test above stays green" is not).

tests/test_recursion_provenance.py
```diff
@@ -0,0 +1,174 @@
...
+        with pytest.raises(RecursionError) as caught:
+            preflight_config(tmp_path, warn=warned.append)
+        frames = traceback.extract_tb(caught.value.__traceback__)
+        assert "_runaway" in [f.name for f in frames]
+        # Re-RAISED by raise_if_defect, not merely never caught by
+        # anything. Narrowing REJECTIONS to (ValueError, TypeError)
+        # makes raise_if_defect unreachable for this exception; the
+        # RecursionError still comes out of preflight_config, so every
+        # other assertion here holds. Measured before this line and the
+        # two like it existed: that mutation left the file 3 of 3 green.
```
0xfauzi, line 133: **"that mutation left the file 3 of 3 green" does not reproduce either.**

Planted `REJECTIONS = (ValueError, TypeError)` on this head and ran the full suite with `tests/test_recursion_provenance.py` ignored entirely (`__pycache__` purged, `PYTHONDONTWRITEBYTECODE=1`):

```
FAILED tests/test_serve_cli.py::TestServeOnce::test_an_invalid_config_exits_two
FAILED tests/test_tui_config_guard.py::TestSharedGuard::test_every_module_that_spells_the_domain_base_is_pinned
FAILED tests/test_tui_config_guard.py::TestSharedGuard::test_a_domain_runtimeerror_is_still_reported
3 failed, 5561 passed, 37 skipped, 38 xfailed
```

All three files predate this branch, so the suite was not green under that mutation at the first review head either. The comment may be true of *this file* in isolation, but it is written as, and the PR body reads as, a statement about the suite ("STILL GREEN before this round's fix ... it is the whole reason the `raise_if_defect` frame assertions exist").

The assertion is still worth keeping - it is the only one that distinguishes a deliberate re-raise from nothing having caught it. Just restate the measurement to what it actually shows: three sibling tests catch the narrowing, none of them says *where* the re-raise happened, and this line does. H4.

tests/test_recursion_provenance.py
```diff
@@ -0,0 +1,174 @@
...
+        assert "raise_if_defect" in [f.name for f in frames]
+        # A DISCLOSURE, not a control. This fixture writes VALID_TOML,
+        # so tomllib is never entered and no production mutation can
+        # turn this line red. It records the direction the frames point
+        # in; the kills are the two assertions above.
+        assert not [f for f in frames if Path(f.filename).is_relative_to(_TOMLLIB_DIR)]
```
0xfauzi, line 140: **`assert not [...]` disclosed by comment is the shape CLAUDE.md says must be a strict xfail.**

*"`assert hits(...) == []` is not a control. It passes when the walk is correctly narrow AND when the walk has been switched off, which is the same shape as the defect. Express a genuine disclosed miss as an explicit blind-spot record plus a strict xfail, so a later widening XPASSes and fails loudly instead of silently agreeing."*

The comment above is exactly that disclosure, and it is honest, but a comment is not the mechanism the rule asks for. This repo already has the pattern in the file this PR leans on: `tests/test_toml_readers.py::TestTheDisclosedLimits`, *"Under `xfail(strict=True)`, so the day a walk gets stronger the row XPASSes and the docstring has to be edited in the same diff. A disclosure with no test behind it rots silently."*

Concretely: if `VALID_TOML` were ever changed to something that does reach tomllib, this line quietly becomes a real assertion nobody chose, and if `_TOMLLIB_DIR` ever resolves to something no frame can be relative to, it quietly becomes vacuous in a second way. Neither shows up. Move it to its own `xfail(strict=True)` row, or drop it and keep the comment.

kstrl/config_preflight.py
```diff
@@ -129,6 +129,37 @@ def raise_if_defect(exc: BaseException) -> None:
...
     we modelled and so not something we can honestly blame a file for.
     ``tests/test_tui_config_guard.py`` pins both halves.
+
+    ``RecursionError`` is the one that needs an argument rather than a
+    rule, and the argument is layering rather than inspection (#323). A
+    kstrl.toml with 600 nested arrays exhausts the stack at the default
+    recursion limit inside tomllib's recursive descent, and that is the
+    operator's file, not a defect of ours. It never arrives here. Every
+    tomllib parse in ``kstrl/`` ends on a bare ``except Exception``, and
+    the two-layer census in ``tests/test_toml_readers.py`` is what makes
```
0xfauzi, line 139: **The census this paragraph cites closes the parses; nothing closes the re-raise, and the walk that looks closest actually clears it.**

`tests/test_tui_config_walk.py::guarding_handler` decides a config load inside `kstrl/tui/` is safe when its handler's clause origins include `kstrl.config_preflight.SURFACE_REJECTIONS`. It never looks at the handler *body*, so a new screen that writes

```python
try:
    cfg = SomeConfig.load(root)
except SURFACE_REJECTIONS:
    cfg = None
```

is CLEARED by that walk, passes the whole suite, and silently swallows exactly the defect `raise_if_defect` exists to re-raise.

CLAUDE.md, from the #324 sweep: *"A guard that CLEARS must be narrow, because over-matching converts a resolution into a clearing and deletes the mechanism ... If a clearing guard cannot PROVE a site is compliant, it must flag."* That file's own docstring already states the direction it is allowed to be wrong in ("a spurious offender is a line somebody reads, a missing one is #289") - the `raise_if_defect` half is the part it cannot prove today.

Requiring a `raise_if_defect` call in the cleared handler's body is a one-predicate widening and is the same guard the census in the sibling comment needs.

kstrl/config.py
```diff
@@ -564,6 +566,12 @@ def load_toml_document(path: Path) -> dict[str, Any]:
     that no handler can promise the interpreter has the headroom left to
     render the message.
 
+    Provenance follows (#323): the parse runs 9 to 16 frames deep under
+    ``ks status``, 13 to 52 on the Textual event loop, so a
+    ``RecursionError`` here is the document's. Not by construction:
+    measured, a CALLER at 991 of 1000 frames gets a valid file called
+    unparseable, and IS the runaway. ``raise_if_defect`` has the rest.
```
0xfauzi: **This lands `kstrl/config.py` at exactly 800 lines, the ratchet's ceiling, for prose the new test module already carries.**

`wc -l kstrl/config.py` = 800; `scripts/precommit/file_length_ratchet.py` has `MAX_LINES = 800` and fails a file that "was at or under the limit and is now over it". The hook passes at 800 and fails at 801, so the next line anyone adds to this file - in this branch or in a sibling one - is a split they did not come to do.

CLAUDE.md records this exact merge failure: *"PR #341 merging #342 found a file at 750 lines on one side and 800 on the other merging to 827 against an 800-line ratchet, a failure present in neither branch and attached to no hunk."* `config.py` is a hot file and main is moving, so a textually clean merge can land over the ceiling with no conflict on any hunk here.

The PR body already acknowledges the zero headroom and says it is why the fuller residual statement lives in the test module docstring. The same argument applies to this paragraph: `tests/test_recursion_provenance.py`'s "THE RESIDUAL" section states all of it, with the 991/992/993 band, and this repeats it in five lines. Cross-reference the test module and keep the file under the ceiling.

(The 991/992 relabel and 993 escape reproduce exactly here, so the measurement itself is good.)

tests/test_recursion_provenance.py
```diff
@@ -0,0 +1,174 @@
...
+    def test_a_parse_recursion_is_the_files_fault_and_names_the_file(self, tmp_path: Path) -> None:
+        toml = tmp_path / "kstrl.toml"
+        toml.write_bytes(DEEP_NEST_TOML)
+        with pytest.raises(ConfigError) as caught:
+            preflight_config(tmp_path, warn=lambda m: None)
+        assert str(toml) in str(caught.value)
+        assert BROAD_FRAGMENT in str(caught.value)
+        # The cause is kept, so the boundary is visible in the report
+        # rather than inferred from the message alone.
+        assert isinstance(caught.value.__cause__, RecursionError)
```
0xfauzi, line 116: **Assertion-for-assertion the same test as `tests/test_config_toml.py:255`, one frame up.**

`test_a_recursion_error_is_reported_not_raised` already writes `DEEP_NEST_TOML` and asserts the path is in the message, `BROAD_FRAGMENT` is in the message, and `caught.value.__cause__` is a `RecursionError` - all three, in that order. The module docstring above names it and says direction (a) "lives at the loader" there, then restates it here against `preflight_config` instead of `load_toml_document`.

The PR's own mutation data agrees: narrowing the broad clause to `ValueError` is **1 of 4 red**, and the one is this test - the same kill `test_config_toml.py` already had.

What is genuinely new here is only that `collect_config_problems` does not re-catch and relabel the `ConfigError` on the way out. If that is the claim, make it the assertion (the `except OSError` at `config_preflight.py:317` is the thing that could swallow it) and drop the three restated ones.

tests/test_recursion_provenance.py
```diff
@@ -0,0 +1,174 @@
...
+:func:`config_problem_lines`.
+Direction (a) lives at the loader in ``test_config_toml.py``
+(``test_a_recursion_error_is_reported_not_raised``), one call below
+:func:`preflight_config`, and at the seam as the ``deep_nest`` row of
+``TOML_PARSE_FAULTS`` crossed with ``["status"]`` in
+``test_config_preflight.py`` - which is why the seam test below covers
+only direction (b) rather than restating that cell. Direction (b)
+lives at :func:`load_or_report` in ``test_tui_config_guard.py``
+(``test_every_runtimeerror_kstrl_did_not_define_is_re_raised``) and
+raises a CONSTRUCTED ``RecursionError``, so no stack is ever
```
0xfauzi, line 61: **The rule this file states is not applied to the one surface that still breaks it.**

Lines 58-62 name `load_or_report`'s coverage in `test_tui_config_guard.py::test_every_runtimeerror_kstrl_did_not_define_is_re_raised` and note that it "raises a CONSTRUCTED `RecursionError`, so no stack is ever exhausted" - and the docstring's own premise is that direction (b) uses a real self-calling function "because a stub would pass in every broken state".

`load_or_report` is the fourth reporting surface and the only one a long-lived TUI screen reaches with no seam in front of it. This PR adds real-exhaustion coverage at three surfaces and leaves the stub at the one it identifies as weakest. `_runaway` is already in this file; adding a fourth case that calls `load_or_report(EvolutionConfig.load, tmp_path, blame_env=False)` under the same monkeypatch is three lines.

(For the record, `config_preflight.py:405` *is* pinned today - the constructed-`RecursionError` parametrize does catch its deletion. The gap is fidelity, not coverage.)

tests/test_recursion_provenance.py
```diff
@@ -0,0 +1,174 @@
...
+        # two like it existed: that mutation left the file 3 of 3 green.
+        assert "raise_if_defect" in [f.name for f in frames]
+        # A DISCLOSURE, not a control. This fixture writes VALID_TOML,
+        # so tomllib is never entered and no production mutation can
+        # turn this line red. It records the direction the frames point
+        # in; the kills are the two assertions above.
+        assert not [f for f in frames if Path(f.filename).is_relative_to(_TOMLLIB_DIR)]
+        # [evolution] is the one degrading section. Its degrade path must
+        # NOT swallow this: a warning line here would hide the cycle.
+        assert warned == []
```
0xfauzi, line 143: **`warned == []` is dominated by the `pytest.raises` two lines above it, so it is not the control the module docstring bills it as.**

The docstring calls it one of the three things "not in any of those" prior tests: *"the only assertion anywhere that the one DEGRADING section must not turn a defect of ours into a warning line."*

`warn` is only ever called at `config_preflight.py:335`, in the `else` of the same `if section.fatal` that follows `raise_if_defect(exc)`. `[evolution]` is the only non-fatal section and the only one failing in this fixture. So any mutation that lets `warn` fire also stops the `RecursionError` propagating, and `pytest.raises(RecursionError)` fails first with DID NOT RAISE. The PR's own data shows it: deleting `raise_if_defect` from `collect_config_problems` is 3 of 4 red, and this assertion is never the reason.

The one mutation it does uniquely catch is `warn(...)` being hoisted above `raise_if_defect`. That is a real ordering property and worth keeping - it is just not "a defect must not be degraded into a warning", it is "the defect check runs before the degrade path". Say that instead.

tests/test_recursion_provenance.py
```diff
@@ -0,0 +1,174 @@
...
+        ``RecursionError`` a second time, so deleting only ITS
+        ``raise_if_defect`` turns the runaway into a returned string
+        while the test above stays green.
+        """
+        (tmp_path / "kstrl.toml").write_text(VALID_TOML, encoding="utf-8")
+        monkeypatch.setattr(EvolutionConfig, "load", _runaway)
+        with pytest.raises(RecursionError) as caught:
+            config_problem_lines(tmp_path, warn=lambda m: None)
+        assert "config_problem_lines" in _frame_names(caught.value)
+        assert "raise_if_defect" in _frame_names(caught.value)
```
0xfauzi, line 161: `_frame_names(caught.value)` runs `traceback.extract_tb` over the whole traceback each call, and this test's traceback is ~1000 `_runaway` frames deep - `extract_tb` builds a `FrameSummary` per frame and pulls each one's source line through `linecache`. Both this line and 159 (and 169/170 in the seam test) do it twice.

Bind it once, the way `test_a_kstrl_recursion_is_ours_and_keeps_its_traceback` already does with `frames`:

```suggestion
        names = _frame_names(caught.value)
        assert "config_problem_lines" in names
        assert "raise_if_defect" in names
```
(with line 159 removed)

tests/helpers/bad_toml.py
```diff
@@ -73,11 +73,15 @@
...
 #: #318 round 3.
 #:
-#: Measured by binary search on this machine: 496 standalone, 495 through
-#: the CLI (the seam's own frames cost one level). Scaled off the live
-#: recursion limit rather than fixed at 600, so a run with a raised limit
-#: still nests past it: any depth above the limit exhausts the stack
-#: whatever the limit is, which is also why there is no ``max(600, ...)``
+#: Measured by binary search on this machine at the default limit: the
+#: first depth that raises is 497 from a one-frame caller and 490
+#: through ``ks status``. There is no one threshold. The caller's own
```
0xfauzi, line 78: **Five measured constants in prose with no mechanism to notice when they stop being true, and no interpreter version recorded.**

I reproduced all of these on 3.12.8: 497 from a one-frame caller, 490 through `ks status` (binary search on `UNPARSEABLE_TOML_MESSAGE` in `CliRunner().invoke(cli, ["status"])`), 331 and 264 for inline tables. So the numbers are right today.

Two things they are missing. (1) The interpreter. The PR body says 3.12.8 and notes the 3.13 half was taken under `uvx --python 3.13`; the source says only "on this machine" and "at the default limit". These are CPython-version-specific and CI pins 3.12, so a version bump moves them silently. (2) A mechanism. `_NEST_DEPTH` scales off `sys.getrecursionlimit()`, so nothing goes red when 497 becomes 430 - the comment just becomes false. CLAUDE.md: *"Re-derive a census by running it; never by diffing the literal."*

`kstrl/config.py:549-551` carries the same numbers with the same exposure. Either name the interpreter beside them, or (better, and cheap) make the one-frame threshold a computed helper the deep-nest fixture asserts is above.

tests/test_recursion_provenance.py
```diff
@@ -0,0 +1,174 @@
...
+from kstrl.config_preflight import config_problem_lines, preflight_config
+from kstrl.evolution import EvolutionConfig
+from tests.helpers.bad_toml import BROAD_FRAGMENT, DEEP_NEST_TOML
+
+VALID_TOML = "[agent]\ntype = 'claude-code'\n"
+
+#: Where the standard library's TOML parser lives, resolved rather than
+#: spelled: ``_parser.py`` is CPython-private, so a rename or a split
+#: would make a filename check go quiet instead of red.
+_TOMLLIB_DIR = Path(tomllib.__file__).parent
```
0xfauzi, line 94: `Path(tomllib.__file__).parent` runs at import time. `__file__` is optional on a module object - a frozen or embedded stdlib, or any loader that does not set it, gives `AttributeError`, and `__file__ = None` gives `TypeError` from `Path`. Either takes the whole module out as a collection error rather than skipping the one assertion that uses it.

The comment above says the point is to resolve rather than spell `_parser.py`, which is right. `getattr(tomllib, "__file__", None)` with the line 139 assertion skipped when it is None keeps that property and fails soft. Low severity - CI is CPython 3.12 - but it is a module-scope side effect in a file whose other 173 lines are pure.

tests/test_recursion_provenance.py
```diff
@@ -0,0 +1,174 @@
...
+        assert "raise_if_defect" in _frame_names(caught.value)
+
+    def test_the_seam_re_raises_ours_instead_of_reporting_a_file(
+        self, tmp_path: Path, monkeypatch: pytest.MonkeyPatch
+    ) -> None:
+        monkeypatch.chdir(tmp_path)
+        (tmp_path / "kstrl.toml").write_text(VALID_TOML, encoding="utf-8")
+        monkeypatch.setattr(EvolutionConfig, "load", _runaway)
+        result = CliRunner().invoke(cli, ["status"])
+        assert isinstance(result.exception, RecursionError)
```
0xfauzi, line 170: The seam test asserts the exception type and the absence of both output registers, but not `result.exit_code`. The module docstring's claim for this test is "the CLI seam, where the operator gets the traceback rather than a reported line" - the exit code is half of what an operator or a script actually observes, and it is the half that distinguishes "crashed" from `_PREFLIGHT_EXIT`'s 1 or 2 for a reported config fault.

Its sibling in `tests/test_config_guard_survey.py:102` (`test_ks_config_show_names_the_section_instead_of_a_traceback`) asserts `result.exit_code == 1` alongside `"Traceback" not in result.output` for exactly this reason. Add `assert result.exit_code == 1`.

kstrl/config_preflight.py
```diff
@@ -129,6 +129,37 @@ def raise_if_defect(exc: BaseException) -> None:
...
+    that a closed set rather than a count restated here - layer 1,
+    ``EXPECTED_TOMLLIB_SPELLINGS``, resolves nothing and so cannot miss
+    a reader that spells ``tomllib`` at all. ``load_toml_document``
+    re-raises ``ConfigError`` naming the path, and the pyproject.toml
+    and ruff.toml readers in ``verify`` and ``feedforward`` fall back to
```
0xfauzi, line 144: On the `feedforward` half of this sentence, and the handoff in the PR body: `kstrl/feedforward.py:540` and `:586` are both `tomllib.loads(path.read_text(encoding="utf-8"))` **inside** the `try`, so a read `OSError` lands in the same silent `return` as a parse fault - rule (3) of the four CLAUDE.md records from #318, "ALL the I/O outside the guarded block". `verify.py:934` and `config.py:635` hoist the read; these two do not.

The PR body files this as a measured handoff, which is the right call under "one defect class, one PR", and it does not touch this PR's claim (an `OSError` is not a `RecursionError`). Two things worth attaching to it so it is not lost:

- The reason it is invisible is nameable: `handler_verdict` in `tests/test_toml_readers.py` judges unreadable / too wide / too narrow / out of order, and I/O position is not one of the four - so the guard whose failure message *states* the rule cannot enforce it. One predicate, one guard function, both offenders.
- It is the same shape as the sibling comment on line 139: a guard that clears on the part it can see.

Not a blocker on this diff.

PR #349, opened 2026-09-06
feat(pipeline): hard mode halts when the adversarial budget is exhausted (#226)

### LINKED ISSUE #226: R10.5: Adversarial budget exhausted: hard mode halts instead of merging unreviewed
Part of #235 (R10: Control Loop tracking issue). Item R10.5.

Design section: 5.9 (second half). Sizing: S (one PR, under 100 lines). Dependencies: none.

## Why

When `max_adversarial_calls` is exhausted, kstrl drops the reviewer and keeps the cheap checks.

`_phase_review` (`kstrl/pipeline.py:2465-2478`): in hard mode, when the review is not chunked and `adversarial_budget_ok()` is false, `review_mode` is downgraded to `SKIP`, a warning is printed, and the component proceeds on mechanical verification alone. The chunked hard-mode path a few lines later (`pipeline.py:2479-2510`) does the opposite and correct thing: it fails closed with `Finding.infrastructure_error(phase="review", ...)` and a `PhaseFailure(action=FailureAction.FAIL, ..., check="adversarial_budget")`. `_phase_security` has the same downgrade at `pipeline.py:2696-2701`.

In this repository's journal the reviewer accounts for 14 of the 17 recorded failure signatures. Under budget pressure kstrl sheds the sensor doing most of the catching.

Apollo 11's guidance computer, overloaded during descent, shed low-priority jobs and kept navigation, guidance and engine control. Overload shedding needs a priority policy decided in advance, and kstrl's currently drops a loop-critical sensor.

Two fair qualifications, which the PR body should repeat: the degrade is loud (it warns and records `phase_skipped`, so `len(findings) == 0` still means every phase ran), and the default `max_adversarial_calls = 0` means unbounded, so this path only fires when an operator sets a cap. This is a latent defect, not an active one.

### `kstrl/pipeline.py`, `_phase_review`

Replace the branch at `pipeline.py:2465-2478` so that when `review_mode == ReviewMode.HARD` and the adversarial budget is exhausted (unchunked), the phase fails closed exactly like the chunked path:

- `comp.review_passed = False`
- `self._add_findings(comp, [Finding.infrastructure_error(phase="review", explanation="adversarial LLM budget (<cap>) exhausted before review; hard mode refuses to merge unreviewed")])`
- emit the existing `ChunkBudgetInsufficient` event? No: that event is about chunks. Emit `PhaseSkipped`? No: nothing was skipped, the phase refused. Emit nothing new; the `Finding` and the `PhaseFailure` are the record (doctrine 6: no new outcome vocabulary).
- return `ReviewPhaseResult(ran=False, failure=PhaseFailure(action=FailureAction.FAIL, error="Review infrastructure error: adversarial LLM budget exhausted", phase="review", check="adversarial_budget", signatures=["review:budget-exhausted"]))`

Keep the **advisory** behaviour unchanged: advisory mode with an exhausted budget still downgrades to `SKIP` with `phase_skipped` recorded, because advisory never blocks. Write that condition explicitly: `if review_mode == ReviewMode.ADVISORY and not self.adversarial_budget_ok(): <today's downgrade>`.

### `kstrl/pipeline.py`, `_phase_security`

Same change at `pipeline.py:2696-2701`: hard mode fails closed with `Finding.infrastructure_error(phase="security", ...)`, `check="adversarial_budget"`, signature `security:budget-exhausted`; advisory keeps the downgrade.

### The distiller (`pipeline.py:2974`)

Leave the distiller's budget skip as it is. The distiller is not a merge gate; skipping it under budget pressure is the correct priority.

### Docs

- `kstrl.toml.example`, the `max_adversarial_calls` comment (line 109): change to state that when the cap is reached, hard-mode review and security **halt the component** rather than proceeding unreviewed, and advisory mode skips with a recorded `phase_skipped`.
- `docs/runbook.md` section "Adversarial budget exhausted mid-run": rewrite the recovery to match (raise the cap or set the mode to advisory deliberately; the factory will no longer merge unreviewed in hard mode).
- CHANGELOG `### Changed`, with the word **Breaking** if any user could rely on the old behaviour: yes, an operator with a cap set and hard mode on previously got merged components past the cap; now they get halted components. Say so.

## Out of scope

- Changing what counts against the budget, or the cap's default.
- Any new event or enum.

## Tests to write

In `tests/test_review_gates.py` (it already constructs `FactoryConfig`, `run_factory`, and stubs the reviewer; follow its fixtures) or `tests/test_pipeline.py`:

1. `test_hard_mode_budget_exhausted_halts_component`: two components, `max_adversarial_calls = 1`, `review_mode = "hard"`, a reviewer stub that passes. First component completes. Second component ends `failed` with `failed_check == "adversarial_budget"`, its findings contain one `infrastructure_error` for phase `review`, and it is **not** marked completed.
2. `test_advisory_mode_budget_exhausted_still_skips`: same setup with `review_mode = "advisory"`. Second component completes; its findings contain a `phase_skipped` for `review` and no `infrastructure_error`.
3. `test_security_hard_mode_budget_exhausted_halts`: `[security] mode = "hard"`, review advisory, `max_adversarial_calls = 1` consumed by the first component's review. Second component ends failed with `failed_check == "adversarial_budget"` and an `infrastructure_error` for phase `security`.
4. `test_unbounded_default_unchanged`: `max_adversarial_calls = 0`, three components in hard mode all complete (proves the default path is untouched).
5. `test_budget_exhausted_signature_reaches_journal`: after test 1, the evolution journal entry for the second component has `review:budget-exhausted` in `failure_signatures` (read the JSONL the test run wrote).

## Acceptance (all must hold)

- [ ] Tests 1 through 5 pass.
- [ ] With a cap set below the run's need and hard mode on, zero components merge unreviewed (test 1 and 3 prove it for review and security).
- [ ] With the default cap, behaviour is byte-identical: every pre-existing pipeline and review-gate test passes unmodified.
- [ ] `kstrl.toml.example` and the runbook describe the new behaviour; the CHANGELOG entry is marked Breaking.
- [ ] `uv run ks config show` still renders `max_adversarial_calls` in the limits family (`kstrl/config_report.py:158`); nothing there changes.

0xfauzi:
## Summary

When `max_adversarial_calls` runs out, a hard-mode review or security phase used to downgrade itself to a skip and let the component merge on mechanical verification alone. In this repository's own journal the reviewer accounts for 14 of the 17 recorded failure signatures, so the phase shed under budget pressure was the one doing most of the catching. It now halts the component instead.

- `_budget_refusal` in `kstrl/pipeline.py` is the shared refusal for both phases: `comp.review_passed = False` where it applies, a `Finding.infrastructure_error` for the phase, and a `PhaseFailure(action=FAIL, check=ADVERSARIAL_BUDGET_CHECK, signatures=["adversarial_budget:<phase>"])`. FAIL and not RETRY_OR_FAIL, because the budget only shrinks and a retry would burn engineer iterations against the same exhausted cap.
- `ks serve` treats such a run as terminal. `_budget_halt_outcome` in `kstrl/serve.py` reads `Component.failed_check` and returns the EXISTING `BUDGET_HALT` verdict, whose `may_retry` is False. The timeout branch asks the same question through `_timeout_outcome`, so a halt followed by a hang is terminal too.
- No new event, no new verdict and no new enum member, per the issue's doctrine-6 constraint. The Finding, the `PhaseFailure` and the existing verdict are the whole record.
- Both mode splits sit inside the budget check rather than as a second condition beside it, so each stays closed over its mode enum, and each keys on ADVISORY so the arm it closes onto is the refusal rather than the downgrade that merges unreviewed.
- Advisory still records its `phase_skipped` and the component continues, EXCEPT under `setpoint_agreement = "block"`, where the R10.3 set-point gate fails a component whose reviewer never ran and whose story is still marked passes=true. The knowledge distiller is still skipped rather than halted, because it is not a merge gate.
- `_review_did_not_run`, `_security_budget_branch` and `_timeout_outcome` are extractions, not new behaviour: each was lifted out so an added branch did not push its caller past the cyclomatic ratchet. `classify_run` was split the same way, measured cyclomatic 15 to 13 and cognitive 36 to 18.

**Breaking.** Two qualifications from the issue, repeated here because they bound how much this matters: the old degrade was loud (it warned and recorded `phase_skipped`, so `len(findings) == 0` still meant every phase ran), and `max_adversarial_calls = 0` is the default and means unbounded, so this path only ever fired for an operator who had set a cap. An operator who has set one and runs hard mode will now get halted components where they previously got merged ones. That is the intended change. Three consequences that come with it: a halt cascades to every dependent component, each halt files an inbox item against `[inbox] open_item_cap`, and because the verdict is terminal `ks serve` poisons the queue item, so three under-budgeted runs in a row trip `[serve] max_consecutive_poison` (default 3) and stop the daemon admitting work.

A docs correction found while writing the operator arithmetic: `_phase_distill` consumes an adversarial call exactly like the two review phases do, and `KnowledgeConfig.enabled` defaults to `True`. So the "2 calls per component" figure the first docs pass shipped under-budgets a real run. Instrumenting `adversarial_budget_consume` over two components with hard review, hard security and knowledge on counted 6 consumes, not 4. The recommendation is `3 * components` with knowledge enabled, `2 * components` without, and all five sites that carried the number move together here. What is NOT a rule, and is now stated with its component count attached everywhere: at two components a cap of `2 * components` halts the SECOND at security, and at three components the same multiplier halts the THIRD at review, one phase earlier.

Closes #226

## The signature is not the one the issue specified, and that is deliberate

Issue #226 specifies `signatures=["review:budget-exhausted"]` and names an acceptance test asserting that exact string. This PR emits `adversarial_budget:<phase>` instead, and updates that test. The reason is measured.

`autonomy_replay.RunRecord.decisive` reads the SIGNATURE prefix: everything before the first colon is the check name, and its infrastructure prefixes are derived from `evolution._CATEGORY_BY_CHECK`. Under a `review:` prefix a run whose reviewer never ran scored `decisive=True`, a verdict about the factory's judgement, and moved the autonomy ladder. `factory._infra_casualty` asks the FINDING question about the same run, sees the `infrastructure_error` the refusal attaches, and scores it `decisive=False`. One reads the finding and is right; the other reads the name and is wrong. The comment above `_CATEGORY_BY_CHECK` states the rule this breaks in the repository's own words: a taxonomy that answers a question twice will eventually answer it two ways.

With the check-led spelling and `adversarial_budget` enrolled as infrastructure, both consumers read the same answer out of one table. Measured on this head: `adversarial_budget:review`, `adversarial_budget:security` and `adversarial_budget:setpoint` all give `infra_aborted=True`, `decisive=False`; the three old spellings all give `decisive=True`. The pre-existing sibling `review:setpoint-budget-exhausted` is swept into the same shape, because it is the same defect one branch over.

The sweep costs one thing and it is disclosed rather than buried. The `#339` census above `_CATEGORY_BY_CHECK` counts the signatures the two consumers answer differently about. It moves from six to seven. The two `_budget_refusal` signatures stop diverging, because that function attaches the infrastructure finding. `adversarial_budget:setpoint` starts diverging, because the set-point gate's only finding is the `phase_skipped` trace: the replay now calls it plumbing while the live side still calls it judgement. Before the sweep those two agreed by both being wrong. The new row is in the census and both halves are asserted by `test_the_setpoint_refusal_is_a_disclosed_divergence`, so the disclosure fails if it stops being true. Reconciling the live half means changing what `factory._infra_casualty` counts, which is `#332`'s file.

**Tested.**

- Seventeen new tests, counted from `git diff origin/main...HEAD -- tests/`. Eight in `tests/test_review_gates.py`: the halt itself (hard review, advisory unchanged, hard security, unbounded default, journal signature), the advisory security skip, the terminal `ks serve` verdict read off the manifest a real halted run wrote, and a consume counter that pins the 3-per-component arithmetic by counting rather than restating it. Seven in `tests/test_serve.py` for the classifier branch alone, including three mixed manifests and both directions of the timeout branch's manifest read. One in `tests/test_setpoint_agreement.py` asserting both halves of the divergence the sweep exposes, and one in `tests/test_phase_e.py` for the second component stopping at review.
- **Red first**: tests 1, 3 and 5 fail against unpatched `main`. Tests 2 and 4 pass unpatched and are controls, which is the point of including them. The `ks serve` tests fail on this branch's own earlier rounds, which the review measured for round 2 and which the timeout mutations below re-measure for round 3.
- **Mutation-tested**, 33 mutations, each run with `__pycache__` deleted and `PYTHONDONTWRITEBYTECODE=1` because bytecode staleness has produced a false pass in this repo before, each plant asserting its anchor matched exactly once so a silent no-op plant cannot be scored as a pass, each pytest bounded at 600s in its own process group with `os.killpg` on expiry so a hang is reported as a hang, and the outcome classified from the pytest exit code rather than from a scraped summary line. All 33 CAUGHT: none still green, none hung, none that failed to plant. Eighteen are the original set, four are the round-2 reviewer's, and eleven are new to round 3: four aimed at the timeout branch (never reads the manifest, halts whatever the manifest says in both the pre- and post-fix shapes, branch never consulted, and the hang dropped from the halt's evidence), two at the sibling's cause, three at the signature and its enrolment, and one at the mode split. Baseline for every run was 479 passed across `test_review_gates`, `test_phase_e`, `test_setpoint_agreement`, `test_pipeline`, `test_serve`, `test_context`, `test_check_name_enrolment`, `test_infrastructure_category_consumers` and `test_signature_spellings`; raw stdout per run is kept in the lane's `r3_mut_raw/`.
- The distiller arithmetic is measured, not reasoned: `adversarial_budget_consume` was patched with a counting spy over a real `run_factory`, and the three-component case was run to observe which component halts and at which phase rather than predicted from reading the code.
- Full suite 5609 passed, 37 skipped, 38 xfailed in 311.40s, on the tree merged with `origin/main` dee0c94, which a fresh fetch confirms is still main's head and already an ancestor of this branch. `mypy kstrl/ --strict`, `ruff check kstrl/ tests/`, `ruff format --check`, `scripts/gen_docs.py --check` and the whole pre-commit set green on each commit, including both complexity ratchets. The first round-3 attempt at the S1 fix was REJECTED by `complexipy` (`classify_run` 18 to 20, cyclomatic 13 to 14), which is why the timeout branch is a function rather than two inline branches.

**Assumed / not done.**

- No real-LLM factory run: every test stubs the reviewer, the security reviewer and the engineer, which is deliberate here (it makes the cap the only variable that can end a component) but means this is not evidence about live agent behaviour.
- The `ks serve` loop itself was not driven end to end. The classifier is tested directly, against a manifest the real factory wrote; the requeue decision that consumes the verdict is `Verdict.may_retry`, which is one line and already tested. The poison-streak consequence now documented in the runbook was verified by reading all four sites, not by running three consecutive under-budgeted runs.
- The N8 change of which mode arm each budget check closes onto is a posture change no test can exercise: `ReviewMode` and `SecurityMode` are closed by their `__post_init__` validators and SKIP returns above both checks, so there is no third member to reach the arm. It is verified only by the suite staying green.

## Measured handoff: the halt is paid for after the engineer loop

The review asked for a fourth branch in `factory._refused_before_launch` so a component certain to be refused is failed before its engineer loop runs. It is not in this PR, and here is the measurement so the next person does not have to redo it.

The refusal fires inside Phase 2, which runs after the engineer loop for that component. Measured with hard review and hard security, independent components, no dependencies:

- cap 1, 2 components: 2 engineer loops, 0 completed, 2 discarded.
- cap 1, 3 components: 3 engineer loops, 0 completed, 3 discarded.
- cap 1, 5 components: 5 engineer loops, 0 completed, 5 discarded.
- cap `2 * n`: 1 discarded at n=2 and n=3, 2 at n=5.

The same script run on 568bca4 gives the SAME engineer-loop counts, 2, 3 and 5, with every component completed. So this PR does not make an under-budgeted run cost a dollar more in engineer spend: the loops were already being paid, and what changes is that their output is discarded instead of merged with the reviewer shed. A pre-launch refusal would spend less than either commit, which makes it an improvement over both rather than a repair of a regression here. The reason for not doing it in this PR: the pre-launch gate necessarily attributes the failure to the scheduling phase, and the issue's own acceptance criterion, this PR's tests, the runbook symptom string and the CHANGELOG all name a Phase 2 or Phase 2.5 failure with an `infrastructure_error` for that phase. The cheaper shape, and the better one, is a startup warning when `max_adversarial_calls > 0 and cap < needed`, computed the way `factory.setpoint_gate_unreachable_warning` already warns about a control the operator believes is on: it changes no phase, no symptom string and no acceptance criterion, and the runbook this PR writes already states the arithmetic it would use.

## Checklist

## Provenance

No adversarial prompt body changed, so neither box applies. The `kstrl.toml` scaffold comment in `DEFAULT_KSTRL_TOML` is not a hashed template: `SCAFFOLDED_TEMPLATES` enrols only `prompt.md`, `understand_prompt.md` and `feature_understand_prompt.md`, and `tests/test_prompt_versions.py` enrols only `*_PROMPT` constants. No version bump and no ledger row are owed.


tests/test_review_gates.py
```diff
@@ -616,6 +719,125 @@ def test_budget_exhaustion_emits_finding_and_journal_event(
...
+        comp_b = run.component("comp-b")
+        assert "comp-b" not in run.result.completed
+        assert comp_b.status == "failed"
+        assert comp_b.failed_phase == "review"
+        assert comp_b.failed_check == "adversarial_budget"
+        infra = [f for f in comp_b.findings if f.is_infrastructure_error]
+        assert [f.phase for f in infra] == ["review"]
+        # FAIL, not RETRY_OR_FAIL: a retry cannot recover budget, so the
+        # component must not burn engineer iterations against the wall.
+        assert comp_b.retries == 0
```
0xfauzi, line 774: **The `FAIL` vs `RETRY_OR_FAIL` decision has no test. Measured: the mutation survives the whole suite.**

`_factory_config` (tests/test_review_gates.py:513) pins `max_retries=0`, and `_run_with_budget` never overrides it. With `max_retries=0`, `retry_or_fail` takes `0 < 0` -> False and falls straight into `fail()` with the same phase, check and signatures, leaving `comp.retries` at 0. So every assertion in this test - `status == "failed"`, `failed_phase`, `failed_check`, the infra finding, and this line - passes identically whether `_budget_refusal` returns `FAIL` or `RETRY_OR_FAIL`.

Measured, `__pycache__` cleared and `PYTHONDONTWRITEBYTECODE=1`, at `a61edc2` with `action=FailureAction.FAIL` changed to `RETRY_OR_FAIL` in `_budget_refusal`:

```
5569 passed, 37 skipped, 38 xfailed in 365.50s
```

Identical to the PR body's own baseline. The design decision stated three times in this diff (the `_budget_refusal` docstring, the CHANGELOG, the PR body) is the one thing not measured, and the comment two lines above claims it is. Under an operator's real `max_retries` (default 3) the mutant burns three full engineer loops per component against a deterministic wall, which is exactly the outcome the docstring says it prevents.

Fix: give `_run_with_budget` a `max_retries` above zero for at least the hard-mode halt test, then assert `retries == 0`. That makes the line a control instead of a tautology.

kstrl/pipeline.py
```diff
@@ -2910,6 +2910,147 @@ def _coverage_failure(
...
+        """Phase 2.5's exhausted-budget branch, both modes.
+
+        The mode split lives here rather than beside the budget check in
+        ``_phase_security`` so that check stays closed over
+        ``SecurityMode``: every non-hard mode reaching an exhausted
+        budget takes the recorded skip, which is what it did before
+        #226. ``SecurityConfig.__post_init__`` rejects any mode outside
+        skip|advisory|hard, and skip already returned above.
+        """
+        if sec_config.mode == SecurityMode.HARD.value:
```
0xfauzi: **The mode split this method exists for is untested. Measured: deleting it survives the whole suite.**

The docstring says the split lives here so the budget check "stays closed over `SecurityMode`: every non-hard mode reaching an exhausted budget takes the recorded skip". No test enters the else branch. `test_security_hard_mode_budget_exhausted_halts` is the only test that reaches `_security_budget_wall`, and it configures `SecurityMode.HARD`; grepping `tests/` for a Phase 2.5 budget skip finds nothing (`tests/test_review_gates.py:371` builds the Finding directly, it does not drive the phase).

Measured at `a61edc2`, `__pycache__` cleared, `PYTHONDONTWRITEBYTECODE=1`, with this line replaced by `if True:` so every security mode halts:

```
5569 passed, 37 skipped, 38 xfailed in 335.28s
```

So a future edit that makes an advisory security phase start halting components has no test standing in front of it, and advisory-never-blocks is the invariant the whole mode is for. Phase 2's equivalent split IS covered (`test_advisory_mode_budget_exhausted_still_skips` catches the same mutation on the review side), so this is an asymmetry, not a shared gap.

Fix: one test with `security_config=SecurityConfig(mode="advisory")` and the cap spent, asserting the component completes and carries a `phase_skipped` finding for `security`.

kstrl/pipeline.py
```diff
@@ -3021,6 +3162,32 @@ def _phase_review(
         if review_mode == ReviewMode.SKIP:
             review_skip_reason = "review disabled (mode=skip)"
         elif not self.adversarial_budget_ok():
```
0xfauzi, line 3202: **The wall fires after the engineer has been paid for, not before. `_refused_before_launch` is the right depth and this diff does not use it.**

Once the cap is spent and `review_mode == "hard"`, no remaining component can pass Phase 2. But the scheduler does not know that: `_run_scheduling_pass` calls `_refused_before_launch` (kstrl/factory.py:2528), which does not consult the adversarial budget, then launches a full engineer loop for each remaining component - up to `max_iterations` LLM calls - and only then this line refuses it.

That is the defect class `_refused_before_launch` was written for, and its own comment measures it:

> #294: the LAST point before spend ... Without this gate such a component runs a full engineer loop - up to max_iterations LLM calls - and only then does Phase 1 refuse it. Measured ... roughly 4.83 dollars and 14 minutes on the table per reset.

The economics invert: `max_adversarial_calls` is a cost knob, and after this change exceeding it costs MORE than before, because the engineer spend is now thrown away instead of merged. The PR body applies exactly this reasoning to retries ("a retry would burn engineer iterations against the same deterministic wall") and then does not apply it to the scheduler, where the same wall is equally deterministic and the same spend is at stake.

Two places to put the fix, both cheaper than the docs: a fourth refusal in `_refused_before_launch` when `review_mode`/`security mode` is hard and `adversarial_budget_ok()` is false, and a preflight refusal, since `components * (1 + hard_security + knowledge)` against the cap is fully known before the run starts. The five doc edits in this PR hand that arithmetic to the operator instead of computing it.

kstrl/pipeline.py
```diff
@@ -2910,6 +2910,147 @@ def _coverage_failure(
...
+        self._add_findings(
+            comp,
+            [Finding.infrastructure_error(phase=phase, explanation=reason)],
+        )
+        return PhaseFailure(
+            action=FailureAction.FAIL,
+            error=error,
+            phase=phase,
+            check="adversarial_budget",
+            signatures=[f"{phase}:budget-exhausted"],
```
0xfauzi: **`ks serve` will classify this halt as retryable infrastructure and re-run the same work against the same cap.**

`_budget_refusal` records an `infrastructure_error` Finding. `serve.classify_run` ends at:

```python
return Outcome(
    Verdict.RETRY_INFRA,
    "every failed component carried an infrastructure_error finding: " + ...
)
```

A budget-halted component has findings (so it is not `unevidenced`) and is an `_infra_casualty` (so it is not `judged`), which lands it in exactly that branch. The factory exits 1 (`factory.py:994`), so the branch is reachable.

The terminal verdict for a configured ceiling is `Verdict.BUDGET_HALT`, and `budget_halt_reason` (serve.py:880) derives it from the typed `ev.BudgetExceeded` event only. Doctrine 6 forbade a new event, so this halt is invisible to it. Its own reason string is the case:

> retrying would re-run the same work against the same limit at a higher cost

Sometimes the retry does help, because a fresh run resets `_adversarial_calls` and only the unfinished components remain. Sometimes it cannot: `max_adversarial_calls = 1` with hard security halts the first component at Phase 2.5 on every run forever, and the daemon pays a full engineer loop each time until the poison breaker trips.

Worth deciding explicitly rather than by omission: either emit the halt somewhere `budget_halt_reason` can see it, or add an arm to `classify_run` that recognises `failed_check == "adversarial_budget"` and returns `BUDGET_HALT` when the cap cannot cover the remaining components.

docs/runbook.md
```diff
@@ -168,11 +168,20 @@ To stop it failing components, set `[divergence] mode = "advisory"` (the default
...
+Phase 2.5 FAILED for <comp_id>: Security review infrastructure error: adversarial LLM budget (N) exhausted before the phase ran; hard mode refuses to merge unreviewed
+```
+
+It is recorded as `failed_check = adversarial_budget` and journalled as `review:budget-exhausted` or `security:budget-exhausted`, with an `infrastructure_error` finding for the phase. It does not retry: the budget only shrinks, so a retry would burn engineer iterations against the same wall.
+
+**Diagnose**: `FactoryConfig.max_adversarial_calls` is set and the count of review + security + distillation calls has hit the cap. Every one of those phases that runs costs one call per component, and the knowledge distiller spends from the same cap even though it gates nothing, so with hard review, hard security and `[knowledge] enabled = true` (the default) a run costs `3 * components`. Measured on a two-component run: a cap of `2 * components` completes the first component and halts the second at security with `failed_check = adversarial_budget`, because component A's distiller spent the call component B's security needed. Budget `3 * components`, or `2 * components` with `[knowledge] enabled = false`.
```
0xfauzi: **The priority policy is stated but not implemented: the non-gate still spends the gates' budget, and first.**

The design note says the policy is "the merge gates halt, the non-gate sheds". What the code does is let the non-gate SPEND before the merge gates get their turn: `_phase_distill` calls `adversarial_budget_consume()` (pipeline.py:3841) from the same counter, and component A's distiller runs before component B's review. This paragraph is the proof, measured in this PR: "component A's distiller spent the call component B's security needed".

That is a priority inversion, and the response here is to document the arithmetic across five surfaces rather than remove it. A shed that happens after the resource is consumed is not a shed. Cheaper and more general fixes, both of which delete a number the operator has to carry:

- reserve the merge gates' calls: check `adversarial_budget_remaining()` against what the remaining components' gates need before letting the distiller consume, or
- stop charging the distiller to `max_adversarial_calls` at all, since by this PR's own argument it gates nothing and it already has `[knowledge] enabled` as its off switch.

Either one makes `2 * components` correct again and removes the case this section has to explain.

kstrl/pipeline.py
```diff
@@ -3021,6 +3162,32 @@ def _phase_review(
...
+            # Finding and the PhaseFailure are the record.
+            #
+            # The mode split is INSIDE the budget check, not a second
+            # condition beside it, so the check stays closed over
+            # ReviewMode: a mode added later falls into the downgrade
+            # below (today's behaviour) instead of falling past both
+            # branches and out of the budget check entirely. Phase 2.5
+            # has the same shape for the same reason.
+            if review_mode == ReviewMode.HARD:
+                comp.review_passed = False
```
0xfauzi, line 3223: **`review_passed = False` says the reviewer rejected the component, for a reviewer that never ran, and nothing reads it.**

The advisory path for the identical condition sets `None` (`_review_did_not_run`, pipeline.py:2970), which is the honest value: not measured. `False` is the value a real hard-mode rejection writes at pipeline.py:3287, so the manifest's `reviewPassed: false` now means two different things and a reader cannot tell them apart from that field.

It is also not load-bearing. The only consumer is `knowledge.distill_facts`, which tests `review_passed is not True`, so `False` and `None` behave identically there, and the component fails before the distiller anyway. Measured at `a61edc2` with the line deleted:

```
tests/test_review_gates.py tests/test_phase_e.py tests/test_setpoint_agreement.py tests/test_pipeline.py
186 passed in 4.25s
```

Nothing observes it. Given this repo's own rule that a gate must never record what it could not measure as a value, `None` is the correct write here, or drop the line.

kstrl/pipeline.py
```diff
@@ -2910,6 +2910,147 @@ def _coverage_failure(
             signatures=[f"{phase}:coverage-unverified:{reason}"],
         )
 
+    def _budget_refusal(
+        self,
+        comp: Component,
+        *,
+        phase: str,
+        banner: str,
```
0xfauzi, line 2929: **The comment that documents this counter still describes the old behaviour, three lines above the counter itself.**

pipeline.py:534-537:

```python
# E4: adversarial-call counter shared across review / security /
# knowledge phases. When max_adversarial_calls is 0 the budget is
# unbounded; otherwise the LLM phase is skipped once the budget
# is exhausted, with an informational log line.
```

After this PR the phase is not skipped and there is no informational log line for a hard-mode review or security phase: it is `ui.err` and a halted component. This diff moved the five sites carrying the NUMBER (README, gen_docs, kstrl.toml.example, init_cmd, runbook) and missed the one carrying the BEHAVIOUR, which is the one a reader of `adversarial_budget_ok()` will actually find.

scripts/gen_docs.py
```diff
@@ -494,7 +494,17 @@ def identity_keys(cfg_cls: Any, names: list[str]) -> dict[str, str]:
...
+    (
+        "factory",
+        "max_adversarial_calls",
+    ): "cap on review+security+distill LLM calls; 0 = unbounded. "
+    "At the cap a hard-mode review or security phase HALTS the component "
+    "(failed_check = adversarial_budget) instead of merging it on mechanical "
+    "checks alone; an advisory phase records a phase_skipped and continues; the "
+    "distiller is always skipped rather than halted. Every phase that runs costs one call, and the "
+    "distiller spends from the same cap although it gates nothing, so hard "
+    "review plus hard security plus knowledge costs 3 per component - a cap of "
```
0xfauzi: **"a cap of 2 per component halts the second one at security" is a two-component measurement stated as a general rule, and it is false for three or more.**

Trace `adversarial_budget_ok` (`calls < cap`) with three components at `cap = 6`:

```
A: review 1, security 2, distill 3
B: review 4, security 5, distill 6
C: review  -> 6 < 6 is False -> HALT at REVIEW
```

The second component completes; the THIRD halts, and at review, not security. Four components at `cap = 8` gives the same shape one component along. So both halves of the sentence, which component and which phase, hold only for n = 2.

docs/runbook.md:182 and kstrl.toml.example both qualify it correctly ("Measured on a two-component run" / "MEASURED on two components"). This string, the README line it generates, and kstrl/init_cmd.py:459 drop the qualifier, so an operator with three components looks for a Phase 2.5 failure on component B and finds a Phase 2 failure on component C.

Either carry the qualifier into all three, or state the rule that does generalise: the run halts at the first phase call after `3 * components` minus the distiller calls the cap could not cover.

kstrl/pipeline.py
```diff
@@ -3021,6 +3162,32 @@ def _phase_review(
...
+            #
+            # The mode split is INSIDE the budget check, not a second
+            # condition beside it, so the check stays closed over
+            # ReviewMode: a mode added later falls into the downgrade
+            # below (today's behaviour) instead of falling past both
+            # branches and out of the budget check entirely. Phase 2.5
+            # has the same shape for the same reason.
+            if review_mode == ReviewMode.HARD:
+                comp.review_passed = False
+                return ReviewPhaseResult(
```
0xfauzi, line 3224: **`ran=False` with `skip_reason=None` breaks the dataclass's stated invariant, and `skip_reason` has no reader.**

`ReviewPhaseResult`'s docstring (pipeline.py:184) says "``ran=False`` records the skip reason". This return is the first `ran=False` that records nothing, and `_security_budget_wall`'s hard branch does the same. A reader who trusts the docstring will treat `ran=False and skip_reason is None` as impossible.

Worth noting while the field is being moved: grepping `kstrl/` for `.skip_reason` finds exactly one reader, `distill.skip_reason` at pipeline.py:2292. `ReviewPhaseResult.skip_reason` and `SecurityPhaseResult.skip_reason` are written by five sites and read by none, and `_review_did_not_run` now threads it through a new parameter to keep doing that. Either delete the two fields or fix the docstring to say what `ran=False` now means.

docs/runbook.md
```diff
@@ -168,11 +168,20 @@ To stop it failing components, set `[divergence] mode = "advisory"` (the default
...
-**Diagnose**: `FactoryConfig.max_adversarial_calls` is set and the count of review + security + distillation calls has hit the cap.
+**Symptom, hard mode** (R10.5): the component halts instead.
 
-**Resolve**: increase the cap, or accept that later components run without adversarial phases. The mechanical pipeline (Phase 1) still gates them.
+```
+Phase 2 FAILED for <comp_id>: Review infrastructure error: adversarial LLM budget (N) exhausted before the phase ran; hard mode refuses to merge unreviewed
+Phase 2.5 FAILED for <comp_id>: Security review infrastructure error: adversarial LLM budget (N) exhausted before the phase ran; hard mode refuses to merge unreviewed
+```
+
+It is recorded as `failed_check = adversarial_budget` and journalled as `review:budget-exhausted` or `security:budget-exhausted`, with an `infrastructure_error` finding for the phase. It does not retry: the budget only shrinks, so a retry would burn engineer iterations against the same wall.
```
0xfauzi: **Two other docs still record the superseded rule, and they are the requirement records this halt replaces.**

- docs/remediation-sessions.md:326: "Skipped and budget-exhausted phases emit a synthetic Finding (category "phase_skipped", is_infrastructure_error stays False ...) and a journal event"
- docs/remediation-roadmap.md:226: the same R1.2 bullet

After this PR a budget-exhausted phase in hard mode emits an `infrastructure_error` Finding and no `phase_skipped`, which is the exact opposite of both sentences on the exact field they name. The runbook, the design note and the tracker row all moved in this diff; these two did not, and they are the ones a reader consults to learn what a `phase_skipped` finding is supposed to cover.

kstrl/pipeline.py
```diff
@@ -2910,6 +2921,150 @@ def _coverage_failure(
...
+        self._add_findings(
+            comp,
+            [Finding.infrastructure_error(phase=phase, explanation=reason)],
+        )
+        return PhaseFailure(
+            action=FailureAction.FAIL,
+            error=error,
+            phase=phase,
+            check=ADVERSARIAL_BUDGET_CHECK,
+            signatures=[f"{phase}:budget-exhausted"],
```
0xfauzi: **A budget halt is scored as a decisive judgement by the autonomy replay.**

The signature prefix here is the phase name, so the journal writes `review:budget-exhausted` / `security:budget-exhausted`. `evolution._CATEGORY_BY_CHECK` maps check `review` -> category `review` and `security` -> `security`, so neither lands in `INFRASTRUCTURE_CHECKS`, so neither reaches `autonomy_replay.INFRA_FAILURE_PREFIXES`.

Concretely: cap set, hard review, comp-b halts. `evolution.record_run` picks the dominant signature and writes `common_failure = "review:budget-exhausted"` into `experiments.tsv`. `RunRecord.infra_aborted` does `common_failure.startswith(INFRA_FAILURE_PREFIXES)` -> False, so `RunRecord.decisive` -> True, and a run in which the reviewer never ran is counted as "a verdict about the factory's judgement". That is exactly the defect #315 fixed for `provisioning:`, `token_budget:`, `pr:` and `aborted:` - "a verdict about an agent that could not have prevented any of them".

Two further consequences of the same choice:

- `evolution._summarize_findings` counts the refusal under `by_phase["review"]` and `infrastructure_errors`, so the evolve screen shows reviewer-phase trouble for a phase that never started.
- The `#339` census comment in `kstrl/evolution.py` ("Six signatures, in both directions", second bullet: "a finding is attached, but `review`/`security` are not infrastructure prefixes, so the replay counts them as evidence") is now stale at eight. Per the repo's own census rule, that number moves with the diff or it stops being a control.

The fix is a distinct check name rather than the phase name - e.g. `signatures=[f"adversarial_budget:{phase}"]` with `"adversarial_budget": "infrastructure"` enrolled in `_CATEGORY_BY_CHECK` - which reaches both consumers in one edit, as that table's docstring says it is designed to.

kstrl/pipeline.py
```diff
@@ -2910,6 +2921,150 @@ def _coverage_failure(
             signatures=[f"{phase}:coverage-unverified:{reason}"],
         )
 
+    def _budget_refusal(
```
0xfauzi, line 2924: **A cap that is certain to halt the run is not refused before the run starts, and the repo already has the surface for it.**

Everything needed to predict this halt is known at run start: `max_adversarial_calls`, `review_mode`, `security_config.mode`, `knowledge_config.enabled`, and `len(manifest.components)`. The runbook this PR writes states the rule exactly - `3 * components`, or `2 * components` with knowledge off - so `max_adversarial_calls > 0 and cap < needed` is a computable predicate, not a judgement.

With no such check, an operator who sets `max_adversarial_calls = 4` on a 5-component spec in hard mode now: launches 5 engineer loops, merges at most one component, halts the rest, cascades the dependents, files up to 5 inbox items, and gets the queue item poisoned. Before this PR the same config merged everything with the reviewer shed - loudly bad, but cheap. The new failure mode is expensive and arrives only after the spend.

`kstrl/factory.py` already has the pattern for exactly this shape: `setpoint_gate_unreachable_warning` (line 548) and `merge_gate_unreachable_warning` warn at startup when "a governance control silently fail[s] open ... because the operator believes it is on". This is the mirror case - a control that is on and guaranteed to fire - and it deserves the same startup warning, or a `_report_preflight` refusal, before the first engineer call.

kstrl/pipeline.py
```diff
@@ -3021,6 +3176,36 @@ def _phase_review(
...
+            # and no event is invented for it (doctrine 6): the
+            # Finding and the PhaseFailure are the record.
+            #
+            # The mode split is INSIDE the budget check, not a second
+            # condition beside it, so the check stays closed over
+            # ReviewMode: a mode added later falls into the downgrade
+            # below (today's behaviour) instead of falling past both
+            # branches and out of the budget check entirely. Phase 2.5
+            # has the same shape for the same reason.
+            if review_mode == ReviewMode.HARD:
```
0xfauzi: **The mode split is closed in the unsafe direction: any mode that is not `HARD` merges unreviewed.**

The comment above argues this keeps the check "closed over ReviewMode: a mode added later falls into the downgrade below (today's behaviour) instead of falling past both branches". It is closed, but the default it is closed onto is the fail-open one. Add `ReviewMode.STRICT` (or any mode stricter than hard) tomorrow and it silently takes the downgrade: the component merges on mechanical checks alone, which is precisely the behaviour this PR exists to remove, and nothing goes red.

Only one mode wants the skip, and it can be named: `if review_mode == ReviewMode.ADVISORY: <downgrade> else: <refuse>`. Then an unrecognised mode refuses, which is the direction this repo's own doctrine picks - "If a clearing guard cannot PROVE a site is compliant, it must flag", and the halt-over-heroics rule in CLAUDE.md.

`_security_budget_branch` at line 3045 has the identical inversion (`if sec_config.mode == SecurityMode.HARD.value: refuse else: skip`) and wants the same flip.

kstrl/serve.py
```diff
@@ -1071,6 +1081,64 @@ def classify_run(
...
+    # every budget-halted component carries an infrastructure_error
+    # finding and would otherwise reach the RETRY_INFRA return at the
+    # bottom of _merits_outcome.
+    budget = _budget_halt_outcome(failed, run.returncode)
+    if budget is not None:
+        return budget
+    return _merits_outcome(failed, run.returncode)
+
+
+def _budget_halt_outcome(failed: Sequence[Any], returncode: int) -> Outcome | None:
```
0xfauzi: **`Finding.infrastructure_error` is the wrong marker for a deliberate refusal, and this function is the patch for that.**

The docstring here is candid about the mechanism: "every budget-halted component carries an infrastructure_error finding and would otherwise reach the RETRY_INFRA return". That is the tell. `Finding.infrastructure_error` is documented in `kstrl/findings.py` as "the given role failed to execute (timeout, parse failure, agent crash)", and `Finding.is_phase_skip` right below it is documented as the marker for deliberate non-execution: "Deliberately NOT an infrastructure error: nothing broke, the phase was skipped on purpose." A budget refusal is the second kind - nothing broke - filed as the first.

The cost is not just this branch. `is_infrastructure_error` has three other live consumers, and none of them is patched here:

- `factory._infra_casualty` (line 2774) drops budget-halted components out of `judged_failures`, which feeds `decisive` in the L1/L2 autonomy accounting.
- `evolution._summarize_findings` increments `infrastructure_errors` and `by_phase["review"]`.
- `serve._infra_casualty` (line 936), which is only saved by the new branch running first.

The deeper fix is a marker that says what happened - a refusal category, or `is_phase_skip` plus a severity the hard-mode gate honours - so that every consumer of "was this infrastructure?" gets the right answer from one place, instead of one consumer getting a special case and three keeping the wrong one.

docs/runbook.md
```diff
@@ -168,11 +168,24 @@ To stop it failing components, set `[divergence] mode = "advisory"` (the default
...
+
+**Diagnose**: `FactoryConfig.max_adversarial_calls` is set and the count of review + security + distillation calls has hit the cap. Every one of those phases that runs costs one call per component, and the knowledge distiller spends from the same cap even though it gates nothing, so with hard review, hard security and `[knowledge] enabled = true` (the default) a run costs `3 * components`. Budget `3 * components`, or `2 * components` with `[knowledge] enabled = false`.
+
+Anything less halts SOME component, but which one and at which phase depends on the component count, so do not read the two-component case as a rule. Measured: at two components and a cap of 4 (`2 * components`), comp-a completes and comp-b halts at security, because comp-a's distiller spent the call comp-b's security needed. At three components and a cap of 6 (`2 * components` again), comp-a and comp-b complete and comp-c halts at REVIEW, one phase earlier.
```
0xfauzi, line 186: **The refusal is paid for after the engineer loop, and `_refused_before_launch` is the existing gate for exactly this.**

The PR body measures this and hands it off: cap 1 / 5 components = 5 engineer loops, 0 completed, 5 discarded. That inverts the knob: a cap set to bound LLM spend now causes strictly more spend than not setting it, because the engineer loop (up to `max_iterations` calls per component) is far more expensive than the adversarial call the cap was counting.

The stated reason for deferring is that a pre-launch gate would make the documented Phase 2 / Phase 2.5 symptom unreachable. That is a solvable framing, not a blocker: `_refused_before_launch` calls `pipeline.fail(...)` with a caller-chosen `phase` and `check`, so the pre-launch branch can emit the same `phase="review"`, `check=ADVERSARIAL_BUDGET_CHECK`, `signatures=["review:budget-exhausted"]` and the same `infrastructure_error` finding that `_budget_refusal` emits, keeping every published symptom string intact.

The precedent is in that function's own comments: #294 added the third branch for the identical arithmetic - "Without this gate such a component runs a full engineer loop - up to max_iterations LLM calls - and only then does Phase 1 refuse it", with dollars attached. The docstring's objection to a fourth branch is complexity, and the function was extracted precisely so a fourth branch is cheap.

Not blocking the PR, but the handoff understates it: this is the same defect class the function already exists to close.

kstrl/pipeline.py
```diff
@@ -3021,6 +3176,36 @@ def _phase_review(
...
+                comp.review_passed = False
```
0xfauzi, line 3223: **`review_passed = False` records a rejection that never happened.**

`review_passed` is the reviewer's verdict, persisted to `manifest.json` as `reviewPassed` and read back by `knowledge.knowledge_confidence_ceiling`. `False` means "the reviewer looked and rejected". Here the reviewer never started.

The test's justification - "It is False rather than None because the component is failing" - is a property of the component, not of the reviewer, and the field is the reviewer's. The advisory path four lines down writes `None` for exactly this reason, and it is the same "ran clean" vs "never ran" distinction R1.2, `is_phase_skip` and `Finding.infrastructure_error` all exist to preserve. `None` is the accurate value and no consumer distinguishes it from `False` today (`knowledge_confidence_ceiling` tests `review_passed is not True`), so this costs nothing to correct and stops a wrong fact being written to the manifest a dashboard or a later run reads.

docs/runbook.md
```diff
@@ -168,11 +168,24 @@ To stop it failing components, set `[divergence] mode = "advisory"` (the default
...
+Phase 2.5 FAILED for <comp_id>: Security review infrastructure error: adversarial LLM budget (N) exhausted before the phase ran; hard mode refuses to merge unreviewed
+```
+
+It is recorded as `failed_check = adversarial_budget` and journalled as `review:budget-exhausted` or `security:budget-exhausted`, with an `infrastructure_error` finding for the phase. It does not retry: the budget only shrinks, so a retry would burn engineer iterations against the same cap. `ks serve` reads that `failed_check` and classifies the whole run as `budget_halt`, which is terminal: the queue item is not requeued, because the cap starts again at zero on the next attempt and the run would stop at the same component.
+
+Two consequences worth knowing before you set a cap. Every component that depends on a halted one is SKIPPED (`cascade_skip`), so one halt near the root of the DAG can end most of the run. And each halted component files an inbox item keyed `halted:<comp>:<phase>:adversarial_budget`; `[inbox] open_item_cap` (default 50) stops `ks serve` admitting new queue work once the inbox is that full, so a cap far below the run's need can quietly reach it.
```
0xfauzi: **A third consequence is missing here: the serve poison-streak breaker.**

This paragraph documents two consequences (cascade skip, `open_item_cap`). There is a third and it is the one that stops the daemon rather than slowing it.

`BUDGET_HALT` has `may_retry = False`, so `_finish_run` takes the `queue.poison(...)` branch and calls `ledger.record_terminal(poisoned=True)`, which increments `consecutive_poison`. `ServeConfig.max_consecutive_poison` defaults to **3**. So three consecutive under-budgeted runs trip the breaker and `ks serve` stops admitting work with "something systemic is wrong" - which is true, but the operator's inbox will say `budget_halt` three times and nothing will say the daemon has stopped because of it.

Worth a sentence here beside the other two, since a cap far below the run's need reaches this faster than it reaches the 50-item inbox cap.

kstrl/serve.py
```diff
@@ -69,6 +69,7 @@
 from typing import Any, Final, Protocol
 
 from kstrl.agents.base import ARCHITECT_COMPONENT, ARCHITECT_ROLE
+from kstrl.manifest import ADVERSARIAL_BUDGET_CHECK
```
0xfauzi: **This top-level import makes the deferred `Manifest` import inside `classify_run` (line 1055) dead weight.**

`from kstrl.manifest import Manifest` sits inside `classify_run` rather than at module scope, which reads as a deliberate deferral (import cost, or a cycle someone once hit). Importing `ADVERSARIAL_BUDGET_CHECK` from the same module at line 72 fully executes `kstrl.manifest` at `kstrl.serve` import time, so whatever the deferral bought is now gone while the function-local import remains to suggest it is still in force.

Pick one: either the deferral matters and the constant should be imported lazily too (or re-homed somewhere `serve` already imports), or it never mattered and the local import at 1055 should be hoisted in the same diff. Leaving both is a control that has stopped being true with nothing failing.

kstrl/serve.py
```diff
@@ -1071,6 +1081,64 @@ def classify_run(
...
+            "budget_halted": halted,
+            "other_failures": others,
+        },
+    )
+
+
+def _merits_outcome(failed: Sequence[Any], returncode: int) -> Outcome:
+    """Whether named failures were the spec's fault, unproven, or infra.
+
+    Split out of ``classify_run`` unchanged by #226 round 2, which added
```
0xfauzi: **"unchanged" is not accurate, and it is standing in for a test.**

Two things moved in the extraction:

1. The `SPEC_FAILURE` evidence is now `{**evidence, "judged_failures": judged}`, so `judged_failures` moves from second key to last. Dict order is what lands in the inbox item's JSON.
2. `sibling_note` was guarded by `if unevidenced:` and is now guarded by `if detail else ""`. Equivalent today only because `f"{comp.id}: {comp.error or 'no error recorded'}"` can never render empty - a real invariant, but an unstated one, and the guard now depends on it.

Neither is a live bug. The problem is that "lifted out unchanged" is the reason given for not testing the extracted function, and here it is the claim rather than a test carrying the weight. Either restore the literal shape, or add an assertion on the mixed `SPEC_FAILURE` evidence key set so a future edit to `_unevidenced_evidence` cannot quietly change what an operator reads.

kstrl/init_cmd.py
```diff
@@ -451,6 +451,9 @@
 # review_mode = "hard"             # hard | advisory | skip
 # merge_timeout = 300.0            # seconds to wait for PR merge confirmation
 # max_adversarial_calls = 0        # 0 = unbounded; caps review+security+distill LLM calls per run
+# At the cap, hard-mode review and security HALT the component with
+# failed_check=adversarial_budget; advisory skips, and can still fail it
```
0xfauzi: **This prose line survives the scaffold test by accident.**

`tests/test_config_control_plane.py::_uncomment_scaffold` uncomments any scaffold line that starts `# `, contains `" = "`, and whose first character after the `# ` is not uppercase. Line 455 is the only one of the three new lines that starts lowercase (`failed_check=...`), so it is separated from being uncommented into `[factory]` by one thing: the absence of spaces around its `=`.

Write it the way the rest of the repo writes that field - `# failed_check = adversarial_budget` - and `_uncomment_scaffold` turns it into a bare `failed_check = adversarial_budget` key inside `[factory]`, which `test_scaffold_keys_are_consumed_by_a_loader` then fails as an unconsumed scaffold key.

Cheapest fix: capitalise the continuation lines (`# Failed_check ...` reads badly, so reflow so each line begins with a capital), or indent them as `#   ` continuations the way `kstrl.toml.example` does for its multi-line notes - the `stripped.startswith("# ")` test still matches after `.strip()`, so indentation alone does not save it; capitalisation does.

kstrl/pipeline.py
```diff
@@ -2910,6 +2921,150 @@ def _coverage_failure(
             signatures=[f"{phase}:coverage-unverified:{reason}"],
         )
 
+    def _budget_refusal(
+        self,
+        comp: Component,
+        *,
+        phase: str,
+        banner: str,
```
0xfauzi, line 2929: **`banner` and `role` are derivable from `phase`, and passing all three lets them disagree.**

There are exactly two call sites and each passes a fixed triple: `("review", "Phase 2", "Review")` and `("security", "Phase 2.5", "Security review")`. Three parallel parameters means nothing stops a third caller passing `phase="security", banner="Phase 2"`, which would put the wrong banner in the operator-facing line and the runbook symptom string the tests pin.

A module-level `dict[str, tuple[str, str]]` keyed on `phase` collapses this to one argument and makes the pairing a fact rather than a convention - and gives the pipeline the phase-to-banner map it currently spells out as string literals at every `ui.info`/`ui.err` site.

Issue #151, opened 2026-07-21
R8.4: Factory health trending - EWMA/SPC over run metrics, demotion triggers


0xfauzi:
> **Stabilisation queue: P3, rank 14 of 21** (set 2026-09-09; the full ordered list and the coordinator's rules are in #369). Makes the health-breach demotion seam real. On hold until the owner lifts the P3 hold in #369.

## Why

The autonomy ladder's demotion triggers need trend detection over run metrics, and the operator needs an evidence surface (`ks health why-demoted`). The evolution journal and `experiments.tsv` already record the raw data; nothing trends it.

## Approach: hand-roll detection (stdlib), DuckDB as optional query layer

No maintained lightweight SPC library exists (pyspc is stale, river is a heavy import for this). EWMA (lambda ~0.2) plus two Western Electric rules (1 point beyond 3 sigma; 2-of-3 beyond 2 sigma same side) is ~30 lines and designed exactly for small persistent shifts. DuckDB reads JSONL natively (`read_json_auto`) with a single self-contained wheel - right for exploratory `ks health query "<sql>"`, but it stays an optional extra: autonomy safety logic must not depend on an optional dependency.

## Key design points

- Metrics: retry rate, cost per merged component, `infrastructure_error` rate, calibration detection deltas, human-edit rate.
- Control limits computed from the repo's own baseline period, never fixed constants; minimum n >= 8 decisive runs before any automatic transition; hysteresis via the ladder's cool-down.
- Calibration detection rate stays in the demotion basis deliberately: it is adversarial ground truth that cannot be improved by weakening verification (anti-Goodhart).
- `ks health` renders per-metric EWMA vs control limits with sparklines; `ks health why-demoted` replays triggering evidence.

## Done when

`ks health` renders trends from real journal data; trigger rules are unit-tested on synthetic drift series; a replay over historical data is documented (thresholds measured, not assumed); the duckdb optional extra is wired for the query subcommand.

## Dependencies

None hard. Feeds R8.2's demotion triggers.

Full plan: `docs/dark-factory-roadmap.md` (R8.4).

0xfauzi:
R10 review (2026-08-24, tracking issue #235): this item stays and is the missing sensor for the ladder's `HEALTH_BREACH` trigger, which today has no emitter.

Two changes to how it should land, from the control-loop design (`docs/control-loop-design.md` section 5.11, PR #221):

[...]

PR #350, opened 2026-09-06
feat(autonomy): wire the calibration-regression demotion trigger and the health-breach seam (#232)

### LINKED ISSUE #232: R10.11: Wire the dead demotion triggers: calibration regression now, health-breach seam for R8.4
Part of #235 (R10: Control Loop tracking issue). Item R10.11.

Design section: 5.11 (and gap 3.6). Sizing: M (one PR for the calibration trigger and the seam; the health trigger's emitter is #151 and stays there). Dependencies: none for the calibration half; **#151 (R8.4)** for the health half.

## Why

The autonomy ladder is gain scheduling: a controller whose aggressiveness is selected by operating regime. Gain scheduling with unwired inputs schedules on partial information.

`DemotionTrigger` (`kstrl/autonomy.py:90-101`) declares five causes: `POLICY_VIOLATION`, `CALIBRATION_REGRESSION`, `HEALTH_BREACH`, `HUMAN_REJECTED_AUTO_MERGE`, `MANUAL`. Only `POLICY_VIOLATION` fires automatically (`kstrl/factory.py:2042-2043`). `CALIBRATION_REGRESSION` is referenced only by the offline replay tool (`kstrl/autonomy_replay.py:213-214`), which by design "reports rather than decides". `HEALTH_BREACH` has no emitter because the module that would compute it does not exist (`ls kstrl/ | grep -iE "health|trend|ewma"` is empty; that is #151, pending). `grep -rn CALIBRATION_DRIFT kstrl/*.py` finds the inbox kind (`kstrl/inbox.py:52`, `ItemKind.CALIBRATION_DRIFT`) and no emitter.

So a factory that is quietly getting worse without violating policy keeps its autonomy, and a calibration regression that a human has already measured changes nothing.

Every threshold in the ladder is labelled an unmeasured placeholder (`autonomy.py:38-42`). That is honest, and it is why both triggers ship **advisory first**: an inbox item, with demotion behind an explicit switch.

### 1. One demotion applier, shared

Extract the demotion-apply block from `_record_autonomy_outcome` (`kstrl/factory.py:2040-2072`: `state.demote(...)`, the at-floor branch, `commit_transition(...)`, and the `Inbox.add(ItemKind.DEMOTION_NOTICE, ...)` call) into a function in `kstrl/autonomy.py`:

```python
def apply_demotion(
    root_dir: Path, trigger: DemotionTrigger, reason: str, *,
    evidence: dict[str, Any], run_id: str, ui: UI, bus: EventBus | None = None,
) -> Transition | None:
    """Demote one level for `trigger`, persist, emit, and open the inbox
    notice. Returns None when already at the floor (violation still counted).
    """
```

`_record_autonomy_outcome` calls it for `POLICY_VIOLATION` with no behaviour change. The existing autonomy tests are the proof; none may be modified.

### 2. Calibration regression: emitter

`compare_baselines` (`kstrl/calibration.py:460`) already returns a `Comparison` with `.passed`. The CLI entry `python -m kstrl.calibration compare` (`calibration.py:709-734`) prints and exits 1 on regression. Add to that entry:

- A `--root PATH` option (default cwd) so it can find `kstrl.toml` and the control directory.
- After computing the comparison, if `AutonomyConfig.load(root).enabled`:
  - Always, on `passed == False`: open an inbox item `ItemKind.CALIBRATION_DRIFT` with title `Calibration regression: <n> failing threshold(s)`, detail = the comparison's failure lines, `dedupe_key = f"calibration:{new_baseline_timestamp}"`, evidence = the failure dicts. This is the advisory tier.
  - Additionally, when `[autonomy] demote_on_calibration_regression = true` (new `AutonomyConfig` field, default `False`, env `KSTRL_AUTONOMY_DEMOTE_ON_CALIBRATION`): call `apply_demotion(root, DemotionTrigger.CALIBRATION_REGRESSION, reason=<first failure line>, evidence={...}, run_id="calibration:<timestamp>", ui=...)`.
  - When autonomy is disabled: no inbox item, no demotion, and one printed line `autonomy ladder disabled; regression recorded in the report only`.
- The exit code is unchanged (1 on regression) so CI usage is unaffected.

### 3. Health breach: the seam only

Define the contract #151 must implement, so the two PRs meet cleanly:

```python
# kstrl/health.py (created by #151)
@dataclass(frozen=True)
class HealthBreach:
    metric: str        # e.g. "retry_rate"
    rule: str          # e.g. "WE1: 1 point beyond 3 sigma"
    value: float
    limit: float
    window_runs: int

def health_breaches(root_dir: Path) -> list[HealthBreach]: ...
```

In this PR, add to `_record_autonomy_outcome` a call guarded by `try: from kstrl.health import health_breaches except ImportError: breaches = []`, after the policy-violation handling. For each breach: open `ItemKind.DEMOTION_NOTICE`? No: a breach is not yet a demotion. Open a new inbox item using the **existing** `ItemKind.CALIBRATION_DRIFT`? No, wrong meaning. Use `ItemKind.HALTED_RUN`? No. Doctrine 6 forbids minting an outcome enum, but `ItemKind` is the inbox's kind vocabulary and its docstring says "adding a kind means adding an emitter, never just a label" (`kstrl/inbox.py:52-58`). This PR adds the emitter, so add `HEALTH_BREACH = "health_breach"` to `ItemKind`. Then, when `[autonomy] demote_on_health_breach = true` (default `False`, env `KSTRL_AUTONOMY_DEMOTE_ON_HEALTH`), call `apply_demotion(..., DemotionTrigger.HEALTH_BREACH, ...)`.

Until #151 lands, the import fails, `breaches` is empty, and nothing fires. Write a test that proves the import guard works by monkeypatching a fake `kstrl.health` module.

### 4. Replay stays advisory for the rules

`ks autonomy replay` (`kstrl/cli.py:3457`) is the advisory mode for the health rules: it reports would-have-fired counts without mutating state. Add a sentence to its docstring and to `docs/dark-factory-roadmap.md` R8.4 section saying so, and carry the published false-alarm arithmetic into #151 by comment: one point beyond three sigma fires by chance about once in 370 observations; all four Western Electric rules together, about once in 92 (https://handwiki.org/wiki/Western_Electric_rules). Since a health breach can demote, a false alarm costs a level and a ten-run cooldown (`DEMOTION_COOLDOWN_RUNS = 10`, `autonomy.py:121`). Pick the rule set with that arithmetic in front of you.

### 5. Docs

- `kstrl.toml.example` `[autonomy]`: the two new keys, default false, with the sentence "advisory first: an inbox item always; demotion only when true".
- `docs/runbook.md` "Calibration suite reports a regression": add what now happens with the ladder on.
- Regenerate README. CHANGELOG `### Added`.

## Out of scope

- Implementing `kstrl/health.py` (that is #151).
- `HUMAN_REJECTED_AUTO_MERGE`: no detector exists (an auto-merged PR later reverted by a human). File separately after R8.7 defines release identity.
- Changing any ladder threshold.

## Tests to write

`tests/test_autonomy.py` (extend) and `tests/test_calibration.py` (the non-LLM parts; calibration's LLM suite is opt-in and must not be required here):

1. `test_apply_demotion_matches_previous_behaviour`: the existing policy-violation demotion test still passes unmodified, and a direct call to `apply_demotion` with `POLICY_VIOLATION` produces the same transition record and inbox item as the previous inline code.
2. `test_apply_demotion_at_floor_returns_none_and_counts`.
3. `test_compare_regression_opens_inbox_item`: two baseline files where the new one drops a role below `MIN_ROLE_DETECTION_RATE`; autonomy enabled; after `compare`, the inbox has one `calibration_drift` item; the ladder level is unchanged.
4. `test_compare_regression_demotes_when_enabled`: same with `demote_on_calibration_regression = true` and level L2; level becomes L1, transition trigger is `CALIBRATION_REGRESSION`, and both a `calibration_drift` and a `demotion_notice` item exist.
5. `test_compare_regression_dedupes`: running compare twice on the same pair yields one inbox item.
6. `test_compare_no_regression_writes_nothing`.
7. `test_compare_autonomy_disabled_writes_nothing_and_says_so`.
8. `test_health_import_guard`: without `kstrl.health`, `_record_autonomy_outcome` completes and writes no health item; with a monkeypatched `kstrl.health.health_breaches` returning one breach, an inbox item of kind `health_breach` is written and the level is unchanged; with `demote_on_health_breach = true`, the level drops one and the trigger is `HEALTH_BREACH`.
9. `test_config_flags_env_beats_toml` for both new keys.

## Acceptance (all must hold)

- [ ] `python -m kstrl.calibration compare <old> <new> --root <tmp>` with a regressing pair and the ladder on opens a `calibration_drift` inbox item and, by default, does not demote (test 3).
- [ ] With `demote_on_calibration_regression = true` it demotes exactly one level with trigger `CALIBRATION_REGRESSION` (test 4).
- [ ] The health seam is inert until `kstrl.health` exists and fires when it does (test 8).
- [ ] `ks autonomy replay` output is unchanged (its tests pass unmodified).
- [ ] All existing autonomy tests pass unmodified; tests 1 through 9 pass; baseline suite passes; gen_docs check passes.
- [ ] `docs/dark-factory-roadmap.md` R8.4 section carries the false-alarm note and a pointer to this issue.

0xfauzi:
## Summary

`DemotionTrigger` has declared five causes since R8.2 and exactly one of them
had an emitter. A factory getting quietly worse without breaching the policy
envelope kept its level, and a calibration regression a human had already
measured changed nothing. This wires the two the issue names, advisory first.

- **One applier.** `autonomy.apply_demotion`, extracted from
  `factory._record_autonomy_outcome`, is now the single path for every
  automatic demotion, so the four writes it performs (ladder state, evolution
  journal, run event stream, inbox notice) cannot drift apart per trigger.
  `ks autonomy demote` is the fourth call site and still does its own thing;
  folding it in needs an `actor` parameter and is out of scope here.
- **Calibration regression.** `python -m kstrl.calibration compare` gains
  `--root` and, when the ladder is enabled, opens a `calibration_drift` inbox
  item on a regression. `[autonomy] demote_on_calibration_regression` (env
  `KSTRL_AUTONOMY_DEMOTE_ON_CALIBRATION`) additionally revokes one level, once
  per COMPARISON however often that comparison is re-run. The dedupe is on the
  pair of baselines: a comparison is a pair, so `compare old_A new` and
  `compare old_B new` are different regressions and must not collapse onto one
  item. When EITHER file has no `timestamp` key the identity falls back to a
  digest of both baselines' detection-rate tables, because `Baseline.timestamp`
  fills a missing key with one literal shared by every such file, which is a
  fill-in and not an identity. The digest is over the rates rather than over the
  failure lines because a floor failure quotes only the NEW rate, so two
  different old baselines can produce byte-identical failure text.
- **Health breach: the R8.4 seam only.** `kstrl/health.py` does not exist yet
  (#151), so `factory._record_health_breaches` is the meeting point, plus a
  `health_breach` inbox kind and `[autonomy] demote_on_health_breach` (env
  `KSTRL_AUTONOMY_DEMOTE_ON_HEALTH`). The import guard swallows exactly one
  thing: `ModuleNotFoundError` whose `.name` is `kstrl.health`. Both the import
  AND the `health_breaches` lookup are inside the guarded block, so a module
  that exists and raises for its own missing dependency, and a module that
  exists with the function renamed, both reach the caller's "Autonomy state
  update failed" warning instead of reading as "#151 has not landed". Health
  demotion is additionally suppressed during cool-down, out loud: a breach is a
  windowed trend, so an ungated trigger would cost a level per run down to L1
  before the operator had read the first notice, and a level that does not move
  for ten runs with no line explaining why is the failure this ladder exists to
  show.

Both switches default off, and neither is accepted unless written as an
unquoted boolean. Every entry threshold in the ladder is still an unmeasured
placeholder, so revoking autonomy by default would act on a number nobody has
measured; and `bool("false")` is `True`, so a typo that ARMS a revocation switch
is worse than one that disarms it. `[autonomy] enabled` is read the same strict
way, which is a BEHAVIOUR CHANGE worth stating plainly: a config that spells it
`"false"` or `"0"` used to switch the ladder on and now reports a configuration
problem instead. It is in scope because it is the switch that arms the two keys
this PR adds, and because it can only ever fail in the arming direction. The
other `[autonomy]` keys keep the repo-wide `bool(section[...])` coercion;
tightening all 29 sites changes how existing configs load and is its own change
with its own guard.

Exit codes are unchanged (0 pass, 1 regression) with one addition: on a
REGRESSION, a `kstrl.toml` that will not load or cannot be read exits 2 instead
of being read as "ladder off". A passing comparison never consults the config,
so a repo with a typo'd `kstrl.toml` does not fail every green CI run. Anything
that fails after the config resolves - the inbox write, the demotion itself -
is reported on stderr and leaves the exit code alone.

The roadmap note carries the arithmetic #151 must choose its rule set against,
where the choice gets made rather than in a design doc nobody reads at that
moment: one point beyond three sigma fires by chance about once in 370
observations and all four Western Electric rules together about once in 92,
against a cost of one level plus ten decisive runs of locked re-promotion. It
is written there once, and the code docstring points at it.

`kstrl/calibration_ladder.py` is a new module because the consequence half
should not drag the control plane into every importer of a comparison, and
because folding it back into `kstrl/calibration.py` does not fit. Measured the
way a reader can repeat it: `calibration.py` is 752 lines and this module's body
below its last import (line 51 of 291) is 240, so the fold is at least 992
against the 800-line ratchet.

Round 2 addressed the first review's 10 should-fix and 10 nits and closed all
eight mutations it had measured as still green.

Round 3 addresses the second review's 5 should-fix and 15 nits. What that review
found, stated plainly because it is the useful part: the round-2 fixes were
applied to the sites the first review NAMED rather than swept across the class
each one belonged to. Four separate findings were the same shape.

- The degraded-state guard was on the demotion branch; the branch an ordinary
  run takes still saved unconditionally, so a damaged `autonomy.json` was
  replaced by a fresh L1 on the first clean run and the next load found nothing
  to report. `AutonomyState.save` now owns that refusal, which is the one
  function that writes the file.
- `ControlStateError` was caught at four inbox sites. `Inbox._append` takes the
  control lock on EVERY write, so the hole was at every write: three more sites
  had the same enumeration, and one of them, the TUI's approve/reject/snooze,
  was found by the new census guard rather than by reading. There an uncaught
  `ControlLockedError` reaches the Textual event loop.
- `InboxConfig.load` casts per key, so a TOML date raises `TypeError`. Three of
  the guards written in round 2 to close exactly that shape one function over
  did not catch it. At the demotion notice it arrived after the demotion had
  already been saved.
- `[autonomy] enabled` still coerced, two lines above the two keys it gates,
  which are read through the helper written to refuse coercion
  (`kstrl/autonomy.py` lines 755, 757 and 762).

The guard for instance eight is `tests/test_inbox_write_guards.py`, and it is
built the way CLAUDE.md asks: a census of where an `Inbox` is OBTAINED, which is
closed by construction, under an inventory of what each write catches. Three
declines are recorded with measurements below.

Three false statements in this PR's own text are also corrected, found by
re-measuring rather than by the review. `AutonomyState.save`'s docstring cited a
guard file that does not exist; two comments still called the sweep six inbox
sites when the census this PR ships finds seven; and the fold-back arithmetic in
`kstrl/calibration_ladder.py`, which round 2's N3r2 already corrected once, had
gone stale again on the next edit to the same file. It now states the derivation
(`calibration.py` 752 lines, last import line 51 of 291) rather than a total.

Closes #232

**Tested.**

- Full suite on the tree merged with `origin/main` 1754357, bytecode deleted
  first and `PYTHONDONTWRITEBYTECODE=1`: 5709 passed, 37 skipped, 38 xfailed in 344s, against 5564 / 37 / 38 on
  `origin/main` at 568bca4, 5590 / 37 / 38 at the round-1 head 9e105c5 and
  5662 / 37 / 38 at the round-2 head 6dea6f9. `uv run mypy kstrl/ --strict`
  clean (130 files), `uv run ruff check kstrl/ tests/` clean,
  `uv run python scripts/gen_docs.py --check` current. The `pre-commit` hook set
  ran on every commit here, including both ratchets and the em-dash check; the
  hooks that fetch remote repos ran too, since these were ordinary commits.
- **The static-guard census was RUN, not read.** The 34 test files under
  `tests/` that AST-walk the package collect 1122 / 9 / 37 / 0 (passed / skipped /
  xfailed / failed) before the `origin/main` merge, 1123 / 9 / 37 / 0 after it and
  1124 / 9 / 37 / 0 at the head, over the same 34 files. Both deltas are attributed to
  a named file: `tests/test_astwalk_nets.py` 68 -> 69, which is #351's change arriving
  with the merge, and `tests/test_inbox_write_guards.py` 28 -> 29, the layer-3 control
  added here. Measured per file rather than inferred, since a code move in another file
  changes what a walk sees without touching yours.
- The four new test files collect 95 cases: `tests/test_demotion_triggers.py`
  (31, the applier and the config surface), `tests/test_calibration_ladder.py`
  (22, the compare emitter), `tests/test_health_seam.py` (13, the R8.4 seam) and
  `tests/test_inbox_write_guards.py` (29, the static census of inbox writes),
  over one `tests/helpers/demotion.py`. The first three are one file split three
  ways because it reached 1015 lines and the 800-line pre-commit ratchet fails
  that. The seam is exercised against a monkeypatched `kstrl.health` module and,
  for the import guard, against a `MetaPathFinder` on the real import machinery,
  rather than assumed inert.
- **The red-first evidence here is a COLLECTION error, not per-test.** On
  568bca4 the new files fail to import (`cannot import name 'apply_demotion'`,
  `ItemKind has no attribute 'HEALTH_BREACH'`), so every case is red for one
  reason. That is the expected shape for a feature PR; the per-property
  evidence is the mutation testing below.
- **Round 3's four fixes are each red first on the round-2 head**, which is the
  evidence a feature PR's collection error cannot give. Measured before the fix:
  a truncated `autonomy.json` came back as a fresh L1 payload after one clean
  run; `[inbox] open_item_cap = 1979-05-27` produced a `TypeError` traceback out
  of `apply_demotion`, out of `compare` and out of the health seam; two
  timestamp-less comparisons against different old baselines produced one
  demotion where two were due; and `[autonomy] enabled = "false"` loaded as
  `True`. The cool-down pair is the exception and is disclosed as such: both
  cases pass on the shipped code, because they exist to pin a boundary that
  nothing crossed. The seed-2 case is the one that discriminates, and the
  off-by-one plant (`> 0` written `> 1`) fails it.
- **That the `apply_demotion` extraction changed nothing** is carried by the
  two pre-existing tests that pass UNMODIFIED:
  `test_autonomy_ladder.py::TestFactoryIntegration::test_policy_violation_demotes_and_journals`
  and `test_inbox.py::TestFactoryEmission::test_demotion_emits_notice_with_evidence`.
  One exception, disclosed: the AT-FLOOR warning text changed (it names the
  trigger and carries the reason, because the applier serves five triggers
  where the inline block served one). Nothing covered the old string; the new
  one is pinned by `test_floor_message_names_the_trigger_and_the_reason`.
  `test_autonomy_ladder.py` is untouched; `tests/test_inbox.py` gains one
  parametrize row for the `HEALTH_BREACH` priority ledger and nothing else.
- **Mutation testing, round 3: 79 plants, 72 caught, 2 still green, 5 retargeted and then caught, 0 hung.** Three harnesses in detached
  worktrees at the merged head: this round's 25 plants, the round-2 reviewer's
  own 17 re-run unchanged, and round 2's 37 re-run unchanged. Every run deleted
  `__pycache__` before AND after, set `PYTHONDONTWRITEBYTECODE=1`, bounded the
  child at 420s, spawned it with `start_new_session=True` and killed it by
  `os.killpg` on timeout, and refused to score a plant whose anchor did not
  occur exactly once. The two survivors are the
  same disclosed property (`M21`, moving the `health_breaches` lookup back
  outside a clause left narrow, which is behaviour-neutral; `M37`, that move
  plus the widening, is caught) scored by two harnesses. Five plants did not
  measure what they were aimed at and are reported as their own outcome rather
  than folded into a pass: four whose anchors this round's fixes moved
  (`M9`, `M24`, `M28`, `M35`, all reported COULD-NOT-PLANT by the dry pass) and
  one that applied but REORDERED the digest material instead of removing it
  (`F15`). All five were retargeted at the property and all five are then
  caught. One plant found a real hole: the refused-save branch of
  `commit_transition` had no reachable caller, so removing it changed nothing
  until a test drove the function directly. No run exceeded 45s against the
  420s bound, so nothing was scored as caught by a hang. One plant, `M36`, is scored by
  `mypy --strict` rather than by pytest, because the thing it mutates is a
  static guard, and six (`F10` to `F13`, `F19`, `F20`) are scored against
  the new census guard one layer at a time, because a guard you did not mutate
  is a guard you did not test, per layer. The last two exist because that rule
  found layer 3 without a planted control of its own.
- The exit-2 path is parametrized on `tests/helpers/bad_toml`'s shared #318
  table (four document faults) plus two faults that table cannot reach: a
  `chmod 000` file, which raises `OSError` because `load_toml_section`
  deliberately does not normalise it, and `max_level = 1979-05-27`, a valid
  TOML date whose per-key cast raises `TypeError`.
- `control_lock` is not reentrant within one process, measured: two nested
  `control_lock(root)` calls block, killed by a 5s alarm. See the handoff below.
- `lanes/232/r2_verify_head.py` in the review scratchpad drives the four round-3
  subjects against a worktree from outside the suite: damaged bytes byte-
  identical after two consecutive clean runs and reported on both, a `TypeError`
  absorbed at each of the three emitters with the demotion still saved and the
  DEMOTED line still printed, both sides of the cool-down boundary, and the
  census counts. It exits non-zero on any of them.
- `kstrl/statedir.py` gains `ControlStateError` as the base of
  `ControlLockedError` and `ControlUnavailableError`. Both stay `RuntimeError`
  subclasses, so nothing that catches one by name changes. The `RuntimeError`
  census in `tests/test_tui_config_guard.py` moves with it (`statedir.py` 2 ->
  1) and its walk now follows subclasses transitively, with the closure
  asserted rather than listed. That walk's exactness rests on the package being
  imported, which its docstring claimed and its code did not do: the
  comprehension imported one module per NAME it found, 11 of this package's 130,
  and 27 modules stayed unimported even in a process already holding 100. The
  import is now its own line and `test_the_domain_walk_imports_the_whole_package`
  asserts it.
- **Who catches what, when writing to the inbox, is now a static census rather
  than a habit.** `tests/test_inbox_write_guards.py` pins every `Inbox(...)`
  construction in `kstrl/` (9, across 7 modules, by row AND by count because the
  row form deduplicates), every mutating call reached from one (14 rows: 9
  guarded across 7 functions, 5 propagating with a stated reason each), and
  every `InboxConfig.load` (8 rows). A guarded row must have an enclosing
  handler resolving to `kstrl.statedir.ControlStateError` BY ORIGIN, not by
  spelling, since a name is not an identity. Each layer carries a control,
  because a matching inventory is also what a switched-off walk returns.

**Assumed, and flagged as such.**

- The `HealthBreach` field names (`metric`, `rule`, `value`, `limit`,
  `window_runs`) are taken from the issue text. #151 has not been written, so
  nothing here verifies the eventual dataclass agrees on the NAMES. It is
  checked on the SHAPE: `_HealthBreach`'s members are declared read-only
  properties, and `factory._health_breach_seam_accepts_the_contract` (a
  `TYPE_CHECKING`-only function) hands mypy a `@dataclass(frozen=True)` spelled
  as the issue spells it where a `_HealthBreach` is expected. The obvious
  spelling of that Protocol - plain `metric: str` members - demands SETTABLE
  attributes and rejects the frozen dataclass the contract mandates; restoring
  it turns `mypy --strict` red, which is the guard.
- The 1-in-370 and 1-in-92 false-alarm figures are quoted from the linked
  reference, not re-derived here.
- Nothing exercises a real `kstrl/health.py`, because there is not one. The
  demoting path is proven only against a monkeypatched module.
- SK8's concurrency race is reasoned, not raced: the factory's own comment says
  its ladder update runs inside the factory lock so two runs cannot interleave
  read-modify-write, and the compare path holds no equivalent.
- The five `cli.py` inbox writes and the two unguarded `InboxConfig.load` reads
  are enrolled in the new guard as PROPAGATING with a stated reason rather than
  changed. The reasoning is that an operator typed the command and is watching
  it, and that `[inbox]` is a preflight section so a malformed value is named
  before either is reached. Neither claim is exercised by a test here.

**Handoffs, measured, not filed as issues.**

- `Inbox.path` is an unmemoised property: it re-derives the control dir per
  `add`, each derivation spawning git subprocesses. Measured on the advisory
  health path with a stub module, counting `subprocess.run` and
  `subprocess.Popen` at their own layers: 1 breach = 106 ms / 24 spawns, 10 = 604 ms / 141, 20 = 1122 ms /
  271, linear in breaches. It is pre-existing and
  inert today (no `kstrl.health`), but this PR ships the repo's first loop over
  `Inbox.add`, and #151's rule set is what makes N > 1. The fix is to resolve
  the path once in `Inbox.__init__`, which `statedir.xdg_state_home` already
  does for the same reason.
- **The demotion notice does not fire the notify push.** `Pipeline._inbox_add`
  ends with `notify.fire_inbox_item` for notifiable items, `inbox.notifiable()`
  names `DEMOTION_NOTICE`, and `apply_demotion` has no `Notifier`. So an
  operator with notifications on is not told about an automatic demotion. This
  predates the PR (the inline factory block did not either), and fixing it means
  threading a `Notifier` through five call sites.
- **The inbox emission block is written seven times** (`autonomy.py`,
  `calibration_ladder.py`, `factory.py`, `pipeline.py` twice, `serve.py`,
  `tui/screens/inbox.py`), disagreeing on the error channel and on the notify
  push. The thing that made that dangerous - seven copies of an enumerated
  exception tuple, none of them naming the control lock's own error - is closed
  here, at all seven, and `tests/test_inbox_write_guards.py` is what fails when
  an eighth appears. The duplication itself is a follow-up: an `inbox.emit`
  helper would have to reconcile three different error channels and two
  different answers about the notify push, which is a design question rather
  than a defect.
- **No lock is held across load / check-history / demote / save on the compare
  path**, so a concurrent factory demotion is lost. Not closed here, with a
  measurement: `control_lock` is the obvious wrapper and it self-deadlocks. It
  is a `flock` on a fresh file description, and `AutonomyState.save` and
  `Inbox._append` each take it themselves, so an outer hold blocks the inner
  one in the same process (measured: two nested calls block, killed by a 5s
  alarm). Closing it needs lock-holding variants of both callees and a change
  to every caller including the factory. The race and that measurement are a
  comment at the site.

## Checklist

## Provenance

AI code review will be run on this PR as an assist at the owner's request; the
owner remains the gating reviewer (H1).

No adversarial prompt body changed, so neither box applies. The nine enrolled
prompts in `tests/test_prompt_versions.py::_PROMPTS` are untouched and their
snapshot tuples are unchanged in this diff.


kstrl/calibration_ladder.py
```diff
@@ -0,0 +1,141 @@
...
+        print(f"error: {exc}", file=sys.stderr)
+        return 2
+    if comparison.passed:
+        return None
+    if not config.enabled:
+        print(LADDER_DISABLED_LINE)
+        return None
+
+    failures = list(comparison.failures)
+    new_baseline = comparison.new.timestamp
```
0xfauzi: **The dedupe identity is `"unknown"` whenever a baseline has no `timestamp`, which permanently disarms the trigger.**

`load_baseline` (calibration.py:407) does `timestamp=str(data.get("timestamp", "unknown"))`, and v1 baselines are explicitly supported with no timestamp at all. Nothing here validates that the identity it deduped on is a real identity. The first regression against any timestamp-less baseline writes `new_baseline: "unknown"` into the transition record; every later regression against a *different* timestamp-less baseline matches it and is silently dropped.

Reproduced on this branch (`enabled = true`, `demote_on_calibration_regression = true`, ladder at L3, two structurally different regressed baselines, neither carrying a `timestamp` key):

```
run1: 1   level after run1: 2
autonomy: baseline unknown already cost a level; not demoting again
run2: 1   level after run2: 2   demotions: 1
```

The advisory tier collapses the same way: both regressions land on `dedupe_key="calibration:unknown"`, and `Inbox.add` on an open match updates only `occurrences`/`last_seen_at`/`detail`, so the item keeps the *first* regression's title and evidence.

This is the CLAUDE.md rule verbatim: a gate that keys on parsed output must reject what it cannot parse, never count it as zero. Either refuse to demote when `comparison.new.timestamp == "unknown"` (with a stated reason on stdout, not silence), or key the dedupe on a content digest of the new baseline rather than on a field the loader invents a default for.

kstrl/autonomy.py
```diff
@@ -844,6 +886,74 @@ def commit_transition(
...
+    inbox notice - cannot drift apart per trigger. An inbox failure is
+    warned about, never fatal: losing the notice must not strand the
+    ladder with a saved demotion and no record of why.
+    """
+    if state is None:
+        state = AutonomyState.load(root_dir)
+    trigger_text = trigger.label.replace("_", " ")
+    record = state.demote(trigger, reason, evidence=evidence)
+    if record is None:
+        state.save(root_dir)
```
0xfauzi: **The unconditional save at the floor destroys a degraded ladder state and its entire history.**

`AutonomyState.load` fails closed by returning a *fresh* L1 state with `degraded_reason` set. A degraded state is therefore always at the floor, so it always reaches this branch, and this `state.save(root_dir)` writes the blank L1 payload over the file that failed to parse - level, history and the degraded signal `safemode.safe_mode_reasons` reads all gone.

Reproduced on this branch with `python -m kstrl.calibration compare` and an `autonomy.json` holding `level: 3` plus one demote record and a non-string `since`:

```
load -> 1 degraded: autonomy: rejected ladder state .../autonomy.json (since must be a string); failing closed to L1 Supervised
exit: 1
AFTER: {"schema_version":1,"level":1,"since":"2026-09-06T...",...,"history":[]}
```

The docstring justifies the save with "the caller's own counters ... were mutated before the demotion was attempted" - true for `factory._record_autonomy_outcome`, false for the new `calibration_ladder.report_to_ladder` caller, which passes a state it just loaded and never touched. The extraction generalised a factory-only assumption to a measurement command the runbook describes as changing nothing but exit codes.

Save on this path only when the caller actually passed a `state`, and never when `state.degraded_reason is not None`.

kstrl/calibration_ladder.py
```diff
@@ -0,0 +1,141 @@
...
+
+    ``None`` means "leave the compare command's own exit code alone",
+    which is every path but one: a ``kstrl.toml`` that will not load
+    returns 2. Reading an unloadable config as "autonomy disabled" would
+    be a silent removal of the mechanism - the operator switched the
+    ladder ON and a typo elsewhere in the file switched it off again.
+    """
+    try:
+        config = AutonomyConfig.load(root_dir)
+    except (ValueError, AutonomyError) as exc:
```
0xfauzi: **The config guard is one exception type short: a non-numeric `max_level` raises `TypeError`, which escapes and produces a raw traceback.**

`AutonomyConfig.load` does `int(section["max_level"])` (autonomy.py:677). TOML values that are not numbers or strings raise `TypeError`, not `ValueError`:

```
'[autonomy]\nmax_level = [1]\n'          -> TypeError: int() argument must be ... not 'list'
'[autonomy]\nmax_level = 1979-05-27\n'   -> TypeError: int() argument must be ... not 'datetime.date'
report_to_ladder -> TypeError int() argument must be ... not 'datetime.date'
```

The last line is `report_to_ladder` called directly on this branch. `python -m kstrl.calibration compare` has no `config_preflight` in front of it, so this reaches the user as an unhandled traceback.

The `TOML_PARSE_FAULTS` table the test parametrises on covers *document* faults, which `load_toml_document` normalises to `ConfigError(ValueError)`. It does not cover the per-key casts one field over - the same shape as the #260 round-2 finding recorded in CLAUDE.md. Either widen this to `except Exception` (broad clause last, all I/O outside), or have `AutonomyConfig.load` own its cast taxonomy.

kstrl/calibration_ladder.py
```diff
@@ -0,0 +1,141 @@
...
+            return
+        Inbox(root_dir, inbox_config).add(
+            ItemKind.CALIBRATION_DRIFT,
+            f"Calibration regression: {len(failures)} failing threshold(s)",
+            detail="\n".join(failures),
+            run_id=f"calibration:{new_baseline}",
+            dedupe_key=f"calibration:{new_baseline}",
+            evidence=evidence,
+        )
+    except (OSError, ValueError) as exc:
```
0xfauzi: **`ControlUnavailableError` is a `RuntimeError`, so it escapes this guard and every sibling copy of it.**

`statedir.control_lock` raises `ControlUnavailableError(RuntimeError)` when the control directory cannot be created or locked (statedir.py:638, 605). `Inbox._append` calls `control_lock` unconditionally, so this `except (OSError, ValueError)` cannot catch its most likely failure. Same hole in `autonomy.apply_demotion:947` and `factory._open_health_breach_items:2787`.

Confirmed on this branch by making `control_lock` raise: the compare command dies with an unhandled `kstrl.statedir.ControlUnavailableError` traceback. In the factory the outer `except Exception` absorbs it; from `calibration compare` nothing does, which contradicts this function's own docstring - "a failed write must not change the compare command's exit code, which is the measurement's answer".

Add `ControlUnavailableError` to the tuple in all three places (or catch it in `report_to_ladder`).

kstrl/calibration_ladder.py
```diff
@@ -0,0 +1,141 @@
...
+        "failures": failures,
+        "new_baseline": new_baseline,
+        "old": str(comparison.old.path),
+        "new": str(comparison.new.path),
+    }
+    _open_drift_item(root_dir, evidence)
+
+    if not config.demote_on_calibration_regression:
+        return None
+    state = AutonomyState.load(root_dir)
```
0xfauzi: **Load / check-history / demote / save with no lock held across the sequence: a concurrent factory run's demotion can be lost.**

`AutonomyState.save` and `Inbox._append` each take `control_lock` for their own write, but nothing holds it across this read-modify-write. `factory.run_factory` is explicit about the opposite ("Runs INSIDE the factory lock, before it is released, so two runs cannot interleave read-modify-write", factory.py:4310).

Interleaving: compare loads L3; a factory run finishing at the same moment demotes L3 -> L2 for a policy violation and saves; compare then demotes its own stale L3 copy to L2 and saves, and the policy violation's transition record is gone from `history` along with the level it cost. The control dir is keyed on the origin URL, so this is cross-checkout, not just cross-process in one tree.

Wrap the load-through-`apply_demotion` sequence in `control_lock(root_dir)`, or take the factory lock.

kstrl/autonomy.py
```diff
@@ -652,11 +675,30 @@ def load(cls, root_dir: Path | None = None) -> AutonomyConfig:
         defaults = cls()
         enabled = bool(section["enabled"]) if "enabled" in section else defaults.enabled
         max_level = int(section["max_level"]) if "max_level" in section else defaults.max_level
+        demote_calibration = (
+            bool(section["demote_on_calibration_regression"])
```
0xfauzi: **`bool(...)` on the TOML value makes both new switches fail OPEN: `= "false"` loads as `True`.**

Measured on this branch:

```
'[autonomy]\nenabled = true\ndemote_on_calibration_regression = "false"\n'
  -> AutonomyConfig(enabled=True, ..., demote_on_calibration_regression=True, ...)
'[autonomy]\nenabled = true\ndemote_on_health_breach = "off"\n'
  -> AutonomyConfig(enabled=True, ..., demote_on_health_breach=True)
```

Every non-empty string is truthy. The PR's whole safety argument is "both switches default off, because every entry threshold in the ladder is still an unmeasured placeholder" - and a quoted `"false"` is the single most common way to write a TOML boolean wrong. The failure is silent and revokes autonomy.

It follows the existing `enabled` line, so it is a pre-existing pattern, but these are the two keys whose wrong value costs a level. Reject a non-bool with a named `ConfigError` the way `enum_field_error`/`required_field_error` do in `decisions.py`, rather than coercing it.

kstrl/factory.py
```diff
@@ -2731,11 +2732,145 @@ def _expired_futures(
...
+        if exc.name != "kstrl.health":
+            raise
+        return
+    breaches: list[_HealthBreach] = list(health.health_breaches(root_dir))
+    if not breaches:
+        return
+    _open_health_breach_items(root_dir, breaches, run_id=run_id, ui=ui)
+    if not autonomy_config.demote_on_health_breach:
+        return
+    if state.cooldown_runs_remaining > 0:
```
0xfauzi, line 2922: **The cool-down suppression is a silent return: nothing tells the operator a demotion they configured was skipped.**

The sibling path in `calibration_ladder.report_to_ladder:130` prints `autonomy: baseline <ts> already cost a level; not demoting again` for exactly this situation, and `LADDER_DISABLED_LINE` exists so "the absence of an inbox item is a stated outcome rather than silence". Here, with `demote_on_health_breach = true` and a real breach, the level does not move and no line is written, no journal entry, nothing on the `HEALTH_BREACH` item recording that it was suppressed.

Since a breach is a windowed trend the suppression can hold for the full `DEMOTION_COOLDOWN_RUNS = 10` decisive runs, so this is the *common* case, not the edge. CLAUDE.md: "Don't add silent code paths - if it's worth deciding, it's worth recording."

`ui.warn` it, and put `suppressed_by_cooldown: <n>` on the breach item's evidence.

kstrl/autonomy.py
```diff
@@ -844,6 +886,74 @@ def commit_transition(
...
+        ui.warn(f"Autonomy: {trigger_text} recorded ({reason}); already at L1, nothing to revoke")
+        return None
+    commit_transition(state, record, root_dir, bus=bus, run_id=run_id)
+    # R8.2 promised this and R8.3 delivers it: a demotion is exactly the
+    # boundary condition an over-the-loop operator must see, and the
+    # triggering evidence is perishable - it belongs on the item.
+    try:
+        inbox_config = InboxConfig.load(root_dir)
+        if inbox_config.enabled:
+            Inbox(root_dir, inbox_config).add(
```
0xfauzi, line 1045: **The single applier for every automatic demotion never fires the notification hook, so `[inbox] notify_action_required` stays a no-op for the one kind `notifiable()` singles out.**

`inbox.notifiable()` is `item.kind.action_required or item.kind is ItemKind.DEMOTION_NOTICE` - the `or` clause exists solely for this item. `pipeline._inbox_add:1674` does the push ("Without this the [inbox] notify knob was a documented no-op"); this call site does not.

The PR body flags it as pre-existing, which is true of the inline block. What changed is that this is now the *only* emitter of `DEMOTION_NOTICE` for all three triggers, so it is the one place that can fix it, and the fix is three lines already written next door. A demotion is `Priority.HIGH` with perishable evidence; it is the item most worth interrupting a human for and it is the only notifiable kind that never does.

kstrl/calibration_ladder.py
```diff
@@ -0,0 +1,141 @@
...
+    if _already_demoted_for(state, new_baseline):
+        print(f"autonomy: baseline {new_baseline} already cost a level; not demoting again")
+        return None
+    apply_demotion(
+        root_dir,
+        DemotionTrigger.CALIBRATION_REGRESSION,
+        failures[0],
```
0xfauzi: **Nothing checks that `new` is actually newer than `old`, and the command now revokes an autonomy level.**

`compare_baselines` just diffs its two arguments; argument order was previously a reporting concern with an exit code attached. It is now load-bearing: `compare <new> <old>` reports the *improvement* as a regression (every recovered fixture reads as "newly missed"), opens a `calibration_drift` item, and with the switch on takes a level off the ladder and starts a ten-run cool-down - for a measurement that got better.

`comparison.new.timestamp` is already read two lines up. Refuse (or at minimum warn loudly) when `comparison.new.timestamp <= comparison.old.timestamp` and both parse, rather than acting on a transposition.

kstrl/calibration.py
```diff
@@ -718,6 +730,14 @@ def main(argv: Sequence[str] | None = None) -> int:
             return 2
         comparison = compare_baselines(old, new)
         print(format_comparison(comparison))
+        # Imported here rather than at module scope so this module, the
+        # non-LLM measurement half, does not drag the control plane
+        # (autonomy, inbox, UI) into every importer of a comparison.
+        from kstrl.calibration_ladder import report_to_ladder
+
+        override = report_to_ladder(comparison, (args.root or Path.cwd()).resolve())
```
0xfauzi, line 744: **A `--root` that points nowhere reads as "ladder off" - the exact silent-disable this PR argues against for a broken config.**

`resolve_config_file` on a nonexistent or wrong directory yields no `kstrl.toml`, `load_toml_section` returns `{}`, `enabled` falls to the `False` default, and `report_to_ladder` prints `autonomy ladder disabled; regression recorded in the report only` and returns `None`. An operator who typo'd the path, or ran from the wrong cwd with the default, is told the ladder is off when it is on.

The asymmetry is the point: a `kstrl.toml` that will not load exits 2 because "the config is broken must not read as the ladder is off", but a root with no `kstrl.toml` at all gets no such treatment. Validate that `root_dir` exists and is a directory before calling, and say "no kstrl.toml at <path>" rather than "ladder disabled".

kstrl/factory.py
```diff
@@ -2731,11 +2732,145 @@ def _expired_futures(
...
+    Honours ``[inbox] enabled``; a failed write warns and never fails the
+    run, because losing the notice must not also lose the demotion the
+    caller goes on to apply.
+    """
+    try:
+        inbox_config = InboxConfig.load(root_dir)
+        if not inbox_config.enabled:
+            return
+        inbox = Inbox(root_dir, inbox_config)
+        for breach in breaches:
```
0xfauzi: **One `try` around the whole loop means breach 3 failing silently drops breaches 4..N; and each iteration re-derives the control dir and re-scans the log.**

Two costs in one block. (1) The guard is outside the loop, so a single `OSError` on one `inbox.add` aborts every remaining breach with one "Inbox write failed" line that names no breach. Move the `try` inside, or accumulate failures.

(2) `Inbox.add` -> `find_by_dedupe_key` -> `_folded()` -> `scan()` -> `self.path` -> `control_file` -> `control_dir` -> `repo_id` -> a `git` subprocess, per breach, plus a full re-read of the log. The PR body measures it: 1 breach = 54 ms / 12 spawns, 20 = 1356 ms / 259, with 1.148 s of a 1.207 s run in `statedir.control_dir`.

This PR ships the repo's first loop over `Inbox.add`, so it is the caller that makes the pre-existing unmemoised `Inbox.path` bite. Resolving the path once in `Inbox.__init__` (as `statedir.xdg_state_home` already caches for the same reason) is a few lines and belongs here rather than in a handoff paragraph, per CLAUDE.md's one-defect-class-one-PR convention.

kstrl/autonomy.py
```diff
@@ -844,6 +886,74 @@ def commit_transition(
...
+        state = AutonomyState.load(root_dir)
+    trigger_text = trigger.label.replace("_", " ")
+    record = state.demote(trigger, reason, evidence=evidence)
+    if record is None:
+        state.save(root_dir)
+        ui.warn(f"Autonomy: {trigger_text} recorded ({reason}); already at L1, nothing to revoke")
+        return None
+    commit_transition(state, record, root_dir, bus=bus, run_id=run_id)
+    # R8.2 promised this and R8.3 delivers it: a demotion is exactly the
+    # boundary condition an over-the-loop operator must see, and the
```
0xfauzi, line 1040: **Fourth copy of the load-config / check-enabled / add / catch-(OSError, ValueError) block, with a different error channel in each.**

The four: this one, `calibration_ladder._open_drift_item:78`, `factory._open_health_breach_items:2764`, `pipeline._inbox_add:1653`. They disagree on where the failure goes (`ui.warn` here and in factory, `print(..., file=sys.stderr)` in calibration_ladder), on whether the notify push fires (only pipeline), and on caching the `Inbox` (only pipeline). They agree on the exception tuple, which is how the same `ControlUnavailableError` hole got copied four times.

The PR's own thesis is "one applier so the four writes cannot drift apart per trigger" - the same argument applies one level down. An `inbox.emit(root_dir, kind, title, *, on_error)` helper owning config load, the enabled check, the notify push and the exception taxonomy would let all four call sites shrink to the call, and would make the taxonomy fix a one-line change instead of four.

kstrl/factory.py
```diff
@@ -2731,11 +2732,145 @@ def _expired_futures(
     return [f for f in running if not f.done() and f in deadlines and now >= deadlines[f]]
 
 
+class _HealthBreach(Protocol):
```
0xfauzi, line 2737: **`_HealthBreach` cannot fail: `importlib.import_module` returns `Any`, so the annotation on line 2836 checks nothing at either end.**

`breaches: list[_HealthBreach] = list(health.health_breaches(root_dir))` assigns from `Any`, which mypy accepts unconditionally - the PR body says so. So this Protocol documents an intent no tool enforces and no test can violate, and the five field names are, as the body admits, taken from issue text nothing verifies.

The deeper shape costs about the same: put the protocol in a small module (`kstrl/health_contract.py`) that both this seam and #151 import, and have `kstrl/health.py` declare `def health_breaches(root_dir: Path) -> list[HealthBreach]`. Then `--strict` checks the implementation against the contract when #151 lands, instead of the mismatch surfacing at runtime as an `AttributeError` inside a warning.

kstrl/autonomy.py
```diff
@@ -844,6 +886,74 @@ def commit_transition(
...
+    warned about, never fatal: losing the notice must not strand the
+    ladder with a saved demotion and no record of why.
+    """
+    if state is None:
+        state = AutonomyState.load(root_dir)
+    trigger_text = trigger.label.replace("_", " ")
+    record = state.demote(trigger, reason, evidence=evidence)
+    if record is None:
+        state.save(root_dir)
+        ui.warn(f"Autonomy: {trigger_text} recorded ({reason}); already at L1, nothing to revoke")
```
0xfauzi, line 1036: **This warning is not the verbatim move the commit claims: the floor message swapped a violation count for a reason string, and nothing asserts either.**

Before: `f"Autonomy: policy violation recorded ({len(violations)} component(s)); already at L1, nothing to revoke"`. Now: `({reason})`, which for the policy trigger is `"policy violation in comp-a, comp-b"` - so the line reads `Autonomy: policy violation recorded (policy violation in comp-a, comp-b); already at L1`, repeating the trigger and dropping the count.

The commit message says "moved verbatim. Two substitutions: the trigger's own label replaces the hardcoded 'policy violation' in both warnings (identical bytes for POLICY_VIOLATION)". That is not identical bytes on this line. I grepped `tests/test_autonomy_ladder.py` on this branch for `nothing to revoke`, `already at L1` and `component(s)` and found nothing, so the two pre-existing tests offered as the extraction's proof never covered this string in either form.

Small on its own; it matters because the audit trail for this refactor is the claim that nothing changed.

docs/control-loop-design.md
```diff
@@ -1012,7 +1012,7 @@ Cycle: R10, milestone [R10: Control Loop](https://github.com/0xfauzi/kstrl/miles
 | 8 | 5.4 golden patterns | [#229](https://github.com/0xfauzi/kstrl/issues/229) | `[ ]` |
 | 9 | 5.7 memory file | [#230](https://github.com/0xfauzi/kstrl/issues/230) | `[ ]` |
 | 10 | 5.8 polled steering | [#231](https://github.com/0xfauzi/kstrl/issues/231) | `[ ]` |
-| 11 | 5.11 wire the dead demotion triggers | [#232](https://github.com/0xfauzi/kstrl/issues/232) | `[ ]` |
+| 11 | 5.11 wire the dead demotion triggers | [#232](https://github.com/0xfauzi/kstrl/issues/232) | `[x]` merged in #350 |
```
0xfauzi, line 1031: **Row 11 is marked `[x] merged in #350` on the branch of the not-yet-merged #350.**

If the PR is revised or closed, the tracker on `main` never carried the false state - but on this branch, and in any review or reland built from it, the doc asserts a merge that has not happened. The neighbouring convention is a tracker updated in the same PR, so the row belongs here; the *merged* claim does not.

`[x]` alone, or the PR number without "merged", keeps the row honest at every point where it is read.

kstrl/pipeline.py
```diff
@@ -1688,7 +1689,11 @@ def _inbox_add(
                     item.title,
                     component_id=component,
                 )
-        except (OSError, ValueError) as exc:
+        except (OSError, ValueError, ControlStateError) as exc:
```
0xfauzi: The `ControlStateError` widening stopped one method short. `_inbox_resolve` at `pipeline.py:1634` is the other inbox WRITER in this class and still catches `(OSError, ValueError, InboxError)`.

`Inbox.resolve` -> `_decide` -> `_append` -> `control_lock`, so it raises exactly what this hunk was widened for. `ControlUnavailableError` / `ControlLockedError` are `ControlStateError` (a `RuntimeError`) and are not `InboxError`, so they walk straight past that tuple.

Scenario: the XDG control dir is not writable (read-only `XDG_STATE_HOME`, a full volume, a root-owned `control.lock` left by a sudo run). A component reaches `merge_state == "merged"` at `pipeline.py:1971`; the manifest already records the merge, and `self._inbox_resolve(f"merge:{comp.id}", ...)` then raises out of the transition and fails the run on bookkeeping. That is exactly what `_inbox_add`'s own docstring promises cannot happen ("Bookkeeping must not be able to fail a run").

Same hole, third site: `serve.py:2218` catches `(OSError, ValueError, KeyError)` under a docstring that says "An inbox write must not be able to undo a queue transition that already happened". The PR body's "four copies ... closed here" count is short by at least two.

Fix: add `ControlStateError` to the tuple at line 1634 (and to `serve.py:2218`).

<sub>via /code-review xhigh</sub>

kstrl/calibration_ladder.py
```diff
@@ -0,0 +1,250 @@
...
+    failures = list(comparison.failures)
+    comparison_id = _comparison_id(comparison)
+    evidence: dict[str, Any] = {
+        "failures": failures,
+        "comparison_id": comparison_id,
+        "new_baseline": comparison.new.timestamp,
+        "old": str(comparison.old.path),
+        "new": str(comparison.new.path),
+    }
+    _open_drift_item(root_dir, evidence)
```
0xfauzi, line 259: `_open_drift_item` is called here with no `try` around it, and its own guard on line 165 catches `(OSError, ValueError, ControlStateError)` - no `TypeError`.

`InboxConfig.load` does `int(section["open_item_cap"])` and `float(section["snooze_hours"])`. A TOML date, array or table in either raises `TypeError`, not `ValueError`. `report_to_ladder` catches `TypeError` twenty-five lines up (line 190) for exactly this reason - its comment says "a per-key cast raises TypeError for a TOML date or array" - and the same reasoning is not applied one function over.

Scenario:
```toml
[autonomy]
enabled = true
[inbox]
open_item_cap = 1979-05-27
```
`python -m kstrl.calibration compare old new --root .` on a regression: `AutonomyConfig.load` succeeds, `_open_drift_item` -> `InboxConfig.load` -> `TypeError: int() argument must be a string ... not 'datetime.date'` -> escapes `report_to_ladder`, escapes `calibration.main`, raw traceback.

This breaks the module docstring's stated contract ("Everything AFTER the config resolves is reported and never changes the exit code") and the runbook paragraph this PR adds. Add `TypeError` to line 165's tuple.

<sub>via /code-review xhigh</sub>

kstrl/factory.py
```diff
@@ -2794,58 +3011,29 @@ def _infra_casualty(comp_id: str) -> bool:
...
+            DemotionTrigger.POLICY_VIOLATION,
+            f"policy violation in {', '.join(sorted(violations))}",
+            evidence={"components": sorted(violations), "run_id": run_id},
+            run_id=run_id,
+            ui=ui,
+            bus=bus,
+            state=state,
+        )
+    else:
         state.save(root_dir)
```
0xfauzi: This `state.save(root_dir)` defeats the degraded-state protection added in this PR.

`AutonomyState.load` fails closed to a fresh L1 with `degraded_reason` set, and `apply_demotion` (autonomy.py:967) now refuses to write that fresh state back "because saving a fresh L1 over damaged bytes destroys the only thing an operator could have repaired". But the guard sits only on the policy-violation-AND-at-floor path. The `else` branch here - a run with no policy violations, the overwhelmingly common one - loads the degraded state, counts a decisive run and merged components onto it, and saves it straight over the damaged file.

Scenario: `autonomy.json` is truncated by a crash. The next `ks factory` run has no policy findings: `load` warns "rejected ladder state ... failing closed to L1", `record_decisive_run()` and `record_merged_component()` mutate the fresh L1, and this line writes `{"level": 1, "decisive_runs_at_level": 1, ...}` over the damaged bytes. The operator now has nothing to repair, the history is gone, and `safemode` stops reporting the degradation because the next `load` finds a clean file. The new mechanism never fires on the path that actually reaches it.

The fix belongs one level down - `AutonomyState.save` refusing to write a state whose `degraded_reason` is set - rather than in one branch of one caller; at minimum repeat the `if state.degraded_reason is None` check here.

<sub>via /code-review xhigh</sub>

kstrl/autonomy.py
```diff
@@ -844,6 +919,98 @@ def commit_transition(
...
+                run_id=run_id,
+                dedupe_key=f"demotion:{run_id}:{record.to_level}",
+                evidence={
+                    "trigger": record.trigger,
+                    "from_level": record.from_level,
+                    "to_level": record.to_level,
+                    **record.evidence,
+                },
+            )
+    except (OSError, ValueError, ControlStateError) as exc:
```
0xfauzi: Same `TypeError` hole as `calibration_ladder.py:165`, and here it costs the property this function exists to hold.

`InboxConfig.load` raises `TypeError` (not `ValueError`) for a per-key cast fault - `[inbox] open_item_cap = 1979-05-27`, `snooze_hours = [1]`. That escapes this tuple, but `commit_transition` has ALREADY saved the demotion three lines above.

Scenario: `[inbox] open_item_cap` is a TOML date, a policy violation fires. `state.demote()` -> `commit_transition` writes L3 -> L2 to disk and the journal -> `InboxConfig.load` raises `TypeError` -> escapes to the factory's `except Exception` at factory.py:4403, which warns "Autonomy state update failed (non-fatal)". The level is revoked, no notice was opened, and the `Autonomy DEMOTED L3 -> L2` line at line 1006 never printed. The docstring's "An inbox failure is warned about, never fatal: losing the notice must not strand the ladder with a saved demotion and no record of why" is exactly what happens.

`factory._open_health_breach_items` has the identical omission at both of its guards (factory.py:2823 and 2848), where it silently drops every health item and skips the demotion entirely. One fix: add `TypeError` to all three tuples.

<sub>via /code-review xhigh</sub>

kstrl/calibration_ladder.py
```diff
@@ -0,0 +1,250 @@
...
+
+    So: the two timestamps when both are real, and otherwise a digest of
+    the regression itself. The digest says in the key how it was derived,
+    and it dedupes on what was MEASURED, which is the property the
+    timestamps were standing in for.
+    """
+    old_ts = comparison.old.timestamp
+    new_ts = comparison.new.timestamp
+    if UNKNOWN_TIMESTAMP in (old_ts, new_ts):
+        digest = hashlib.sha256("\n".join(comparison.failures).encode("utf-8")).hexdigest()
```
0xfauzi: The digest branch violates the pair-identity property stated four lines above it.

`_comparison_id`'s docstring: "A comparison is a pair, so `compare old_A new` and `compare old_B new` are different regressions and must not collapse onto one item." That holds in the timestamp branch. The digest branch hashes only `comparison.failures`, which names roles, rates and thresholds but neither baseline, so any two comparisons whose failure text matches share one id.

Scenario: two captures of the same prompt, both saved without a `timestamp` key (which is the whole reason this branch exists), both regressing security to 0.60 against the same old baseline. `compare old new_a --root .` demotes and opens the item. `compare old new_b --root .` produces byte-identical failure lines, `_already_demoted_for` matches, and the second genuine regression neither demotes nor gets its own inbox row - it just bumps the first one's occurrence count. Same collapse for `compare old_A new` / `compare old_B new` when the two old baselines have equal rates.

`test_timestampless_baselines_do_not_share_an_identity` passes only because it plants two structurally different regressions; it does not discriminate this case.

Fix: fold identity into the hash - the two baselines' file digests, or `str(comparison.old.path)`/`str(comparison.new.path)` alongside the failure lines.

<sub>via /code-review xhigh</sub>

kstrl/inbox.py
```diff
@@ -529,6 +542,8 @@ def add(
             existing.last_seen_at = _iso(now)
             if detail:
                 existing.detail = detail
+            if evidence:
+                existing.evidence = evidence
```
0xfauzi, line 549: `evidence` now refreshes alongside `detail`, which is right, but `run_id` is the third description of the same observation and it still holds the FIRST occurrence.

The reasoning in the new docstring - "They are two descriptions of the same observation, and refreshing only the prose left the structured half ... reporting the first occurrence while the text reported the latest" - applies verbatim to `run_id`. `title` is argued for explicitly (it is the row's label); `run_id` is not mentioned at all, so this reads as an omission rather than a decision.

Scenario, from this PR's own emitter: `_open_health_breach_items` dedupes on `health:{metric}:{rule}` with no run id, and passes `run_id=run_id` on every repeat. Run 1 records `retry_rate` at 0.4 over 8 runs, `run_id="run-1"`. Run 20 records 0.9 over 20 runs; `detail` and `evidence` update, `run_id` still says `run-1`. `ks inbox show` then points the operator at a nineteen-run-old run for numbers that came from the latest one - and `run-1`'s artifacts may well be garbage-collected.

Either refresh `run_id` with the rest, or state in the docstring why it is pinned to the first occurrence the way `title` is.

<sub>via /code-review xhigh</sub>

kstrl/autonomy.py
```diff
@@ -652,11 +710,28 @@ def load(cls, root_dir: Path | None = None) -> AutonomyConfig:
         defaults = cls()
         enabled = bool(section["enabled"]) if "enabled" in section else defaults.enabled
```
0xfauzi: Two definitions of "how `[autonomy]` reads a boolean" now live in one method, and the weaker one is on the key that gates everything.

`_strict_bool`'s own argument is "a typo that ARMS a safety switch is worse than one that disarms it". `enabled` is the switch that arms the whole ladder - `demote_on_calibration_regression` and `demote_on_health_breach` do nothing without it, and `report_to_ladder` / `_record_autonomy_outcome` are both gated on it. Yet `enabled = "false"` still evaluates `bool("false")` -> `True` here.

Scenario: an operator writes
```toml
[autonomy]
enabled = "false"
demote_on_health_breach = true
```
intending the ladder off while they experiment with the health rules. `enabled` coerces to True, `demote_on_health_breach` is a real boolean and passes `_strict_bool`, and once #151 lands the factory starts revoking levels in a repo the operator believes has the ladder switched off. The exact failure the new helper was written to prevent, one key over.

The deferral to "29 sites repo-wide" is reasonable for the other 28; this one is in the same seven-line block, guards the two keys that were just tightened, and can only ever fail in the arming direction (a bogus value reads True). Route `enabled` through `_strict_bool` too.

<sub>via /code-review xhigh</sub>

kstrl/calibration_ladder.py
```diff
@@ -0,0 +1,250 @@
...
+        if _already_demoted_for(state, comparison_id):
+            print(f"autonomy: comparison {comparison_id} already cost a level; not demoting again")
+            return None
+        apply_demotion(
+            root_dir,
+            DemotionTrigger.CALIBRATION_REGRESSION,
+            failures[0],
+            evidence=evidence,
+            run_id=f"calibration:{comparison_id}",
+            ui=PlainUI(no_color=True),
```
0xfauzi, line 283: `PlainUI` defaults to `file=sys.stderr`, so `apply_demotion`'s `Autonomy DEMOTED L3 -> L2 ... cool-down 10 decisive run(s)` line lands on stderr - while every other ladder outcome in this module is a bare `print` to stdout: `LADDER_DISABLED_LINE` (line 200), "arguments look reversed" (line 203), "already cost a level" (line 234).

Scenario: CI runs `python -m kstrl.calibration compare old new --root . > compare.log`. `compare.log` records "autonomy ladder disabled" on the runs where nothing happened, and says nothing at all on the run that actually revoked a level - that line went to stderr, which the redirect dropped. The most consequential of the four outcomes is the only one missing from the channel the other three chose.

The module docstring frames all four as one vocabulary ("a stated outcome rather than silence"). Either pass `PlainUI(no_color=True, file=sys.stdout)` here, or route the three `print` calls to stderr - but not one of each.

<sub>via /code-review xhigh</sub>

kstrl/factory.py
```diff
@@ -2731,11 +2733,223 @@ def _expired_futures(
...
+        inbox_config = InboxConfig.load(root_dir)
+    except (OSError, ValueError, ControlStateError) as exc:
+        ui.warn(f"Inbox write failed (non-fatal): {exc}")
+        return
+    if not inbox_config.enabled:
+        return
+    inbox = Inbox(root_dir, inbox_config)
+    for breach in breaches:
+        try:
+            inbox.add(
```
0xfauzi, line 2837: This is the repo's first loop over `Inbox.add`, and `Inbox.path` is an unmemoised property.

`Inbox.path` -> `control_file` -> `control_dir` -> `repo_id` -> `_origin_url`, which runs `subprocess.run(["git", "-C", ..., "remote", "get-url", "origin"], timeout=5)`. Every `add` derives that path at least four times (`scan().exists`, `scan().read_bytes`, `_append`'s `ensure_control_state`, `_append`'s `open`), and `find_by_dedupe_key` -> `_folded` -> `scan` re-reads and re-parses the WHOLE inbox log per breach, so the loop is O(N) git spawns and O(N^2) line parses in the number of breaches.

The PR body already measures it: 1 breach = 106 ms / 24 spawns, 10 = 604 ms / 141, 20 = 1122 ms / 271, linear in breaches - and then leaves it as a handoff. #151's rule set is precisely what makes N > 1, and this loop is what turns a pre-existing per-call cost into a per-run cost on the factory's terminal path, inside the factory lock.

The fix named in the PR body is one line: resolve the path once in `Inbox.__init__` (as `statedir.xdg_state_home` already caches for the same reason). Doing it in the PR that creates the loop is cheaper than doing it after.

<sub>via /code-review xhigh</sub>

kstrl/autonomy.py
```diff
@@ -844,6 +919,98 @@ def commit_transition(
...
+        else:
+            # ``load`` fails closed to a fresh L1 when the stored record
+            # is damaged, and those bytes are the only thing an operator
+            # could repair. Saving here replaces them with an empty
+            # ladder, and the counters that save would carry were lost
+            # with the file they were read from.
+            ui.warn(
+                f"Autonomy: ladder state is degraded ({state.degraded_reason}); not overwriting it"
+            )
+        ui.warn(f"Autonomy: {trigger_text} recorded ({reason}); already at L1, nothing to revoke")
```
0xfauzi, line 1036: `inbox.notifiable()` returns action-required kinds **plus `DEMOTION_NOTICE`** - the one informational kind singled out for paging - and `pipeline._inbox_add` honours that with `notify.fire_inbox_item` when `config.notify_action_required` is set. This block has no `Notifier` and never pushes.

Before this PR the gap covered one trigger. `apply_demotion` is now the single applier for POLICY_VIOLATION, CALIBRATION_REGRESSION and HEALTH_BREACH, so it covers every automatic demotion the factory can make, and the docstring's claim that this function exists so "the four writes it performs ... cannot drift apart per trigger" is what makes the missing fifth write conspicuous.

Scenario: an operator sets `[inbox] notify_action_required = true` and a notify hook. A health breach revokes L3 -> L2 overnight. They are paged for a `halted_run` an hour later but never for the demotion, which is the boundary condition the R8.2 design says an over-the-loop operator must see.

The PR body names this as a handoff ("threading a `Notifier` through five call sites"). An `ui`-style optional `notifier: Notifier | None = None` parameter defaulted to None costs one keyword at four of the five sites.

<sub>via /code-review xhigh</sub>

kstrl/calibration_ladder.py
```diff
@@ -0,0 +1,250 @@
...
+        # ValueError) and deliberately does NOT normalise OSError; a
+        # per-key cast raises TypeError for a TOML date or array; and
+        # __post_init__ raises AutonomyError, a RuntimeError, for a
+        # max_level outside the ladder.
+        print(f"error: {exc}", file=sys.stderr)
+        return 2
+    if not config.enabled:
+        print(f"{LADDER_DISABLED_LINE} (root: {root_dir})")
+        return None
+    if _arguments_reversed(comparison):
```
0xfauzi, line 243: `_arguments_reversed` returning True skips everything below it - including `_open_drift_item`, which is the ADVISORY tier that fires with the demotion switch off.

The justification given is specifically about the level: "That was harmless while the command only printed; it costs a level now." That argues for skipping the demotion. It does not argue for withholding the record.

Scenario: an operator with the ladder on and `demote_on_calibration_regression = false` (the default, and the tier this PR calls "advisory first") types the arguments the wrong way round. The command still exits 1, still prints the regression report, and the only durable trace - the `calibration_drift` item - is never written. `test_reversed_arguments_do_not_consult_the_ladder` pins `inbox_items(tmp_path) == []`, so the behaviour is deliberate but the docstring never says why the advisory half is included.

Either move the check down to guard only the demotion block at line 220, or state in the docstring why an order artefact should not be recorded at all.

<sub>via /code-review xhigh</sub>

kstrl/factory.py
```diff
@@ -2731,11 +2733,223 @@ def _expired_futures(
...
+    _open_health_breach_items(root_dir, breaches, run_id=run_id, ui=ui)
+    if not autonomy_config.demote_on_health_breach:
+        return
+    if state.cooldown_runs_remaining > 0:
+        ui.warn(
+            f"Autonomy: {len(breaches)} health breach(es) recorded; not demoting during "
+            f"cool-down ({state.cooldown_runs_remaining} decisive run(s) remaining)"
+        )
+        return
+    first = breaches[0]
```
0xfauzi, line 2928: `first = breaches[0]` names one breach in the durable reason (`f"{first.metric}: {first.rule}"`), which reaches the transition record, the evolution journal, the run event stream and the inbox notice title.

Nothing in the #232/#151 contract says `health_breaches(root_dir)` returns an ordered sequence. `_HealthBreach` describes the record, not the collection, and `list(breaches_of(root_dir))` accepts any iterable - a `set` comprehension over breached metrics is the obvious implementation and its order varies with `PYTHONHASHSEED`.

Scenario: #151 returns `{retry_rate breach, review_latency breach}` as a set. Two runs on the same persistent trend record two different reasons, and an operator comparing the journal to the inbox (which has one item per breach, all of them) cannot tell whether the metric changed or the ordering did.

The evidence payload already carries all of them. Either sort deterministically (`min(breaches, key=lambda b: (b.metric, b.rule))`) or make the reason name the count plus the metrics, e.g. `f"{len(breaches)} breach(es): " + ", ".join(sorted(b.metric for b in breaches))`.

<sub>via /code-review xhigh</sub>

kstrl/calibration_ladder.py
```diff
@@ -0,0 +1,250 @@
...
+    ``evidence`` dict rather than passed separately, so an item cannot
+    describe the regression three ways. That holds only because the key
+    is the identity of the comparison the payload came from, and because
+    ``Inbox.add`` refreshes ``evidence`` alongside ``detail`` on a
+    repeat: with either half missing, one item ends up carrying the title
+    of one comparison and the numbers of another.
+    """
+    failures: list[str] = evidence["failures"]
+    comparison_id: str = evidence["comparison_id"]
+    try:
```
0xfauzi, line 190: Four near-identical "load config, check enabled, construct, add, swallow" blocks now exist - `pipeline._inbox_add`, `autonomy.apply_demotion`, `factory._open_health_breach_items` and this one - and this PR wrote two of them.

They have already diverged on three axes:
- caught set: `(OSError, ValueError, ControlStateError)` here and in three others, `(OSError, ValueError, InboxError)` in `_inbox_resolve`, `(OSError, ValueError, KeyError)` in `serve.py:2218`, and none of the six catches the `TypeError` `InboxConfig.load` can raise
- channel: `ui.warn` in three, `print(..., file=sys.stderr)` here
- notify push: `pipeline._inbox_add` fires `notify.fire_inbox_item`, the other three do not

That is not a hypothetical cost - two of the findings on this PR are one copy having a guard the others lack. The PR body records the duplication as a follow-up on the grounds that "the one thing that made that dangerous ... is closed here by giving them one shared name to catch". It was not closed: `_inbox_resolve` and `serve.py` never got the shared name.

One `inbox.emit(root_dir, kind, title, *, detail, run_id, dedupe_key, evidence, warn, notifier=None) -> None` in `kstrl/inbox.py` collapses all six call sites and makes the exception set a single edit. Per CLAUDE.md's "One defect class, one PR", this is the fix that stops the class rather than the sixth instance of it.

<sub>via /code-review xhigh</sub>

docs/dark-factory-roadmap.md
```diff
@@ -319,7 +319,10 @@ run reaches the journal only). Opt-in via
 `[autonomy] enabled` (default false) because L1 is STRICTER than today's
 defaults - it forces the merge gate on. R8.4 will enrich the demotion
 triggers with health-metric breaches; the `HEALTH_BREACH` trigger already
-exists for it to fire.
+exists for it to fire, and since
```
0xfauzi, line 322: This PR adds `ItemKind.HEALTH_BREACH` and edits this file in two places, but leaves the R8.3 item-kind enumeration at line 427 stale:

> item kinds policy_exception / merge_gate / halted_run / budget_overrun / demotion_notice / calibration_drift / test_adequacy

That list is the doc's canonical inventory of the kinds, and the paragraph under it ("Emitters are wired, not declared") enumerates the emitter for each one - which is precisely the claim `health_breach` now needs to make, along with the fact that its emitter is inert until #151. `docs/env-vars.md` and `kstrl.toml.example` were both updated for the config keys; this is the one surface that names the kinds, and it now under-reports by one.

Cost: the next reader auditing "every kind has an emitter" against this list will not find `health_breach`, and CLAUDE.md's audit-trail rule ("if it's worth deciding, it's worth recording") is what this line is for.

<sub>via /code-review xhigh</sub>

tests/test_calibration_ladder.py
```diff
@@ -0,0 +1,374 @@
...
+
+        code = run_compare(tmp_path, old, new)
+
+        assert code == 2
+        err = capsys.readouterr().err
+        assert err.startswith("error:")
+        assert fragment in err
+        assert inbox_items(tmp_path) == []
+
+    @pytest.mark.skipif(os.geteuid() == 0, reason="root reads a 000 file")
```
0xfauzi: `@pytest.mark.skipif(os.geteuid() == 0, ...)` is evaluated at import time, and `os.geteuid` does not exist on Windows - so this raises `AttributeError` during COLLECTION, taking the whole module down rather than skipping one test.

The repo already has the convention in two places and it is spelled the other way in both:
- `tests/test_journal_torn_tail.py:417`: `hasattr(os, "geteuid") and os.geteuid() == 0`
- `tests/test_progress_scope.py:1618`: same

(`tests/test_knowledge.py:283` calls it inside a test body, where an `AttributeError` costs one test rather than a module.)

CLAUDE.md's gotchas section is explicit that POSIX-only machinery gets a skip rather than a crash ("`fcntl.flock` ... is POSIX-only; tests skip on Windows"). Cost is a collection error instead of a skip, and the fix is the four-token guard the two siblings already use.

<sub>via /code-review xhigh</sub>

PR #351, opened 2026-09-06
test(events): the event-name guard walks tests/ too, and the test-side copy of the spec-audit reader goes (#337)

### LINKED ISSUE #337: Journal event-name guard does not cover tests/, and a test helper re-implements the window rule it is asserting on
Found while reviewing PR #336 (issue #314). Both parts are pre-existing and were deliberately left out of that PR's scope.

## The event-name guard's scope is narrower than its claim

`tests/test_evolution.py` grew a guard whose class docstring says the journal event name is "spelled once, in `kstrl/evolution.py`, and imported everywhere else". The walk it runs covers `package_sources()`, which is `kstrl/` only. It never looks at `tests/`.

Four literal sites remain there, found by running the guard's own detector over the test tree:

```
tests/test_decompose.py:805
tests/test_decompose.py:1037
tests/test_decompose.py:1132
tests/test_evolution.py:1055
```

Either the claim should be narrowed to `kstrl/`, or the walk should cover `tests/` and these four should import the constant.

## `_read_spec_issue_events` re-implements the rule PR #336 consolidated

`tests/test_decompose.py:802-805` is the sharper half. PR #336 exists to give the journal window rule one home. This helper is a second implementation of the selection rule sitting in the test tree, so a test asserting on production behaviour is partly asserting on its own copy of that behaviour. It should read through `EvolutionJournal.get_spec_audits` like production does.

It also reads a file kstrl wrote as `journal.read_text()` with no `encoding="utf-8"`.

Per the encoding contract in CLAUDE.md, a reader of any file kstrl writes must name `encoding="utf-8"` and must catch `ValueError` alongside `OSError`, because `UnicodeDecodeError` is a `ValueError` and escapes a fail-closed `except OSError`. This is test code, so the blast radius is a confusing failure under a non-UTF-8 locale rather than a production fault, but it is the same defect `#320` tracks in the production readers.

## Verification

- Run the `literal_event_names` detector from `tests/test_evolution.py` over `tests/` rather than `package_sources()`; it reports the four sites above.
- `grep -n 'read_text()' tests/test_decompose.py`

0xfauzi:
## What

The journal event-name guard's layer 2 now walks `tests/` as well as `kstrl/`, the six literal sites the widened corpus found import `SPEC_ISSUES_EVENT`, and the test tree's copy of a production reader's selection rule is gone. Round 1 of review then closed the corpus itself, corrected three claims that were wider than their walk, and put back the two properties the conversions removed.

- `tests/test_event_names_have_one_home.py`: layer 2 loops over `package_sources() + astwalk.test_sources(exclude=Path(__file__))`, having first asserted that a positive number of the modules it was handed are under `tests/`. Layer 1 stays on `kstrl/`, and each layer's prose is now exactly the corpus it walks.
- Six sites import the constant: `test_decompose.py`, `test_evolution.py:1047` and `test_journal_torn_tail.py:88` and `:142`, plus the read inside the helper deleted below.
- `tests/test_decompose.py`: `_read_spec_issue_events` is deleted. It re-implemented `EvolutionJournal.get_spec_audits`, carrying its own copy of the journal path, the JSONL parse and the `event_type == SPEC_ISSUES_EVENT` selection rule. Its callers read through `journal_at(tmp_path).get_spec_audits()`, except the two questions that are about the bytes on disk, which read the bytes.
- `tests/helpers/journal.py`: `audits_in(path)` becomes `audits_in(journal)` and reads through `get_spec_audits`. `repair_rows_in` takes the journal too, for a different reason: there is still no production reader to route it through, but a path parameter made six call sites reach through `journal.config.journal_path`.
- `tests/helpers/astwalk/corpus.py`: `test_sources` carries `__test__ = False`, which deletes the collection hazard instead of documenting it.

## Why

#337. The guard's docstring claimed an enrolled event name is "spelled once, in `kstrl/evolution.py`, and imported everywhere else", and the walk behind that claim covered `package_sources()` only, so a claim about the repository was enforced over a third of it. `Closes #337`.

The sharper half is the second one. A test that spells the event name for itself while asserting on production behaviour is asserting partly on its own copy of that behaviour, and `_read_spec_issue_events` was the worst case: a second implementation of the reader whose consolidation #336 exists to deliver, sitting in the tests that check it.

Layer 1 was deliberately not widened. Measured over `tests/` it counts 46 spellings in 13 modules, 24 of them in `test_decompose.py` where they are the architect's own JSON key inside fixture payloads. A pinned per-module dict over that moves on every fixture that adds an architect payload, which is the guard-that-gets-silenced its own `_assignment_hits` docstring warns about. The enrolment comment now says that, instead of claiming a coverage over `tests/` that neither layer provides.

## What round 1 of review changed

Round 1 found the PR's own subject unprotected and three claims wider than their walk. All seven were measured before and after.

1. **The corpus had no control.** Deleting `astwalk.test_sources()` from layer 2's loop left the guard file at 2 passed and the full suite at 5564 passed, byte-identical. `assert found == {}` reads the same whether the walk is clean or switched off, which is CLAUDE.md's rule (2) and this repo's eleventh instance of the class. Layer 2 now names its corpus, asserts a positive count of `tests/` modules in the list it is about to iterate, then iterates that same list: closed by construction, no threshold to maintain, and a code move elsewhere cannot drift it.
2. **The enrolment comment claimed `tests/`, which neither layer enforces.** Layer 1 is `kstrl/`-only and layer 2 sees only the shape attached to the `event_type` column, so an indirect spelling in a test module is invisible to both. Measured: the plant in `test_evolution.py` gives 2 passed, the byte-identical plant in `kstrl/observability.py` gives 1 failed. The comment now states each layer's actual scope.
3. **`event_name_spellings` was a third raw `ast.walk` that layer 1 never runs.** `assert_census` counts through `astwalk.census`, and this function's only callers are the shape controls next door, so the controls were certifying a copy. It is one row of `census` now. The discriminating mutation is below.
4. **The PR removed the last behavioural pin on the wire value.** Four of the six converted sites used to cross the constant/literal boundary; after the conversion both sides of each are the constant. One test pins the value against the bytes on disk, in a shape layer 2 does not flag.
5. **The tolerant reader replaced a strict one.** `get_spec_audits` reads through `read_progress_events`, which skips a line it cannot parse; the deleted helper raised on one. Two tests whose subject is the writer now pin the row count off the file as well as the parsed count, and the `run_id` test reads the bytes, because a reader that dropped an unknown key could answer its question wrongly.
6. **`repair_rows_in`'s parameter**, and **layer 2's mixed labelling roots** (`label(source_file, REPO_ROOT)`, so `evolution.py` no longer keys as a repo-root file that does not exist).
7. **Two stale measurements**, both re-derived by running: the alias sweep's "1 of 127 modules in `kstrl/`" sized a walk that now covers both corpora, and the memo timing was wall clock on a loaded machine.

## What round 2 of review changed

Round 2 verified all nineteen round-1 items and found four more, all in the guard file and its helpers; the orchestrator applied them directly.

1. **The corpus control was closed over one of three failure shapes.** The round-1 form asserted a positive count over the `sources` binding and then iterated it: it caught the corpus being deleted, but round 2 measured that slicing the corpus to one module (`[:1]`) and repointing the loop at `package_sources()` while the binding stayed whole both left the guard green. The loop now increments a counter for every `tests/` module it opens and the assertion after it compares that counter with the corpus size derived independently; all three shapes are red (deleted: 1 failed; sliced: 1 failed; repointed: 1 failed), and the docstring no longer says "closed by construction" of a check that was coupled by adjacency.
2. **The opening docstring sentence was false in two ways.** The 46 spellings are 44 of `spec_issues` and 2 of `journal_repair`, and five of them are the TUI's artifact label or a guard's own fixture argument rather than the architect's JSON key. Rewritten to the measured breakdown.
3. **`test_sources.__test__ = False` had no control**; deleting the line failed nothing across seven guard files. `tests/test_astwalk_nets.py` now asserts it, next to the exclude test: deleting the line is 1 failed.
4. **`_journal_rows` said "parsed STRICTLY" and accepted a non-object line**, so a planted `[1, 2]` row left the `run_id` test green (`"run_id" not in [1, 2]` is true). The helper now raises on a line that is not an object, and the planted row is red.

Every plant above was run with `__pycache__` deleted and `PYTHONDONTWRITEBYTECODE=1`, and the tree restored from byte copies. Full suite after round 2, on the tree merged with `origin/main` at 50c965e: 5611 passed, 37 skipped, 38 xfailed in 337s, exit 0; mypy strict, ruff, gen_docs and the four precommit ratchets clean.

## Tested vs assumed

Tested. Every number below was produced by running the thing, in this worktree, with `__pycache__` deleted first and `PYTHONDONTWRITEBYTECODE=1`.

Gates on the merged tree (`origin/main` dee0c94 merged in, clean):

- `uv run pytest tests/ -q -p no:cacheprovider`: 5593 passed, 37 skipped, 38 xfailed, exit 0, 325 s. On the branch before the merge: 5565 / 37 / 38, exit 0, 333 s, against the 5564 / 37 / 38 baseline at 568bca4. The delta of one is the wire-value pin, the only test this PR adds.
- `uv run mypy kstrl/ --strict`: success, 129 source files. `uv run ruff check kstrl/ tests/`: clean. `uv run ruff format --check kstrl/ tests/`: 340 files already formatted. `uv run python scripts/gen_docs.py --check`: current.
- `scripts/precommit/`: the full hook set ran on each commit and passed, including the file-length ratchet, the cyclomatic ratchet, complexipy, codespell, vulture and the em-dash check. Zero em dashes added by the diff. The guard file is 686 lines against the 800 ratchet; `test_decompose.py` is 2910.

Guard census, re-derived by executing the detectors in both worktrees, never by diffing a literal:

| corpus | modules | layer 2 sites | layer 1 spellings | modules binding an alias |
|---|---|---|---|---|
| `kstrl/` at 568bca4, at the PR head and merged | 129 | 0 | 6 in 3 | 1, 6 names |
| `tests/` at 568bca4 | 204 | 6 (3 writes, 3 reads) | 53 in 14 | 4, 10 names |
| `tests/` after this PR, before the merge | 204 | 0 | 46 in 13 | 4, 10 names |
| `tests/` after the merge with dee0c94 | 206 | 0 | 46 in 13 | 4, 10 names |
| `tests/` after the merge with 50c965e (#349) | 206 | 0 | 46 in 13 | 5, 11 names |

The first merge added two test modules and neither spells an enrolled name or binds an alias. The second merge (#349) grows `tests/test_review_gates.py`, which now binds `matches` as an event-type alias; the alias sweep's docstring was re-derived on that tree. Layer 1 and layer 2 are unchanged across both. The `component_result` counts in the handoff below come from the guard's own layer-2 detector, not from grep, which gives 5 and 27 for the same population.

Mutations, one per mechanism, each plant asserting its anchor occurred exactly once so a silent no-op would report COULD-NOT-PLANT rather than a pass. None hung, none failed to plant, and each was restored from a byte copy taken before the edit rather than with `git checkout`:

| mutation | layer | outcome |
|---|---|---|
| M0, clean | both | 46 passed |
| M1, `"spec_issues"` back at `test_evolution.py:1047` | 2 | CAUGHT, message names line 1047 and the direction |
| M2, both literals back at `test_journal_torn_tail.py:88` | 2 | CAUGHT, message names line 88 and the list |
| M3, M1 planted AND the corpus narrowed to `package_sources()` | 2, corpus | CAUGHT. Before this round it was STILL GREEN, which is what the corpus control changes |
| M4, the corpus narrowed alone, nothing else | 2, corpus | CAUGHT, 1 failed 45 passed, message reports 129 modules and names the missing directory |
| M5, a row of layer 1's pinned dict changed | 1 | CAUGHT |
| M6, `spells_an_enrolled_event()` stubbed to always-False | 1 and the shape controls | CAUGHT, 9 failed 37 passed |
| M7, `astwalk.census` stubbed to return `{}` | the shape controls, over the executed path | CAUGHT, 8 failed 36 passed |
| M7b, the same census stub with the OLD raw-`ast.walk` `event_name_spellings` restored | same | STILL GREEN, 44 passed. This pair is the finding: the controls used to certify a copy |
| M9, a `from tests.helpers.astwalk import test_sources` | collection | 2 collected with `__test__ = False`, 3 without, and the third PASSES on a warning |
| M11, the writer emits a malformed row after every entry | strict row read | CAUGHT, 2 failed |
| M11b, the same writer mutation with the strictness pin removed | same | STILL GREEN, 2 passed. The tolerant reader skips the row and the count still comes out right |
| M12, the writer emits `run_id` and the reader strips it | byte read | CAUGHT, 1 failed |
| M12b, the same with the test reading through `get_spec_audits` | same | STILL GREEN, 1 passed |
| M13, `SPEC_ISSUES_EVENT` renamed to `spec_audit_row` | wire-value pin | CAUGHT, 1 failed |
| M14, the production tear repair deleted | torn-tail behaviour | CAUGHT, 15 failed 14 passed 1 xfailed, the same count and the same names as before the reroute |
| M15, `get_spec_audits` returns every entry unfiltered | the `audits_in` reroute | CAUGHT, 3 failed, which is coverage this PR gains |

M1, M4, M7 and M13 were re-run on the merged tree and all four stay CAUGHT.

Assumed, not tested: nothing. No claim in this body is relayed from the plan or from review without being re-run here.

## Disclosed, measured, not fixed here

Layer 2's corpus contains `tests/test_event_name_shapes.py`, whose whole subject is the shapes this layer forbids. `exclude=` takes one path and the exposure is a sibling, so it is disclosed in the test's docstring instead: those controls are clean today only because each is a source string that `folded_str` folds whole, and the first control written as real code makes this guard fail on its own positive-control file. That is the point at which layer 2 needs a named allowlist, and the same is true from the other direction the day `component_result` is enrolled, because `tests/test_decompose.py:2460` writes a deliberate legacy on-disk row with it. Enrolling the five names this file commits to is a 43-site job by the guard's own detector: `component_result` alone is 9 in `kstrl/` and 23 across 8 test modules. Two fixtures in `tests/test_evolve_screen_encoding.py` also spell event names inside bytes literals, which `folded_str` folds to `None`, so neither layer sees them; there are zero such fixtures in `kstrl/`.

The 46 layer-1 spellings in 13 test modules are unenforced in every shape but the column-attached one. Widening layer 1 to `tests/` would mean a 46-entry per-module dict that moves on every new architect fixture, so the claim was narrowed instead of the walk widened.

Drive-by, stated rather than implied: the three `read_text(encoding="utf-8")` edits in `tests/test_decompose.py` are the issue's own text, not a sweep of #320's class. The issue names that reader specifically, which is why they are here; they are not an attempt to take the class. 172 bare `read_text()` calls remain in `tests/`, and `tests/helpers/encodingwalk.py` still scans `package_sources()` only, so none of them is a site any guard can report. The issue also asked for `ValueError` beside `OSError`: none of the three sites has an `except` clause at all, so there is nothing for a `UnicodeDecodeError` to escape and that half of the contract does not apply.

## Behaviour changes

None in `kstrl/`. Production code is untouched; this is test infrastructure and test helpers only. Three test-facing signature changes: `audits_in` and `repair_rows_in` take an `EvolutionJournal` rather than a `Path`, and `_read_spec_issue_events` no longer exists. One failure message is worse: a missing journal used to fail as `assert journal.exists(), "journal event was not written"`. The two tests where that mattered now read the file first, so a missing journal fails on the read.

No CHANGELOG line: a test guard's corpus and two test helpers are not user-visible. Same precedent as #324, #336, #341, #342 and #344.

### DIFF
--- a/tests/helpers/astwalk/corpus.py
+++ b/tests/helpers/astwalk/corpus.py
@@ -82,16 +82,27 @@ def test_sources(exclude: Path | None = None) -> list[Path]:
     otherwise scan itself. Never empty, for the reason above: excluding
     the caller cannot empty a directory that holds the caller.
 
-    Reach this through the module, ``astwalk.test_sources()``, and never
-    by a from-import: pytest collects any module-level name matching
-    ``test*``, so the import binds a third "test" in the calling module
-    and pytest runs the helper as one. Measured in
-    ``test_event_names_have_one_home.py``: 2 collected becomes 3.
+    The hazard this name carries is a MODULE PYTEST COLLECTS, not a
+    from-import: pytest collects any module-level name matching
+    ``test*``, so a from-import binds an extra "test" there and pytest
+    runs the helper as one. ``__test__ = False`` below is what makes
+    that safe rather than a rule every caller has to remember, which
+    is why the rule is stated as the hazard and not as the import
+    style. ``astwalk/__init__.py`` from-imports this name and must;
+    pytest does not collect ``tests/helpers/``. Measured in
+    ``test_event_names_have_one_home.py``: with a from-import and no
+    ``__test__``, 2 collected became 3 and the third PASSED, emitting
+    only a ``PytestReturnNotNoneWarning`` into a suite that already
+    emits eight. With ``__test__ = False`` the same from-import
+    collects 2.
     """
     skip = exclude.resolve() if exclude is not None else None
     found = [path for path in sorted(TESTS_DIR.rglob("*.py")) if path.resolve() != skip]
     assert found, _EMPTY.format(what="tests/", root=TESTS_DIR)
     return found
+
+
+test_sources.__test__ = False  # type: ignore[attr-defined]
 
 
 #: What an empty corpus means, said once. It is never a real answer: both
--- a/tests/helpers/astwalk/corpus.py
+++ b/tests/helpers/astwalk/corpus.py
@@ -81,6 +81,12 @@ def test_sources(exclude: Path | None = None) -> list[Path]:
     A guard naming the shapes it forbids in its own fixtures would
     otherwise scan itself. Never empty, for the reason above: excluding
     the caller cannot empty a directory that holds the caller.
+
+    Reach this through the module, ``astwalk.test_sources()``, and never
+    by a from-import: pytest collects any module-level name matching
+    ``test*``, so the import binds a third "test" in the calling module
+    and pytest runs the helper as one. Measured in
+    ``test_event_names_have_one_home.py``: 2 collected becomes 3.
     """
     skip = exclude.resolve() if exclude is not None else None
     found = [path for path in sorted(TESTS_DIR.rglob("*.py")) if path.resolve() != skip]
--- a/tests/helpers/journal.py
+++ b/tests/helpers/journal.py
@@ -23,9 +23,15 @@ weakest case for this module and worth revisiting if it stays that way.
 ``audits_in`` takes the journal rather than its path and reads through
 ``EvolutionJournal.get_spec_audits``, so a test asserting on what the
 journal reads back is not asserting on a second copy of that reader's
-selection rule (#337). ``repair_rows_in`` still takes a path, because
-no production reader returns repair rows to route it through:
-``get_repair_count`` returns an int.
+selection rule (#337). ``repair_rows_in`` takes the journal too, for a
+separate reason that round 1 of review on #337 separated out: there is
+no production reader to route it THROUGH, because ``get_repair_count``
+returns an int, but the PARAMETER is a different question. Handing it a
+path made six call sites in ``test_journal_torn_tail.py`` write
+``repair_rows_in(journal.config.journal_path)`` next to
+``audits_in(journal)``, and reaching through ``config.journal_path`` at
+a call site is what ``EvolutionJournal.get_spec_audits``' own docstring
+calls the defect.
 """
 
 from __future__ import annotations
@@ -122,5 +128,15 @@ def audits_in(journal: EvolutionJournal) -> list[str]:
     return [str(entry.get("project")) for entry in journal.get_spec_audits()]
 
 
-def repair_rows_in(path: Path) -> list[dict[str, Any]]:
-    return [e for e in read_progress_events(path) if e.get("event_type") == JOURNAL_REPAIR_EVENT]
+def repair_rows_in(journal: EvolutionJournal) -> list[dict[str, Any]]:
+    """Every repair row this journal's file holds, in order.
+
+    Reads the file rather than a production reader: no reader returns
+    repair rows. The selection rule is this helper's own and is the one
+    copy of it in the suite.
+    """
+    return [
+        entry
+        for entry in read_progress_events(journal.config.journal_path)
+        if entry.get("event_type") == JOURNAL_REPAIR_EVENT
+    ]
--- a/tests/helpers/journal.py
+++ b/tests/helpers/journal.py
@@ -19,6 +19,13 @@ ones with a single caller today. The RECORD BUILDERS and READERS
 entries as the cost file; ``component_result``, ``audits_in`` and
 ``repair_rows_in`` have one caller each at the moment, which is the
 weakest case for this module and worth revisiting if it stays that way.
+
+``audits_in`` takes the journal rather than its path and reads through
+``EvolutionJournal.get_spec_audits``, so a test asserting on what the
+journal reads back is not asserting on a second copy of that reader's
+selection rule (#337). ``repair_rows_in`` still takes a path, because
+no production reader returns repair rows to route it through:
+``get_repair_count`` returns an int.
 """
 
 from __future__ import annotations
@@ -110,12 +117,13 @@ def journal_at(tmp_path: Path) -> EvolutionJournal:
     return EvolutionJournal(EvolutionConfig.load(tmp_path))
 
 
-def audits_in(path: Path) -> list[str]:
-    return [
-        str(entry.get("project"))
-        for entry in read_progress_events(path)
-        if entry.get("event_type") == SPEC_ISSUES_EVENT
-    ]
+def audits_in(journal: EvolutionJournal) -> list[str]:
+    """The projects of every spec audit the journal reads back, in order.
+
+    Through ``EvolutionJournal.get_spec_audits``, not a private copy of
+    its selection rule (#337).
+    """
+    return [str(entry.get("project")) for entry in journal.get_spec_audits()]
 
 
 def repair_rows_in(path: Path) -> list[dict[str, Any]]:
--- a/tests/helpers/journal.py
+++ b/tests/helpers/journal.py
@@ -118,11 +118,7 @@ def journal_at(tmp_path: Path) -> EvolutionJournal:
 
 
 def audits_in(journal: EvolutionJournal) -> list[str]:
-    """The projects of every spec audit the journal reads back, in order.
-
-    Through ``EvolutionJournal.get_spec_audits``, not a private copy of
-    its selection rule (#337).
-    """
+    """The projects of every spec audit the journal reads back, in order."""
     return [str(entry.get("project")) for entry in journal.get_spec_audits()]
 
 
--- a/tests/test_decompose.py
+++ b/tests/test_decompose.py
@@ -1432,11 +1432,17 @@ def _journal_rows(tmp_path: Path) -> list[dict[str, Any]]:
     and drops the copy.
     """
     journal = tmp_path / ".kstrl" / "evolution.jsonl"
-    return [
-        json.loads(line)
-        for line in journal.read_text(encoding="utf-8").splitlines()
-        if line.strip()
-    ]
+    rows: list[dict[str, Any]] = []
+    for line in journal.read_text(encoding="utf-8").splitlines():
+        if not line.strip():
+            continue
+        row = json.loads(line)
+        # Strict in both dimensions: a line that parses but is not an
+        # object would satisfy `"run_id" not in row` and pass silently.
+        if not isinstance(row, dict):
+            raise AssertionError(f"journal line is not an object: {line!r}")
+        rows.append(row)
+    return rows
 
 
 class TestSpecIssuesPersistence:
--- a/tests/test_decompose.py
+++ b/tests/test_decompose.py
@@ -1419,6 +1419,26 @@ def _journal_with(tmp_path: Path, entries: list[dict[str, object]]) -> Evolution
     return journal
 
 
+def _journal_rows(tmp_path: Path) -> list[dict[str, Any]]:
+    """Every line of the journal file, parsed STRICTLY, in file order.
+
+    Not a second copy of any selection rule - it selects nothing. It is
+    here because ``get_spec_audits`` reads through
+    ``read_progress_events``, which is deliberately tolerant and SKIPS a
+    line it cannot parse. A test whose subject is what the WRITER put on
+    disk cannot ask that reader: a malformed row would be skipped and
+    the count would still come out right. The helper #337 deleted parsed
+    strictly as a side effect of being a copy; this keeps the strictness
+    and drops the copy.
+    """
+    journal = tmp_path / ".kstrl" / "evolution.jsonl"
+    return [
+        json.loads(line)
+        for line in journal.read_text(encoding="utf-8").splitlines()
+        if line.strip()
+    ]
+
+
 class TestSpecIssuesPersistence:
     """R1.7: red-team output becomes a durable artifact + journal event."""
 
@@ -1516,6 +1536,7 @@ class TestSpecIssuesPersistence:
                 root_dir=tmp_path,
             )
 
+        assert len(_journal_rows(tmp_path)) == 1, "the writer put more than the audit on disk"
         events = journal_at(tmp_path).get_spec_audits()
         assert len(events) == 1
         assert events[0]["halted"] is True
@@ -1527,10 +1548,42 @@ class TestSpecIssuesPersistence:
             tmp_path,
             _single_component_output([_story()], spec_issues=[MINOR_ISSUE]),
         )
+        assert len(_journal_rows(tmp_path)) == 1, "the writer put more than the audit on disk"
         events = journal_at(tmp_path).get_spec_audits()
         assert len(events) == 1
         assert events[0]["halted"] is False
         assert events[0]["counts"] == {"blocker": 0, "major": 0, "minor": 1}
+
+    def test_the_event_type_on_disk_is_the_wire_value(self, tmp_path: Path) -> None:
+        """The one deliberate literal in the test tree, and why it is one.
+
+        Every other site writes the row and reads it back through
+        SPEC_ISSUES_EVENT, so renaming the constant renames both sides
+        and nothing goes red. Measured in round 1 of review on #337:
+        with SPEC_ISSUES_EVENT set to "spec_audit_row" and the two
+        guard-side literals updated the way somebody doing the rename
+        would update them, 437 passed and 1 xfailed while the journal
+        wrote an event_type no journal already on disk carries. Rows
+        already written are what makes the wire value not the constant's
+        to change, so it is pinned here, once, against the bytes.
+
+        Asserted as a substring of the file rather than as
+        ``row["event_type"] == ...`` because that second shape is
+        exactly what layer 2 of the event-name guard forbids, and one
+        deliberate site is not worth an allowlist in a layer that has
+        none.
+        """
+        self._run(
+            tmp_path,
+            _single_component_output([_story()], spec_issues=[MINOR_ISSUE]),
+        )
+
+        raw = (tmp_path / ".kstrl" / "evolution.jsonl").read_text(encoding="utf-8")
+
+        assert '"event_type":"spec_issues"' in raw, (
+            "the spec audit on disk no longer carries the wire value spec_issues. "
+            f"Renaming SPEC_ISSUES_EVENT does not rename rows already written: {raw}"
+        )
 
 
 class TestPrdValidationInsideRetryLoop:
@@ -2444,11 +2497,22 @@ class TestSpecConvergenceThroughDecompose:
     def test_journal_entries_still_carry_no_run_id(self, tmp_path: Path) -> None:
         """The report reads entries the run-windowed reader drops; if a
         run_id ever appears here, that reader would start windowing
-        spec audits by factory run and this feature would go quiet."""
-        self._run(tmp_path, [BLOCKER_ISSUE])
-        events = journal_at(tmp_path).get_spec_audits()
+        spec audits by factory run and this feature would go quiet.
 
-        assert events and all("run_id" not in e for e in events)
+        Read off the bytes rather than through ``get_spec_audits``: the
+        subject is what decompose PUT on disk, so a reader that dropped
+        or renamed an unknown key could answer this question "no run_id"
+        about rows that carry one, which is the failure the paragraph
+        above names. Every row this run writes is a spec audit
+        (``_record_spec_issues_event`` is decompose's only journal
+        write), so nothing is selected here and no selection rule is
+        copied.
+        """
+        self._run(tmp_path, [BLOCKER_ISSUE])
+        rows = _journal_rows(tmp_path)
+
+        assert rows, "the run wrote no journal rows, so this assertion pins nothing"
+        assert all("run_id" not in row for row in rows), rows
 
     def test_a_legacy_entry_without_run_id_does_not_break_the_read(
         self,
--- a/tests/test_decompose.py
+++ b/tests/test_decompose.py
@@ -1419,13 +1419,6 @@ def _journal_with(tmp_path: Path, entries: list[dict[str, object]]) -> Evolution
     return journal
 
 
-def _read_spec_issue_events(tmp_path: Path) -> list[dict[str, object]]:
-    journal = tmp_path / ".kstrl" / "evolution.jsonl"
-    assert journal.exists(), "journal event was not written"
-    entries = [json.loads(line) for line in journal.read_text().splitlines() if line.strip()]
-    return [e for e in entries if e.get("event_type") == SPEC_ISSUES_EVENT]
-
-
 class TestSpecIssuesPersistence:
     """R1.7: red-team output becomes a durable artifact + journal event."""
 
@@ -1460,7 +1453,7 @@ class TestSpecIssuesPersistence:
         assert artifact.exists()
         assert exc_info.value.artifact_path == artifact
 
-        content = json.loads(artifact.read_text())
+        content = json.loads(artifact.read_text(encoding="utf-8"))
         assert content["project"] == "test"
         assert content["specFile"] == "spec.md"
         assert content["halted"] is True
@@ -1482,7 +1475,7 @@ class TestSpecIssuesPersistence:
             _single_component_output([_story()], spec_issues=[MINOR_ISSUE]),
         )
         assert artifact.exists()
-        content = json.loads(artifact.read_text())
+        content = json.loads(artifact.read_text(encoding="utf-8"))
         assert content["halted"] is False
         assert content["counts"] == {"blocker": 0, "major": 0, "minor": 1}
         assert content["issues"][0]["summary"] == "Edge case unspecified"
@@ -1496,7 +1489,7 @@ class TestSpecIssuesPersistence:
             _single_component_output([_story()], spec_issues=[]),
         )
         assert artifact.exists()
-        content = json.loads(artifact.read_text())
+        content = json.loads(artifact.read_text(encoding="utf-8"))
         assert content["halted"] is False
         assert content["issues"] == []
 
@@ -1523,7 +1516,7 @@ class TestSpecIssuesPersistence:
                 root_dir=tmp_path,
             )
 
-        events = _read_spec_issue_events(tmp_path)
+        events = journal_at(tmp_path).get_spec_audits()
         assert len(events) == 1
         assert events[0]["halted"] is True
         assert events[0]["counts"] == {"blocker": 1, "major": 0, "minor": 0}
@@ -1534,7 +1527,7 @@ class TestSpecIssuesPersistence:
             tmp_path,
             _single_component_output([_story()], spec_issues=[MINOR_ISSUE]),
         )
-        events = _read_spec_issue_events(tmp_path)
+        events = journal_at(tmp_path).get_spec_audits()
         assert len(events) == 1
         assert events[0]["halted"] is False
         assert events[0]["counts"] == {"blocker": 0, "major": 0, "minor": 1}
@@ -2453,7 +2446,7 @@ class TestSpecConvergenceThroughDecompose:
         run_id ever appears here, that reader would start windowing
         spec audits by factory run and this feature would go quiet."""
         self._run(tmp_path, [BLOCKER_ISSUE])
-        events = _read_spec_issue_events(tmp_path)
+        events = journal_at(tmp_path).get_spec_audits()
 
         assert events and all("run_id" not in e for e in events)
 
--- a/tests/test_decompose.py
+++ b/tests/test_decompose.py
@@ -1423,7 +1423,7 @@ def _read_spec_issue_events(tmp_path: Path) -> list[dict[str, object]]:
     journal = tmp_path / ".kstrl" / "evolution.jsonl"
     assert journal.exists(), "journal event was not written"
     entries = [json.loads(line) for line in journal.read_text().splitlines() if line.strip()]
-    return [e for e in entries if e.get("event_type") == "spec_issues"]
+    return [e for e in entries if e.get("event_type") == SPEC_ISSUES_EVENT]
 
 
 class TestSpecIssuesPersistence:
@@ -1668,7 +1668,7 @@ class TestSpecConvergenceReport:
         spec_file: str = "spec.md",
     ) -> dict[str, object]:
         """One prior journal entry, in the shape decompose writes."""
-        entry: dict[str, object] = {"event_type": "spec_issues", "spec_file": spec_file}
+        entry: dict[str, object] = {"event_type": SPEC_ISSUES_EVENT, "spec_file": spec_file}
         if issues is not None:
             entry["issues"] = _issue_dicts(issues)
         return entry
@@ -1763,7 +1763,7 @@ class TestSpecConvergenceReport:
         operator hand-edited, must read rather than raise."""
         history: list[dict[str, object]] = [
             {
-                "event_type": "spec_issues",
+                "event_type": SPEC_ISSUES_EVENT,
                 "issues": [
                     "not a dict",
                     {"severity": "blocker"},
--- a/tests/test_event_names_have_one_home.py
+++ b/tests/test_event_names_have_one_home.py
@@ -236,10 +236,10 @@ def literal_event_names(
     through ``all_nodes``, which memoises it, so the second name reuses
     the first name's walk instead of repeating it.
 
-    Measured as process CPU time over the 332-module corpus this layer
+    Measured as process CPU time over the 334-module corpus this layer
     walks, three sweeps each, parse cache warmed first so the number is
-    about the walk: 1.26-1.27 s a sweep with a raw ``ast.walk``,
-    0.75-0.76 s with the memo warm. The earlier wall-clock version of
+    about the walk: 1.36-1.37 s a sweep with a raw ``ast.walk``,
+    0.80-0.81 s with the memo warm. The earlier wall-clock version of
     this number could not be reproduced without knowing what else the
     machine was running, which is why it is stated as CPU here.
     """
@@ -537,7 +537,7 @@ def event_type_aliases(tree: ast.Module) -> frozenset[str]:
     (#337 round 1: the earlier sentence gave the ``kstrl/`` number only,
     which was two thirds short of the walk). ``kstrl/``: 1 of 129
     modules binds any alias, 6 names, all in ``evolution.py``.
-    ``tests/``: 4 of 204 modules, 10 names, in ``test_factory.py``,
+    ``tests/``: 4 of 206 modules, 10 names, in ``test_factory.py``,
     ``test_journal_torn_tail.py``, ``test_resume_ergonomics.py`` and
     ``test_usage_meter.py``. Three of those names are generic
     (``entry``, ``rows``, ``results``), so an unrelated ``entry ==
--- a/tests/test_event_names_have_one_home.py
+++ b/tests/test_event_names_have_one_home.py
@@ -17,6 +17,9 @@ prompt the split, because it reports rather than fails once a file is
 already over, and ``test_evolution.py`` was at 1407 lines.
 
 TWO LAYERS, because one of them is a net and the other is a message.
+They walk different corpora, and the claim each makes is the corpus it
+walks: layer 2 covers ``kstrl/`` and ``tests/``, layer 1 covers
+``kstrl/``.
 
 LAYER 1, :func:`event_name_spellings`, counts every expression in
 ``kstrl/`` whose folded value IS an enrolled event name, per module. It
@@ -27,9 +30,19 @@ and enumerates no node types, which is the point. #324 records that this
 repo has about eleven AST guards each re-implementing that resolution and
 each holed independently, and layer 2 below was the twelfth: round 1 of
 review on #336 measured six ordinary shapes walking straight past it.
+It stops at ``kstrl/`` deliberately (#337). Measured over ``tests/`` it
+counts 46 spellings in 13 modules, 24 of them in ``test_decompose.py``
+alone, where they are the architect's own JSON key inside a fixture
+payload. A pinned dict over that would move on every fixture that adds
+an architect payload, which is the guard that gets silenced
+``_assignment_hits`` warns about.
 
 LAYER 2, :func:`literal_event_names`, enumerates shapes and names the
-offending line and its direction. It is not redundant. Layer 1 can only
+offending line and its direction. It walks ``kstrl/`` AND ``tests/``
+(#337): a test that spells the name for itself while asserting on
+production behaviour is asserting partly on its own copy of it, and the
+six sites the widened corpus found were all of that kind. It is not
+redundant. Layer 1 can only
 say "this module's spelling count moved", which is the wrong message when
 the answer is "you wrote a journal row with a bare literal, import the
 constant". Layer 1 in turn catches what layer 2 cannot, and after #336
@@ -52,6 +65,7 @@ from collections.abc import Iterable
 from pathlib import Path
 
 from kstrl.evolution import JOURNAL_REPAIR_EVENT, SPEC_ISSUES_EVENT
+from tests.helpers import astwalk
 from tests.helpers.astwalk import (
     Sees,
     assert_census,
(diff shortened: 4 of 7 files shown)


tests/test_event_names_have_one_home.py
```diff
@@ -551,9 +568,16 @@ def test_nobody_spells_an_enrolled_event_name_for_themselves(self) -> None:
...
+
+        Over ``kstrl/`` AND ``tests/`` (#337); the module docstring
+        carries why. Widening the corpus found six sites in the test
+        tree, three writes and three reads, across ``test_decompose.py``,
+        ``test_evolution.py`` and ``test_journal_torn_tail.py``. None was
+        deliberate, so this layer needs no allowlist.
+        """
         found: dict[str, list[str]] = {}
-        for source_file in package_sources():
+        for source_file in package_sources() + astwalk.test_sources():
```
0xfauzi: The widened corpus is the whole point of this PR and nothing pins it. `assert found == {}` over `package_sources() + astwalk.test_sources()` is green when the walk is correctly clean AND when the walk has been narrowed, which is the same shape as the defect — CLAUDE.md rule (2) of the guard-design list, and `assert_census`'s own docstring says it directly: "no control can speak for it: a control parses a string, so it fires whether `sources` holds 128 modules or none."

Measured here, not relayed: with `+ astwalk.test_sources()` deleted from this line, `tests/test_event_names_have_one_home.py tests/test_event_name_shapes.py tests/test_astwalk_nets.py tests/test_astwalk_mechanisms.py tests/test_astwalk.py` gives 186 passed, 13 xfailed, exit 0 (bytecode deleted, `PYTHONDONTWRITEBYTECODE=1`). The two per-corpus emptiness asserts do not help: dropping the `+` leaves 129 non-empty modules.

The PR body discloses this and defers it as "adds a mechanism rather than removes one". That reading inverts the priority: the rule about removing mechanisms applies to the FIX, and the mechanism being removed silently here is the corpus itself. The one-line form already has precedent in this suite at `tests/test_procgroup.py:737`.

```suggestion
        sources = package_sources() + astwalk.test_sources()
        assert len(sources) > 300, (
            f"layer 2 is walking {len(sources)} modules; it is meant to walk "
            "kstrl/ AND tests/, and a narrowed corpus reports clean."
        )
        for source_file in sources:
```

tests/test_event_names_have_one_home.py
```diff
@@ -63,8 +77,8 @@
 )
 
 #: The journal event names that have been hoisted to a constant, and so
-#: must never be spelled as a literal in ``kstrl/`` again. Five more are
-#: still bare literals (``component_result`` nine times in
+#: must never be spelled as a literal in ``kstrl/`` or ``tests/`` again.
```
0xfauzi: This comment now claims `tests/` as well, and neither layer enforces it there. Layer 1 (the net that catches a spelling in ANY shape) deliberately stays on `kstrl/`, and layer 2 only sees a spelling attached to the `event_type` column, so an indirect spelling in a test module is invisible to both.

Measured, planted into `tests/test_evolution.py` with bytecode deleted:

```python
_SPEC = "spec_issues"


def _legacy_row() -> dict[str, str]:
    return {"event_type": _SPEC}
```

`uv run pytest tests/test_event_names_have_one_home.py` -> 2 passed. `folded_str` on the `Name` value returns `None`, so `_pair_hits` misses it, and layer 1 never opens the file. Over `kstrl/` the same plant is caught, by layer 1.

That is a claim covering a corpus the walk does not cover, which is the exact defect this PR exists to fix (`#337`), reintroduced one comment over. 46 spellings in 13 test modules are unenforced today. Either say what is actually true here ("must never be spelled as a literal ATTACHED TO THE `event_type` COLUMN in `tests/`"), or widen layer 1.

tests/helpers/astwalk/corpus.py
```diff
@@ -81,6 +81,12 @@ def test_sources(exclude: Path | None = None) -> list[Path]:
     A guard naming the shapes it forbids in its own fixtures would
     otherwise scan itself. Never empty, for the reason above: excluding
     the caller cannot empty a directory that holds the caller.
+
+    Reach this through the module, ``astwalk.test_sources()``, and never
+    by a from-import: pytest collects any module-level name matching
+    ``test*``, so the import binds a third "test" in the calling module
+    and pytest runs the helper as one. Measured in
```
0xfauzi: This documents a footgun instead of removing it, and pytest already has the mechanism. `test_sources.__test__ = False` makes the helper uncollectable however it is imported, which deletes the rule, the module-qualified call convention, and the comment in the calling guard all at once.

Measured, both directions, bytecode deleted:

- `from tests.helpers.astwalk import test_sources` added to `tests/test_event_names_have_one_home.py`: `pytest --co -q` -> `3 tests collected`, including `tests/test_event_names_have_one_home.py::test_sources`. Confirms the docstring's number.
- Same from-import left in place, plus `test_sources.__test__ = False` in this module: `2 tests collected`.

Second problem with the rule as written: "never by a from-import" is false for the file that re-exports it — `tests/helpers/astwalk/__init__.py:82` from-imports `test_sources` and must, and it is safe because pytest does not collect `tests/helpers/`. A reader following this docstring concludes the package's own `__init__` is violating it. The hazard is "a module pytest collects", not "a from-import".

```suggestion
    Never from-imported into a module pytest collects: pytest collects
    any module-level name matching ``test*``, and the import would bind a
    third "test" there. ``__test__ = False`` below is what makes that
    safe rather than a rule every caller has to remember; measured in
    ``test_event_names_have_one_home.py``, which collects 2 either way.
    """
```

(with `test_sources.__test__ = False  # type: ignore[attr-defined]` after the function).

tests/test_event_names_have_one_home.py
```diff
@@ -551,9 +568,16 @@ def test_nobody_spells_an_enrolled_event_name_for_themselves(self) -> None:
...
 
     def test_no_module_names_an_enrolled_event_as_a_literal(self) -> None:
-        """Layer 2, the message: name the offending line and its direction."""
+        """Layer 2, the message: name the offending line and its direction.
+
+        Over ``kstrl/`` AND ``tests/`` (#337); the module docstring
+        carries why. Widening the corpus found six sites in the test
+        tree, three writes and three reads, across ``test_decompose.py``,
+        ``test_evolution.py`` and ``test_journal_torn_tail.py``. None was
+        deliberate, so this layer needs no allowlist.
```
0xfauzi: "this layer needs no allowlist" is true of today's six sites, but the corpus this commit adds contains `tests/test_event_name_shapes.py`, whose entire purpose is to hold the shapes this layer forbids. They survive only because every control there is a source STRING that `folded_str` folds whole. `tests/helpers/astwalk/corpus.py::test_sources` offers `exclude=` for exactly this case and says so ("A guard naming the shapes it forbids in its own fixtures would otherwise scan itself"); two peers use it, `tests/test_procgroup.py:719` and `tests/test_process_scoping.py:203`. This call passes nothing.

Concretely: the first control someone writes as real code rather than a snippet string — a module-level fixture row, or `hits()` refactored to build the tree from a real function instead of `ast.parse(source)` — makes this guard fail on its own positive-control file, and the only green route is to weaken the control. `exclude=Path(__file__)` does not reach it either, because the exposure is the SIBLING file, so this needs a named exclusion of `test_event_name_shapes.py` with the reason attached, or the disclosure that the sibling is in the corpus.

The PR body's second handoff already names a live case of the missing allowlist from the other direction: `tests/test_decompose.py:2460` is a deliberate on-disk legacy row, and enrolling `component_result` makes it a 43-site job against a layer with nowhere to record a deliberate site.

tests/test_event_names_have_one_home.py
```diff
@@ -181,14 +195,17 @@ def literal_event_names(
...
+    0.46 s with this and the node walk hoisted. It is O(names x modules)
+    on a computation that does not depend on names, and this file commits
+    to five more names. The traversal is still per name, but it goes
+    through ``all_nodes``, which memoises it, so the second name reuses
+    the first name's walk instead of repeating it: measured over the
+    333-module corpus, the guard test went 1.53 s to 1.1 s.
     """
     resolved = event_type_aliases(tree) if aliases is None else aliases
-    return [hit for node in ast.walk(tree) for hit in _hits_at(node, event_name, resolved)]
+    return [hit for node in all_nodes(tree) for hit in _hits_at(node, event_name, resolved)]
```
0xfauzi, line 249: This commit's message says layer 2's walks now go "through `astwalk.all_nodes`, the memo layer 1 already uses". Layer 1 does not use that memo: `event_name_spellings` at line 130 is still `sum(1 for node in ast.walk(parsed(source_file)) if sees(node))`, a raw full-module walk of the same trees this line now has memoised. It is the third sibling and it was left behind.

Two more facts about that function, both checkable by grep:

1. **Layer 1's assertion never calls it.** `test_nobody_spells_an_enrolled_event_name_for_themselves` calls `assert_census(sources=..., sees=spells_an_enrolled_event(), ...)`, and `assert_census` counts through `astwalk.census`. The only callers of `event_name_spellings` are `tests/test_event_name_shapes.py:286` and `:378`. So the module docstring's "LAYER 1, :func:`event_name_spellings`" names a function layer 1 does not run, and `TestLayerOneCatchesWhatLayerTwoCannot` proves its claims against a second implementation rather than against the code the guard executes.
2. **It duplicates `astwalk.census`.** `census` is `for node in all_nodes(parsed(source_file)): if sees(node)` keyed by label — the same body, memoised, for a corpus instead of one file. `event_name_spellings(path)` is `census([path], spells_an_enrolled_event()).get(label(path), 0)`.

They agree today because both use `spells_an_enrolled_event()` and `ast.walk`/`all_nodes` return the same nodes. The cost is that the controls certify a copy, which is the same shape as the `_read_spec_issue_events` deletion this PR is otherwise making.

tests/test_event_names_have_one_home.py
```diff
@@ -478,7 +495,7 @@ def event_type_aliases(tree: ast.Module) -> frozenset[str]:
     to be wrong in. Measured: 1 of 127 modules in ``kstrl/`` binds any
     alias at all.
     """
-    nodes = list(ast.walk(tree))
+    nodes = list(all_nodes(tree))
```
0xfauzi: The docstring three lines up still ends "Measured: 1 of 127 modules in ``kstrl/`` binds any alias at all", and that number is now the justification for a per-MODULE alias collection running over 333 modules, two thirds of which it never measured. Same shape as the module-level claim `#337` fixes: the sentence covers a corpus the function no longer stops at.

Measured over the corpus this PR gives it, bytecode deleted:

| corpus | modules | binding an alias | alias names |
|---|---|---|---|
| `kstrl/` | 129 | 1 (`evolution.py`) | 6 |
| `tests/` | 204 | 4 | 11 |

The four are `test_factory.py` (`comp_entries`, `entry`), `test_journal_torn_tail.py` (`parsed_types`), `test_resume_ergonomics.py` (`component_results`, `final_findings`, `journaled`, `superseded`), `test_usage_meter.py` (`results`, `roles`, `rows`). `entry`, `rows` and `results` are generic enough that a future unrelated `entry == "<an enrolled name>"` anywhere in `test_factory.py` reads as an event-type comparison. That is the over-report direction, which the docstring correctly calls the allowed one — but the sentence should carry the number for the corpus it now runs over, so the next reader is not sizing this residual from a third of the data.

Also on this line: `list(...)` copies an immutable tuple the memo just handed back, per module, for nothing — `_alias_sweep` only iterates it. `all_nodes` returns a tuple precisely so callers can share it. Widen `_alias_sweep`'s parameter to `Sequence[ast.AST]` and drop the copy.

```suggestion
    nodes = all_nodes(tree)
```

tests/helpers/journal.py
```diff
@@ -19,6 +19,13 @@
 entries as the cost file; ``component_result``, ``audits_in`` and
 ``repair_rows_in`` have one caller each at the moment, which is the
 weakest case for this module and worth revisiting if it stays that way.
+
+``audits_in`` takes the journal rather than its path and reads through
+``EvolutionJournal.get_spec_audits``, so a test asserting on what the
+journal reads back is not asserting on a second copy of that reader's
+selection rule (#337). ``repair_rows_in`` still takes a path, because
+no production reader returns repair rows to route it through:
```
0xfauzi: The stated reason is about the READER (`get_repair_count` returns an int, so there is no production reader to route through) but the change made is to the PARAMETER, and those are separable. `repair_rows_in` can take the `EvolutionJournal` and keep its own `read_progress_events(journal.config.journal_path)` body.

The cost of not doing that is six call sites in `test_journal_torn_tail.py` that now read `repair_rows_in(journal.config.journal_path)` next to `audits_in(journal)` (lines 175-176 and 212-213 are the pair, one line apart), plus 471, 508, 523, 543. Reaching through `journal.config.journal_path` at a call site is the exact pattern `EvolutionJournal.get_spec_audits`' docstring calls the defect — "a caller holding the path would quietly return less than the journal holds" — so this PR removes that reach-through from one helper and leaves the other one handing it to every caller.

tests/test_decompose.py
```diff
@@ -1523,7 +1516,7 @@ def test_journal_event_on_halt(self, tmp_path: Path) -> None:
                 root_dir=tmp_path,
             )
 
-        events = _read_spec_issue_events(tmp_path)
+        events = journal_at(tmp_path).get_spec_audits()
```
0xfauzi, line 1546: The deleted `_read_spec_issue_events` parsed strictly: `json.loads(line) for line in ... if line.strip()` raised `JSONDecodeError` on a malformed row. `get_spec_audits` -> `_read_all_entries` -> `read_progress_events` is deliberately tolerant and skips one.

So the narrowing is not only the failure message the PR body discloses. A run that wrote one well-formed audit and one corrupt row used to fail loudly on the corrupt bytes; now the corrupt row is skipped, `len(events) == 1` holds and the test passes. The same applies at line 1530.

Nothing else in this class covers it: `TestSpecIssuesPersistence` reads the artifact JSON, not the journal bytes. If the strict read was load-bearing, assert the row count from the file as well as the parsed count; if it was not, the disclosure in the PR body should say the reader got tolerant, not only that the message got worse.

tests/test_decompose.py
```diff
@@ -2453,7 +2446,7 @@ def test_journal_entries_still_carry_no_run_id(self, tmp_path: Path) -> None:
         run_id ever appears here, that reader would start windowing
         spec audits by factory run and this feature would go quiet."""
         self._run(tmp_path, [BLOCKER_ISSUE])
-        events = _read_spec_issue_events(tmp_path)
+        events = journal_at(tmp_path).get_spec_audits()
```
0xfauzi: This test's subject is the WRITER: its docstring says "if a run_id ever appears here, that reader would start windowing spec audits by factory run and this feature would go quiet". The evidence now passes through `get_spec_audits`, a production reader, so a reader change can decide the answer to a question about what decompose put on disk.

The concrete path: any normalisation added to `get_spec_audits` or `_read_all_entries` that drops or renames unknown keys makes `all("run_id" not in e ...)` true while the rows on disk carry `run_id`, and `_read_journal_entries` starts windowing them — the exact failure the docstring names, now invisible to the test that names it. The old `_read_spec_issue_events` could not be fooled that way because it opened the file.

Consolidating the SELECTION rule was right; this one assertion is about the bytes, so it wants the raw read. `_journal_with`-style direct reads are already the pattern at lines 2105, 2457 and 2649 in this same file.

tests/test_event_names_have_one_home.py
0xfauzi: Second point on this line: the failure dict now mixes two labelling roots. `label(source_file)` defaults to `KSTRL_PACKAGE`, so a package file keys as `evolution.py` while a test file falls through to `REPO_ROOT` and keys as `tests/test_evolution.py`. The message a reader gets is `{"evolution.py": [...], "tests/test_evolution.py": [...]}`, where only one of the two paths can be opened as written, and `evolution.py` reads as a repo-root file that does not exist.

`tests/test_astwalk_mechanisms.py` already passes the root explicitly when its corpus is `tests/` (`astwalk.label(source_file, astwalk.REPO_ROOT)`). Layer 1's pinned dict is package-keyed and should stay that way; layer 2 is the one whose corpus went mixed, so it is the one that wants `label(source_file, REPO_ROOT)`.

tests/test_decompose.py
```diff
@@ -1460,7 +1453,7 @@ def test_artifact_written_on_halt(self, tmp_path: Path) -> None:
         assert artifact.exists()
         assert exc_info.value.artifact_path == artifact
 
-        content = json.loads(artifact.read_text())
+        content = json.loads(artifact.read_text(encoding="utf-8"))
```
0xfauzi, line 1482: These three `encoding="utf-8"` additions are a partial fix of a different defect class in a PR about event names, and the PR body says so ("`#320`'s class, not this one ... 172 bare `read_text()` calls remain in `tests/`").

CLAUDE.md's convention is "**One defect class, one PR.** Fix every instance plus the guard that catches instance N+1". Three of ~175 sites is the opposite end of that: it clears the one file a reviewer is most likely to spot-check while the class stays open, and `tests/helpers/encodingwalk.py` still walks `package_sources()` only, so no guard can report the remaining 172. A later sweep now has one file that looks done for a reason unrelated to the sweep.

Either drop them from this PR, or take the class: widen the encoding walk to `tests/` and fix all 172 in a PR of its own.

tests/test_event_names_have_one_home.py
```diff
@@ -551,9 +623,53 @@ def test_nobody_spells_an_enrolled_event_name_for_themselves(self) -> None:
...
+        assertion fail on its own positive-control file, and the only
+        green route would be to weaken the control. That is the point at
+        which this layer needs a named allowlist, and the same is true
+        from the other direction the day ``component_result`` is
+        enrolled: ``tests/test_decompose.py`` writes a deliberate legacy
+        on-disk row with it.
+        """
+        sources = package_sources() + astwalk.test_sources(exclude=Path(__file__))
+        walked_tests = [path for path in sources if path.is_relative_to(TESTS_DIR)]
+        assert walked_tests, (
```
0xfauzi: **The corpus control catches deletion but not narrowing, which is the shape the defect actually takes.**

`assert walked_tests` is a truthiness check, so it fires only when `astwalk.test_sources(...)` disappears from line 663 entirely. Measured on this head, `__pycache__` cleared and `PYTHONDONTWRITEBYTECODE=1`: changing line 663 to `... + astwalk.test_sources(exclude=Path(__file__))[:1]` leaves this file at **2 passed**, with layer 2 covering one test module out of 205. `test_sources` already asserts its own non-emptiness, so the only case this control adds is total deletion.

The two peers the docstring cites pin a floor instead: `tests/test_procgroup.py:737` and `tests/test_astwalk_nets.py:626` both `assert len(...) > 100`. The docstring says "closed by construction, no threshold to maintain", which a truthiness check is not: it is closed over one shape of the mutation and open over every partial one. Counting what was handed in against the corpus it came from would close it in the direction the guard is wrong in.

tests/test_decompose.py
```diff
@@ -2494,11 +2540,22 @@ def test_second_run_compares_against_the_first(self, tmp_path: Path) -> None:
...
+        (``_record_spec_issues_event`` is decompose's only journal
+        write), so nothing is selected here and no selection rule is
+        copied.
+        """
         self._run(tmp_path, [BLOCKER_ISSUE])
-        events = _read_spec_issue_events(tmp_path)
+        rows = _journal_rows(tmp_path)
 
-        assert events and all("run_id" not in e for e in events)
+        assert rows, "the run wrote no journal rows, so this assertion pins nothing"
```
0xfauzi, line 2563: **This assertion no longer anchors on a spec audit, and `origin/main` did.**

Before this PR the line read `assert events and all("run_id" not in e for e in events)`, where `events` came from `_read_spec_issue_events`, i.e. rows filtered to `event_type == "spec_issues"`. `_journal_rows` selects nothing, so `assert rows` now passes on any row of any type.

Measured on both trees, `__pycache__` cleared and `PYTHONDONTWRITEBYTECODE=1`, with `kstrl/decompose.py:1409` planted as `"event_type": "architect_audit"` so decompose writes no spec audit at all:

| tree | outcome |
|---|---|
| `origin/main` | **FAILED**, `assert ([])` |
| this head | **passed** |

The message on 2557 says "this assertion pins nothing", which is what it then does about spec audits. The byte-level read the docstring argues for is still available: pin the row count the way lines 1539 and 1551 already do, so a run that writes something other than the audit fails here rather than reading clean.

tests/helpers/astwalk/corpus.py
```diff
@@ -81,13 +81,30 @@ def test_sources(exclude: Path | None = None) -> list[Path]:
...
+    emits eight. With ``__test__ = False`` the same from-import
+    collects 2.
     """
     skip = exclude.resolve() if exclude is not None else None
     found = [path for path in sorted(TESTS_DIR.rglob("*.py")) if path.resolve() != skip]
     assert found, _EMPTY.format(what="tests/", root=TESTS_DIR)
     return found
 
 
+test_sources.__test__ = False  # type: ignore[attr-defined]
```
0xfauzi, line 105: **Nothing in the suite fails if this line is deleted.**

`grep -rn "__test__" tests/ kstrl/` returns this assignment and three prose mentions, nothing else. The M9 measurement in the PR body (2 collected with it, 3 without, and the third passes on a warning) was made by hand and did not become a check, and no module from-imports `test_sources` today, so deleting the line moves no count anywhere in the suite. That is the repo's dominant defect class one layer up: a mechanism with no control, going quiet rather than red.

The mechanism itself is real, verified here on pytest 9.1.1: a module with `helper.__test__ = False; test_marked = helper` collects 1 test; without the attribute it collects 2, and the extra one passes.

One line in `tests/test_astwalk_nets.py` closes it, next to the `exclude` control already there:

    assert astwalk.test_sources.__test__ is False

tests/test_event_names_have_one_home.py
```diff
@@ -181,14 +228,23 @@ def literal_event_names(
...
+    Measured as process CPU time over the 334-module corpus this layer
+    walks, three sweeps each, parse cache warmed first so the number is
+    about the walk: 1.36-1.37 s a sweep with a raw ``ast.walk``,
+    0.80-0.81 s with the memo warm. The earlier wall-clock version of
+    this number could not be reproduced without knowing what else the
+    machine was running, which is why it is stated as CPU here.
     """
     resolved = event_type_aliases(tree) if aliases is None else aliases
-    return [hit for node in ast.walk(tree) for hit in _hits_at(node, event_name, resolved)]
+    return [hit for node in all_nodes(tree) for hit in _hits_at(node, event_name, resolved)]
```
0xfauzi, line 249: **`all_nodes` never releases anything it caches, so routing this walk through it makes throwaway trees immortal.**

`_NODES` in `corpus.py:132` is a `WeakKeyDictionary` keyed on the tree, but the value is `tuple(ast.walk(tree))` and `ast.walk` yields the root node first, so every value holds a strong reference to its own key. Measured on this head: 50 trees built with `ast.parse`, handed to `all_nodes`, dropped, then `gc.collect()`, leave **50 entries**; and `all_nodes(t)[0] is t` is `True`. `corpus.py:135` claims the opposite: "a tree a guard built with a bare `ast.parse` and then dropped takes its row with it".

This line and line 548 are new consumers of that. `tests/test_event_name_shapes.py::hits` calls `literal_event_names(ast.parse(source), ...)` on about forty trees the caller means to drop, and each is now retained for the session. Over the 335-module corpus this layer walks, 684,831 nodes are retained.

Test-only memory, but the docstring's claim is the load-bearing part and it is false. Hanging the memo off the tree (`tree._all_nodes`) makes the documented property true, because the cache then dies with its key.

tests/test_event_names_have_one_home.py
```diff
@@ -111,9 +149,18 @@ def event_name_spellings(source_file: Path) -> int:
...
+    controls in ``tests/test_event_name_shapes.py``. Round 1 of review
+    on #337 found it here as a third raw ``ast.walk``, so the controls
+    were certifying a copy of the code the guard executes rather than
+    the code itself. That is the same defect as the test-side copy of
+    the selection rule this PR deletes, one file over.
     """
-    sees = spells_an_enrolled_event()
-    return sum(1 for node in ast.walk(parsed(source_file)) if sees(node))
+    key = label(source_file)
+    return census([source_file], spells_an_enrolled_event()).get(key, 0)
```
0xfauzi, line 165: **`.get(key, 0)` turns a key-derivation mismatch into "zero spellings", which is the clearing direction.**

The caller derives `key = label(source_file)` and `census` derives its row name with `label(source_file)` as well, so the two agree only because both happen to spell the default. If `census`'s default keyer moves, or `label`'s default root does, this returns 0 for every file rather than failing, and `tests/test_event_name_shapes.py:378`'s `== 0` blind-spot control goes on passing. CLAUDE.md: "A gate that keys on parsed output must reject what it cannot parse, never count it as zero."

Summing the census is exactly equivalent for a one-file corpus and has no key to get wrong:

```suggestion
    return sum(census([source_file], spells_an_enrolled_event()).values())
```

tests/test_decompose.py
```diff
@@ -1534,11 +1548,43 @@ def test_journal_event_on_success(self, tmp_path: Path) -> None:
...
+        none.
+        """
+        self._run(
+            tmp_path,
+            _single_component_output([_story()], spec_issues=[MINOR_ISSUE]),
+        )
+
+        raw = (tmp_path / ".kstrl" / "evolution.jsonl").read_text(encoding="utf-8")
+
+        assert '"event_type":"spec_issues"' in raw, (
```
0xfauzi, line 1589: **The substring pins the journal's JSON separators as well as the wire value, and the message names only one of them.**

`kstrl/evolution.py:652` writes `json.dumps(entry, separators=(",", ":"))`. Change that to the default separators, a formatting decision with nothing to do with event names, and this assertion fails saying "the spec audit on disk no longer carries the wire value spec_issues". That is a wrong diagnosis from the one test in the tree whose entire job is to diagnose a wire-value rename correctly, and it is the failure mode the docstring above spends a paragraph arguing this test exists to prevent.

Building the expected fragment from the same separators the writer uses keeps the layer-2 invisibility (the constant still folds to a longer string) while making a formatting change move both sides, so only a rename fails here.

tests/helpers/journal.py
```diff
@@ -110,13 +123,20 @@ def journal_at(tmp_path: Path) -> EvolutionJournal:
...
-        str(entry.get("project"))
-        for entry in read_progress_events(path)
-        if entry.get("event_type") == SPEC_ISSUES_EVENT
-    ]
+def audits_in(journal: EvolutionJournal) -> list[str]:
+    """The projects of every spec audit the journal reads back, in order."""
+    return [str(entry.get("project")) for entry in journal.get_spec_audits()]
+
 
+def repair_rows_in(journal: EvolutionJournal) -> list[dict[str, Any]]:
```
0xfauzi, line 131: **`repair_rows_in(journal)` now reads identically to `audits_in(journal)` at the call site while doing the opposite thing.**

`audits_in` goes through `EvolutionJournal.get_spec_audits`. This one opens `journal.config.journal_path` behind the journal's back. The docstring explains the difference, but the seven call sites no longer show it, and `get_spec_audits`' own docstring names the failure that hides: "if the journal ever compacts, rotates or gains a second segment, this method is what changes, while a caller holding the path would quietly return less than the journal holds". After this change `audits_in` adapts to that and `repair_rows_in` silently does not, and nothing at a call site distinguishes them any more.

Changing the parameter type treats the symptom the review round named (six call sites reaching through `config.journal_path`) rather than the cause (no reader returns repair rows). A `get_repair_rows()` on `EvolutionJournal` beside the existing `get_repair_count()` is a few lines, removes this helper's selection rule entirely, and makes the two helpers mean the same thing at a call site.

tests/test_decompose.py
```diff
@@ -1534,11 +1548,43 @@ def test_journal_event_on_success(self, tmp_path: Path) -> None:
...
+        exactly what layer 2 of the event-name guard forbids, and one
+        deliberate site is not worth an allowlist in a layer that has
+        none.
+        """
+        self._run(
+            tmp_path,
+            _single_component_output([_story()], spec_issues=[MINOR_ISSUE]),
+        )
+
+        raw = (tmp_path / ".kstrl" / "evolution.jsonl").read_text(encoding="utf-8")
```
0xfauzi, line 1587: **A new hand-derived journal path, in the PR that deletes a test-side copy of a production reader.**

`journal_at` is already imported in this file (line 40), and `journal_at(tmp_path).config.journal_path` is the surface `EXPECTED_JOURNAL_PATH_SITES` in `tests/test_journal_one_writer.py` exists to concentrate. Nothing reports these because that census walks `package_sources()` only, so `tests/` already carries fifteen such derivations and this adds a sixteenth; `_journal_rows` at 1434, rewritten in this diff, is a seventeenth that could have moved at the same time.

Reading the bytes rather than a reader is the point of this test and is unaffected: `journal_at(tmp_path).config.journal_path.read_text(encoding="utf-8")` reads the same bytes without a second copy of the on-disk layout.

tests/test_decompose.py
```diff
@@ -1419,11 +1419,24 @@ def _journal_with(tmp_path: Path, entries: list[dict[str, object]]) -> Evolution
...
+    the count would still come out right. The helper #337 deleted parsed
+    strictly as a side effect of being a copy; this keeps the strictness
+    and drops the copy.
+    """
     journal = tmp_path / ".kstrl" / "evolution.jsonl"
-    assert journal.exists(), "journal event was not written"
-    entries = [json.loads(line) for line in journal.read_text().splitlines() if line.strip()]
-    return [e for e in entries if e.get("event_type") == "spec_issues"]
+    return [
+        json.loads(line)
```
0xfauzi: **"parsed STRICTLY" does not check that a line is an object, and the annotation says it does.**

`json.loads` accepts any JSON value, so a line holding an array or a scalar comes back as a `list` or an `int` while the return type says `dict[str, Any]` (tests are outside `[tool.mypy] files`, so nothing catches that). It matters at line 2558: `"run_id" not in [1, 2]` is `True`, so an array line passes the check silently, and a scalar line raises `TypeError` rather than failing readably. `read_progress_events`, the tolerant reader this helper exists to avoid, does skip non-object lines, so the strictness gained here is narrower than the docstring claims in exactly the direction the docstring is worried about.

An `isinstance(row, dict)` check per line, raising with the offending line, closes it and makes the annotation true.

tests/test_event_names_have_one_home.py
```diff
@@ -551,9 +623,53 @@ def test_nobody_spells_an_enrolled_event_name_for_themselves(self) -> None:
...
+        STRING that ``folded_str`` folds whole. The first control
+        somebody writes as real code instead of a snippet makes this
+        assertion fail on its own positive-control file, and the only
+        green route would be to weaken the control. That is the point at
+        which this layer needs a named allowlist, and the same is true
+        from the other direction the day ``component_result`` is
+        enrolled: ``tests/test_decompose.py`` writes a deliberate legacy
+        on-disk row with it.
+        """
+        sources = package_sources() + astwalk.test_sources(exclude=Path(__file__))
```
0xfauzi, line 671: **`exclude` takes one path, and the known next failure is documented instead of fixed.**

The docstring below states that `tests/test_event_name_shapes.py` is in this corpus, that it is clean today only because every control there is a source string `folded_str` folds whole, and that "the first control somebody writes as real code instead of a snippet makes this assertion fail on its own positive-control file, and the only green route would be to weaken the control". That is a trap set for the next author of a shape control, and disclosing it does not disarm it.

The mechanism is two lines in `tests/helpers/astwalk/corpus.py`: widen `exclude` to `Path | Iterable[Path] | None` and resolve each. Both existing callers pass a single `Path` and are unaffected. The same widening covers the `component_result` direction the docstring raises from the other side, and it removes the "weaken the control" route rather than describing it.

tests/test_event_names_have_one_home.py
```diff
@@ -48,23 +67,42 @@
...
+from collections.abc import Iterable, Sequence
 from pathlib import Path
 
 from kstrl.evolution import JOURNAL_REPAIR_EVENT, SPEC_ISSUES_EVENT
+
+# Reached through the module for readability rather than for safety:
+# ``test_sources`` carries ``__test__ = False``, so a from-import of it
+# would not be collected here either. Its docstring carries both
+# measurements.
+from tests.helpers import astwalk
```
0xfauzi, line 81: **This module import exists for a hazard the comment above it says is already gone.**

The comment states that `test_sources` carries `__test__ = False` and that "a from-import of it would not be collected here either", so the file now carries two import paths into the same package, the second one for a stated non-reason. Line 663 then reads `astwalk.test_sources` while `package_sources`, `TESTS_DIR` and `REPO_ROOT` on the surrounding lines are bare names from the block below.

Adding `test_sources` to the existing from-import block and dropping this import and its comment leaves one import path and nothing to explain.

Issue #291, opened 2026-08-31
Atomic writes over operator-owned files silently downgrade permissions to 0600 and use the locale encoding


0xfauzi:
Found while fixing #286 (PR #290, merged). That PR added its own atomic write, review caught that it downgraded a file's permissions, and the agent then measured the two existing copies of the same pattern and found they do it too. Out of scope there; filing it here.

## The defect

`tempfile.mkstemp` creates its temporary file with mode 0600, and `os.replace` carries that mode onto the destination. So an atomic write over an existing file silently tightens its permissions from whatever the operator had, typically 0644 at a default umask, to 0600.

Two call sites do this to files that are git-tracked and operator-owned:

- `workqueue.atomic_write`
- `decompose._atomic_write_json`

Neither pins an encoding either, so both also use the locale encoding for a file whose content may later be compared byte-for-byte.

## Why it matters for these two specifically

The pattern is correct where kstrl writes its own state, which is why it has been fine in `manifest.py` and `knowledge.write_facts`. It is not correct for a file that:

- lives in the operator's repository,
- is expected to be group- or world-readable,
- or is read by a different uid than the one that wrote it.

The concrete failure PR #290 measured for its own case: a container image or CI job that runs one kstrl command as one uid and the factory as another gets a `PermissionError` on a file that worked before, with nothing to indicate why the mode changed.

## The fix PR #290 used

`os.fchmod(fd, mode)` on the descriptor `mkstemp` already returned. No second path lookup, and no umask read, which matters because `run_init` also executes inside the Textual wizard where mutating the process umask is not available.

Where the target already exists, copying `target.stat().st_mode` onto the temp file before the replace preserves whatever the operator had rather than imposing a default.

## Scope

Two functions, plus an encoding pin on each. It touches the work queue writer and the manifest writer, which is why it was not folded into a prompt-scaffolding PR: both are on paths that carry run state, and a mistake there is worse than the bug.

Worth deciding at the same time whether a single shared helper should own this, given there are now four copies of the pattern with two different correct behaviours depending on who owns the file.

PR #352, opened 2026-09-06
fix: six more appenders lose the record after a torn write, the journal lock, and the evolve repair line (#331, #330, #333)

### LINKED ISSUE #330: append_entries probes and appends without a lock: state and shrink the residual
Split out of PR #327 (issue #312) rather than fixed there, at the request of round 1 of review on that PR: "Either lock the probe and append together, or state the residual precisely in the docstring and file an issue. Do not leave a docstring implying atomicity you do not have."

The docstring now states it. This is the issue.

## What is not atomic

`EvolutionJournal.append_entries` reads the last byte of the journal and then appends, both through one `"a+b"` file description. Using one file description removes the path-level races (the file cannot be replaced or a symlink retargeted between the probe and the write), but it is not a lock, and nothing else in this module takes one either.

Three residuals, in order of how much they cost:

1. **A stale probe.** Between this process's tail read and its write, another process can append. If that other write is a complete line, nothing is lost. If it crashed mid-line inside that window, this append lands on the fragment and the pair is unreadable, which is the #312 outcome in a narrower window. Not a regression: at cbdff7c there was no probe at all and the same two writers produced the same result.
2. **A doubled repair record.** Two processes repairing one tear each write a newline and a `journal_repair` row, so a single incident can be recorded twice. Both the blank line and the extra row are skipped by every reader and counted by no aggregate, so this costs accuracy in `get_repair_count` and nothing else. `docs/evolution-metrics.md` says so.
3. **Interleaved lines.** `O_APPEND` makes each `write` land at the end, but a line larger than the buffer can be split across several `write` calls, so concurrent appenders can interleave mid-line. This predates #312 and is not specific to the repair.

## Who can be concurrent

Worth establishing rather than assuming, because the fix depends on it:

- `record_run` (factory), `_record_contract_event` (factory), `journal_superseded_findings` (pipeline), `_record_spec_issues_event` (decompose) and `commit_transition` (autonomy) all write.
- Autonomy CLI commands do not hold the factory lock, so `ks autonomy promote` during a running factory is a real two-writer case.
- `kstrl/inbox.py` appends under `control_lock` and `workqueue` has `queue_lock`; the evolution journal has neither.

## Why it was not just locked in #327

`fcntl.flock` is POSIX-only. The repo already skips its flock tests on Windows (Phase A4 worktree lock), so a lock delivers the invariant on some platforms and not others, and a change that makes concurrency safe on Linux and macOS only should say which platforms it covers rather than being slipped into a data-loss fix. It also touches a path shared by the daemon, the CLI and factory worker threads, so it wants its own tests for lock contention and for the case where the lock file itself cannot be created.

## Suggested shape

One decision to make first: whether the journal gets the same `control_lock` treatment as `inbox`, or whether the repair is made idempotent instead (for example, a repair row that names the byte offset it repaired, so a second repairer can see the first one's work). The second is portable and the first is conventional here.

Whichever is chosen, `tests/test_journal_torn_tail.py` is where the coverage goes, and the residual paragraph in `EvolutionJournal.append_entries` is what should shrink.

### LINKED ISSUE #331: Four other JSONL appenders lose the entry after a torn write, like #312
#312 was the evolution journal: a crash mid-write leaves an unterminated tail, the next append concatenates onto it, and the tolerant reader drops BOTH lines, so the record written after the crash is lost.

Four other JSONL appenders in `kstrl/` have the same defect. Each was reproduced during review of PR #327 with a real tear and a real append, not read and reasoned about:

| Appender | File | After a tear, the reader returns |
|---|---|---|
| `observability.py:127` `ProgressLog.emit` | `progress.jsonl` | `['alpha']`, the entry after the tear lost |
| `events.py:847` `JsonlSink` | `events.jsonl` | `['r1']`, r2 lost |
| `workqueue.py:854` | the queue journal | `['a']`, b lost |
| `inbox.py:504` | the inbox log | `['first']`, second lost |

`kstrl/proposals.py:160` is NOT one of them: it appends Markdown and already pads a missing newline, which is the same idiom arrived at independently.

## The mechanism is already shared

PR #327 hoisted the generic half to `observability.handle_ends_without_newline`, which takes an OPEN BINARY HANDLE rather than a path. That shape was chosen with these four sites in mind:

- `JsonlSink` holds a long-lived handle, so it needs to ask once at open time, which a path-taking predicate cannot serve.
- Opening `"a+b"` rather than `"a"` is what makes an unreadable file raise instead of being probed as "not torn" and appended to blind. Note the consequence before copying it: a file this process can write but not read stops being appendable. That was the right trade for the evolution journal; it should be decided per file, not assumed.

## What is NOT shared, and why this is four decisions

What each appender should WRITE on finding a tear differs:

- the evolution journal writes a `journal_repair` row, which every aggregate ignores by construction, and `ks evolve --status` counts.
- `progress.jsonl` and `events.jsonl` have a different row schema (`ts`/`event`/`run_id`/`data`, and a typed `Event` subclass respectively), and the reducer replays them, so a synthetic row has to be one the reducer can see and ignore.
- `inbox` appends under `control_lock` and `workqueue` under `queue_lock`, so their repair is inside a lock the journal does not have (#330).
- Some of these files are read by `ks status` on every invocation, so the cost of the extra `"a+b"` open matters more than it does on the journal's roughly 10 to 40 appends per run.

## Suggested shape

One PR per file, or one PR with four clearly separated commits. Each needs its own answer to "what does a repaired tear look like in THIS file", its own test with a real torn file, and its own note on whether the `"a+b"` read requirement is acceptable there.

`tests/test_journal_torn_tail.py` is the worked example: reproduce the loss first, plant the fix, then plant a mutation that removes it and watch the test fail.

### LINKED ISSUE #333: Evolve TUI screen does not report journal repairs (#312 follow-up)
## What

`ks evolve --status` reports repaired journal writes; the evolve TUI screen does not.

#312 (PR #327) made `EvolutionJournal.append_entries` repair a journal whose tail lost its newline, and record the repair as a `journal_repair` row so the incident is not silently healed. The read surfaces added there are `EvolutionJournal.get_repair_count()`, a `logger.warning`, and one line in `ks evolve --status` via `kstrl/cli.py::_echo_journal_repairs`.

The argument for the durable row is that under the TUI the logger warning goes to `orchestrator.log` where nobody is looking. That argument names the TUI operator as the underserved one, and the TUI operator is still underserved: `EvolveScreen._load_patterns_and_trends` (`kstrl/tui/screens/evolve.py:182`, reached from `kstrl/tui/screens/home.py:386`) already builds an `EvolutionJournal` and reads the journal for `get_cross_run_patterns()`, and shows nothing about repairs.

## Why it was not done in #327

Scope. #327 is the write-side fix plus its CLI surface; the TUI is a separate surface with its own tests, and `_echo_journal_repairs` lives in the click module so it cannot be reused as is. `_echo_journal_repairs`'s docstring records this explicitly rather than implying the TUI is covered.

## Cost

`get_repair_count` is click-free and on the journal. Measured on every real journal on this machine: 0.02 ms (1 line) to 1.38 ms (989 lines, 326 KiB). The screen already carries a `ConfigProblemBanner`, so a repair line has somewhere to go, and the read is zero if the count is taken off entries the screen already reads.

## Acceptance

The evolve screen shows a non-zero repair count and stays silent at zero, with a test that fails if the wiring is removed.

0xfauzi:
## Summary

A crash mid-write leaves a record file with a tail that has no newline. The next append concatenates onto that tail, and the tolerant reader then drops BOTH lines: the fragment, which was never readable, and the record written after it, which was. #312 fixed the evolution journal. #331 found six more appenders with the same defect, and every one was reproduced with a real tear, a real append and the file's own production reader rather than read and reasoned about.

| Appender | Reader | What the reader returned after a tear |
|---|---|---|
| `observability.ProgressLog.emit` (progress.jsonl) | `read_progress_events`, `reducer.load_run_state` | `['alpha']`, the next entry lost, and a component left `running` when the lost row was its `component_completed` |
| `events.JsonlSink` (events.jsonl, engineer.jsonl) | `read_events` then `reducer.fold` | `['factory_started']`, the next event lost, and no components at all in the folded state |
| `workqueue.Queue._journal` | `journal_entries` | `['a']`, the transition after it lost |
| `inbox.Inbox._append` | `Inbox.items`, `Inbox.scan` | `['first']`, the item after it lost, one unparseable line |
| `knowledge.record_dependency_scope_gap` | `read_dependency_scope_telemetry` | `['alpha']`, the row after it lost |
| `evolution.record_run` (experiments.tsv) | `get_experiment_trends`, `autonomy_replay.load_runs` | the run after it lost AND the run before it RENDERED with shifted columns |

A tail that lost only its terminator is worse than a fragment: it is a COMPLETE record, so appending onto it destroys two records for one interrupted write. experiments.tsv is the worst of the seven because its readers display the corruption instead of dropping it.

The mechanism is one mechanism, so it has one home: `kstrl/appendio.py`, a sibling of `atomicio.py` that imports stdlib only. It opens `"a+b"`, probes and appends through ONE file description, and writes the repair and the payload in ONE write. What each file should WRITE on finding a tear is the caller's, because it is that file's schema: four write a `journal_repair` row, and three take a bare pad for a measured reason. The inbox takes a pad because a valid-JSON row it cannot parse is counted by `scan().unparseable_count()`, which `serve` charges against the #190 admission cap, so a repair row would consume admission capacity until compaction. experiments.tsv takes a pad because TSV has no marker a reader would not render as a run, and records the incident in the log instead. The two Markdown-ish files added in review round 1, the proposal stamp and the `.gitignore` block, take a pad because a blank line is what a paragraph break already is there.

**The lock (#330).** The evolution journal's probe and append now hold `fcntl.LOCK_EX` on the journal's OWN descriptor, not on a sibling lock file. `control_lock` and `queue_lock` need an external file because each serializes MANY files; this is one file with one writer function, so the descriptor already open is the lock. That removes a mechanism rather than adding a check, and it makes #330's "the lock file cannot be created" case not exist, since an unopenable journal already raises `OSError` out of the open. Measured, two processes x 150 appends of 200 KB lines with 74 torn fragments planted, eight runs of each arm: unlocked 244 to 269 of 300 records readable and 76 to 86 repair rows for 74 tears; locked 300 of 300 and exactly 74 rows every run, at 0.11 to 0.14 s against 0.08 to 0.11 s. All three residuals #330 lists close together. Best effort, in two directions: without `fcntl` there is no exclusion, and a `flock` that RAISES takes the same path, because some FUSE, 9p and DrvFs mounts answer `ENOLCK` or `EINVAL` and an unguarded acquisition made every journal append raise on such a mount where 568bca4 wrote the entry.

**The TUI line (#333).** #312's argument for writing a durable `journal_repair` row at all was that under the TUI the logger warning goes to `orchestrator.log` where nobody is looking. The evolve screen was the surface that argument named, and it built an `EvolutionJournal`, read the journal for its patterns tab and said nothing about the rows. A `Static` above the tabs now carries the count, the path, and which of the two outcomes the line above each row is, in `ks evolve --status`'s own words. Silent at zero, and it goes back to silent on reload when the count does.

**The guard.** `tests/test_append_opens_have_one_home.py`, three layers. Layer 1 is closed by construction: it inventories every place in `kstrl/` where an append handle is OBTAINED with a raw `open`, keyed by module, callee and mode, so a new appender in any shape is an unexplained census delta whatever it does with the handle. A mode that does not FOLD is counted rather than skipped, because "the walk could not decide" and "not an append" are different answers, which closes the gap `os.open(p, flags)` would otherwise sit in. Layer 2 attributes each site to its innermost scope and demands a row in `ALLOWED_APPEND_OPENS` with a closed-vocabulary reason, each reason carrying a check that can FAIL a row whose stated reason stopped being true. Layer 3, added in review round 1, censuses the other half: every call INTO `appendio`, keyed by scope, pinning what it passes for `lock`, because a correctly routed appender is invisible to layers 1 and 2 while the exclusion argument defaults to off. All three FLAG; nothing here clears a site on its own evidence. Layer 1's census is 17 sites in 15 rows, down from 25 at 568bca4: nine were routed through `appendio` and one was added, `appendio`'s own.

Closes #331
Closes #330
Closes #333

## Review round 1

An independent review and the code-review skill between them raised 0 blockers, 8 should-fix and 14 nits. What follows is what changed; the declined items and their reasons are at the end of this section.

**Two data-loss paths the PR itself opened, both measured.**

- `JsonlSink` assigned `self._fh` and then wrote through it, so a first write that RAISED left the sink bound and in the no-probe branch for the rest of the run. `EventBus.emit` catches per sink into a counter with no production reader, so nothing said anything. Measured on a torn `events.jsonl` with the first write made to raise once: the second emit wrote onto the unterminated tail and `read_events` returned NOTHING, because the concatenated line does not parse at all. The handle is now bound only after the write returns, and closed if it does not.
- `fcntl.flock` was called outside any `try`. Measured with `flock` raising `ENOLCK`: `append_entries` raised `BlockingIOError` and left ZERO bytes on disk, where the same call at 568bca4 wrote the entry. The lock added by #330 to protect the record was the thing losing it. It degrades to the no-`fcntl` path now: `OSError` exactly rather than an enumeration of the codes we thought of, no retry (`ENOLCK` on a mount is not a transient), and the release stays unguarded because it is reached only when the acquisition succeeded and the bytes are already flushed.

**A check that could not fail for the reason it named.** `Reason.PADS_ITSELF` covered `proposals.mark_applied` and `init_cmd._ensure_gitignore` and asked whether anything in the scope started with a newline. Measured: `mark_applied`'s f-string has two satisfying constants, the pad and its own terminator, so deleting the pad left the guard green; `_ensure_gitignore` computes its separator into a variable, so none of the three constants satisfying its row was the pad. Narrowing was tried and measured too: requiring the constant to be inside a `write` call takes the satisfying-scope count in `kstrl/` from 116 to 9 and still misses the deleted pad, for the same f-string reason, and fails the second row outright. Both sites are routed through `appendio` and the reason is deleted, so the census closes over them by construction. The gitignore site gains a real fix on the way: it decided its separator from a SEPARATE read and then opened the file again, and with the state read reporting the file empty while it held an unterminated rule, the block header was appended onto the user's last rule on one line.

**Silent paths that are now recorded.** A repaired `experiments.tsv` write recorded nothing anywhere, and the comment claimed the read side surfaced it, which the pad had made false. `autonomy_replay.load_runs` turned an unreadable file into "no runs", which reached the operator as "VERDICT: INSUFFICIENT DATA" for a permission or decode error. Both are fixed; the replay exit code is unchanged at 2, because nothing was replayed either way, and the line above it names the cause.

**Claims corrected because they were measured false.** "A row whose width this writer never emits" is not true: the row is built by f-string tab-joining, so a tab or a newline in `project_name` reaches the file unescaped. `experiment_rows` now states three measured residuals instead (a legitimately emitted row dropped; a non-prefix header silently dropping every row this writer appends; a fragment torn inside the first field passing the width check with a fabricated `run_id`) and the fourth thing to know, that there is no counter, so `[]` means three different things. `load_runs`'s docstring said "the population a promotion is decided on"; it has one caller and that command never mutates ladder state. And the H4 section below.

**Smaller ones applied.** `append_terminated` now requires `repair`'s terminator as well as `payload`'s, because an unterminated repair line loses the record it precedes. The queue journal's repair row dropped `ensure_ascii=False`, which #291's rule names. One queue assertion compared a call to itself and passed for any result. The root skip has one home in `tests/helpers/rootperms.py` instead of four copies, one of which was missing. `tests/test_append_torn_tail.py` was at 796 lines against the 800 ratchet, so the experiments.tsv class moved into its own file. The progress-log repair row is stamped from the event's own timestamp, since it is written above that event.

**Declined, one line each.**

- A warning on the first `EventBus` sink drop. `events` has no logger and the bus has no UI, so the surface it would reach is `orchestrator.log`, which is the surface #333 exists because nobody reads; giving `dropped` a real reader is the fix and it belongs in the run summary. The decision is written down at `JsonlSink` rather than left implicit.
- `repair: str | Callable[[], str]` to build the payload lazily. It adds a second shape for one parameter to save a measured 2.04 us of a 35.82 us `emit`, at each of four sites plus once per sink, and CLAUDE.md's tie-break prefers removing a mechanism to adding one. The timestamp half of that finding is fixed directly instead.
- A content check on `run_id` in the width filter. A row shape check keyed on `mint_run_id`'s format would refuse a hand-edited or foreign row that is otherwise fine, which is a different contract from "this file's writer emits rows of these widths"; the residual is disclosed instead.
- A blank-line separator decided from the live handle in `_ensure_gitignore`. The empty-versus-non-empty half still comes from the marker read, because that read is not removable: it is what stops `ks init` duplicating the block.

## Review round 2

Round 2 raised 0 blockers, 6 should-fix and 8 nits, and the code-review skill contributed nine findings of which two became should-fix items this round would not otherwise have had. Three of the six were a mechanism this PR CLAIMED and did not have. All six are applied.

**F1. experiments.tsv was read on a dialect its writer does not write.** `experiment_rows` parsed with `csv`'s default quoting against a writer that joins its fields on a tab and never quotes or escapes. A `"` at the start of any field opens a quoted region that swallows every byte to the next one. Measured on the head before the change: a three-run file whose FIRST run was named `"proj` returned ZERO rows, and a 20,001-run one raised `_csv.Error: field larger than field limit (131072)`. `_csv.Error` is neither an `OSError` nor a `ValueError` (both `isinstance` checks measured False), so it escaped `get_experiment_trends`, whose guard is around the read and not the parse, and `ks autonomy replay`'s handler as well. Not a regression: 50c965e had the same shape. What was this PR's was the DISCLOSURE, since round 1 replaced a false claim with three measured residuals and the quote was not one of them, and it is different in kind from the other two: a tab or a newline damages its own row, a quote consumes every row after it.

`ExperimentsDialect` is now the one place the file's on-disk shape is written down, with `QUOTE_NONE` and no `escapechar`. The reader parses with it, `EXPERIMENTS_HEADER` is joined on its delimiter, and `record_run` builds its row the same way. Byte-identical on both, verified rather than argued: the header against the previous literal, and the row over 3000 randomised rows including tabs, newlines and quotes. Not through a `csv.writer`, because under `QUOTE_NONE` with no escapechar a writer RAISES on a field holding a tab and `record_run` catches only `OSError`, which would turn residual 1 from a dropped row into a lost run.

The field-size limit fires under `QUOTE_NONE` too, so the parse is guarded: `except Exception` exactly, per CLAUDE.md's rule from #318 that a parser's error taxonomy belongs to the parser, converted to a `ValueError`, which is the refusal path both callers already take, with all the I/O outside the guarded block. `ks autonomy replay` names it and exits 2; `get_experiment_trends` logs it with the path and the cause and returns `[]`, because it runs on the Textual event loop where a raise is #289's crash again, and because `[]` already means three things and a fourth silent one would be a mechanism removed. The cost of materialising the records under one guard rather than streaming under two is measured and disclosed: 18.9 ms and 21.1 MB against 15.0 ms and 17.1 MB on a 20,000-run file.

**F2. The sink's flush was one line below its binding, so the ENOSPC the K1 fix names still left it bound.** `open_for_append` returns a `BufferedRandom` with an 8 KiB buffer and an event line is smaller than that: measured, a 121-byte write leaves 0 bytes on disk until the flush. So an out-of-space error surfaces at the FLUSH, and the flush sat after `self._fh = self._probe_and_write(line)`. The K1 defect re-entered through the one line the K1 fix did not move. The flush is inside `_probe_and_write`'s `try` now, immediately before the handle is returned, and only the no-probe branch of `emit` flushes.

**F3. Layer 3 of the guard missed the alias shape layer 1 explicitly closes.** Measured against the predicate on five snippets: `append_records(...)` and `appendio.append_records(...)` were seen; `from kstrl.appendio import append_records as _ar` plus `_ar(...)`, the same for `appending`, and `_w = append_records` plus `_w(...)` were not. Review's mutation R15, a new appender in `knowledge.py` reached through an aliased import, was still green over 692 tests. This is the repo's dominant defect class, in the round that deleted instance eleven, with the worked example of the fix sitting in the same file. `appendio_entry_names` resolves both binding shapes over the pooled corpus to a fixed point, over-matching in the way layer 1 already documents.

Two controls were missing as well as the resolution. There is no attribute-form call into `appendio` anywhere in `kstrl/`, so that disjunct was uncontrolled and review's R17, narrowing the predicate to `ast.Name`, was also still green; the census assertion has an attribute-form control now. The alias axis cannot be controlled from that assertion, because its predicate builds its names from the package rather than from the control snippet, which is exactly why layer 1's `_o = open` row lives in the detect class; layer 3's alias rows are there too. Two disclosed limits that a resolver this suite does not have cannot close, an entry point stored on an attribute and one reached through `functools.partial`, are strict xfails rather than silence.

The detect class moved to `tests/test_append_guard_detects.py`. Unsplit, the guard file would be 853 lines against the 800 ratchet; it is 685 and 168 now, and this is the seam `tests/test_journal_guard_detects.py` already took for the same reason.

**F4. The two sites round 1's own routing added made three claims in `appendio` false.** Round 1's S2 fix asserted that "each caller says so where it opens" is now true and the S1 fix in the same round made it false again, which is the one-line-apart failure this repo keeps logging. Measured by grep over the eight routed callers: `proposals.py` carried no `"a+b"` clause, the other seven did; `repair=""` is passed at four sites, not two; eight callers do not ask for the lock, not six. The clause is at `proposals.mark_applied` now and names what its two callers actually do with the raise, which differ: `apply_proposal` catches `OSError` and says retrying is safe, `cli`'s own loop does not. The `.kstrl/` sentence names `.gitignore` as its one exception and why it is safe, checked rather than assumed: `_gitignore_state` reads through `_read_text_or_none` and answers `("keep", None)` for a file it cannot read, and `_ensure_gitignore` returns on that answer before the append.

**F5. `_show_repairs`'s stated reason for reading the journal twice was false.** It said the patterns read could not be shared "because that one takes a lookback window and this count is over the whole file". `get_cross_run_patterns` goes through `_read_journal_entries`, which calls `_read_all_entries()` and applies the window in memory, so both reads take the whole file. The second read is kept and the real reason is written down: folding them means either passing entries into a method whose own docstring rejects taking the count from somewhere else, or giving `EvolutionJournal` a cached read that changes what every other caller sees. The doubling is stated with a fresh measurement rather than an arithmetic doubling of round 1's figure: 20 calls each on a 10,000-line journal give 9.3 ms and 9.2 ms, so the screen pays 18.5 ms where one read would cost 9.3.

**F6. The branch conflicted with `origin/main` for the second consecutive round.** It is merged here at `origin/main` 1754357 (#351), and then again at 0522501 (#350), which landed while this round was measuring. Two conflicts, both in files #351 rewrote and this branch also touches, which is CLAUDE.md's "merging into a moving main is where the defects are". In `tests/test_event_names_have_one_home.py`: main's widened docstring and its `from tests.helpers import astwalk` import kept, this branch's `from kstrl.appendio import JOURNAL_REPAIR_EVENT` kept beside `from kstrl.evolution import SPEC_ISSUES_EVENT`. Main's first line said the one place an enrolled name is spelled is `kstrl/evolution.py`, which this branch makes false for `JOURNAL_REPAIR_EVENT`, whose home is `appendio.py`; the sentence names both modules now. The `tests/` figures in that paragraph were re-derived post-merge by running the walk: 46 total, 44 `spec_issues` and 2 `journal_repair`, exactly as written. In `tests/test_journal_torn_tail.py`: this branch's imports kept, the evolution import widened to carry main's `SPEC_ISSUES_EVENT`; every branch-added line survives, checked line by line against the merge base. The second merge's two conflicts are both additive and both keep BOTH sides; H4 names them and what was re-run because of them.

**Nits applied.** N1: `experiment_rows` said blank lines are skipped because the writer's pad is one, and it is not; measured on both tear shapes, the pad writes `"\n" + payload` and leaves no blank line. N2: the `record_run` comment still named `csv.DictReader`; the outcome it promises does hold, checked on a legacy prefix header. N3: `docs/evolution-metrics.md` documented the doubled-repair-row degradation as Windows-only, and `_flock` takes the same path on any mount whose `flock` raises `OSError`. N4: five lines of operator prose were duplicated between `cli._echo_journal_repairs` and `EvolveScreen._show_repairs` with the CLI docstring claiming the wording was deliberately the same on both and nothing keeping it so; `EvolutionJournal.repair_summary()` is the one home, returning `None` at zero so the file is still read once. N8: `_flock` said the release is unguarded because "by then the bytes are flushed", and the flush is the statement immediately above it in the same `finally`, so a raising flush skips `LOCK_UN` and `appending`'s `handle.close()` drops the exclusion instead.

N4 caught something that was not its subject. The full suite on the merged tree went red on `tests/test_journal_one_writer.py`: hoisting the sentence took `evolution.py: self.config.journal_path` from 3 to 4 and left the `cli.py` and `tui/screens/evolve.py` rows with nothing behind them. Two modules fewer get hold of the journal path, which is the direction that guard exists to push, and no targeted run would have shown it.

**Nits declined.** N5, the two contention tests running the same 4 x 200 x 200 KB workload: real duplication, measured at 0.71 s and 0.69 s, and folding them costs a shared-state seam between two tests that currently share nothing. N6, the PR body implying the repair payload is serialised at one site: corrected to four in this section rather than by changing the decline, which stands on CLAUDE.md's tie-break.

N7 is declined WITH A MEASUREMENT, because its suggested fix does not work. Review found that `append_open_census()` and `routed_append_census()` have no caller and that mutations R6 and R13, each replacing one with `return dict(<its own pin>)`, were still green, and suggested `assert append_open_census() == EXPECTED_APPEND_OPENS` inside the census test. That assertion was written and both mutations were re-run with it in place: still green. A function replaced BY the pin, compared TO the pin, agrees by construction, so while the pin matches the walk the two mutants are equivalent and there is nothing to observe. What could observe one is a caller that runs the census over a corpus whose answer is not the pin, and this suite has no such corpus. A check that cannot fail for the reason it names is the shape the whole guard exists to end, so the assertion was removed and the disclosure is on `append_open_census` instead.

## Behaviour changes

- The six appenders above no longer lose the record written after an interrupted write. Four write a `journal_repair` row before it; three pad the tail with no row.
- An UNREADABLE record file is now refused where it used to be appended to blind, at three sites that did not say so. The `"a+b"` open is what buys the refusal (#327 round 1 found the path-taking probe fail-OPEN on a mode-0200 journal), and the cost differs by site: `Inbox.add` raises and `pipeline._inbox_add` warns through the UI; `ProgressLog.emit` and `JsonlSink` are on the same `EventBus`, which catches per sink into a counter with no production reader, so a mode-0200 `progress.jsonl` or `events.jsonl` loses that whole stream with no message on any surface where 568bca4 wrote it. Reaching it needs a deliberate `chmod 0200` or an ACL on a file kstrl created itself at the umask default. All three now say so where they open.
- `ks evolve --status` and the TUI trends tab no longer render a corrupted run from experiments.tsv. Both readers of that file drop a row whose width does not fit the file's header. Two widths are legal, not one: a file written before R3.1 has a shorter header and this writer appends full-width rows onto it, so a filter that took the current header's width as the only legal one would answer a rendering defect by deleting every row of a legacy file.
- A repaired experiments.tsv write is logged. It is the one record file that cannot carry a repair row, so the log is the only place the incident can go.
- `ks autonomy replay` no longer counts a fabricated run, and no longer reports an unreadable file as too little history. `autonomy_replay.load_runs` was the second reader of experiments.tsv and was still using a bare `csv.DictReader`: a concatenated row gave a run id off the fragment, this run's fields shifted along, and `_as_int` turned each shifted value into 0 without raising. It also swallowed the read error; both exceptions reach the command now, which names the file and the error above the same exit code 2.
- `EvolutionJournal.append_entries` takes an exclusive lock on POSIX, best effort. Nothing else in the pipeline changes: the lock is a leaf, and `flock` is per open file description, so two threads in one process serialize too. It blocks unconditionally with no switch, matching `control_lock`'s default and not `queue_lock`'s, so a writer stopped inside the critical section stalls every other writer on that file where before they could not wait at all.
- `EvolutionJournal.get_repair_count` is no longer an upper bound as well as a lower one. Under `fcntl`, one row is one tear.
- The evolve screen shows one new line when the journal holds repairs, and nothing when it does not. A repaired `events.jsonl` still has no live surface: `tui/widgets/activity.py::humanize` returns `None` for an unrecognised event, so the typed `JournalRepaired` row is durable, greppable and rendered nowhere. That is the gap #333's next issue is about, and it is named here rather than left for a reader to find.
- `evolution.experiment_rows` is new and public, because the file has two readers. It never existed on main under any name.
- `proposals.mark_applied` writes byte-identical bytes to an intact file, and one newline more to a torn one, where a blank line is a paragraph break.
- experiments.tsv is read on the dialect its writer writes. A `"` at the start of any field used to open a quoted region that swallowed every row after it, so a project named `"proj` hid every later run from `ks evolve --status`, the trends tab and `ks autonomy replay`, and past 128 KiB it raised `_csv.Error` out of all three. A field longer than `csv`'s own limit is a refusal now: `ks autonomy replay` names it and exits 2, the trends tab logs it and shows no runs.
- An out-of-space error on the FIRST event of a run no longer leaves the sink bound and writing onto a torn tail. The error surfaces at the flush rather than the write, because an event line is smaller than the 8 KiB buffer, and the bind now happens only after both land.
- The repair sentence `ks evolve --status` and the evolve screen print has one home, `EvolutionJournal.repair_summary`. The rendered text is unchanged on both surfaces; what changed is that there is one copy of it, and that two modules stop getting hold of the journal path.

Every number in this section was re-measured at the current head. `origin/main` moved TWICE during this round, and the tree measured here is merged with **0522501** (#350), the second of them. Nothing below is carried over from round 1.

**Gates**, bytecode deleted first and `PYTHONDONTWRITEBYTECODE=1` on every run.

| gate | result |
|---|---|
| `pytest tests/` on the merged tree | **5815 passed, 37 skipped, 41 xfailed, 0 failed** |
| `origin/main` 0522501 in a throwaway worktree, same protocol | **5709 passed, 37 skipped, 38 xfailed** |
| delta | **+106 cases and +3 xfails.** Round 1 added 86 and 1; round 2 adds 20 and 2 |
| the same pair against the FIRST merge, 1754357 | 5717/37/41 against 5611/37/38, the same +106 and +3, measured before #350 landed and not re-run since |
| `premerge.sh 352` at the final head | mergeable=MERGEABLE, state=CLEAN, all five checks SUCCESS, behind main by 0, 34 commits with 0 missing `Co-Authored-By`, 0 em dashes in body, diff and commit messages |
| `uv run mypy kstrl/ --strict` | Success, 131 source files |
| `uv run python scripts/gen_docs.py --check` | README generated sections are current |
| full `pre-commit run --files <the 37 changed files>` | every hook Passed, ruff-format included |

Both suite counts were measured twice, at the head and again after the mutation batch, and they agree. Wall time does not, and is not a signal here: the branch took 371.1s and 482.4s and main took 356.9s and 374.0s, the longer of each pair measured while another suite was running on this machine.

`pre-commit` failed once on `ruff format` (a test signature that fitted on one line) and was fixed rather than skipped. `check_no_mocks.py` deserves one sentence, because run by hand it looks red: fed the whole changed-file list it reports rc=1 on `tests/`, and the hook's own `exclude` drops every test path, which is what "mocks stay in tests" means. Run the way the hook runs it, over the twelve changed non-test files, it is rc=0. `file_length_ratchet.py` and `cyclomatic_ratchet.py` are rc=0, and `complexipy` passes: `appendio_entry_names` came in at 26 against the limit of 15 on its first draft and was split into two named helpers rather than given an escape hatch.

**Line lengths, because two files moved against the 800 ratchet this round.** `tests/test_append_opens_have_one_home.py` would be 853 lines unsplit, and the detect class moved to `tests/test_append_guard_detects.py`, leaving 685 and 168. `tests/test_append_torn_tail.py` is at 781 with 19 lines of headroom, and `kstrl/observability.py` is still at exactly 800, which the ratchet passes and the next line in that file will not.

**Two merges, because main moved twice.** `gh pr view 352` reported `mergeable=CONFLICTING` against 1754357, with two conflicts, both in files #351 rewrote and this branch also touches: `tests/test_event_names_have_one_home.py` and `tests/test_journal_torn_tail.py`. Neither is the three CHANGELOG/`evolution.py`/`test_cli.py` conflicts the previous version of this section named, which were round 1's and are stale. Then 0522501 (#350) landed while this round was measuring, with two more conflicts, both additive and both resolved by keeping BOTH sides: CHANGELOG.md, where neither side's entries are about the other's subject, and `autonomy replay`'s docstring in `kstrl/cli.py`, which gained a paragraph on each side. `kstrl/inbox.py` auto-merged, and that one is worth naming: main widened the exception pair every inbox caller catches while this branch routed `_append` through `appendio`. Everything below was re-run on the twice-merged tree, including all 34 mutations, for exactly that reason.

**Censuses, executed at four revisions and diffed, never read from a pinned literal.** Every test file under `tests/` that parses or walks source was RUN at each: 34 files giving 1112 passed / 9 skipped / 38 xfailed before the first merge, 1113 / 9 / 38 after it (the one new case is #351's own), and 36 files giving 1155 / 9 / 40 on the twice-merged tree. The census script re-derives seventeen inventories by executing each walk.

- **Pre-merge against post-merge:** the ONLY delta is `event_name_spellings` over `tests/`, on the three files #351 rewrote (`test_decompose.py` 27 to 24, `test_evolution.py` 1 to gone, `test_journal_torn_tail.py` 4 to 1). Every `kstrl/` inventory is byte-identical across the merge.
- **Post-merge against the final head:** the only delta is `journal_path_sites`, and it is round 2's own: `cli.py: journal.config.journal_path` gone, `tui/screens/evolve.py: journal.config.journal_path` gone, `evolution.py: self.config.journal_path` 8 to 9. That is the N4 hoist, and the pin carries the reason. Two modules fewer get hold of the journal path.
- **Across the second merge:** the only delta is #350's own new code, `tests/helpers/demotion.py` arriving with 3 event-name spellings and two new signature spellings in `kstrl/autonomy.py` and `kstrl/calibration_ladder.py`. No inventory this branch owns moved.
- Unchanged at all four revisions: `append_opens` (17 sites in 15 keys), `routed_appends` (11 in 11), `event_name_spellings` over `kstrl/`, `journal_writes_outside_append_entries` (empty), `mkstemp_seen` (empty), `mkstemp_undecided`, the four `tests/test_encoding_readers.py` pins, `tomllib_spellings`, `await_sites`, `signature_container_sites`, `signature_spellings`, and the five machine-scope pins.

**Mutations. 34 planted, 31 caught, 3 still green, 0 hung, 0 could-not-plant, and the whole batch run THREE times, once on each merged tree and once more at the final head, with identical outcomes every time.** Every one planted alone, `__pycache__` deleted first, `PYTHONDONTWRITEBYTECODE=1`, a 420s bound per run started in its own session so a timeout kills the process group this harness created rather than `uv` while `pytest` survives underneath, and every plant asserting its anchor matched EXACTLY ONCE before writing, so a mutation aimed at a moved anchor is COULD-NOT-PLANT rather than scored as a pass. Baseline over the 23-file target set: 768 passed, 5 xfailed. Every anchor was re-validated after the second merge and all 34 still matched exactly once.

The set is review round 2's twenty (R1 to R20), round 1's four that round 2 did not repeat (M-G, 10d, S2, and `lock=False` as R10b), and one per fix and disclosed limit added this round (F1a to F1d, F2a, F2b, F3a to F3c, N4a).

| # | mutation | outcome |
|---|---|---|
| R1 | bind `self._fh` before `append_terminated` returns | CAUGHT, 2 failed |
| R2 | delete the `try/except OSError` around `fcntl.flock` | CAUGHT, 3 failed |
| R3 | `proposals.mark_applied` back to a raw append | CAUGHT, 5 failed |
| R4 | `_ensure_gitignore` back to the stale-read separator | CAUGHT, 6 failed |
| R5 | drop the width filter from `experiment_rows` | CAUGHT, 4 failed |
| R6 | `append_open_census()` returns its own pin | **still green**, and equivalent: see the N7 note above |
| R7 | delete the `repair` terminator check | CAUGHT, 1 failed |
| R8 | `load_runs` swallows the read error again | CAUGHT, 1 failed |
| R9 | make the experiments.tsv repair warning unreachable | CAUGHT, 1 failed |
| R10 | the journal `append_records` call drops `lock=True` | CAUGHT, 6 failed |
| R10b | the same call passes `lock=False` | CAUGHT, 6 failed |
| R11 | `handle_ends_without_newline` always returns False | CAUGHT, 52 failed |
| R12 | `flush()` after `LOCK_UN` instead of before | CAUGHT, 3 failed |
| R13 | `routed_append_census()` returns its own pin | **still green**, equivalent for R6's reason |
| R14 | `_probe_and_write` stops closing the handle on a raise | CAUGHT, 2 failed |
| R15 | a new appender in `knowledge.py` through an aliased import | CAUGHT, 1 failed (**was still green in review**) |
| R16 | layer 1's `sees` returns False for every node | CAUGHT, 12 failed |
| R17 | layer 3's predicate narrowed to `ast.Name` | CAUGHT, 2 failed (**was still green in review**) |
| R18 | `lock_argument` always answers "default (False)" | CAUGHT, 1 failed |
| R19 | an unrouted `io.open(p, "ab")` appender in `knowledge.py` | CAUGHT, 4 failed |
| R20 | delete the payload terminator check | CAUGHT, 2 failed |
| F1a | the reader's dialect back to `QUOTE_MINIMAL` | CAUGHT, 14 failed |
| F1b | the parse guard narrowed from `Exception` to `ValueError` | CAUGHT, 3 failed |
| F1c | `get_experiment_trends` stops guarding the parse | CAUGHT, 1 failed |
| F1d | `record_run` joins its row on `"\|"` instead of the dialect's delimiter | CAUGHT, 6 failed |
| F2a | delete the flush inside `_probe_and_write` | CAUGHT, 1 failed |
| F2b | move `emit`'s flush back out of the `else` | **still green**, and benign: see below |
| F3a | layer 3 stops resolving `import ... as` aliases | CAUGHT, 2 failed |
| F3b | layer 3 stops resolving local rebinds | CAUGHT, 2 failed |
| F3c | an appender in `knowledge.py` through a local rebind | CAUGHT, 1 failed |
| N4a | `repair_summary` returns a sentence at zero repairs | CAUGHT, 4 failed |
| M-G | the queue repair row borrows an `item_id` | CAUGHT, 2 failed |
| 10d | an `ALLOWED_APPEND_OPENS` row deleted | CAUGHT, 1 failed |
| S2 | `load_runs` back to a bare `csv.DictReader` | CAUGHT, 1 failed |

**All three still-greens are equivalent mutants, and each was checked rather than asserted.** R6 and R13 are the N7 note above: a function replaced BY its pin, compared TO that pin, agrees while the pin matches the walk, and the assertion review suggested was written and re-run and did not catch either. F2b was checked by planting it and running the head probe: the sink is still unbound after a raising first flush and the probe still repeats, because the flush inside `_probe_and_write` already ran. Planting F2a on top of it fails both checks. So the mechanism is the inner flush and the `else` scoping is redundancy, which is now written on `_probe_and_write` so the next reader does not have to re-measure it.

**Red-first, for each behaviour change this round.** F1: four of the six new experiments tests and the new CLI test were run on the head before the fix and failed, `experiment_rows` returning `[]` on a three-run file and `_csv.Error` escaping to `exit_code=1` in the CLI runner. F2: the new sink test failed on `self._fh is None` with the handle bound. F3: the layer 3 predicate was measured against the five snippets directly. F4, F5, N1, N2, N3, N8 are claim corrections with no behaviour to make red, and each correction was measured (the pad producing no blank line on either tear shape, the eight callers grepped, both `_read_all_entries` paths read, the `flock` release read).

**A reusable probe** is at `scratchpad/lanes/331/r2_verify_head.py`. Run from the worktree with `uv run python`, it checks all four behaviour claims against the shipped code in about a second and exits non-zero naming any that fail: the quote-leading project name returning every row, the oversize field refusing without escaping `get_experiment_trends` or the CLI handler, the sink staying unbound and re-probing when the first flush raises, and layer 3 seeing all six call shapes while staying quiet on `append_terminated`.

**Assumed, and named as assumed.**

- **The no-`fcntl` path.** Every lock test is `importorskip("fcntl")`, so the Windows degradation is stated in three docstrings and tested on no machine here. The `flock`-raises path IS tested by patching `fcntl.flock`, and R2 is red against it.
- **The mode-0200 half of the read refusal** is skipped under root; the undecodable-byte half runs everywhere.
- **The repair line is not asserted as PIXELS**, only as the widget's `display` flag and its content string.
- **The per-appender loss figures in the summary table** were measured by the implementing lane against 568bca4 and re-measured by review against both trees. The production code they describe has not changed since, and `tests/test_append_torn_tail.py` and `tests/test_experiments_tsv_torn_tail.py` re-assert them, but the standalone repro scripts were not re-run this round.
- **The four-writer contention sizing** was not re-measured this round. Review reproduced 10 of 10 red on the mutant and 10 of 10 green on the head at the shipped size, and R10, R10b and R12 are red here.
- **The `serve` admission-cap reachability** behind the inbox's bare pad is confirmed by reading `serve.py`, not by running the daemon.

**Decisions the owner might disagree with.**

1. **`ExperimentsDialect` is a shared constant that the writer uses only for its delimiter.** The writer does not go through `csv` at all and cannot, because under `QUOTE_NONE` a `csv.writer` raises on a field holding a tab and `record_run` catches only `OSError`. So the sharing is real for the delimiter and descriptive for the quoting rules. The alternative was to leave the reader's dialect inline and say the same thing in a comment.
2. **A parse refusal in `get_experiment_trends` is logged and swallowed, where an unreadable file is swallowed silently.** Both return `[]`. The split is that a missing file is the ordinary state of a project with no runs while an unparseable one is an incident, and that this runs on the Textual event loop.
3. **Four writers, not two**, in the contention test, carried over from round 1.
4. **A silently dropped run stream is left silent**, carried over from round 1, with the decision written at `JsonlSink`.
5. **`load_runs` raises where it returned `[]`**, carried over from round 1, and it now raises for a third reason (an unparseable file) as well as the two it already had.

## Checklist

## Provenance

None. No `*_PROMPT` constant and no text reaching an LLM role's prompt is touched, so H2 and H3 do not apply.


kstrl/appendio.py
```diff
@@ -0,0 +1,273 @@
...
+    written out here rather than left to ``close()``: the handle is a
+    ``BufferedRandom``, and unlocking first lets this process's bytes
+    land after the exclusion has dropped.
+    """
+    try:
+        import fcntl
+    except ImportError:
+        yield
+        return
+    fcntl.flock(handle.fileno(), fcntl.LOCK_EX)
```
0xfauzi: **`_flock` blocks forever and lets `OSError` out, unlike both house locks it cites.**

`fcntl.flock(handle.fileno(), fcntl.LOCK_EX)` is unconditionally blocking and unguarded. The two locks the docstring names as the shape being copied both do two things this does not:

- `workqueue.queue_lock` defaults to `blocking=False` and ORs in `LOCK_NB`; `statedir.control_lock` offers the same switch.
- Both wrap the `flock` call in `try/except OSError` and raise a named error (`QueueLockedError` / `ControlLockedError`).

Two consequences on the new path:

1. `flock` is not universally supported. On a mount where it returns `ENOLCK`/`EINVAL` (some FUSE and 9p/DrvFs mounts, e.g. a WSL2 checkout under `/mnt/c`), the `OSError` propagates out of `_flock` -> `appending` -> `append_records` -> `EvolutionJournal.append_entries`, which deliberately lets `OSError` out. Every journal append then raises where it succeeded before this PR, and the module's own stated policy is "healing forward rather than raising, because every caller is a record-keeper". PEP 475 covers `EINTR`, so that one is fine; `ENOLCK` and unsupported-filesystem are not.
2. A stalled or crashed-but-not-reaped holder blocks every other writer indefinitely with no timeout and no diagnostic. `append_entries` is on the factory's commit path.

The measured contention numbers do not cover either case (both arms ran on a local disk with cooperating writers). Suggest matching the house shape: a `LOCK_NB` variant or a bounded wait, and an `except OSError` that degrades to no exclusion with a logged warning rather than failing the append.

kstrl/evolution.py
```diff
@@ -64,6 +70,55 @@
...
+    """
+    reader = csv.reader(io.StringIO(text), delimiter="\t")
+    try:
+        header = next(reader)
+    except StopIteration:
+        return []
+    columns = EXPERIMENTS_HEADER.split("\t")
+    widths = {len(header)}
+    if header == columns[: len(header)]:
+        widths.add(len(columns))
```
0xfauzi, line 248: **The width filter misses a tear inside the first field, which is the run_id.**

A concatenated line has `k + 14 - 1` fields, where `k` is the fragment's field count. That is 15 or more for every tear past the first tab, and `widths` drops it. But a fragment torn *inside* the first field has `k == 1`, so the concatenation has exactly **14** fields, passes `len(fields) in widths`, and renders.

Concretely: `record_run` is interrupted after writing `run_20260101-ab` of `run_20260101-abcdef\t2026-...`, then the next run appends. The surviving line is `run_20260101-abrun_20260101-ghijkl\t2026-...\t...` - 14 fields, a plausible `run_id`, and every other column holding the *real* values of the second run. `get_experiment_trends` renders it, and `autonomy_replay.load_runs` feeds it to the ladder as a run that never happened with a `run_id` that never existed.

`run_id` is column 1, so this is not a corner: it is whatever share of the row's bytes the run id occupies. The PR body says "`ks evolve --status` and the TUI trends tab no longer render a corrupted run from experiments.tsv", which overstates what a width check can do. The docstring should either disclose the residual or the filter should add a cheap content check on column 1 (this writer's run ids have a known shape - `mint_run_id`), since a legal-width row whose `run_id` does not match that shape is exactly the case width cannot see.

kstrl/events.py
```diff
@@ -837,21 +867,53 @@ def close(self) -> None:
...
-        line = event.to_json_line()
+        line = event.to_json_line() + "\n"
         with self._lock:
             if self._fh is None:
-                self._fh = open(self.path, "a", encoding="utf-8")
-            self._fh.write(line + "\n")
+                # First emit: open, probe and write in one call, so the
+                # repair row and the event it protects land in ONE
+                # write and nothing can get between them.
+                self._fh = open_for_append(self.path)
```
0xfauzi: **`self._fh` is assigned before the probe, so a failed first write disables the probe for the whole run.**

```python
self._fh = open_for_append(self.path)
append_terminated(self._fh, line, repair=...)
```

If `append_terminated` raises - the probe's `seek`/`read(1)` hitting `EIO`, or the write hitting `ENOSPC` - `self._fh` is already set. `EventBus.emit` swallows it (`except Exception: self.dropped += 1`), and every later `emit` takes the `else` branch, which writes straight through with the comment "The probe is not repeated". So the torn tail is never repaired, the next event is concatenated onto it, and `read_events` drops both - the exact defect this class was changed to prevent, reached through a path nothing reports.

Assign after the call succeeds:

```suggestion
                handle = open_for_append(self.path)
                append_terminated(
```

(and set `self._fh = handle` after `append_terminated` returns, using `handle` in the call). Then a failed first emit leaves `_fh is None` and the next emit re-probes.

kstrl/evolution.py
```diff
@@ -932,12 +982,31 @@ def record_run(
...
+            #
+            # needs_header is unchanged and still asked before the
+            # append: a file that is missing or empty cannot have a torn
+            # tail, so the two never both fire.
+            #
+            # The other side of the two-sided contract: utf-8 is pinned
+            # in the helper, and this is the file get_experiment_trends
+            # decodes as utf-8.
+            payload = (EXPERIMENTS_HEADER + "\n" if needs_header else "") + row + "\n"
+            append_records(self.config.experiments_path, payload, repair="")
```
0xfauzi: **A repaired experiments.tsv write is now completely silent.**

The return of `append_records` is discarded, `repair=""` writes no row, and there is no `logger.warning` on this path (unlike `append_entries`, `ProgressLog.emit` and `Queue._journal`, all of which log or write a row). The comment above says "The incident is surfaced on the read side instead, by get_experiment_trends dropping a row whose width is not one this writer emits" - but after this change the pad means there *is* no concatenated row to drop. The fragment left behind is a short row that `experiment_rows` drops silently, with no counter and no message anywhere.

So the state after the fix is: a crash tore experiments.tsv, a record was lost, and nothing in the product ever says so. That is the silent code path CLAUDE.md rules out ("Don't add silent code paths - if it's worth deciding, it's worth recording"), and the argument for the `journal_repair` row in the first place (#312: the operator has no other trace) applies identically here.

The bare-pad decision is well argued; the missing signal is separable from it. One line closes it:

```suggestion
            payload = (EXPERIMENTS_HEADER + "\n" if needs_header else "") + row + "\n"
            if append_records(self.config.experiments_path, payload, repair=""):
                logger.warning(
                    "experiments.tsv did not end in a newline, so a crash tore it: "
                    "the tail was padded onto its own line and the row above this "
                    "run may be a fragment (%s)",
                    self.config.experiments_path,
                )
```

kstrl/appendio.py
```diff
@@ -0,0 +1,273 @@
...
+    contract in ``atomicio``: what lands is utf-8, and every reader of
+    it names ``encoding="utf-8"`` and catches ``ValueError`` beside
+    ``OSError``.
+
+    Healing forward rather than raising, because every caller is a
+    record-keeper: refusing to append would answer the loss of one
+    record by losing every later one.
+    """
+    if payload and not payload.endswith("\n"):
+        raise ValueError("append_terminated payload must end in a newline")
```
0xfauzi, line 189: **`payload`'s terminator is enforced; `repair`'s is not, and an unterminated `repair` recreates the defect.**

`append_terminated` refuses an unterminated `payload` precisely because "a caller that hands over an unterminated payload is re-creating the defect this module exists for". Two lines down, `payload = "\n" + repair + payload` accepts `repair` verbatim. A caller passing a repair line without its trailing `\n` glues the repair row onto the first record, producing one unparseable line and losing that record - the same defect, one field over.

The four repair-writing callers all happen to append `"\n"` today, and nothing enforces it: not the type (`str` covers both), not a test (`tests/test_appendio.py` covers `""` and a terminated row, never an unterminated one), and not the new census guard, which only inventories append-mode *opens*. It also only fires on the tear path, which no caller exercises in normal running, so a future caller would ship the bug and see it years later in a torn file.

```suggestion
        raise ValueError("append_terminated payload must end in a newline")
    if repair and not repair.endswith("\n"):
        raise ValueError("append_terminated repair must end in a newline")
```

tests/test_append_torn_tail.py
```diff
@@ -0,0 +1,796 @@
...
+        an operator looking for the incident would go.
+        """
+        queue = self.queue_at(tmp_path)
+        self.add(queue, "a")
+        tear(queue.journal_path)
+        self.add(queue, "b")
+
+        assert repair_rows(queue.journal_entries())
+        assert [e.get("item_id") for e in queue.journal_entries("b")] == ["b"]
+        assert queue.journal_entries("a") == queue.journal_entries("a")
```
0xfauzi: **This assertion is a tautology - it compares one call to itself and passes for any result.**

```python
assert queue.journal_entries("a") == queue.journal_entries("a")
```

It passes for `[]`, and it passes if the repair row *does* carry `item_id == "a"`, which is the thing the test name and docstring say it establishes. The tear is planted after item `a`, so `a`'s history is exactly where a repair row that borrowed an id would show up, and the assertion above it only covers `b`.

Same shape as the rule in CLAUDE.md about `assert hits(...) == []` not being a control: it holds when the mechanism works and when the mechanism is gone. Presumably intended:

```suggestion
        assert [e.get("item_id") for e in queue.journal_entries("a")] == ["a"]
```

kstrl/appendio.py
```diff
@@ -0,0 +1,273 @@
...
+    fcntl.flock(handle.fileno(), fcntl.LOCK_EX)
+    try:
+        yield
+    finally:
+        handle.flush()
+        fcntl.flock(handle.fileno(), fcntl.LOCK_UN)
+
+
+@contextmanager
+def appending(path: Path, *, lock: bool = False) -> Iterator[IO[bytes]]:
```
0xfauzi, line 276: **`lock=False` is the default, and nothing in the new guard can see a routed caller that omits it.**

`test_append_opens_have_one_home.py` layer 1 is a census over places an append handle is *obtained* - `open`, `os.open`, `fdopen`, `NamedTemporaryFile`. A call to `append_records(path, payload, repair=..., lock=True)` obtains nothing, so a new appender that is routed correctly through this module is invisible to both layers. Layer 1's docstring says "a new appender in any shape shows up as an unexplained census delta whatever it does with the handle afterwards", which is true for a raw `open` and not for the routed call the guard is trying to steer people towards.

The result is that the mechanism's own safety parameter defaults to off and has no census. A new record file gets no exclusion, and the diff that adds it contains nothing a reviewer has to justify - the opposite of the `ALLOWED_APPEND_OPENS` design, where adding a row is the point.

The PR body records skipping the keyword-required form because it would add `lock=False` at six sites with "no detection". The detection is available at the same cost as layer 2: a second table keyed `module.py:scope` over `append_records`/`appending` call sites with the `lock=` value each one passes, so an omission is an unexplained row rather than a default.

kstrl/observability.py
```diff
@@ -124,13 +126,29 @@ def emit(
...
+        # ``running`` when the lost row was its ``component_completed``.
+        #
+        # utf-8 is pinned inside the helper, matching the reader below.
+        # ``json.dumps`` leaves ensure_ascii at its default, so what
+        # lands here is pure ASCII today and the locale cannot corrupt
+        # it (#291's two-sided contract, #320's sweep).
+        repaired = append_records(
+            self._path,
+            json.dumps(event) + "\n",
+            repair=json.dumps(self._repair_event()) + "\n",
```
0xfauzi: **The repair payload is built on every append, for a value used only when a tear is found.**

`json.dumps(self._repair_event()) + "\n"` is an argument, so `_iso_now()`, the dict and the serialisation run on every `emit`. The same shape is in three other callers this PR adds: `append_entries` (`_journal_line(_repair_entry())`, which calls `_timestamp_now()`), `Queue._journal` (`json.dumps(_journal_repair_entry(), ...)`), and `knowledge.record_dependency_scope_gap` (a second `_telemetry_timestamp()` call). A tear is a once-per-crash event, so this is ~100% waste by construction.

The PR measured 2.04 us of a 35.82 us `emit` here and correctly called it clean for progress.jsonl. The point is the API shape rather than this one number: `repair: str` forces the work at every call site, and the version that does not is one annotation:

```python
def append_terminated(handle, payload, *, repair: str | Callable[[], str]) -> bool:
    ...
    if repaired:
        payload = "\n" + (repair() if callable(repair) else repair) + payload
```

That also makes the `repair` terminator check cheap to run only on the path that uses it.

kstrl/observability.py
```diff
@@ -142,6 +160,28 @@ def emit(
                     f"progress sink {type(sink).__name__} failed on {event_type}: {exc} (non-fatal)"
                 )
 
+    def _repair_event(self) -> dict[str, Any]:
```
0xfauzi: **The repair row carries a timestamp later than the event it is written above.**

`event["ts"]` is stamped at the top of `emit`; `_repair_event()`'s `ts` is stamped when the argument is evaluated, after it; and the repair row is written *before* the event in the file. So progress.jsonl now contains a non-monotonic pair whenever a repair fires - a row dated after the row that follows it.

Nothing in the readers appears to depend on ordering (`summarize_events` records timestamps, `reducer` takes `max`), so this is a correctness-of-the-record issue rather than a crash: an operator reading the file to reconstruct when the crash happened gets a timestamp from *after* the recovery, describing an event from before it. The evolution journal has the same shape and had it before this PR.

Cheapest fix is to stamp the repair row's `ts` from the same `_iso_now()` value the event uses, or to build it lazily (see the note on line 143) so it is stamped at the moment the tear is actually detected, which is the honest time for it.

tests/test_appendio.py
```diff
@@ -0,0 +1,165 @@
...
+    def test_open_for_append_raises_on_a_file_it_cannot_read(self, tmp_path: Path) -> None:
+        """The point of ``"a+b"``: write-only is refused, not probed blind.
+
+        Round 1 of review on #327 found the opposite shape fail-OPEN: a
+        path-taking probe that answered False on every ``OSError``
+        reported a mode-0200 file as "not torn" and appended to it.
+        """
+        path = tmp_path / "log.jsonl"
+        path.write_bytes(b'{"a":1}\n')
+        path.chmod(0o200)
```
0xfauzi, line 169: **This chmod test has no root skip, unlike the sibling file added in the same PR.**

`tests/test_append_torn_tail.py` defines

```python
skip_as_root = pytest.mark.skipif(
    hasattr(os, "geteuid") and os.geteuid() == 0,
    reason="root ignores the mode bits this test sets",
)
```

and applies it to its two permission tests. This one sets `0o200` and asserts `pytest.raises(OSError)` with no such guard, so it fails outright when the suite runs as root - a container CI default. The two files landed together, so the inconsistency is visible in one diff.

Add the same marker (or hoist `skip_as_root` into `tests/helpers/` so the two cannot drift).

kstrl/evolution.py
```diff
@@ -64,6 +70,55 @@
...
+    reader = csv.reader(io.StringIO(text), delimiter="\t")
+    try:
+        header = next(reader)
+    except StopIteration:
+        return []
+    columns = EXPERIMENTS_HEADER.split("\t")
+    widths = {len(header)}
+    if header == columns[: len(header)]:
+        widths.add(len(columns))
+    return [
```
0xfauzi, line 249: **A dropped row leaves no trace, so "no runs" and "every row dropped" look identical.**

The comprehension silently discards anything that is not one of the two legal widths. That is the right call for a concatenation, but it also silently discards:

- every row of a file whose first line is not a header this writer wrote (a truncated or replaced header makes `widths` garbage and drops the whole file);
- any row whose content adds a field - a tab reaching `project` or `common_failure` - which `csv.DictReader` used to return with the overflow under the `None` key.

In both cases `get_experiment_trends` returns `[]` and the trends tab renders empty, which is exactly what it renders for a project that has never run. `load_runs` returns `[]` and the ladder sees a population of zero, which reads as "not enough runs to promote" rather than "the file is unreadable".

Returning the count of dropped rows (or logging it once) would separate the two states. This is the read-side half of the same audit-trail gap as the silent pad on line 1009.

kstrl/autonomy_replay.py
```diff
@@ -170,14 +169,27 @@ def _as_float(value: str) -> float:
...
+    """
     if not path.exists():
         return []
     runs: list[RunRecord] = []
-    with path.open(encoding="utf-8", newline="") as handle:
-        for row in csv.DictReader(handle, delimiter="\t"):
-            if not row.get("run_id"):
-                continue
+    try:
+        text = path.read_text(encoding="utf-8")
```
0xfauzi: **`newline=""` was dropped from the read.**

The previous code was `path.open(encoding="utf-8", newline="")`, which is what the `csv` module documents as required for a file handed to a reader. `read_text()` applies universal-newline translation, so a `\r\n` inside a quoted field reaches `csv.reader` as `\n` and a lone `\r` becomes a record separator.

Narrow for a TSV this writer produces, and the switch to `read_text` is what buys the `ValueError`-beside-`OSError` handling that the encoding pin needs, so this is not an argument for reverting. But `evolution.experiment_rows` builds `io.StringIO(text)` and both readers now go through it, so if the pair is meant to keep the csv contract it belongs there: `io.StringIO(text, newline="")` keeps the reader's view of the bytes intact and costs nothing.

The `EXPECTED_CLEARED_READS` note in `tests/test_encoding_readers.py` argues the new disposition is stronger, which is true about the *encoding* half; it does not mention that the csv half changed.

kstrl/workqueue.py
```diff
@@ -852,12 +866,38 @@ def _journal(self, entry: JournalEntry) -> None:
...
         """
         self.path.mkdir(parents=True, exist_ok=True)
         line = json.dumps(entry.to_dict(), ensure_ascii=False)
         try:
-            with open(self.journal_path, "a", encoding="utf-8") as handle:
-                handle.write(line + "\n")
+            append_records(
+                self.journal_path,
+                line + "\n",
+                repair=json.dumps(_journal_repair_entry(), ensure_ascii=False) + "\n",
```
0xfauzi: **`ensure_ascii=False` on a line kstrl writes.**

CLAUDE.md, on the two-sided encoding contract: "Do not 'improve' that to `ensure_ascii=False`: #291 tried it, and it made kstrl the SOURCE of non-ASCII bytes on ordinary LLM output (one curly quote in a component description), which broke six readers under `LC_ALL=C`."

The entry line above it already spelled it and this new line copies that, and the repair row's own content (`ts` plus the literal `journal_repair`) is ASCII either way, so nothing breaks today. Worth flagging because it is a *new* line spelling the setting the rule names, in a PR whose whole subject is the durability of these files, and because the row is now built by a shared helper (`_journal_repair_entry`) whose output could gain a field later. Dropping the argument here costs nothing and matches every other repair row this PR adds (`observability`, `knowledge` and `evolution` all leave `ensure_ascii` at its default).

kstrl/evolution.py
```diff
@@ -65,6 +71,100 @@
...
+
+    ``newline=""`` on the buffer is what ``csv`` documents as required
+    for a stream handed to a reader. It is only half the contract here
+    and the other half is not reachable: both callers arrive with text
+    from ``Path.read_text``, which has already applied universal-newline
+    translation, and ``read_text`` grew a ``newline`` parameter in 3.13
+    while this project's floor is 3.11. So a lone ``\r`` inside a field
+    is an ``\n`` before it gets here, which residual 1 covers.
+    """
+    reader = csv.reader(io.StringIO(text, newline=""), delimiter="\t")
```
0xfauzi: **`csv.reader` honours quotes that this file's writer never emits, so one `"` at the start of a field silently deletes every later run — or raises `_csv.Error`.**

`row` is built by f-string tab-joining, so nothing quotes or escapes. But `csv.reader` here runs with default `quoting=QUOTE_MINIMAL`, so a field that *starts* with `"` opens a quoted field that consumes the rest of the file. The docstring enumerates a tab and a newline in `project_name` as the residuals; the quote is a third unescaped metacharacter and it costs more than either.

Measured, `--project-name '"quoted'` on run 2 of 4:

```
rows returned: 1   # run-1 only; run-2, run-3, run-4 gone
```

Every run after the offending one vanishes from `ks evolve --status`, the TUI trends tab, and `ks autonomy replay`'s population, with no counter anywhere (the docstring's own fourth point).

On a file larger than the 128 KB field limit past that row it stops being silent and starts being a traceback:

```
file size 267011
RAISED Error field larger than field limit (131072)
```

`_csv.Error` is not a `ValueError`, so it is caught by neither `get_experiment_trends` (whose `try` wraps only the `read_text`) nor the new `autonomy_replay_cmd` handler (`except (OSError, ValueError)`). It escapes out of `EvolveScreen.on_mount` on the Textual event loop — the exact #289 shape `get_experiment_trends`'s docstring says it fixed — and out of `ks autonomy replay` as a traceback.

```suggestion
    reader = csv.reader(io.StringIO(text, newline=""), delimiter="\t", quoting=csv.QUOTE_NONE)
```

Measured with `QUOTE_NONE`: all three runs returned, `project_name` = `'"quoted'` verbatim.

kstrl/events.py
```diff
@@ -837,23 +867,104 @@ def close(self) -> None:
...
-                self._fh = open(self.path, "a", encoding="utf-8")
-            self._fh.write(line + "\n")
+                self._fh = self._probe_and_write(line)
+            else:
+                # Every later emit writes straight through. The probe is
+                # not repeated: this handle has been at the end of the
+                # file since the first emit, so nothing can have torn it
+                # without this process being dead.
+                self._fh.write(line.encode("utf-8"))
             self._fh.flush()
```
0xfauzi: **The flush is outside `_probe_and_write`'s guard, so the ENOSPC case the method exists for still binds the handle.**

`_probe_and_write`'s docstring: *"a first emit that hit `ENOSPC` on the write or `EIO` on the probe's read was silent; the sink then took the no-probe branch for the rest of the run"*. That is still reachable. `append_terminated` writes into a `BufferedRandom`; a JSON line is far under the 8 KB buffer, so no `write(2)` happens inside the `try`. The first real syscall is this `self._fh.flush()`, one line after `self._fh` has already been assigned.

So on a full disk: probe finds the tear, the repair + first line go into the buffer, `self._fh` is bound, `flush()` raises `OSError`, `EventBus.emit` catches it into `dropped` (which nothing reads), and every subsequent emit takes the `else` branch. The sink now believes it repaired a tear it may not have written.

The guard only covers a `write` big enough to force a buffer flush, and the probe's read. Moving the flush inside `_probe_and_write`'s `try` (before `return handle`) makes the docstring's claim true:

```python
        handle = open_for_append(self.path)
        try:
            append_terminated(handle, line, repair=...)
            handle.flush()
        except BaseException:
            handle.close()
            raise
        return handle
```
(and then `emit` flushes only on the `else` branch).

kstrl/tui/screens/evolve.py
```diff
@@ -207,6 +216,63 @@ def _load_patterns_and_trends(self, root_dir: Path) -> None:
...
+        the one to watch if journals ever get that big.
+
+        Not shared with the patterns read, because that one takes a
+        lookback window and this count is over the whole file. Sharing
+        them would make the repair count depend on how many runs the
+        patterns tab happens to be showing, which is a count of
+        something else.
+        """
+        line = self.query_one("#evolve-repairs", Static)
+        if journal is None or not (repairs := journal.get_repair_count()):
```
0xfauzi: **The stated reason for not sharing the read is not true, and the cost is a second full parse of the journal on the Textual event loop.**

The docstring says: *"Not shared with the patterns read, because that one takes a lookback window and this count is over the whole file. Sharing them would make the repair count depend on how many runs the patterns tab happens to be showing."*

`_read_journal_entries` reads the **whole** file and windows afterwards:

```python
def _read_journal_entries(self, lookback_runs: int = 10):
    entries = self._read_all_entries()      # every entry, no window
    ...
    return [e for e in entries if e.get("run_id", "") in allowed]
```

So `get_cross_run_patterns()` on the line below already paid for the full `read_progress_events` + `json.loads` of every line. `get_repair_count()` does it a second time. Sharing one `_read_all_entries()` result between the two would not make the count depend on the window, because the window is applied after the read — the reason given for keeping them separate cannot happen.

That doubles the measured 12.7 ms at 10k lines to 25 ms of blocking parse on the event loop, on mount and on every `r`.

kstrl/tui/screens/evolve.py
```diff
@@ -207,6 +216,63 @@ def _load_patterns_and_trends(self, root_dir: Path) -> None:
...
+            line.display = False
+            return
+        line.display = True
+        line.update(
+            Text(
+                f"▲ journal: {repairs} interrupted write(s) repaired. A crash left "
```
0xfauzi: **Five lines of operator-facing prose duplicated verbatim from `cli._echo_journal_repairs`, with no constant and no guard.**

This string and the one at `cli.py:733` differ only by the `▲ ` prefix and leading whitespace. The docstring calls that deliberate — *"the wording is deliberately the same on both"* — but nothing holds it that way, and this PR is itself the argument against leaving it: `REPAIR_DETAIL` was hoisted into `appendio` precisely because *"it was written out three times, once per appender, and the three had already drifted into 'a complete event' against 'a complete record' for one incident."*

Two copies of a sentence in two modules is the same setup, one copy earlier. `tests/test_event_names_have_one_home.py` exists in this repo because a name spelled twice drifted; there is no equivalent for this sentence.

A module-level template next to `get_repair_count` (which both renderers already call) that takes the count and the path would leave each surface owning only its prefix and its `UI`/`Text` mechanics.

kstrl/appendio.py
```diff
@@ -0,0 +1,326 @@
...
+lines. Writing a newline FIRST repairs the tail into a line of its own.
+That drops a genuine fragment, which was never readable and was never
+going to be, and RECOVERS a tail that lost only its terminator, which
+was a complete record. The second case is why the cost is not one entry:
+appending onto a whole record destroys that record as well as the new
+one.
+
+What is NOT shared is what each file should WRITE on finding a tear, so
+that is the caller's ``repair`` argument and nothing here has an opinion
+about it. Two callers pass "" for a bare pad, and both have a reason
```
0xfauzi: **Two counts in this module's docstring are wrong as of review round 1, and the second one leaves two files with no stated exclusion argument.**

*"Two callers pass `""` for a bare pad"* — four do:

```
kstrl/evolution.py:1108   append_records(..., repair="")
kstrl/inbox.py:535        append_records(path, line, repair="")
kstrl/init_cmd.py:1352    append_records(path, separator + gitignore_block(language), repair="")
kstrl/proposals.py:181    append_records(path, f"\n**Applied**: {applied_at}\n", repair="")
```

The paragraph then names the reasons for only the two it counted (inbox, experiments.tsv).

The same drift is in `appending`'s docstring at line 294: *"The other six appendio callers either hold an outer lock already (`inbox`..., `workqueue`...) or have one writer process per file (`progress.jsonl`, `events.jsonl`, `engineer.jsonl`, the E8 telemetry log, `experiments.tsv`)"*. There are eight callers in nine call sites; the enumeration omits `proposals.mark_applied` and `init_cmd._ensure_gitignore`.

That omission is the substantive half. `.gitignore` is the one file here that a **foreign** process routinely writes (the user's editor, a second `ks init`, a scaffolder), and it is the one caller with no sentence saying why it needs no exclusion. Layer 3 of the guard pins `lock=default (False)` for both rows, so the pin records the choice without anyone having had to argue it.

kstrl/evolution.py
```diff
@@ -955,12 +1075,39 @@ def record_run(
...
+            # journal_repair row, on the one file that cannot carry one.
+            #
+            # needs_header is unchanged and still asked before the
+            # append: a file that is missing or empty cannot have a torn
+            # tail, so the two never both fire.
+            #
+            # The other side of the two-sided contract: utf-8 is pinned
+            # in the helper, and this is the file get_experiment_trends
+            # decodes as utf-8.
+            payload = (EXPERIMENTS_HEADER + "\n" if needs_header else "") + row + "\n"
```
0xfauzi, line 1217: **The comment 50 lines above this line still describes the reader that was just replaced.**

`kstrl/evolution.py:1055-1056`:

```python
# csv.DictReader in get_experiment_trends drops the extra values
# rather than crashing.
```

`get_experiment_trends` no longer uses `csv.DictReader` — it calls `experiment_rows`, and the new width filter does not "drop the extra values", it drops the **whole row** unless its width is exactly `len(header)` or `len(EXPERIMENTS_HEADER.split("\t"))`. That is residual 2 of `experiment_rows`, and the comment now promises the opposite of what the code does for a file whose header is not a prefix of the current one.

Same paragraph, one line up: the R3.1 tolerance claim is still correct only because the old header happens to be a prefix; the comment should point at `experiment_rows`'s prefix rule rather than at a reader that is gone.

docs/evolution-metrics.md
```diff
@@ -55,12 +55,25 @@ did. Reading one of these rows:
...
+  file and adds none.
+
+  It is no longer an UPPER bound as well. Two processes repairing the
+  same tear used to write two rows (#330 residual 2). Since #331 the
+  append holds `fcntl.LOCK_EX` on the journal's own descriptor across
+  the probe and the write, so the second writer probes after the first
+  has finished and finds a terminated file: one row is one tear.
+  Measured, two processes x 150 appends with 74 tears planted, eight
+  runs of each arm: 76 to 86 rows unlocked, exactly 74 locked on every
+  run. On a platform with no `fcntl` (Windows) nothing is excluded and
```
0xfauzi: **The degradation is stated as Windows-only, but the code takes the same path on any mount whose `flock` raises.**

`appendio._flock` has two paths to "no exclusion", and this paragraph documents one:

```python
    try:
        import fcntl
    except ImportError:
        yield; return
    try:
        fcntl.flock(handle.fileno(), fcntl.LOCK_EX)
    except OSError:
        yield; return
```

The second is the one the PR body measured (`ENOLCK`/`EINVAL` on some FUSE, 9p and DrvFs mounts) and the one `_flock`'s own docstring calls out as *"TWO ways the exclusion is unavailable, and both take the same path"*. On such a mount, on Linux, `get_repair_count` is an upper bound again — exactly what this paragraph tells an operator is now impossible off Windows.

`docs/` is where an operator reads this, so the sentence should name the mount case as well as the platform case.

tests/test_journal_lock.py
```diff
@@ -0,0 +1,391 @@
...
+#: writers below takes under a second on this machine; this is the fuse,
+#: not the expected duration, so it is generous by two orders of
+#: magnitude. A child still alive at this point is a deadlock and is
+#: reported as one.
+JOIN_TIMEOUT_S = 90.0
+
+#: Sized by measurement, not by taste: see the class docstring. Rounds
+#: per writer, bytes of padding per entry, how often a writer "crashes"
+#: and leaves a fragment, and how many writers contend.
+ROUNDS = 200
```
0xfauzi, line 59: **The two contention tests run the identical 4 x 200 x 200 KB workload twice, and `_read` already returns everything both of them need.**

`_run_writers` returns `(run_ids, repairs, unparseable)`. `test_no_record_is_lost_to_a_concurrent_writer` uses the first, `test_one_unreadable_line_gets_one_repair_row` uses the other two, and each calls `_run_writers` on its own `tmp_path`.

That is 4 x 200 x 200 KB ~= 160 MB written per test, 320 MB for the class, and `_read` slurps each file whole and then `split(b"\n")`s it, so peak RSS is roughly twice the file for the duration of each read.

A class-scoped (or module-scoped) fixture that runs the writers once and hands both tests the tuple halves the disk and the memory and asserts on exactly the same run. The two assertions are independent views of one experiment, not two experiments.

kstrl/observability.py
```diff
@@ -124,13 +126,35 @@ def emit(
...
+        #
+        # The ``"a+b"`` open widens what can fail: a log this process
+        # can write but not read is refused rather than appended to
+        # blind, so this RAISES where it returned. ``JsonlSink``'s
+        # docstring carries the decision, because the factory reaches
+        # this through a sink on the same bus and both go quiet at once.
+        repaired = append_records(
+            self._path,
+            json.dumps(event) + "\n",
+            repair=json.dumps(self._repair_event(event["ts"])) + "\n",
```
0xfauzi, line 149: **The repair row is serialised on every append at four sites now, not the one the decline was measured on.**

`self._repair_event(...)` builds a dict and `json.dumps` serialises a ~250-character constant on every single `emit`, for a row that is written approximately never. The PR body declines the `repair: str | Callable[[], str]` shape on the measured cost of 2.04 us out of a 35.82 us `emit` and CLAUDE.md's remove-a-mechanism tie-break.

The measurement covered this call site only. The same eager build now happens at three more:

- `evolution.append_entries` — `_journal_line(_repair_entry())`, including a `_timestamp_now()` call, per append;
- `workqueue.Queue._journal` — `json.dumps(_journal_repair_entry())` per queue transition;
- `knowledge.record_dependency_scope_gap` — a dict plus `json.dumps` per row.

Four sites paying for a never-used payload changes the arithmetic the decline rested on. Worth re-deciding, or worth noting in the declined list that the cost was measured at one of four.

Issue #306, opened 2026-08-31
check_mutation_score's read-only skip records a passing check, so a non-measurement counts as evidence


0xfauzi:
Declined during PR #300 (#288) with reasons, and recorded here so the argument is not lost.

## The defect

When `run_mechanical_verification` runs with `read_only=True`, `mutation_testing` is skipped because mutmut works by rewriting source. That skip is recorded as a **passing** `CheckResult`.

So a check that measured nothing reports a pass, and any caller computing `all(check.passed for check in result.checks)` counts a non-measurement as evidence.

## Why it matters, and why PR #300 did not fix it

PR #300 argued exactly this representation point for the diff-dependent checks, and acted on it: `narrow_to_undiffed` makes those checks **absent** from the result rather than present-and-passing, precisely so `all(c.passed)` cannot be fooled. `check_mutation_score`'s read-only skip is the same shape, left inconsistent with the decision made next to it.

It was left alone because changing it alters `ks sense`'s documented `--json` output: the `checks` array would lose an entry that consumers may key on. That is a compatibility question with its own answer, not something to decide inside a fix for a different issue.

It does not clash on the feature path today, because `mutation_testing` is in `DIFF_DEPENDENT_CHECKS` and is therefore already absent there.

## What to decide

Either the skip becomes absent (consistent with #300, breaks the documented `--json` shape), or it stays present with an explicit third state that is neither pass nor fail (consistent with the schema, needs `SENSE_SCHEMA_VERSION` and every consumer to handle it).

The third option, leaving a non-measurement recorded as a pass, is the one this project's own rules argue against: do not add unverifiable claims to results without flagging them.

## Related

- #288 / PR #300, which took the opposite decision for the diff-dependent checks.

0xfauzi:
Verified and reproduced against origin/main at 130cd66. On a throwaway repo, `ks sense --json` returned:

    mutation_testing  True  | Skipped: mutation testing rewrites the files it mutates and cannot run read-only

Scope correction: this is bigger than filed. `check_mutation_score` has four non-measurements that all report a pass - read-only, mutmut not installed, no non-test Python files changed, and timeout. The last three reach `ks factory` too, not just `ks sense`. Fixing only the read-only case leaves three.

[...]

PR #353, opened 2026-09-06
fix(verify): one dead-code row spoke for two phases, and one of them had not run (#335)

### LINKED ISSUE #335: check_dead_code reports a pass on four non-measurements, the same defect as #306
Found during PR #334 (issues #306 and #316), 150 lines from the code that PR fixes. Not fixed there because the honest fix is structural.

## The defect

`check_dead_code` has the same shape #306 described for `check_mutation_score`: paths where nothing is measured and the check records a **pass**. Four of them.

## Why it matters more than a display bug

#306 was filed as a `ks sense` display problem. PR #334 established that the real consequence is elsewhere: `kstrl/review.py:813-816` builds the LLM code reviewer's `verification_summary` from every check, with

```python
status = "PASS" if check.passed else "FAIL"
```

So a non-measurement reaches an adversarial role as evidence of a pass. On `origin/main` that literally emitted:

```
- mutation_testing: PASS - mutmut not installed, skipping
```

The same path carries `dead_code`. An adversarial reviewer being told a gate passed when the gate did not run is the failure mode the whole factory exists to prevent.

## Why PR #334 did not fix it

`check_dead_code` fuses two phases into one row: a ruff phase that ran and a vulture phase that did not. Omitting the row (the fix #334 chose, on the grounds that `passed` is the only field every consumer reads) would discard a real ruff result along with the absent vulture one.

The honest fix is to split the row. That moves `DIFF_DEPENDENT_CHECKS`, `evolution`'s category table, and `feature_verify`'s baselines, which is a wider change than #306 was scoped for.

## The precedent to follow

PR #334 chose **omission** over an explicit not-measured status, for a reason that applies here too: `passed` is the only field every consumer reads (`all(c.passed)`, `report_lines`, `ks sense --json`, `review.build_review_prompt`), `SENSE_SCHEMA_VERSION = 1` has readers the repo cannot enumerate, and widening to `bool | None` makes `all(c.passed)` falsy so the gate fails on a check that never ran. `kstrl/feature_verify.py` already states omission as the repo's written rule: suppressed checks are absent rather than recorded as passing skips.

## Related, also unfiled

The other three `PipelineHooks` fields remain typed `Callable[..., VerificationResult]`-style and unprotected. PR #334 typed only `run_mechanical_verification`, and recorded the measurement for the rest: `run_review` and `run_security_review` have adjacent transposable `Path` arguments, and `distill_facts` has three among ten positional arguments.

0xfauzi:
## Summary

`check_dead_code` fused two phases into one row: a ruff `F401,F811,F841` auto-fix that scans `.`, and a vulture scan (or the operator's own `[verify] dead_code_command`) over the non-test Python files in the diff. One `CheckResult` carried both verdicts, and `kstrl/review.py` renders every row as `PASS` or `FAIL` into the LLM code reviewer's verification summary. So a phase that never ran reached an adversarial role as evidence of a pass.

The issue said four such paths. Driving the fused function through mocks identical to the ones its own tests use found **nine**, every one of them printing `dead_code  pass`: six where the detector did not run (vulture absent, vulture absent after ruff fixed 2, the scan timing out, the scan exiting 2 with no output, nothing scannable in the diff, and vulture absent under `ks sense`, which runs read-only) and three where ruff did not (absent, timed out, exit 2 on a bad `ruff.toml`), the last three silently, inside a row that reported only the vulture verdict. An independent review of round 1 found **three more of the same shape**, all of them measured, and they are fixed here rather than filed: ruff output with no summary line in it, a diff git could not read at all, and a `git commit` that did not land. A second independent review found **two more**, both of them introduced by the round-2 fix and both measured against real ruff: a fixing run whose findings all have unsafe fixes, and a clean tree under ruff 0.2.0 through 0.3.2. The count is now fourteen.

Two of the original nine are why omitting the row (the fix PR #334 chose for `mutation_testing`) is wrong here: with vulture absent and ruff having fixed 2, or having reported 3 auto-removable under `ks sense`, dropping the row would throw a real measurement away with the absent one.

So the row is split, and each half answers for itself:

- `check_dead_code_ruff` -> `dead_code_ruff`, a row when ruff ran AND said what it did, because zero fixes is a measurement. `tool_missing` / `timed_out` / `command_failed` when it did not.
- `check_dead_code` -> `dead_code`, the vulture-or-`dead_code_command` phase, which is the one that can FAIL. `tool_missing` / `no_target` / `timed_out` / `command_failed` when it did not run. It loses `read_only`: vulture and an operator's detector read the tree without changing it, which is the move #334 made for mutmut.
- `_dead_code_checks` owns the order. ruff runs FIRST so the detector scans a tree with the ruff-fixable subset already deleted, and a test asserts that order, because splitting the function made it a property of the caller rather than of one function body.

A phase that could not run records no row and appends a `NotMeasured(check, reason, detail)` to `VerificationResult.not_measured`, which is #306's rule and #334's vocabulary. No new reason token, no new status, no new phase. One toggle, `[verify] dead_code_cleanup`, still owns both phases: splitting the row is not splitting the switch.

**`dead_code` stays on the vulture phase**, which is the reverse of what the issue sketched. Three reasons, in order of weight. `signatures_from_verification` emits `<name>:<code>` only for `passed=False`, and the ruff phase has no failing return, so every `dead_code:*` signature that could ever have been written came from the vulture phase; keeping the name there keeps journals comparable (moot on disk, where the five evolution journals on this machine carry zero `dead_code:*` check signatures, so this is structural rather than observed). `DIFF_DEPENDENT_CHECKS` lists `dead_code` because it reads `git diff`, and only the vulture phase does; ruff scans `.`. And `[verify] dead_code_command` replaces the vulture phase, so a row named `dead_code_vulture` would be misnamed exactly when the operator brings their own detector. `dead_code_ruff` is therefore the new name and no existing consumer moves.

Closes #335

## Round 2: what the review found, and what it changed

Two blockers, four should-fix and ten nits, all measured by the reviewer and all reproduced here before being fixed. The two blockers were both in the code this PR wrote.

**B1. An output ruff did not explain was read as zero fixes.** `_ruff_count` scraped a summary line and returned 0 when nothing matched, and 0 is also what "ruff removed nothing" says. A project setting `[tool.ruff] output-format` to a non-text format makes ruff print its findings with no summary line at all. Reproduced on the round-1 head against the real ruff 0.16.1, with `output-format = "json"` in the measured project's own `pyproject.toml`: exit 1, two unused imports removed from disk, `M mod.py` left uncommitted, and the row reading `dead_code_ruff  pass  ruff auto-fixed 0`. That is the defect class this PR exists to close, inside the gate that closes it, and the reviewer is right that the round-1 docstring described the open path in the course of claiming it was shut.

Two layers, because one is not enough. `--output-format=concise` is pinned on the command line for the same reason `--no-fix` already was: measured on ruff 0.16.1, the flag beats both `[tool.ruff] output-format = "json"` and `RUFF_OUTPUT_FORMAT=json` in the environment, and concise still prints the `Found N errors`, `(M fixed, K remaining)` and `[*] N fixable` lines this phase reads. And `_ruff_count` now returns `int | None`, `None` for any output matching no shape it positively recognises, which the check turns into a `command_failed` gap. The pin closes today's route in; the refusal is what holds when a later ruff renames its summary. The commit is still gated on a count greater than zero, which is now always a real number rather than possibly a fabricated one.

**B2. Diff-derived file names reached `/bin/sh`.** `run_scrubbed` shells out for a string, `_changed_non_test_python` feeds two call sites, and the names come from `git diff --name-only` over a diff an agent wrote, which is the least trusted input in the factory. Measured through the head's own `check_dead_code` with the diff stubbed to `src/my file.py` and `src/$(id).py`: the command built was `vulture src/my file.py src/$(id).py --min-confidence 80`. The space splits the argument, vulture reports two paths it cannot find, and this check reads that error text as findings and FAILs the component for a reason naming the wrong cause; `$(id)` is command substitution the shell executes. vulture is now an argv list, which `run_scrubbed` runs with no shell. The mutation gate keeps its string and quotes instead, because `run_scrubbed` shells out for a string. Round 2 shipped the paths space-separated on the guess that mutmut read them as one option plus one positional argument, and stated that guess in a comment; it is wrong, and round 3 measured it and replaced it with the comma form. See B2-R2 below.

The four should-fix, all measured:

- **A diff git could not read was reported as `no_target`.** `git.get_diff_names` defaults to a lenient read whose own docstring calls it fail-OPEN: a bad base ref, a missing `origin/<base>`, a non-repository and a timeout all arrive as `[]`, and both callers turned `[]` into `no_target`, the one reason token that means "nothing to scan, not a fault". The read is strict now and the failure is `command_failed`, for the dead-code gate and the mutation gate together, since the helper is the one home for both.
- **A `git commit` that did not land was reported as `ruff auto-fixed N`.** `run_scrubbed` returns a `CompletedProcess` and never raises on a non-zero exit, and neither status was read, so a commit refused by a repo hook or aborted for want of a `user.email` left ruff's deletions on disk while the row said they were committed. The round-1 docstring's excuse, that the next check would notice, is false: `check_diff_scope`, `check_bad_patterns` and `check_test_adequacy` all run BEFORE this point. Both statuses are read and the row says `ruff auto-fixed 2, 1 remaining, not committed: <reason>`.
- **`ks feature` suppressed `dead_code_ruff` with no row, no gap and no line in the report.** `narrow_to_undiffed` clears `dead_code_cleanup`, one toggle over two phases, and `dead_code_ruff` is correctly not in `DIFF_DEPENDENT_CHECKS`, so the announcement could not cover it. That was a silent non-measurement the split itself introduced, one file over from the one it was closing. The report now names it on a line of its own, with the toggle as the reason rather than the diff, because appending the name to the diff sentence would state a false one. The first draft of that fix keyed the line on `config.dead_code_cleanup`, which is a branch that can never run: the config reaching `_announce_verification` has already been narrowed. It is unconditional.
- **The census is a ledger, and says so.** Round 2 took it from nine stubbed states to twelve, including the two B1 states and the git-read one, and the docstring stopped claiming a completeness the test does not have. Closure by construction is the better shape and is not available here: the resource being enumerated is what two external tools print. What replaced it was the default, an output matching no known shape being a gap rather than a zero. Round 3 took the ledger to fourteen and found the limit of that default the hard way; see B1-R2 and N1-R2.

The ten nits: seven applied, three declined. Applied: the read-only row now counts the `[*] N fixable` line rather than `Found N errors`, which differ by ruff's unsafe fixes (on one measured tree, `ks sense` said 3 while the factory removed 2, and the word in the message is "removable"); the fixing row carries what ruff left behind; `_last_output_line` strips each stream before choosing, so a whitespace-only stderr no longer returns "no output" while the diagnostic sits in stdout; `test_sense_never_edits_stages_or_commits` asserts something that discriminates and needs no ruff on PATH; `_which_only` takes the full `shutil.which` signature; `ks sense` stops demanding a readable base when no dead-code phase reads one; and this body's numbers are re-measured below. Declined, one line each: the wording of `ruff reports 0 auto-removable` on a tree with no Python files (ruff's own verdict there is `All checks passed!`, and the `dead_code` gap on the same run already tells the operator there was nothing to scan; the alternative is a branch whose only product is prose); the reviewer's summary gaining a `not_measured` line (it changes what an LLM role is shown, which is #334's precedent to revisit, not this PR's); and extracting the dead-code gate into its own module (accurate and worth doing, but a relocation on top of a behaviour fix invalidates the mutation evidence, so it wants its own `refactor:` PR with the set re-run).

## Round 3: what the second review found, and what it changed

Two blockers, five should-fix and six nits, all measured by the reviewer. Both blockers were in code round 2 wrote, and one of them is the same function round 1's blocker was in. Everything below was reproduced here before being changed.

**B1-R2. A shape ruff prints on an ordinary tree was refused, so a healthy run was reported as a tool failure.** Round 2's fixing branch recognised two shapes, `Found N errors (M fixed, K remaining)` and `All checks passed!`. ruff omits the parenthetical entirely when `--fix` had nothing SAFE to fix and prints `Found N errors.` on its own, and `_ruff_count` returned `None`, which the caller turns into `command_failed: ruff check exited 1 and printed no count line`. The condition is one `x = do_thing()` left behind by an engineer iteration. This is round 1's defect inverted: round 1 read an unknown output as zero, round 2 read a known output as unreadable, and the second is what a refusal-by-default costs when the recogniser set was reasoned about rather than measured.

So the recogniser set is now MEASURED. `lanes/335/r2/ruff_shapes.py` ran this phase's exact two commands under ruff 0.2.0, 0.3.0 and the pinned 0.16.1, in both modes, over five trees: clean, only-safely-fixable findings, only-unsafely-fixable findings, a mix, and a tree with no Python file in it. Every captured stdout-plus-stderr is a row in `RUFF_SHAPES` in `tests/test_verify.py`, tagged with the versions that printed it and with what `_ruff_count` must return, and a parametrized test drives the shipped parse over all fifteen. Two of them were shapes round 2 refused:

| tree | mode | ruff | printed | round 2 | now |
|---|---|---|---|---|---|
| three unsafe-fix `F841` | `--fix` | 0.2.0, 0.3.0, 0.16.1 | `Found 3 errors.` + `No fixes available (3 hidden fixes ...)`, exit 1 | `command_failed` gap | `ruff auto-fixed 0, 3 remaining` |
| clean | both | 0.2.0, 0.3.0, 0.3.2 | nothing at all, exit 0 | `command_failed` gap | row, 0 |
| clean | both | 0.3.3+, 0.16.1 | `All checks passed!`, exit 0 | row, 0 | row, 0 |
| no Python file | read-only | 0.2.0, 0.3.0 | `warning: No Python files found ...`, exit 0 | `command_failed` gap | row, 0 |
| two unused imports | read-only | all three | `Found 2 errors.` + `[*] 2 fixable ...`, exit 1 | row, 2 | row, 2 |
| two unused imports | `--fix` | all three | `Found 2 errors (2 fixed, 0 remaining).`, exit 0 | row, 2 | row, 2 |
| mix of both | read-only | all three | `Found 3 errors.` + `[*] 2 fixable ... (1 hidden fix ...)`, exit 1 | row, 2 | row, 2 |
| mix of both | `--fix` | all three | `Found 3 errors (2 fixed, 1 remaining).`, exit 1 | row, 2 | row, 2 |

The silent-clean case is the one that matters most, because ruff 0.2.0 is this phase's own floor: the version at the bottom of the supported range had every clean run reported as a tool failure. It is recognised as a SHAPE and not as a fallback, paired with `returncode == 0`, and exit 1 with nothing printed is still a gap. The refusal itself is unchanged: a JSON body and a renamed summary still return `None`, and three rows in the corpus assert exactly that, so round 1's blocker stays closed.

**B2-R2. The comment justifying the mutmut invocation stated a false fact about mutmut, and that false fact was the stated reason for the shape.** Round 2 wrote that `--paths-to-mutate=a.py b.py` reaches mutmut as one option plus one positional argument, said mutmut was not on the machine to measure it, and shipped the guess as a comment. Measured on mutmut 2.5.1 through `uvx`: `mutmut run --help` prints `Usage: mutmut run [OPTIONS] [ARGUMENT]`, one positional slot, so `mutmut run --paths-to-mutate=a.py b.py c.py --no-progress` exits non-zero with `Error: Got unexpected extra argument (c.py)` having mutated nothing, and the two-file form is accepted only because the second file fills that slot instead of being a path to mutate. Three changed non-test Python files is an ordinary diff, so the shape round 2 shipped is an outage for the common case. It is now `shlex.quote(",".join(py_files))`, which is what `--paths-to-mutate` documents and what mutmut's own `split_paths` splits on, and the comment says what was measured. The gate is opt-in (`[verify] mutation_testing`, default off), so this was never a live outage.

The `now` column is the count `_ruff_count` returns; the message it becomes is `ruff auto-fixed N[, K remaining]` in fixing mode and `ruff reports N auto-removable, not removed` in read-only.

The five should-fix, all measured:

- **`strict=True` was not pinned by any test, only the `except` clause was.** The census row stubbed `git.get_diff_names` to raise unconditionally, so it raised whatever the flag said; flipping the one keyword to `False` reverted round 2's fail-open fix with nothing red. The stub now HONOURS the flag, returning `[]` when lenient, so the mutation changes the outcome from `command_failed` to `no_target` and is caught. Two layers, both tested.
- **The gap for a ruff too old for the flag named no cause.** ruff is a clap CLI: it prints the diagnosis FIRST and `For more information, try '--help'.` last, and `_last_output_line` took the last one. Measured with a 0.1.15 shim first on PATH, round 2's gap read `ruff check exited 2: For more information, try '--help'.` `_tool_failure_line` now takes the first stderr line beginning `error:` and falls back to the last line otherwise, so the gap reads `ruff check exited 2: error: invalid value 'concise' for '--output-format <OUTPUT_FORMAT>'`. `_last_output_line`'s docstring gave "a usage message puts the sentence a reader needs at the bottom" as its reason, which is right for a Python traceback and measurably wrong for the tool this module runs most; it says so now. The **ruff 0.2.0 floor is disclosed** in `CHANGELOG.md` and in `docs/runbook.md`, where the phase is described.
- **Two `ks sense` tests skipped on ruff's presence and needed its version.** Measured by putting ruff 0.1.15 in a scratch checkout's venv: `test_sense_reports_the_dead_code_phases_separately` FAILS with `expected one 'dead_code_ruff' check, got []`, a message naming nothing about a version. The skip is now a capability probe, `ruff check --output-format=concise --help` exiting 0. Mutating that gate took three more commits, and they are the substance of this round's second half. Reverting both decorators to a presence check was STILL GREEN, because the two conditions are the same bool on any machine whose ruff is current, so a census over this module's own skip marks was added: it collects every `skipif` whose reason mentions ruff and fails on one that is not the capability gate. That census reads the mark's REASON, and a condition-only swap leaves the reason alone, so the mark's SOURCE is parsed as well and asserted to call the probe. Both plants are caught now and both were measured green first. The probe's own `shutil.which` guard was deleted in the same pass: `subprocess.run` raises `FileNotFoundError` for a missing binary, which is an `OSError` the probe already catches, so removing the guard was STILL GREEN even against a test that puts an empty directory on PATH. A branch no test can reach is not a second opinion.
- **The argv list closed shell injection but not option injection.** Measured on real vulture 2.16 over a directory holding `a.py` and `-v.py`: `vulture -v.py a.py --min-confidence 80` exits 2 with `vulture: error: unrecognized arguments: -.py`, which `check_dead_code` parses as findings and reports as `N dead code issues remaining`, a FAIL naming the wrong cause. It is now `["vulture", "--min-confidence", "80", "--", *py_files]`, which exits 0 on the same two paths. The comma form removes mutmut's positional slot entirely.
- **`_sense_needs_diff`'s docstring premise was falsified by round 2's own diff.** "Every check that consumes the diff reads it through the LENIENT git helpers" stopped being true of the dead-code phase when `_changed_non_test_python` went strict. The clause now separates the two reasons: `diff_scope` and `bad_patterns` need the preflight to avoid a silent zero, and the dead-code phase is there because an up-front exit 2 naming the base reads better than a gap in the JSON.

Of the six nits, three are applied and three recorded. **N3-R2**: the non-zero-exit guard in `check_dead_code` decided on output EMPTINESS, so a detector that exited non-zero and printed only lines the `__all__` filter drops fell through to `no remaining dead code`. It decides on the exit code now, and the gap carries the tool's last line. **N6-R2**: the discriminating half of `test_sense_never_edits_stages_or_commits` was scoped to the dead-code rows, which are empty on a machine without ruff, so it evaporated exactly where it was needed; it is asserted over `document["checks"]`, which is never empty. **N1-R2**: the census docstring claimed the refusal default's safe half as though it were the whole property, and now says the default is safe in ONE direction, naming B1-R2 as the case it got wrong. Recorded and not adopted, one line each: `_ruff_fix_message` staging and committing under a name that reads like a string is pinned by two mutations today and is a note about the next edit (**N2-R2**); `not f.startswith("test")` misses `src/tests/x.py` and is unchanged in behaviour (**N4-R2**); and the `ks feature` line is unconditional because the `DIFF_DEPENDENT_CHECKS` line above it has the identical property for `mutation_testing`, and changing one would make the two disagree (**N5-R2**).

Three claims round 2's body made are corrected here rather than left standing: `command_failed` did NOT carry ruff's own error line (it does now); "concise still prints the lines this phase reads" was true and incomplete in the way that mattered; and "assumed, not measured: nothing load-bearing" was false, because the mutmut grammar was assumed and shipped as a comment.

## Behaviour changes

- The dead-code gate reports **two rows instead of one**: `dead_code_ruff` and `dead_code`. This shows up in the Phase 1 report table, `ks sense --json`, `ks sense`'s terminal table, and the LLM reviewer's verification summary.
- **A phase that measured nothing gets no row.** Fourteen states that used to print `dead_code  pass` now print `dead_code  not measured  <reason detail>` or `dead_code_ruff  not measured  ...` and appear in `not_measured`. The run's `passed` is unchanged: a gap is a sidecar, seen by the operator and gating nothing. Installing a binary is not something the engineer's next diff can do, so a gap is not a FAIL either.
- **`dead_code`'s message loses the fused ruff note.** It was `ruff auto-fixed 2; vulture not installed`; it is now `no remaining dead code` or `N dead code issues remaining`, with the ruff count on its own row.
- **The ruff row's numbers changed meaning.** Read-only reports the auto-removable subset (`ruff reports 2 auto-removable, not removed`) instead of every finding; the fixing row reports what was left (`ruff auto-fixed 2, 1 remaining`) and says so when the commit did not land.
- **`ruff` older than 0.2.0 now produces a gap rather than a number.** Measured: ruff 0.0.272 rejects `--output-format` outright, ruff 0.1.0 and 0.1.15 have the flag but not the value `concise` (their set was `text, json, json-lines, junit, grouped, github, gitlab, pylint, azure[, sarif]`), and 0.2.0 and 0.3.0 accept it. No single value spans both eras: `text` is invalid on 0.16.1. On an older ruff the phase reports `command_failed` carrying ruff's own diagnosis, exactly `ruff check exited 2: error: invalid value 'concise' for '--output-format <OUTPUT_FORMAT>'`, which is loud rather than silent and is the direction this change is for. The floor is stated in `CHANGELOG.md` and `docs/runbook.md`.
- **Two shapes real ruff prints are rows again rather than gaps.** A fixing run whose every finding has only an unsafe fix prints `Found N errors.` alone and is now `ruff auto-fixed 0, N remaining`; a clean tree on ruff 0.2.0 through 0.3.2 prints nothing at all and is now a measured 0. Both were `command_failed` between the round-2 fix and this one, which is a regression this PR introduced and closed inside its own review cycle. The refusal is unchanged for output the parse genuinely does not know, and exit 1 with no output is still a gap.
- **`ks sense --json` `schema_version` 2 -> 3.** By v2's own stated rule: an absent `dead_code` row gains a second meaning a v2 reader would get wrong, and a new row name appears in `checks`.
- **A ruff exit outside 0 and 1 is no longer read as "auto-fixed 0"**, and **a detector that exits non-zero and reports no finding this check can read is no longer read as a clean tree.** That decision is on the EXIT CODE, not on the output being empty. Measured on vulture 2.16: exit 3 is findings, 1 is invalid input, 2 is a bad command line. The comment in the old code claiming "exit code 1 when it finds dead code" was wrong.
- **The mutation gate sends mutmut one comma-separated argument**, and **the dead-code detector's file list goes behind a `--` separator**. Both are corrections to round 2's fix, measured on mutmut 2.5.1 and vulture 2.16 respectively.
- **`ks feature` names `dead_code_ruff` among the checks it is not running**, and **`ks sense` no longer exits 2 for want of a base when no dead-code phase reads one** (`dead_code_ruff` scans `.`, and with `[verify] dead_code_command` set the detector is the operator's program run without a diff read).
- `check_dead_code`'s signature drops `read_only`; both functions return `CheckResult | NotMeasured`; `_dead_code_command` returns `str | list[str] | NotMeasured`; `_changed_non_test_python` returns `list[str] | NotMeasured`; `_commit_ruff_fixes` returns the reason the commit did not land, or `None`.

Every number below was re-measured at the final head, with `__pycache__` deleted and `PYTHONDONTWRITEBYTECODE=1` on every run, and with the venv's bin FIRST on PATH. That last point is load-bearing and is stated because it cost a wasted measurement: the ambient PATH on this machine has a homebrew ruff 0.0.272, which predates `--output-format`, and round 2's first mutation run scored every verdict against a control that was not green because of it.

Tested, by running it:

- **The round-3 items were reproduced before they were changed.** `lanes/335/r2/ruff_shapes.py` ran this phase's exact two commands under real ruff 0.2.0, 0.3.0 and 0.16.1 over five trees and recorded the raw stdout and stderr as JSON, which is where `RUFF_SHAPES` comes from. `uvx --from mutmut==2.5.1 mutmut run --help` and three invocations answered B2-R2. `uvx vulture==2.16` answered S4-R2 in both directions. A shell shim printing ruff 0.1.15's stderr answered S2-R2.
- **`lanes/335/r2_verify_head.py` re-checks the four load-bearing round-2 items against the shipped functions and exits 0.** The unsafe-only tree under real ruff 0.16.1 gives `ruff auto-fixed 0, 3 remaining` and leaves the three locals on disk; three changed files reach mutmut as `mutmut run --paths-to-mutate=a.py,b.py,c.py --no-progress`; `-v.py` comes back as `["vulture", "--min-confidence", "80", "--", "-v.py", "src/ok.py"]`, which real vulture 2.16 exits 0 on while it exits 2 with `unrecognized arguments: -.py` on the unseparated form; and the old-ruff shim's gap reads exactly `ruff check exited 2: error: invalid value 'concise' for '--output-format <OUTPUT_FORMAT>'`.
- **Full suite** on the merged tree: **5778 passed, 37 skipped, 38 xfailed, exit 0**, 386s. Run twice, once at the head before the last test commit (5777, 402s) and once at the final head, the one case being that commit's own test. Collected counts, which do not depend on how many are skipped: `origin/main` (0522501) collects **5784** and this head collects **5853**, so the branch adds **69** cases.
- **`origin/main` moved to 0522501 (#350) while round 3 was in flight, so it is merged in.** The only conflict was `CHANGELOG.md`, purely additive on both sides, resolved by keeping both entries. Every static guard was re-derived by RUNNING it before and after the merge and the two runs diffed, never by comparing a pin: the 33 test files that AST-walk `kstrl/` or `tests/` give **1094 passed, 9 skipped, 37 xfailed** before, and the 34 after the merge give **1124 passed, 9 skipped, 37 xfailed**. The whole delta is #350's own: `tests/test_inbox_write_guards.py` appears with 29 cases and `tests/test_tui_config_guard.py` goes 26 to 27. The census was run a third time at the final head: **35 files, 1149 passed, 9 skipped, 37 xfailed**. That delta IS this branch's, and it is one file: `tests/test_sense_cli.py` enters the census at 25 cases because the last commit made it parse its own source to pin a skip mark. No other guard's count moved in either direction.
- `uv run mypy kstrl/ --strict` clean (**130** files on the merged tree), `uv run ruff check kstrl/ tests/` clean, `uv run python scripts/gen_docs.py --check` clean. `cyclomatic_ratchet.py`, `file_length_ratchet.py`, `check_no_mocks.py` and `check_agent_docs_sync.py` all exit 0 over the nine changed Python files, and the full `pre-commit` hook set ran on every commit.
- **Complexity**, re-measured with complexipy 7.0.1. `check_dead_code` was cognitive **55** against a gate of 15 on `origin/main`, which is why touching it forced the partition. The whole gate now: `_ruff_count` 12, `check_dead_code` 9, `check_mutation_score` 7, `check_dead_code_ruff` 6, `_dead_code_checks` 5, `_dead_code_command` 4, `_ruff_fix_message` 4, `_changed_non_test_python` 4, `_commit_ruff_fixes` 3, `_ruff_remaining` 2, `_last_output_line` 1, `_tool_failure_line` 1, `_ruff_dead_code_command` 1, `_ruff_said_nothing` 0. `_ruff_count` grew from 9 to 12 for the two restored recognisers and is under the gate of 15. `_sense_needs_diff` is 2 and takes `sense` from 20 to 19. The only complexipy failures left in `kstrl/verify.py` are `check_test_adequacy` 18 and `check_self_critique` 25, both pre-existing and untouched.
- **File length**: `kstrl/verify.py` 2824 to **3271**, `tests/test_verify.py` 1795 to **3071**, `tests/test_sense_cli.py` 467 to **689**. All three were already over the 800-line ratchet, so growth is reported rather than failed. That is a real cost and it is stated, not hidden. `tests/test_feature_verification.py` is at 796 of 800.
- **Forty-six mutations at the final head: 45 caught, 1 still green and disclosed, 0 hung, 0 could-not-plant.** The set is round 1's, plus the round-2 reviewer's, plus one new plant per round-3 item and four for the round-3 test work. Each is planted alone, each anchor is asserted to appear exactly once before the edit and a mismatch is reported as COULD-NOT-PLANT rather than folded into a pass, `__pycache__` is deleted and `PYTHONDONTWRITEBYTECODE=1` set on every run, and each run is bounded at 420s in its own process group and killed with `os.killpg`. Control 467 passed, 4 xfailed; the slowest plant was 16s. The one still green is **R13**, which widens a negative assertion from the dead-code rows to every row (N6-R2): both forms are TRUE in both environments, because `ks sense` genuinely never auto-fixes, and what the narrow form loses is the ability to be FALSE on a machine with no ruff. Vacuity is not something a green-or-red signal can see, so it is recorded as uninformative by construction rather than counted. An earlier run of the same set, at the head before the last three test commits, was 44 caught with one COULD-NOT-PLANT: its anchor was text those commits moved, and it is superseded by two plants that both go red now. **Two plants were STILL GREEN before they produced code**, which is the whole reason the last three commits exist: reverting the two `ks sense` skip decorators to a presence check, and deleting the capability probe's `shutil.which` guard. Both are caught now, the first by a census over this module's skip marks plus a source-level assertion, the second by deleting the branch.
- **The repro script still runs on the final tree.** Nine paths plus a control: each prints exactly one row for the phase that measured and one gap for the phase that did not, no `PASS` line exists for a phase that did not run, and the control still renders `- dead_code: FAIL - 2 dead code issues remaining`.
- **The control exists.** `TestCheckDeadCode` and `TestCheckDeadCodeRuff` each carry a case that must return a row, `TestRuffShapes` asserts twelve measured shapes come back as numbers, and `TestDeadCodeRowsOnlyExistWhenMeasured` asserts a vulture finding still fails the run and that the toggle off records neither a row nor a gap, so the classes cannot be satisfied by a function that returns a gap unconditionally.
- Each not-measured test also asserts the collaborator that would have measured **never ran** (`run.call_count == 0`, no `git` in the recorded commands), so a deleted guard cannot pass by falling through to a later exit.

Not a behavioural red-first, and said plainly rather than counted as one: `tests/test_verify.py` and `tests/test_state_dir_scope.py` cannot be COLLECTED against base production code at all, because they import `check_dead_code_ruff`, which does not exist there. A collection error is not a failing assertion. The red-first evidence for those two files is the repro script and the mutation set.

Assumed, not measured, stated because round 2's version of this line was false: **nothing load-bearing**, and the two things round 2 assumed are now measured. The journal claim in the Summary is structural rather than observed, and says so. `RUFF_SHAPES` carries three rows the tool does not print under these flags (a JSON body in each mode, and a hypothetical renamed summary), which exist to pin the refusal and are labelled as such rather than presented as captures.

## Checklist

## Provenance

None. `verification_summary` is harness DATA interpolated by `REVIEWER_PROMPT.format(...)`: splitting one row into two changes the LINES the reviewer is shown, not the prompt BODY. No `*_PROMPT` constant, no `*_PROMPT_VERSION` and no snapshot tuple is touched, which is the same rule PR #334 followed when it removed a row from this summary. H2 and H3 do not apply.

## Pre-merge check

`scripts` output at the pushed head `c53a25b`, verbatim:

```
== PR #353 (fix/335-dead-code-non-measurement) ==
mergeable=MERGEABLE state=CLEAN
  test: SUCCESS
  spine: SUCCESS
  lint: SUCCESS
  coverage: SUCCESS
  GitGuardian Security Checks: SUCCESS
closes: Closes #335 
em dashes in body: 0
em dashes in diff: 0
commits: 21
commits missing Co-Authored-By: 3
em dashes in commit messages: 0
behind main by: 0
```

The three commits without a `Co-Authored-By` trailer are the three `origin/main` merge commits,
which carry no body.

## Follow-ups, measured and not fixed here

- **A vulture tool ERROR with output is still a false FAIL.** Exit 1 (invalid input, a syntax error or a missing file) and exit 2 (bad command line) both print to stderr, so the detector's error text is parsed as findings and reported as `N dead code issues remaining`. Round 3 closed the half where nothing survives the filter (that is a `command_failed` gap now, keyed on the exit code) and removed one way in (`--` before the paths), but a line that LOOKS like a finding still counts as one. The commonest way to reach it is not a broken tool: any diff that DELETES a Python file puts the deleted path on vulture's command line, and a missing path is exit 1 with `Error: <path> could not be found.` A `[verify] dead_code_command` that is not installed is the same shape, exit 127 through the shell. It needs a way to tell vulture's finding exits from its error exits, and it is left as a named follow-up.
- **The reviewer's verification summary carries no `not_measured` line.** After this change the LLM reviewer sees no dead-code line at all when a phase did not run, and cannot tell "turned off in kstrl.toml" from "asked for, measured nothing", which is the ambiguity the schema bump removes for JSON readers. #334 set that precedent for `mutation_testing`, and changing it changes what an adversarial role is shown, so it is a decision for its own PR.
- **The dead-code gate is a module's worth of code inside a 3271-line file**: fourteen module-level functions plus five module-level constants, one entry point, and `kstrl/fixtures.py` as the precedent for the lazy import that avoids the cycle. Worth extracting, in a `refactor:` PR with the mutation set re-run against the moved code.
- **`_ruff_fix_message` stages and commits the worktree** under a name that reads like a string builder, gated on a count computed for the message (N2-R2). It is pinned today by two mutations that both go red, so this is about the next edit: memoising the message or moving it below the `CheckResult` construction would stop the commit with nothing failing. One line in the caller separates them.
- **`not f.startswith("test")` is a path-prefix test** (N4-R2). It misses `src/tests/x.py`, `kstrl/test_helpers.py` and `pkg/tests/conftest.py`, and wrongly excludes a top-level `testdata/`. Pre-existing and unchanged in behaviour here; worth a `pathlib`-based rule with its own tests.
- `PipelineHooks.run_review`, `run_security_review` and `distill_facts` are still `Callable[..., ...]`-typed. PR #334 typed only `run_mechanical_verification`, so adjacent transposable `Path` arguments in the other three remain unprotected by mypy.


kstrl/verify.py
```diff
@@ -1955,6 +1980,20 @@ def _count_before(word: str, text: str) -> int:
...
+
+    One home for the rule, because two checks act on it and both report
+    ``no_target`` when it comes back empty: the mutation gate mutates
+    exactly these files and the dead-code scan scans exactly these
+    files. Widening what counts as a test file in one place and not the
+    other would make one of the two gaps say something the other does
+    not mean.
+    """
+    changed = git.get_diff_names(base_branch, cwd)
+    return [f for f in changed if f.endswith(".py") and not f.startswith("test")]
```
0xfauzi, line 2018: **`git.get_diff_names` is read leniently here, so a diff that could not be READ is recorded as "nothing to scan".**

`strict=False` is the default, and `kstrl/git.py:798` says what that means: a bad base ref, a missing `origin/<base>`, a non-repository or a timeout all map onto `[]`, which is indistinguishable from "no files changed". Both callers then treat `[]` as `no_target`:

- line 2351: `dead_code` / `the diff changed no non-test Python file, so there was nothing to scan`
- line 2072: `mutation_testing` / `...nothing to mutate`

That is the fail-open shape this PR exists to remove, one layer down. `no_target` is the one reason token that means "not a fault", and it is being used for a measurement failure. The tokens are machine-read (`ks sense --json`, `NotMeasured.as_token` into `events.jsonl`), so a consumer cannot tell the two apart, and `command_failed` already exists for exactly this.

Reachable on the factory path: `check_diff_scope` reads leniently too, so with `[verify] dead_code_cleanup = true` and a base whose remote ref does not resolve (or a git call that times out under a parallel run), Phase 1 prints `dead_code  not measured  the diff changed no non-test Python file` and the run continues green. `ks sense` is protected only because `cli.py:3841` does a separate strict read first.

This helper is new in this PR and its docstring calls itself the one home for the rule, which makes it the place to fix both callers at once: read with `strict=True` and let each caller turn `git.GitDiffError` into `NOT_MEASURED_COMMAND_FAILED`, or return `list[str] | None` with `None` meaning could-not-read.

kstrl/verify.py
```diff
@@ -2120,195 +2167,315 @@ def _no_counts(result: subprocess.CompletedProcess[str]) -> NotMeasured:
...
+    all, and a check they turned on that measured nothing records why.
+    One toggle, ``[verify] dead_code_cleanup``, still owns both phases -
+    splitting the ROW is not splitting the switch.
+
+    Order is load-bearing and is why this is a function rather than two
+    calls inline: ruff runs FIRST so the detector scans a tree with the
+    ruff-fixable subset already deleted. Reversing it changes what
+    vulture reports.
+    """
+    if not config.dead_code_cleanup:
```
0xfauzi, line 2735: **The split makes `dead_code_ruff` disappear from `ks feature` with no row, no gap, and no line in the announcement.**

`ks feature` reaches Phase 1 through `run_undiffed_verification` -> `narrow_to_undiffed`, which sets `dead_code_cleanup=False`. This line then returns `([], [])`. Meanwhile `_announce_verification` (`kstrl/feature_verify.py:191`) prints only `", ".join(DIFF_DEPENDENT_CHECKS)`, and this PR deliberately keeps `DEAD_CODE_RUFF_CHECK` out of that tuple.

So for an operator who set `dead_code_cleanup = true`: `dead_code` is suppressed and named, `dead_code_ruff` is suppressed and named nowhere. Not in `checks`, not in `not_measured`, not in the announcement, not in the `verification_result` event.

Before this PR the fused row was in `DIFF_DEPENDENT_CHECKS` and was announced, so nothing was silently suppressed. This is a new silent non-measurement introduced by the split, and it is the same class the PR is closing one file over. The runbook sentence records it, but the runbook is not the report an operator is reading at the time.

Two ways out, and the second is the one the PR's own reasoning argues for: give the announcement a second tuple for "suppressed for a reason other than the diff" and put `DEAD_CODE_RUFF_CHECK` in it; or leave `dead_code_cleanup` alone in `narrow_to_undiffed` and gate only the vulture phase, since ruff scans `.` and can answer honestly with no base.

kstrl/verify.py
```diff
@@ -2120,195 +2167,315 @@ def _no_counts(result: subprocess.CompletedProcess[str]) -> NotMeasured:
...
+    """How many findings ruff removed, or would remove.
+
+    Two different numbers off two different lines, because the two modes
+    print different things: a fixing run reports ``Found 3 errors (2
+    fixed, 1 remaining).`` and a ``--no-fix`` run reports ``Found 3
+    errors.`` with nothing removed. Last match wins in the fixing case,
+    which is what the fused function did.
+    """
+    if read_only:
+        found = re.search(r"Found (\d+) error", output)
```
0xfauzi: **Wrong line. The read-only row overstates what the factory would remove, by exactly the unsafe-fix count.**

Measured just now on ruff 0.16.1, one `F401` plus one `F841` whose right-hand side is a call (ruff marks that fix unsafe):

```
--no-fix:  Found 2 errors.
           [*] 1 fixable with the `--fix` option (1 hidden fix can be enabled with the `--unsafe-fixes` option).
--fix:     Found 2 errors (1 fixed, 1 remaining).      exit 1, one file changed
```

`Found (\d+) error` returns **2**, so `ks sense` prints `ruff reports 2 auto-removable, not removed` while the factory path removes **1**. `check_dead_code_ruff`'s docstring states that read-only "reports what the factory WOULD have removed"; measured, it does not.

The number that answers the question is on the `[*]` line, and the two modes should be reading counts that mean the same thing. Older ruff prints `[*] 2 potentially fixable with the --fix option.`, so one pattern covers both:

```suggestion
        found = re.search(r"\[\*\] (\d+) (?:potentially )?fixable", output)
```

kstrl/verify.py
```diff
@@ -2120,195 +2167,315 @@ def _no_counts(result: subprocess.CompletedProcess[str]) -> NotMeasured:
...
+    A list, not a string: ``run_scrubbed`` only shells out for a string,
+    and the ``:(exclude)`` pathspec must reach git unmangled.
+
+    A timeout here is swallowed rather than reported as a gap: the
+    measurement already happened, and the row it produced is true. What
+    is lost is the tidying, which the next check would notice for
+    itself.
+    """
+    try:
+        run_scrubbed(
```
0xfauzi: **Neither `git add` nor `git commit` has its exit status read, so the row can say `ruff auto-fixed 2` over a worktree where nothing was committed.**

`run_scrubbed` returns a `CompletedProcess` and never raises on a non-zero exit (line 201). Both calls here discard it. A commit rejected by a repo hook, or aborted because the worktree has no `user.email`, leaves the deletions on disk and uncommitted while `check_dead_code_ruff` returns `CheckResult(passed=True, "ruff auto-fixed 2")`.

The `except subprocess.TimeoutExpired: pass` adds a second path: a timeout on `git add` skips the `git commit` entirely. The docstring's justification, that "what is lost is the tidying, which the next check would notice for itself", does not hold here - `check_diff_scope`, `check_bad_patterns` and `check_test_adequacy` all ran BEFORE this point in `run_mechanical_verification` (lines 2913-2948), and only `mutation_testing` and `self_critique` run after.

This is the shape of the bug the PR is fixing, one step further in: a step that did not happen, reported as done. Reading `returncode` on both and either amending the message or returning `NotMeasured(DEAD_CODE_RUFF_CHECK, NOT_MEASURED_COMMAND_FAILED, ...)` when the commit did not land keeps the row honest with the vocabulary already in the file.

kstrl/verify.py
```diff
@@ -2120,195 +2167,315 @@ def _no_counts(result: subprocess.CompletedProcess[str]) -> NotMeasured:
...
+            "so nothing scanned for dead code",
+        )
+    py_files = _changed_non_test_python(base_branch, cwd)
+    if not py_files:
+        return NotMeasured(
+            DEAD_CODE_CHECK,
+            NOT_MEASURED_NO_TARGET,
+            "the diff changed no non-test Python file, so there was nothing to scan",
         )
+    return f"vulture {' '.join(py_files)} --min-confidence 80"
```
0xfauzi: **Diff-derived file names are interpolated unquoted into a string that runs through a shell.**

`run_scrubbed` passes `shell=isinstance(cmd, str)` (line 162), so this string is handed to `/bin/sh`. `py_files` comes straight out of `git diff --name-only`. The same pattern is at line 2076: `f"mutmut run --paths-to-mutate={paths_arg} --no-progress"`.

Two failures, one noisy and one silent:

- A changed file named `my file.py` splits into two arguments, vulture prints a not-found error to stderr, and the branch at line 2416 turns that error text into `1 dead code issues remaining` and FAILS the component.
- A file named `a;id;.py`, `$(cmd).py` or `` `cmd`.py `` executes. In a factory these names come from an agent-authored diff, which is the least trusted input in the system.

`_changed_non_test_python` is new in this PR and is now the single home for these paths, which is what makes this fixable in one place for both callers: `shlex.quote` each entry, or better, build an argv list. `run_scrubbed` already accepts one and `_commit_ruff_fixes` relies on that exact property for its `:(exclude)` pathspec.

kstrl/verify.py
```diff
@@ -2120,195 +2167,315 @@ def _no_counts(result: subprocess.CompletedProcess[str]) -> NotMeasured:
...
         )
 
     output = (result.stdout + result.stderr).strip()
-    if result.returncode != 0 and output:
-        # vulture returns exit code 1 when it finds dead code
-        lines = output.splitlines()
-        # Filter out common false positives (e.g., __all__, __init__)
+    if result.returncode != 0:
+        if not output:
+            return NotMeasured(
```
0xfauzi: **The new `command_failed` guard only catches a silent crash. A detector that errors *with* output is still reported as findings, and still FAILS the component.**

`if not output` is the whole guard. Anything that exits non-zero and prints something falls through to line 2416, where its stderr becomes `real_issues`, and the row is `N dead code issues remaining` with `passed=False`.

By the docstring's own measurements, that is most of vulture's error surface: exit 1 is invalid input (a syntax error in a scanned file) and exit 2 is a bad command line, and both print to stderr. A `[verify] dead_code_command` that is not installed is worse - the shell exits 127 with `sh: 1: foo: not found`, which becomes exactly one "dead code issue" and fails a component on every retry.

The PR body names this as a follow-up, and it is genuinely a different direction of error. Flagging it because it is the other half of this one row: the PR fixes "a scan that did not run, reported as a pass" and leaves "a scan that could not run, reported as a FAIL", which spends engineer iterations on a diff that cannot change the outcome. For the built-in path the tool already separates them - vulture exit 3 is findings - so `result.returncode != 3` on the vulture branch is enough, and a custom command can keep the current lenient reading.

kstrl/verify.py
```diff
@@ -2090,6 +2126,21 @@ def check_mutation_score(
...
+
+    stderr first because that is where a tool that could not start says
+    so, and the last line because a traceback or a usage message puts
+    the sentence a reader needs at the bottom.
+
+    Capped for the reason git.py caps its stderr at 500: this reaches
+    ``ks sense --json`` and the terminal, and one unbroken line of tool
+    output has no bound.
+    """
+    tail = (result.stderr or result.stdout).strip().splitlines()
```
0xfauzi: **`result.stderr or result.stdout` picks stderr whenever it is truthy, including when it holds only whitespace.**

A tool that writes its error to stdout and a bare newline to stderr gives `tail = "\n".strip().splitlines() == []`, and the function returns `"no output"` while the sentence the reader needs is sitting in stdout.

That reaches both callers as the whole content of the gap: `ruff check exited 2: no output` (line 2301) and `mutmut run exited 3 and reported no counts: no output` (line 2159). The gap exists to carry that diagnostic; this is the one line that can throw it away.

```suggestion
    tail = (result.stderr.strip() or result.stdout.strip()).splitlines()
```

Carried over from the code this helper extracts, so it is pre-existing - but the extraction is the moment it becomes one line to fix rather than two.

kstrl/verify.py
```diff
@@ -2120,195 +2167,315 @@ def _no_counts(result: subprocess.CompletedProcess[str]) -> NotMeasured:
...
+            DEAD_CODE_RUFF_CHECK,
+            NOT_MEASURED_COMMAND_FAILED,
+            f"ruff check exited {result.returncode}: {_last_output_line(result)}",
+        )
+
+    count = _ruff_count(result.stdout + result.stderr, read_only=read_only)
+    if read_only:
+        message = f"ruff reports {count} auto-removable, not removed"
+    else:
+        message = f"ruff auto-fixed {count}"
```
0xfauzi: **The fixing branch drops the count of what ruff found and could not remove.**

Measured on ruff 0.16.1: `ruff check --fix --select F401,F811,F841 .` on a tree with one `F401` and one unsafe-fix `F841` exits **1** and prints `Found 2 errors (1 fixed, 1 remaining).` This row renders as `dead_code_ruff  pass  ruff auto-fixed 1`, and the finding ruff located but declined to remove is not mentioned anywhere.

`passed=True` is right, and the runbook's reasoning for it is right: auto-removing an import is not a verdict. The problem is narrower - an operator reading `ruff auto-fixed 0` cannot tell a clean tree from a tree with three unfixable `F841`s, and the read-only branch two lines up avoids exactly that by reporting a number.

`Found N errors (M fixed, K remaining)` already carries K on the same line the regex is scanning. `ruff auto-fixed 1, 1 remaining` costs one more capture group and makes the row report what it measured rather than only what it changed.

kstrl/cli.py
```diff
@@ -3751,13 +3757,18 @@ def sense(
     policy / adequacy / dead-code / mutation checks from kstrl.toml),
     run by hand with no PRD, no branch, no worktree and no agent spend.
 
+    [verify] dead_code_cleanup produces two rows, `dead_code_ruff` for
```
0xfauzi, line 3810: **`needs_diff` is now too coarse for the gate this PR split.**

Line 3831 still has `or verify_cfg.dead_code_cleanup`, so `ks sense` exits 2 on an unreadable base whenever the dead-code toggle is on. After the split that condition is wrong in two directions, and the PR's own text argues both:

- `dead_code_ruff` scans `.` and, in this PR's words, "has an honest answer with no base to diff against". A run that could have produced `dead_code_ruff  pass` plus `dead_code  not measured` now produces nothing at all and exits 2.
- With `[verify] dead_code_command` set, `_dead_code_command` returns the operator's command before touching the diff, so NO dead-code phase reads a diff, and the preflight still demands one.

The toggle no longer maps one-to-one onto a diff-reading check. `verify_cfg.dead_code_cleanup and not verify_cfg.dead_code_command` is the condition that matches what the code does now.

tests/test_sense_cli.py
```diff
@@ -294,12 +315,65 @@ def test_sense_never_edits_stages_or_commits(tmp_path: Path) -> None:
...
-    # left open on purpose: whether vulture is installed in the running
-    # environment decides that, and this test is about the tree, not the
-    # verdict.
-    dead_code = _check(document, "dead_code")
-    assert "not removed" in dead_code["message"] or "Skipped" in dead_code["message"]
+    # What the two dead-code phases REPORTED is asserted in
+    # test_sense_reports_the_dead_code_phases_separately below, which
+    # needs ruff on PATH to have a measurement to read. This test is
+    # about the tree, so it stays unconditional.
+    assert document["checks"]
```
0xfauzi: **This replaces the assertion that proved `ks sense` reported rather than removed, and the replacement is skip-gated.**

The line removed here was `assert "not removed" in dead_code["message"] or "Skipped" in dead_code["message"]` - the check that the read-only mode did the read-only thing. `assert document["checks"]` is true of any run that produced one row at all, and it would still pass if both dead-code phases were deleted.

The real assertions moved into the two tests below, both carrying `@pytest.mark.skipif(shutil.which("ruff") is None)`. So on a machine without ruff, which is precisely the environment the not-measured half of this PR is about, nothing in this file asserts anything about either dead-code phase. The tree-level assertions above survive, but they are about the tree, and they were already passing before the fix.

An unconditional assertion costs one line: that at least one of `dead_code` / `dead_code_ruff` appears as a row or a gap, and that no row named `dead_code_ruff` carries a fixing message.

tests/test_verify.py
```diff
@@ -805,9 +810,27 @@ def test_no_iteration_heading_falls_back_to_whole_file(
...
+def _completed(
+    args: object = "cmd",
+    returncode: int = 0,
+    stdout: str = "",
+    stderr: str = "",
+) -> CompletedProcess[str]:
+    return CompletedProcess(args=args, returncode=returncode, stdout=stdout, stderr=stderr)
+
+
+def _which_only(*present: str) -> Callable[[str], str | None]:
```
0xfauzi, line 825: **The stub is narrower than the function it replaces, and the patch is global.**

`shutil.which` is `which(cmd, mode=os.F_OK | os.X_OK, path=None)`. This stub takes one positional parameter, and `patch("shutil.which", side_effect=_which_only(...))` is in force for everything running inside the block, including all of `run_mechanical_verification`. A call that passes `mode=` or `path=` raises `TypeError` out of the stub rather than being answered by it.

`_without_vulture` in `tests/test_sense_cli.py:287` takes `*args, **kwargs` for exactly this reason. Matching it here means a future check that resolves a tool with a `path=` argument fails on its own merits instead of on the harness.

kstrl/verify.py
```diff
@@ -2120,195 +2167,315 @@ def _no_counts(result: subprocess.CompletedProcess[str]) -> NotMeasured:
...
+    calls inline: ruff runs FIRST so the detector scans a tree with the
+    ruff-fixable subset already deleted. Reversing it changes what
+    vulture reports.
+    """
+    if not config.dead_code_cleanup:
+        return [], []
+    ruff_outcome = check_dead_code_ruff(
+        cwd,
+        config.subprocess_timeout,
+        read_only=read_only,
```
0xfauzi, line 2740: **Altitude, not a bug: the dead-code gate is now a module's worth of code inside a 2995-line file.**

This PR takes `kstrl/verify.py` from 2824 to 2995 lines, and the gate it just split is seven module-level names - `DEAD_CODE_CHECK`, `DEAD_CODE_RUFF_CHECK`, `_ruff_dead_code_command`, `_ruff_count`, `_commit_ruff_fixes`, `check_dead_code_ruff`, `_dead_code_command`, `check_dead_code`, `_dead_code_checks` - with one entry point into `run_mechanical_verification` and no other caller outside tests.

That is the same shape as `kstrl/fixtures.py`, which already lives in its own module and is imported lazily at the call site to avoid the `CheckResult` / `run_scrubbed` cycle (line 2977). The precedent and the import pattern both exist.

The ratchet reports rather than fails here because the file was already oversize, which the PR body states honestly. Raising it because the split is the moment the extraction is nearly free: the seven names move together, the only import to add is the lazy one that `check_fixtures_from_prd` already models, and the alternative is that the next dead-code change argues about a 3100-line file instead.

kstrl/verify.py
```diff
@@ -2120,195 +2209,432 @@ def _no_counts(result: subprocess.CompletedProcess[str]) -> NotMeasured:
...
+        fixable = _RUFF_FIXABLE.search(output)
+        if fixable:
+            return int(fixable.group(1))
+        if _RUFF_FOUND.search(output):
+            return 0
+        return 0 if _RUFF_CLEAN in output else None
+    fixed = _RUFF_FIXED.findall(output)
+    if fixed:
+        return int(fixed[-1][0])
+    return 0 if _RUFF_CLEAN in output else None
```
0xfauzi, line 2373: **A `--fix` run that found errors and could fix none of them is reported as `command_failed`, not as `ruff auto-fixed 0`.**

The read-only branch three lines up has a `_RUFF_FOUND` fallback; the fixing branch has none. ruff omits the `(N fixed, M remaining)` parenthetical entirely when it fixed nothing, so the ordinary shape `Found 1 error.` matches neither `_RUFF_FIXED` nor `_RUFF_CLEAN` and this returns `None`.

Measured on ruff 0.16.1 with this PR's exact fixing command, on a tree whose only finding is one unsafe-fix `F841`:

```
mod.py:6:5: F841 Local variable `x` is assigned to but never used
Found 1 error.
No fixes available (1 hidden fix can be enabled with the `--unsafe-fixes` option).
exit=1
```

Driving `check_dead_code_ruff` over that repo on this head:

```
NotMeasured(check='dead_code_ruff', reason='command_failed',
  detail='ruff check exited 1 and printed no count line: No fixes available (1 hidden fix can be enabled with the `--unsafe-fixes` option).')
```

ruff ran, measured, and printed a count line. F811 (no fix at all) gives the same result with `Found 1 error.` as the detail, so the gap's own text contradicts itself. Any component whose diff leaves an unfixable F401/F811/F841 -- an unused local assigned from a call is the common case -- now gets a permanent bogus `dead_code_ruff` gap instead of `ruff auto-fixed 0, 1 remaining`.

The test suite cannot see it: `RUFF_FIXED_2` is the mixed case that *does* print the parenthetical, and `test_findings_none_of_them_fixable_report_zero_removable` covers only `read_only=True`. The fixing-mode twin of that test is the missing one.

```suggestion
    return 0 if (_RUFF_FOUND.search(output) or _RUFF_CLEAN in output) else None
```

kstrl/verify.py
```diff
@@ -2027,17 +2088,28 @@ def check_mutation_score(
...
+    # /bin/sh, and the names come out of an agent-authored diff: without
+    # the quoting a file called `$(id).py` executes (#335 round 2).
+    # Quoted rather than converted to an argv list because the shell's
+    # word splitting is load-bearing here: `--paths-to-mutate=a.py b.py`
+    # reaches mutmut as one option plus one positional argument, and an
+    # argv list would send it a single option holding both paths. That
+    # is a different invocation, and mutmut is not on this machine to
+    # measure the difference, so the quoting closes the hole without
+    # changing what mutmut receives for any ordinary name.
+    paths_arg = " ".join(shlex.quote(f) for f in py_files)
```
0xfauzi: **`--paths-to-mutate=a.py b.py c.py` is not "one option plus positional arguments" that mutmut accepts -- three or more changed files is a hard usage error, and two mutates the wrong thing.**

The comment above reasons that the shell's word splitting is load-bearing here. Measured on mutmut 2.5.1:

- `mutmut run` is declared `@click.argument('argument', nargs=1, required=False)`. There is exactly ONE positional slot, and `parse_run_argument` treats it as a mutation id or *a single file to mutate*, not as an extra path.
- `--paths-to-mutate` is a single TEXT value that `split_paths` splits on **commas or colons** -- never on spaces.

So with three non-test Python files in the diff:

```
$ mutmut run --paths-to-mutate=a.py b.py c.py --no-progress
Usage: mutmut run [OPTIONS] [ARGUMENT]
Error: Got unexpected extra argument (c.py)
```

mutmut never starts. `check_mutation_score` then reads no counts and records a `command_failed` gap for every component whose diff touches 3+ non-test Python files, which is most of them. With exactly two files, `paths_to_mutate` is `["a.py"]` and `b.py` becomes the single ARGUMENT, so the gate mutates one file and silently ignores the other.

This predates the PR, but the hunk rewrites this line and states the grouping is deliberate, so it is worth correcting here rather than preserving. The comma form is what `split_paths` is for and it also removes the reason the argv list was rejected:

```suggestion
    paths_arg = shlex.quote(",".join(py_files))
```

kstrl/verify.py
```diff
@@ -2120,195 +2209,432 @@ def _no_counts(result: subprocess.CompletedProcess[str]) -> NotMeasured:
...
+            f"ruff check exited {result.returncode} and printed no count line: "
+            f"{_last_output_line(result)}",
+        )
+    if read_only:
+        message = f"ruff reports {count} auto-removable, not removed"
+    else:
+        message = _ruff_fix_message(cwd, output, count)
+    return CheckResult(
+        name=DEAD_CODE_RUFF_CHECK,
+        passed=True,
```
0xfauzi, line 2565: **A phase whose only durable effect failed still returns `passed=True`, and the PR's own thesis says a row's verdict must answer for what actually happened.**

When `_commit_ruff_fixes` comes back with a reason, ruff has already rewritten files on disk and nothing committed them. `check_diff_scope`, `check_bad_patterns` and `check_test_adequacy` have all run above this point, and nothing later in the pipeline commits a worktree -- only the engineer agent does -- so those deletions never reach the branch or the PR. The verdict column still reads PASS, and that is the column `review.build_review_prompt` renders and an operator scans first.

The message does carry `not committed: <reason>`, which is a real improvement over round 1's silence. But the same argument the PR makes for `dead_code` ("a phase that did not run must not report a pass") applies one step over: a phase that ran and whose result was discarded is not a pass either. Either return a `NotMeasured(dead_code_ruff, command_failed, ...)` when the commit did not land -- the measurement exists but its effect does not -- or say explicitly in the docstring why a lost auto-fix is a PASS.

kstrl/feature_verify.py
```diff
@@ -184,12 +184,34 @@ def _announce_verification(
...
+    would be a branch that can never run - the shape of defect this
+    module keeps finding rather than one to add.
     """
     ui.section(f"Verification report ({phase})")
     ui.info(
         "Report only, not a gate: the exit code is unchanged. Not measured "
         "here (no diff to read, see docs/runbook.md): " + ", ".join(DIFF_DEPENDENT_CHECKS)
     )
+    ui.info(
+        "  also not measured: dead_code_ruff. It reads no diff, but one toggle "
```
0xfauzi, line 211: **The line `ks feature` now prints is unconditional, and `dead_code_cleanup` defaults to False -- so for the default configuration the report gives a false reason for a phase the operator never enabled.**

The docstring justifies being unconditional by saying the config has already been narrowed, so a condition on the toggle "would be a branch that can never run". That is true of the narrowed value but it is not the question. The bit worth reading is the operator's PRE-narrowing setting, which `_announce_verification` does not receive -- so "this flow turns it off" is asserted for everybody, including the majority who left `dead_code_cleanup` at its default of `False` and for whom nothing was turned off by this flow at all.

That is the same defect this function's own docstring names two paragraphs up: "Naming it in the same list would have printed a false reason." `test_the_narration_names_the_dead_code_phase_that_reads_no_diff` writes a `kstrl.toml` with no `dead_code_cleanup` key at all, so the new test pins the false statement rather than catching it.

The pre-narrowing config is available in `resolve_feature_verify_config` (it is the input to `narrow_to_undiffed`); passing that one bit down, or passing the un-narrowed config alongside, lets the line say the true thing and not print at all when the operator never asked.

kstrl/verify.py
```diff
@@ -2120,195 +2209,432 @@ def _no_counts(result: subprocess.CompletedProcess[str]) -> NotMeasured:
...
+    py_files = _changed_non_test_python(base_branch, cwd, DEAD_CODE_CHECK)
+    if isinstance(py_files, NotMeasured):
+        return py_files
+    if not py_files:
+        return NotMeasured(
+            DEAD_CODE_CHECK,
+            NOT_MEASURED_NO_TARGET,
+            "the diff changed no non-test Python file, so there was nothing to scan",
+        )
+    return ["vulture", *py_files, "--min-confidence", "80"]
```
0xfauzi: **The argv list closes injection but not option injection: a diff-authored path beginning with `-` is still parsed by vulture as a flag.**

The hunk's reasoning is that these names are the least trusted input in the factory. `shlex.quote` and the argv list both leave a leading hyphen untouched -- `shlex.quote("-rf.py")` returns `-rf.py` -- so a changed file called `--exclude.py` or `-v.py` reaches vulture's argument parser as an option rather than a path, and the same holds for the quoted mutmut string.

vulture parses with `argparse`, which honours an explicit end-of-options marker (measured: `parse_args(['--min-confidence','80','--','-rf.py','a.py'])` gives `paths=['-rf.py','a.py']`). Putting the options first and the paths after `--` closes it. I could not verify against vulture itself -- it is not installed on this machine -- so the ordering wants one run before landing.

```suggestion
    return ["vulture", "--min-confidence", "80", "--", *py_files]
```

kstrl/verify.py
```diff
@@ -2120,195 +2209,432 @@ def _no_counts(result: subprocess.CompletedProcess[str]) -> NotMeasured:
...
+            timeout=30,
+        )
+        if committed.returncode != 0:
+            return f"git commit exited {committed.returncode}: {_last_output_line(committed)}"
+    except subprocess.TimeoutExpired:
+        return "git add or git commit exceeded 30s"
+    return None
+
+
+def _ruff_fix_message(cwd: Path, output: str, count: int) -> str:
```
0xfauzi, line 2448: **A function named `_ruff_fix_message` stages and commits the worktree.**

The git commit is a side effect of building a string, and the only thing that decides whether it happens is `count > 0` computed for the message. `check_dead_code_ruff` reads as `message = _ruff_fix_message(...)`, so nothing at the call site says the tree is about to move. The failure mode is ordinary maintenance: memoise the message, compute it for a log line, build it before deciding to return a row, or reorder it below the `CheckResult(...)` construction, and the auto-commit silently stops happening with every test still green -- exactly the "guard that goes away with nothing failing" shape CLAUDE.md tracks.

Splitting the two apart costs one line in the caller and removes the coupling:

```python
if read_only:
    message = f"ruff reports {count} auto-removable, not removed"
else:
    problem = _commit_ruff_fixes(cwd) if count > 0 else None
    message = _ruff_fix_message(output, count, problem)
```

The docstring's stated reason for fusing them -- "so the commit and the sentence describing the commit cannot drift apart" -- is preserved by passing `problem` in, since the sentence is then derived from the commit's own result.

kstrl/verify.py
```diff
@@ -1955,6 +1981,43 @@ def _count_before(word: str, text: str) -> int:
...
+    """
+    try:
+        changed = git.get_diff_names(base_branch, cwd, strict=True)
+    except git.GitDiffError as exc:
+        return NotMeasured(
+            check,
+            NOT_MEASURED_COMMAND_FAILED,
+            f"git could not read the diff against {base_branch}: {exc}",
+        )
+    return [f for f in changed if f.endswith(".py") and not f.startswith("test")]
```
0xfauzi, line 2018: **"non-test" here means "the path does not start with the five characters `test`", which misses every test file that is not at the top level.**

The docstring promotes this to "one home for the rule" and warns about the two callers disagreeing, but the rule itself is `f.startswith("test")` against a repo-relative path. It matches `tests/x.py` and `test_x.py` at the root, and misses `src/tests/x.py`, `kstrl/test_helpers.py`, `pkg/tests/conftest.py` -- and it wrongly matches an unrelated top-level directory called `testdata/` or `testing/`.

The consequences differ per caller and both are real: vulture scans test files and reports fixtures and helpers as dead code, which FAILS the component; mutmut mutates test files, which produces survivors nobody can kill. Pre-existing, but the extraction is where the rule now lives and the docstring is what a reader will trust.

A basename-and-segment test is the smaller change than a path prefix:

```python
def _is_test_path(path: str) -> bool:
    parts = PurePosixPath(path).parts
    return any(p in ("tests", "test") for p in parts[:-1]) or parts[-1].startswith("test_")
```

kstrl/cli.py
```diff
@@ -3684,6 +3692,47 @@ def _sense_error(message: str, as_json: bool) -> NoReturn:
...
 
 
+def _sense_needs_diff(
+    verify_cfg: VerifyConfig,
+    policy_cfg: PolicyConfig,
+    adequacy_cfg: AdequacyConfig,
+) -> bool:
+    """Whether a check `ks sense` is about to run reads ``git diff``.
+
+    Every check that consumes the diff reads it through the LENIENT git
```
0xfauzi: **This docstring's premise stopped being true in this same PR.**

"Every check that consumes the diff reads it through the LENIENT git helpers, which map a bad ref, a missing base or a non-repository onto an EMPTY file list" -- that is exactly what `_changed_non_test_python` no longer does. It now calls `get_diff_names(..., strict=True)` and turns a `GitDiffError` into a `command_failed` gap, which is the fix one file over. So for the dead-code phase the preflight is no longer preventing a vacuous pass; the check would have said so itself.

That matters for more than accuracy: the stated reason is what a later reader will use to decide whether a NEW clause belongs in this predicate. `diff_scope` and `bad_patterns` genuinely still read leniently and still need the preflight; the dead-code clause is now there for a different reason (a gap is worse operator experience than an up-front exit 2, which is a defensible but different argument). Worth saying which is which, otherwise the next person migrates another check to strict and deletes its clause here as redundant, or leaves a stale one behind.

tests/test_sense_cli.py
```diff
@@ -294,12 +315,73 @@ def test_sense_never_edits_stages_or_commits(tmp_path: Path) -> None:
...
+    # keeps unconditionally is the read-only contract itself, stated
+    # about the document rather than only about the tree: `assert
+    # document["checks"]` alone is true of any run that produced one row
+    # at all, and would still pass with both dead-code phases deleted -
+    # on a machine without ruff, that left nothing in this file
+    # asserting anything about either of them.
+    dead_code_rows = [c for c in document["checks"] if c["name"].startswith("dead_code")]
+    gaps = [g for g in document["not_measured"] if g["check"].startswith("dead_code")]
+    assert dead_code_rows or gaps
+    assert not any("auto-fixed" in row["message"] for row in dead_code_rows)
```
0xfauzi: **The discriminating half of this assertion is vacuous in exactly the case the comment above says it was written for.**

`assert not any("auto-fixed" in row["message"] for row in dead_code_rows)` is trivially True when `dead_code_rows == []`, and `dead_code_rows` is empty precisely on the machine without ruff that the comment names. What survives there is `assert dead_code_rows or gaps` -- one row OR one gap exists -- which both dead-code phases returning an unconditional `NotMeasured` would satisfy.

It is stronger than what it replaced, so this is not a regression. But the comment claims it "keeps unconditionally... the read-only contract itself, stated about the document", and on a no-ruff machine it does not: it keeps a presence check. Asserting the negative on the whole document instead makes it hold in both environments:

```suggestion
    assert dead_code_rows or gaps
    assert not any(
        "auto-fixed" in c["message"] for c in document["checks"]
    ), document["checks"]
```

kstrl/verify.py
```diff
@@ -2120,195 +2209,432 @@ def _no_counts(result: subprocess.CompletedProcess[str]) -> NotMeasured:
...
-    msg_parts: list[str] = []
-    if ruff_touched:
-        msg_parts.append(ruff_note)
-    msg_parts.append("no remaining dead code")
     return CheckResult(
-        name="dead_code",
+        name=DEAD_CODE_CHECK,
         passed=True,
-        message="; ".join(msg_parts),
+        message="no remaining dead code",
```
0xfauzi, line 2710: **The new non-zero-exit guard covers only *empty* output; output that the false-positive filter empties still falls through to `no remaining dead code`.**

The docstring says a silent non-zero exit is now a `command_failed` gap because "exit 3 is findings, 1 is invalid input and 2 is a bad command line". That reasoning does not stop at empty output: a detector that exited non-zero and printed only lines this filter drops -- anything containing `__all__`, anything starting with `#` -- reaches line 2591 and reports a clean tree over a failed run. The listed exit codes are still saying the tool failed.

It is reachable through the follow-up the PR body already names: a diff that DELETES a Python file puts the deleted path on vulture's command line, vulture exits 1 with `Error: <path> could not be found.`, and if that path happens to contain `__all__` (a file named `__all__.py`, or a `dead_code_command` whose usage text mentions it) the whole output filters away and the row reports a pass.

The narrow fix is to decide on the exit code first and use the filter only to *describe* findings, not to decide whether there were any:

```python
if result.returncode != 0 and not real_issues:
    return NotMeasured(
        DEAD_CODE_CHECK,
        NOT_MEASURED_COMMAND_FAILED,
        f"the dead code scan exited {result.returncode} with nothing this could read: "
        f"{_last_output_line(result)}",
    )
```

Issue #231, opened 2026-08-25
R10.10: Polled steering: /memory and /iterate comments on kstrl PRs reach the next run


0xfauzi:
> **Stabilisation queue: P3, rank 13 of 21** (set 2026-09-09; the full ordered list and the coordinator's rules are in #369). First feature by value: a reviewer's comment reaches no run today. On hold until the owner lifts the P3 hold in #369. Two PRs; #228 and #230 are merged.

Part of #235 (R10: Control Loop tracking issue). Item R10.10.

Design section: 5.8 (and gap 3.7). Sizing: L (two PRs, listed below; each independently mergeable). Dependencies: **R10.7** (`PR_FOOTER_MARKER`, `count_open_kstrl_prs`) and **R10.9** (`scripts/kstrl/memory.md`). Related: #188 (actor allowlist); this issue uses the same author check until #188 lands and then adopts its allowlist.

## Why

A human reviewing a pull request produces the highest-quality signal in the whole system, and it currently reaches the next run by no path at all. PR bodies are write-only outputs, Linear is a one-way outbound sink, and the only inbound channels are the local CLI and the GitHub label trigger.

The steering channel: a maintainer comments on the PR; the loop that created it picks the comment up, writes durable guidance into the memory file, and acts on it. This is how the operator tunes the controller over time.

kstrl cannot use a webhook for this. Inbound HTTP is an explicit non-goal (`docs/dark-factory-roadmap.md:1383-1385`), and that decision is sound for a single-operator tool with no server. But `ks serve` already polls GitHub every 60 seconds (`ServeConfig.poll_interval_seconds`, `kstrl/serve.py:239`) and already reads issue timelines (`kstrl/intake_github.py:572`, `verify_authorization`). Polling is not a webhook. The non-goal survives.

Two commands, both left as comments on a kstrl-authored PR by an authorised human:

- `/memory <text>`: append `<text>` to `scripts/kstrl/memory.md` as standing feedback. This is the durable channel.
- `/iterate <text>`: do `/memory <text>`, then re-queue the work item that produced this PR so the factory runs again with the new memory in context.

## PR 1 of 2: queue items remember the PR they produced

Today a queue item records no PR. `grep -n "pr_url\|pr_number" kstrl/workqueue.py kstrl/serve.py` returns nothing. Without this, `/iterate` cannot find what to re-run.

[...]

PR #354, opened 2026-09-06
feat(serve): bound the pull requests the daemon leaves open (R10.7, #228)

### LINKED ISSUE #228: R10.7: Flow control: `ks serve` stops admitting work while max_open_prs kstrl PRs are open
Part of #235 (R10: Control Loop tracking issue). Item R10.7.

Design section: 5.6 (and gap 3.5). Sizing: S-M (one PR). Dependencies: none. Applies to the `ks serve` daemon only.

## Why

Nothing bounds pull requests in flight. `grep -rniE "max_open|open_prs|wip_|work_in_progress|concurrent_pr" kstrl/` returns nothing. The only bounds are `max_parallel = 4` (`kstrl/factory.py:169`), the daemon running exactly one queue item per cycle (`kstrl/serve.py:1875-1880`), and the inbox open-item cap of 50 (`kstrl/inbox.py:267`), which gates admission of new work rather than output.

The proven default is one open PR per loop: without a bound, a daily loop can generate several unreviewed PRs in a week, producing review fatigue and merge conflicts. Scheduled runs no-op when the bound is met; manual runs bypass it, because a human typing the command is the authorisation. Stated as a rule: a loop may be handed only as much autonomy as its output can be cheaply and reliably verified.

The bound is shipped at 1 and is configurable. It is one line of config, reversible in seconds, and carries a proven default from a team running this in production. Raise it if it chafes.

### 1. Identify kstrl-authored pull requests

Every PR body kstrl writes ends with the line `Generated by [kstrl](https://github.com/0xfauzi/kstrl)` (`kstrl/pr.py:555` and `:711`). Hoist that literal into a module constant `PR_FOOTER_MARKER` in `kstrl/pr.py` and use it at both sites (no behaviour change; the string is identical). The dampener and the polled steering channel (R10.10) will use the same constant.

Add to `kstrl/pr.py`:

```python
def count_open_kstrl_prs(cwd: Path, *, limit: int = 100) -> int:
    """Open PRs whose body carries PR_FOOTER_MARKER.

    Runs ``gh pr list --state open --limit <limit> --json number,body``
    with GH_TIMEOUT and counts bodies containing the marker. Raises
    RuntimeError on any gh failure, timeout, or unparseable output: the
    caller decides what a failed count means.
    """
```

Use the same `subprocess.run([...], cwd=cwd, capture_output=True, text=True, timeout=GH_TIMEOUT)` shape as `create_component_pr` (`pr.py:575-586`).

### 2. Config

Add to `ServeConfig` (`kstrl/serve.py:239-251`), after `allow_uncovered_cost`:

```python
#: Scheduled admission stops while this many kstrl-authored PRs are open.
#: 0 disables the bound. Manual `ks factory` / `ks run` are unaffected.
max_open_prs: int = 1
```

Load from `[serve] max_open_prs` and env `KSTRL_SERVE_MAX_OPEN_PRS`, following how `max_consecutive_poison` is loaded in `ServeConfig.load` (`serve.py:307-352`). Negative values are a config error.

### 3. The gate

Add to `kstrl/serve.py`, next to `check_budget` (`serve.py:1438`):

```python
def check_open_pr_bound(
    config: ServeConfig, root_dir: Path, *, counter: Callable[[Path], int] | None = None,
) -> Admission:
    """Flow control: refuse admission while max_open_prs kstrl PRs are open.

    Skipped (allowed) when max_open_prs is 0, or when the factory will
    not open PRs at all ([factory] create_prs = false), because then
    there is nothing to bound. A failed count refuses admission with the
    error in the reason: an unknown number of open PRs is not zero.
    """
```

`counter` defaults to `count_open_kstrl_prs`; the parameter exists so tests inject a fake without patching `gh`. Return values:

- `max_open_prs == 0`: `Admission(allowed=True, reason="open-PR bound disabled")`, without calling the counter.
- `FactoryConfig.load(root_dir).create_prs` is false: `Admission(allowed=True, reason="open-PR bound not applicable (create_prs = false)")`, without calling the counter.
- count `< max_open_prs`: allowed, reason `"<count> of <max> kstrl PRs open"`.
- count `>= max_open_prs`: `Admission(allowed=False, reason="<count> kstrl PR(s) open (bound <max>); waiting for review")`. This is a **wait**, not a pause: `pause_reason` stays empty so the daemon re-checks next cycle.
- counter raised: `Admission(allowed=False, reason="cannot count open kstrl PRs: <error>")`, also a wait.

Insert it into the gates tuple in `serve_cycle` (`serve.py:1974-1978`) as the **last** gate, after `check_budget`, so cheaper local checks run first and a GitHub call happens only when everything else admits.

`ks serve --dry-run` already lists every gate in evaluation order (`docs/continuous-intake.md`); confirm the new gate appears there and add it to the doc's gate list.

### 4. Docs

- `kstrl.toml.example` `[serve]`: `max_open_prs = 1` with a two-line comment (bound, manual bypass).
- `docs/continuous-intake.md`: add the gate to the admission-gate list and one paragraph headed "Flow control" quoting the one-open-PR rationale and stating that manual runs bypass it.
- Regenerate README. CHANGELOG `### Added`. Mark the entry **Breaking** for daemon users: a repo with an open kstrl PR will stop admitting until it is merged or closed, unless `max_open_prs = 0`.

## Out of scope

- Bounding PRs within one `ks factory` run (that is `max_parallel`).
- Labels on PRs. The footer marker is enough and covers PRs created before this change.
- Counting PRs per queue item or per component.

## Tests to write

`tests/test_serve.py` (extend) or new `tests/test_flow_control.py`:

1. `test_bound_disabled_skips_counter`: `max_open_prs = 0`; counter is a function that raises if called; allowed.
2. `test_bound_not_applicable_when_no_prs`: `create_prs = false` in a temp `kstrl.toml`; counter raises if called; allowed with the "not applicable" reason.
3. `test_under_bound_allows`: counter returns 0, bound 1; allowed.
4. `test_at_bound_refuses_as_wait`: counter returns 1, bound 1; `allowed` false, `pause_reason == ""`, reason contains `1 kstrl PR(s) open`.
5. `test_counter_failure_refuses`: counter raises `RuntimeError("gh: not found")`; not allowed; reason contains `cannot count`.
6. `test_count_open_kstrl_prs_filters_by_marker`: patch `subprocess.run` to return JSON with three PRs, two carrying `PR_FOOTER_MARKER` in the body; result is 2.
7. `test_count_open_kstrl_prs_raises_on_gh_error`: non-zero return code raises `RuntimeError` whose message includes stderr.
8. `test_serve_cycle_includes_gate_last`: drive `serve_cycle` with a queue holding one ready item and a counter returning 1 (bound 1); the item is **not** leased and `result.skipped` mentions the bound. Then counter returns 0; the item is leased. Use the seam pattern in `tests/test_serve_seam.py`.
9. `test_dry_run_lists_gate`: `ks serve --dry-run` output contains the gate's name in the order list.
10. `test_config_env_beats_toml`: `[serve] max_open_prs = 3` plus `KSTRL_SERVE_MAX_OPEN_PRS=0` resolves to 0; a negative value raises.

## Acceptance (all must hold)

- [ ] With the bound set to N and N kstrl PRs open, the next scheduled admission no-ops with a reason naming the count; with fewer than N, it proceeds (tests 4, 3, 8).
- [ ] `ks factory` and `ks run` are unaffected: no code path outside `serve_cycle` calls the gate (a reviewer greps for `check_open_pr_bound` and finds exactly the definition, the gates tuple, the dry-run listing, and tests).
- [ ] Both footer sites in `kstrl/pr.py` use `PR_FOOTER_MARKER`; `tests/test_pr.py` still passes unmodified.
- [ ] On this repository, with PR #221 (or any kstrl-authored PR) open, `uv run ks serve --dry-run` shows the gate refusing; paste the line. If no kstrl-authored PR is open at the time, open a scratch one from a throwaway branch, capture the line, and close it; say which you did.
- [ ] Tests 1 through 10 pass; baseline suite passes; gen_docs check passes; CHANGELOG marked Breaking for daemon users.

0xfauzi:
## Summary

Nothing bounded the pull requests the `ks serve` daemon leaves open. A daily
loop could stack up several unreviewed PRs in a week; the only existing bounds
are `max_parallel` inside one run, one queue item per cycle, and the inbox
open-item cap, all of which bound what the daemon STARTS rather than what it
leaves behind. This adds flow control on the output.

- **`PR_FOOTER_MARKER`** (`kstrl/pr.py:33`). The footer line kstrl writes at the
  end of every PR body was spelled twice as a literal, in `_generate_pr_body`
  and in `create_single_pr`. It is now one constant used at both sites, with no
  change to the string. Its value is `Generated by [kstrl](https://github.com/0xfauzi/kstrl)`.
  A PR is kstrl-authored when its body ENDS with that line. Two error directions,
  both stated in the docstring and in `docs/continuous-intake.md`: it does not
  cover a PR opened before the footer took its current wording (the literal was
  the ralph-era one until `bd19f2b`), and it does count a hand-written body that
  ends with exactly that line. `#231` will read the same constant.

- **`count_open_kstrl_prs(cwd, *, limit=100) -> OpenPrCount`**
  (`kstrl/serve.py:2102`). Runs `gh pr list --state=open --limit <limit> --json
  number,body` through `intake_github.run_gh`, the package's one gh invocation,
  and validates the payload ROW BY ROW before counting anything: a row that is
  not a dict, has no `body` key, or whose `body` is not a string or null, raises
  with its index and a truncated repr. The TYPE check is round 2: presence alone
  let `{"body": 42}` through to `str(row["body"] or "")`, which coerced it to
  `"42"` and counted it as an unmarked PR, which is the fail-open one field over
  that #260 round 2 records. It returns a pair, not an int, because a page filled
  to `--limit` makes the count a lower bound rather than a count. Every failure
  it CONVERTS is a `RuntimeError` - a gh failure, a timeout, a missing binary,
  unparseable JSON, a payload that is not a list, an unreadable row - and the
  docstring now says that is not a promise nothing else escapes: `run_gh`'s
  decode raises `UnicodeDecodeError` and `json.loads` raises `RecursionError`,
  so a caller must still catch `Exception` (#318). `--state=open` is one argv
  token on purpose; see the census note below.

- **`check_open_pr_bound(config, root_dir, *, counter=None, streak=None) -> Admission`**
  (`kstrl/serve.py:2380`). Allowed without calling the counter when
  `max_open_prs` is 0, and allowed without counting when `[factory] create_prs =
  false`, because then there is nothing to bound. Under the bound on an
  unsaturated page it allows with `"<n> of <max> kstrl PRs open"`. At or over the
  bound it refuses with the ordinary reason, which is conclusive even on a full
  page because rows outside the window can only add to the count. Under the bound
  on a FULL page it refuses too, saying the number is a lower bound. Anything the
  block raises refuses with `cannot count open kstrl PRs: <error>`. The handler
  is `except Exception` and the `FactoryConfig.load` read is inside it: every
  outcome that is not a number means one thing here, and an enumeration of the
  types believed reachable is the defect rather than the precaution (#318). All
  refusals are WAITS: `pause_reason` stays empty, so nothing needs resuming and
  the next cycle proceeds on its own once the PR is gone. `streak` defaults to a
  throwaway instance rather than to `None`, which deleted four
  `if streak is not None` branches (cognitive 13 to 7), and BOTH early-allow
  paths now clear the streak: a gate that is switched off is not evidence that
  `gh` is broken, and leaving a half-built streak across a week of
  `max_open_prs = 0` contradicted the "three consecutive polls" contract.

- **`OpenPrCountStreak`** (`kstrl/serve.py:2206`) and `_record_count_failure`
  (`:2641`). A wait is right for a condition that clears itself and wrong for one
  that never does. Three consecutive polls with an unusable count file ONE inbox
  item through the existing `_file_inbox_item` path under the existing
  `halted_run` kind, deduped on a fixed key, and the first good count resets the
  streak.

  **It is persisted, which is round 2's main change.** `ks serve` ships in two
  launchd shapes and only one is a long-lived process: `--plist-mode keepalive`
  runs one polling process, `--plist-mode interval` runs `ks serve --once` on a
  `StartCalendarInterval`, one process per firing. Held in the loop's frame,
  `consecutive` restarted at 0 every interval firing, so the threshold was
  unreachable and the alarm did not exist in the mode most exposed to the failure
  it names. Measured before the fix: five `--once` firings against a failing
  `gh` filed **0** items where five keepalive polls filed **1**. It now lives in
  the control state directory as `pr_count_streak.json` (`kstrl/statedir.py:53`),
  written through `kstrl/atomicio.py` under `control_lock`, carrying
  `schema_version`, `consecutive` and `filed`. Loading fails toward the ALARM:
  a missing file is streak 0, and anything unreadable - bad permissions, bytes
  that are not utf-8, JSON that will not parse, a shape or schema version this
  code does not know - comes back AT the threshold with the cause reported on the
  observer, so the next unusable count files at once instead of re-earning three
  polls of silence. The damaged file is NOT rewritten, the rule #350 applied to
  `autonomy.json`: overwriting it destroys the only thing an operator could
  inspect, and the next load would then find a clean file and report nothing.

  **`filed` is set after the write, not before it.** `should_file()` used to mark
  the streak as its way of saying "this poll is the one", and only then was
  `_file_inbox_item` called, which swallows every failure and returns `""` and
  returns `""` without trying when `[inbox] enabled = false`. One failing write
  therefore disarmed the alarm for the life of the process and put an id naming
  nothing into `CycleResult.inbox_items`. It is a predicate now;
  `_record_count_failure` sets `filed` only on a non-empty id, appends only that
  id, and sets `needs_human` either way. The item's reason comes off the STREAK
  rather than off the caller's refusal string, because any wait gate's reason
  reaches that function and only the bound's belongs in an item about counting
  pull requests.

- **`_wait_gate_refusal`** (`kstrl/serve.py:2830`) holds the three gates that
  make the cycle wait, in cost order: inbox cap (one local file), factory lock
  (one flock), open-PR bound (one GitHub round trip). It returns `str | None`
  rather than `str`, so a gate refusing with an empty reason is not read as an
  admission. The bound is deliberately NOT a member of the `gates` tuple in
  `serve_cycle`: that tuple is built eagerly, so every member is evaluated before
  the loop reads the first refusal, and a `gh` call per poll behind an
  already-refusing budget is a cost with no purchaser. `TestGateOrderIsCost` is
  the control that keeps it out, with one arrangement per boundary and a positive
  control first.

- **`ServeConfig.max_open_prs`**, default `1`. Read from `[serve] max_open_prs`
  in `kstrl.toml` and overridden by `KSTRL_SERVE_MAX_OPEN_PRS`. A negative value
  is a config error (`ServeError`, which `config_preflight` already reports as
  operator input). `kstrl.toml.example`, the generated README config reference
  and `docs/continuous-intake.md` all carry it.

- **`ks serve --dry-run`** lists the gate last, matching evaluation order:
  `gate open-PR bound: ok` when it admits, `gate open-PR bound: BLOCKS - <reason>`
  when it does not.

`ks factory` and `ks run` are unaffected. `grep -rn check_open_pr_bound kstrl/ tests/`
returns 2 hits in `kstrl/serve.py` (the definition and the one call in
`_wait_gate_refusal`), 2 in `kstrl/cli.py` (the lazy import and the dry-run row),
12 in `tests/test_flow_control.py` and 1 in a guard comment. There is no path to
it outside `serve_cycle` and the dry run.

Closes #228.

## Behaviour changes

- **Breaking for daemon users.** A repository with one open kstrl PR admits no
  scheduled work until that PR is merged or closed. Set `[serve] max_open_prs = 0`
  to restore the old unbounded behaviour, or raise it.
- An operator watching `ks serve` sees
  `1 kstrl PR(s) open (bound 1); waiting for review` as a warning, and
  `result.skipped` carries the same string. No item is leased, no attempt is
  charged, no inbox item is filed, the queue is not paused.
- **Anything that is not a usable count refuses.** A missing binary, an auth
  failure, a timeout, unparseable output, a row that is not a PR record, a bad
  `[factory]` value and a locale decode failure all produce
  `cannot count open kstrl PRs: <error>` and admission stops. So does a `gh pr
  list` page filled to `--limit` when the count is under the bound, with a reason
  that says the number is a lower bound. An unknown number of open PRs is not
  zero, and a gate that cannot count must not clear.
- **A count that keeps failing files an inbox item.** After three consecutive
  polls with an unusable count the daemon files one `halted_run` item titled
  "Continuous intake cannot count open pull requests" and sets `needs_human`,
  then keeps waiting. One item per streak, deduped, reset by the first good
  count, in both LaunchAgent modes. This is new in review round 1 and was
  corrected in round 2, which found it unreachable in `interval` mode: before
  it, an expired `gh` token or a
  `gh` missing from launchd's PATH stopped the daemon admitting anything forever
  while every surface an operator checks read healthy - the queue unpaused, `ks
  inbox` empty, `needs_human` False, exit code 0, and one WARN line per poll in
  `serve.err.log` as the only evidence anywhere. A transient failure still clears
  itself and never reaches the threshold.
- **A new control file, `pr_count_streak.json`.** The daemon writes it in the
  XDG control directory beside `autonomy.json`, `spend.json`, `pause.json`,
  `inbox.jsonl` and `github_processed.json`, through `atomicio` under
  `control_lock`. It holds `{"schema_version": 1, "consecutive": <int>,
  "filed": <bool>}` and nothing else. It is created on the first `ks serve`
  cycle at the umask default (measured on this repository: mode 0644), never by
  `--dry-run`, never by `ks factory` or `ks run`. It is added to
  `statedir.CONTROL_FILENAMES`, so the symlink and escape checks in
  `_control_files_compromised` cover it; it is NOT added to
  `legacy_control_paths` or `STATE_FILES`, because it is post-R8.9 and never had
  an in-tree location. Deleting it is safe: the next poll reads it as first run
  and starts the count at zero. `migrate_control_state` now iterates
  `legacy_control_paths` rather than `CONTROL_FILENAMES`, since a file with no
  former location has nothing to migrate.
- **The remedy in the filed item no longer names one cause for two.** Three
  consecutive full pages filed an item telling the operator to check `gh`'s PATH
  and auth while `gh` was healthy. The detail now covers both causes and the
  reason string above it says which one happened.
- New troubleshooting rows in `docs/continuous-intake.md`, and the backstop list
  there goes from four to five.

## Review round 1: what the review found and what changed

An independent reviewer plus the code-review skill returned 0 blockers, 5
should-fix and 6 nits, with 20 mutations of which three stayed green. All five
should-fix items are fixed; four of the six nits are applied and two are
declined below.

| # | Finding | Done |
|---|---|---|
| S1 | The gate counted an unreadable payload as zero and cleared. Six row shapes measured: `["a","b"]`, `[1,2,3]`, `[{"number":1}]`, a nested list, a null row, and one good row with one bad, each giving `allowed=True reason='0 of 1 kstrl PRs open'`. A page saturated at `--limit` did the same. | Rows validated one by one with an indexed message before anything is counted; `OpenPrCount(count, saturated)` carries the truncation, and the gate refuses in both directions, keeping the ordinary bound reason in the conclusive one. Eight new tests. |
| S2 | The marker was matched anywhere in the body, so a PR that QUOTES the footer counted. Measured live: 8 open PRs, exactly one matched, and it was PR #354 itself. | Anchored: `body.rstrip().endswith(PR_FOOTER_MARKER)`. Pinned through the real writer (`_generate_pr_body`) and by a source check that both `lines.append(PR_FOOTER_MARKER)` sites are the last append before the join. Re-measured after the fix, same repository: the substring match counts 2 of 8, the anchor counts 1. |
| S3 | `assert not marker.exists()` was the only ordering control, and it passed with the fake `gh` switched off (mutations R18 and R19). Only one of the three boundaries the docstring claims was covered; moving the bound first inside `_wait_gate_refusal` (R14) left all 308 tests green. | `TestGateOrderIsCost`: a positive control that asserts the fake IS reached and the refusal says "cannot count", then one negative arrangement per boundary (budget, inbox cap at 1, a real `flock` on `factory.lock`). R14, R18 and R19 are all CAUGHT now. |
| S4 | `except RuntimeError` let two measured exceptions out of a daemon with no per-cycle handler: a `ValueError` from `FactoryConfig.load` on `[factory] max_parallel = "two"` (not even inside the `try`), and a `UnicodeDecodeError` from `run_gh`'s locale decode. | `FactoryConfig.load` moved inside the guarded block, handler widened to `except Exception` with the same reason. Both measured escapes are now tests, the decode one through the production path as well as the seam. This also closes the `RecursionError` nit the skill raised, which only failed closed by accident. |
| S5 | A persistently failing count was a permanent silent no-op: no inbox item, `needs_human` False, queue unpaused, exit 0. Neither of the two failures an unattended daemon actually meets clears itself. | `OpenPrCountStreak`, three consecutive unusable counts filing one deduped `halted_run` item. Driven through `serve` over 1, 2, 3 and 6 cycles with a failing fake `gh`, plus a recovery arrangement that proves the reset. **Round 1 kept it in the loop's frame, which is SF1 below**: it never reached launchd `interval` mode. It is persisted now. |
| N1 | "covers every PR kstrl has ever opened" is false: the footer was the ralph-era literal until `bd19f2b`. | Corrected in the docstring, the CHANGELOG and `docs/continuous-intake.md`, and both error directions are now stated rather than one. |
| N2 | A reword of `PR_FOOTER_MARKER` left all 308 tests green (mutation R15). | `test_the_marker_literal_is_pinned` asserts the value, with a comment saying what changing it costs. R15 is CAUGHT now. |
| N3 | `_wait_gate_refusal` used `""` as its no-refusal sentinel. | Returns `str \| None`. Two mutations plant the regression; both CAUGHT. |
| N4 | `limit` is undocumented. | `docs/continuous-intake.md` says the count looks at the newest 100 open PRs and what happens above that, which S1 made a real behaviour rather than a footnote. |
| N5 | `test_limit_reaches_the_argv` asserted membership, not position. | `argv[argv.index("--limit") + 1] == "7"`. |
| N6 | The idle-cost numbers move with the repository's PR count. | Re-measured below. |
| skill 11 | The seventh env override is written differently to duck the cyclomatic ratchet. | Declined. It is a refactor of `ServeConfig.load`'s seven-key ladder, not of this diff, and the semantics are identical. |
| skill 12, 13, 14, 15 | The allowed-path reason is never displayed; the docs never mention the `create_prs = false` exemption; `max_open_prs` bounds admission rather than open PRs; `run_gh` is imported lazily beside a hoisted import. | Not applied. Four separate small changes, three of them to surfaces this PR does not otherwise touch. Recorded here rather than filed, per the one-class-one-PR rule. |

## Review round 2: what the second review found and what changed

The second independent review returned **0 blockers, 2 should-fix and 11 nits**
over 23 mutations, of which one (T5) stayed green. Both should-fix items were in
round 1's own S5 remedy rather than in the gate: the alarm could not fire in one
of the two shipped launchd modes, and it disarmed itself permanently if its
first write failed. The gate itself the reviewer could not break.

| # | Finding | Done |
|---|---|---|
| SF1 | The S5 inbox alarm cannot fire in launchd `interval` mode, and three shipped artifacts say it can. Measured: five `serve(once=True)` calls against a `gh` exiting 99 give `needs_human` `[F,F,F,F,F]` and **0** inbox items; the same five polls in one keepalive loop give `[F,F,T,F,F]` and **1**. | The streak is persisted in the control state directory as `pr_count_streak.json`, so both modes accumulate it. Re-measured on the head: five `--once` firings give `[F,F,T,F,F]` and **1** item, identical to five keepalive polls. A damaged file arms the alarm rather than restarting the count, and is left byte-identical. 21 new tests in `tests/test_open_pr_streak.py`. |
| SF2 | One failed inbox write disarms the alarm permanently: `should_file()` set `filed` before the write, `_file_inbox_item` swallows every failure and returns `""`, and `result.inbox_items` got `('',)`. Measured: ten further failing polls after that filed **0** items. | `should_file()` is a pure predicate; `_record_count_failure` sets `filed` only on a non-empty id and appends only that id. Measured on the head: the crossing poll with the inbox disabled gives `inbox_items == ()`, `needs_human` True and `filed` still False, and the next poll files exactly one item. |
| NIT 1 | Neither early-allow path resets the streak, so `max_open_prs = 0` preserves a half-built one. | Both call `record_conclusive()`. The four `if streak is not None` guards became a null-object default in the same move, so the branch count fell rather than rose. |
| NIT 2 | The row validator never checks the type of `body`, the one field it reads. | `isinstance(row["body"], str \| None)`. Three new parametrized shapes: an int body, a list body, a dict body. |
| NIT 3 | Saturation drives the same streak, so the filed item gives the wrong remedy. | The detail covers both causes; the reason string above it says which happened. |
| NIT 4 | The fixed dedupe key has no test: mutation T5 STILL GREEN. | `TestTwoInstancesFileOneItem`: two streak instances filing on the same failure give one item with `occurrences == 2`. **T5 is CAUGHT now.** |
| NIT 5 | The counter's stated `RuntimeError` contract is false. | The docstring says which failures it converts and that a caller must still catch `Exception`, naming both escapes. |
| NIT 6 | `docs/lessons/verify/pr-221/admission_gates.py` teaches the opposite gate order. | The widget, its JS mirror in `docs/lessons/pr-221.html`, the SVG chain and the caption all move to the shipped order. `node check_widgets.mjs` reports `admission_gates: 110592 rows agree with the Python sweep` and all nine of the script's claims still hold; claim 4 gained the two conditions the new order needs and selects 264 rows, so it is not vacuous. The half about the lesson HTML calling FlowControl "planned" is left alone, as the reviewer rejected it: 17 `planned` states remain across that map. |
| NIT 7 | The encoding census row now counts a non-read, so a compensating pair can hide a regression. | `--state=open` is one argv token, so the literal leaves `kstrl/serve.py`. The layer-1 row stays at 5 and its comment now says five READS: four existing plus the streak load. Layer 2 gained `serve.py path.read_text(encoding='utf-8')`, which is what proves the composition rather than a coincidence. |
| NIT 8 | `tests/helpers/fakegh.py` imports a private name out of a test module. | Moved to `tests/helpers/executables.py` as `write_executable`. A second instance of the same class was found and fixed in the sweep: `tests/helpers/proclifecycle.py` imported `folded_str` from `tests/test_journal_one_writer.py`, which only re-exports it from `tests/helpers/astwalk`. `tests/test_helper_import_direction.py` is the guard for instance N+1, with four controls (absolute from-import, plain import, relative import, and a helper-to-helper import that must NOT hit). |
| NIT 9 | `limit=100` is unreachable from production and a busy repository blocks forever. | The plain consequence is now in `docs/continuous-intake.md` and the CHANGELOG: at 100 or more open pull requests the daemon admits nothing until the bound is switched off. The `[serve] open_pr_scan_limit` key and the server-side `--search` filter are DECLINED here: both are new surfaces on a gate this round is fixing, and the `--search` idea is the handoff below. |
| NIT 10 | The decode plus list-shape check duplicates `intake_github.parse_issue_list`. | DECLINED, and recorded as a handoff. A shared `decode_gh_json_list(payload, command)` beside `run_gh` is the right shape and would also have to change `parse_issue_list`'s per-entry tolerance, which is a deliberate different failure model documented in its own docstring. |
| NIT 11 | Numbers in the PR body and FIX_R1 that do not reproduce. | Every number in this body is re-measured on the round-2 head and listed below. The five that were wrong: mypy file count, `tests/helpers/fakegh.py` length, `tests/test_flow_control.py` length, four of five line citations, and "`serve_cycle` gained no branch", which was wrong in the safe direction. |
| skill 7 | Ordering the bound after `factory_lock_held` widens the courtesy window to 60s. | Not applied, as the reviewer concluded: the lock check's own comment already says it cannot make exit 2 unambiguous, which is why `classify_run` reads the child's output instead. |
| skill 8 | `test_the_other_writer_joins_immediately_after_the_footer` is a clearing source-text guard. | DECLINED as a handoff, as the reviewer allowed. Extracting `_generate_single_pr_body` and asserting behaviourally is the right fix and is not a defect this PR introduced: `create_single_pr` had no body test before it either. |
| skill 15 | The CHANGELOG drops the human-edited-body case. | Not applied. `docs/continuous-intake.md` states both directions explicitly and the CHANGELOG carries the false negative by implication. |

Every number below was measured on the round-2 head, on the tree merged with
current `origin/main` (`414d662`), with `__pycache__` deleted first and
`PYTHONDONTWRITEBYTECODE=1`.

Tested, by running it:

- **Full suite**, bounded at 1800s: **5974 passed, 37 skipped, 41 xfailed in
  510.30s**, exit 0, no failures. Round 1 on `0522501`: 5765 / 37 / 38 in
  333.55s. The reviewer saw one failure in
  `tests/test_timeout_enforcement.py`, a load flake in a file this PR does not
  touch; it passes here in the full run.
- `uv run mypy kstrl/ --strict`: Success, **131** source files. The round-1 body
  said 129 and the round-1 report said 130; 130 was right at the time and 131 is
  right now, because `origin/main` added `kstrl/appendio.py`.
- `uv run ruff check kstrl/ tests/`: All checks passed.
- `uv run python scripts/gen_docs.py --check`: README generated sections are
  current.
- `scripts/precommit/file_length_ratchet.py` and
  `scripts/precommit/cyclomatic_ratchet.py`, with the PR diff staged over
  `origin/main` in a detached worktree, which is the comparison the hooks make:
  both exit 0. The only file over the 800-line limit is `kstrl/serve.py`, 3704
  to 3709 lines, already over and not CROSSED.
- `complexipy --staged --max-complexity-allowed 15` against the same staged
  tree: exit 0, **1 regressed, 2 improved, 106 new**. Regressed:
  `ServeConfig.__post_init__` 4 to 5, from this branch's `max_open_prs`
  validation, under the limit. Improved: `serve_cycle` **51 to 50** and
  `admission_gates.py::main` 36 to 27. **The round-1 body's "`serve_cycle`
  gained no branch" was wrong, in the safe direction**: moving the two wait
  gates out made it better. Cyclomatic, measured with ruff C901 at both
  revisions: `serve_cycle` **32 on `origin/main`, 31 here**.
- Per-function cognitive complexity of everything this PR adds, all under 15:
  `count_open_kstrl_prs` 9, `check_open_pr_bound` **7** (13 in round 1, before
  the null-object streak deleted four branches), `OpenPrCountStreak._from_payload`
  5, `.load` 3, `.save` 2, `.should_file` 1, `_record_count_failure` 3,
  `_wait_gate_refusal` 3, `_load_pr_count_streak` 1, `_save_pr_count_streak` 1,
  `serve` 11.
- `check_no_mocks.py` on `kstrl/serve.py` and `kstrl/statedir.py`: exit 0. An
  em dash, en dash and emoji scan by codepoint category over every added line of
  `git diff origin/main...HEAD`: **0 hits**.
- **File lengths, re-measured**, because three were wrong in the round-1 body:
  `tests/test_flow_control.py` **717** (656 before this round),
  `tests/test_open_pr_counter.py` **333** (307), `tests/test_open_pr_streak.py`
  **391** (new), `tests/test_helper_import_direction.py` **100** (new),
  `tests/helpers/fakegh.py` **103** (the round-1 body said 88; it grew when
  `FAKE_GH_THIRD_CALL_WORKS` was added), `tests/helpers/executables.py` **19**
  (new), `kstrl/serve.py` **3709**, `kstrl/statedir.py` **670**. The persistence
  tests went into their own file rather than into `test_flow_control.py`
  precisely so the 800-line ratchet is not approached again.
- **40 mutation runs, one plant at a time**, `__pycache__` deleted before every
  run, `PYTHONDONTWRITEBYTECODE=1`, each bounded at 420s with
  `start_new_session` plus `os.killpg`, and the anchor asserted to appear
  EXACTLY ONCE before every write so a plant that did not apply is
  COULD-NOT-PLANT rather than scored as caught. Target:
  `tests/test_flow_control.py tests/test_open_pr_counter.py
  tests/test_open_pr_streak.py tests/test_serve_cli.py tests/test_serve.py
  tests/test_statedir.py tests/test_helper_import_direction.py
  tests/test_encoding_readers.py`, **412 passed in 87.21s** clean.
  **37 CAUGHT, 3 STILL GREEN, none hung, none failed to plant.** Harness and raw
  log in the lane. The three still green are the two deliberate controls (C0, a
  no-op docstring edit in `kstrl/serve.py`, and C1, the same in a test file) and
  P1, which blanks the ordering suite's own positive control: deleting a control
  cannot make another test fail, and R19 is the measurement that the control
  discriminates.

  The reviewer's 23, re-targeted at the round-2 source: R14, R18b, R19, V1, V2,
  V3, S1, S2, A1, A2, K1, K2, T1, T2, T3, T4, T6 and N1 all CAUGHT as before,
  and **T5 (the dedupe key varies per poll), which was STILL GREEN in the review,
  is CAUGHT now** (1 failed).

  Seventeen new plants, at least one per round-2 item:

  | # | Mutation | Outcome |
  |---|---|---|
  | V4 | NIT 2: the `body` TYPE check is dropped, so 42 is coerced to "42" | CAUGHT (3 failed) |
  | F1 | SF1: the streak is built fresh per process again | CAUGHT (3) |
  | F2 | SF1: the streak is never persisted, so `--once` cannot accumulate | CAUGHT (2) |
  | F3 | SF1: the decode escapes the load handler (`except OSError` alone) | CAUGHT (3) |
  | F4 | SF1: a damaged file restarts the count at zero instead of arming | CAUGHT (3) |
  | F5 | SF1: a damaged file is overwritten with a fresh count | CAUGHT (2) |
  | F6 | SF1: the schema version is not checked | CAUGHT (2) |
  | F7 | SF1: `True` validates as a count of 1 (`bool` is an `int` subclass) | CAUGHT (1) |
  | F8 | SF2: `filed` is set even when the inbox write returned nothing | CAUGHT (1) |
  | F9 | SF2: an empty id enters `inbox_items` and the alarm disarms | CAUGHT (1) |
  | F10 | SF2: `should_file` drops the reason check, so another gate files | CAUGHT (2) |
  | F11 | NIT 1: `max_open_prs = 0` preserves a half-built streak | CAUGHT (1) |
  | F12 | NIT 1: `create_prs = false` preserves a half-built streak | CAUGHT (1) |
  | F13 | NIT 7: `--state open` goes back to two argv tokens | CAUGHT (2) |
  | F14 | NIT 8: a helper imports from a test module again | CAUGHT (1) |
  | F15 | the import guard stops resolving relative imports (its own blind spot) | CAUGHT (1) |
  | F16 | the new control file leaves `CONTROL_FILENAMES` | CAUGHT (39) |
  | F17 | the migration iterates every control filename again (KeyError) | CAUGHT (166) |

- **Every static guard was EXECUTED before and after the merge and the censuses
  diffed**, rather than pin literals being compared. All **33** files in
  `tests/` that walk `kstrl/` or `tests/` (found by grepping for
  `package_sources()`, `astwalk` or `ast.parse`), plus `tests/test_statedir.py`:
  **1175 passed, 9 skipped, 37 xfailed** before the merge and **1231 passed, 9
  skipped, 40 xfailed** after it. The 33rd file is this PR's own
  `tests/test_helper_import_direction.py`.

  Every census pin was also re-derived by IMPORTING its module at each revision
  and printing the value, **135 constants across 31 modules**. Across the
  `414d662` merge, seven pins differ and every one is attributable to a commit
  merged in: `EXPECTED_READ_SPELLINGS` (nine rows, #331 and #352 routing appends
  through the new `kstrl/appendio.py`), `EXPECTED_DECIDED_OUT`,
  `EXPECTED_EVENT_NAME_SPELLINGS`, `EXPECTED_JOURNAL_PATH_SITES` and
  `EXPECTED_AWAIT_SITES`. None of this branch's own pins moved in the merge.

  Against the review base `c8d6329`, this round moves exactly **four** pins, and
  the census that pins control-state paths is one of them:

  | Pin | c8d6329 | head | Why |
  |---|---|---|---|
  | `statedir.CONTROL_FILENAMES` | 5 names | 6 names | `pr_count_streak.json`, the new control file. It is deliberately NOT in `legacy_control_paths` or `STATE_FILES`: post-R8.9, so it never had an in-tree location, and `migrate_control_state` now iterates the legacy map instead. Mutation F16 (drop it from the tuple) is CAUGHT at 39 failed, F17 (iterate `CONTROL_FILENAMES` again) at 166 |
  | `statedir.CONTROL_PR_COUNT_STREAK` | absent | `pr_count_streak.json` | the constant itself |
  | `test_encoding_readers.EXPECTED_CLEARED_READS` | no `serve.py path.read_text` row | has one | the streak load |
  | `test_helper_import_direction.EXPECTED_TEST_MODULE_IMPORTS` | absent | `{}` | the new guard |

  **`test_encoding_readers.EXPECTED_READ_SPELLINGS["serve.py"]` did NOT move,
  and that is the one to look at.** It was 5 before and is 5 now, but it means
  something different: NIT 7 removed the argv literal `"open"` (5 to 4) and the
  streak load added a real read (4 to 5). A stationary number across two
  compensating changes is exactly the shape CLAUDE.md warns about, so it is not
  the evidence here: layer 2's `EXPECTED_CLEARED_READS` gained the new read as
  its own row, mutation F13 puts the argv literal back and is CAUGHT, and the
  row's comment now says five READS rather than four reads plus a word.
  `_EXPECTED_STATE_DIR_SPELLINGS` and `EXPECTED_RUNTIMEERROR_SPELLINGS` did not
  move at all: the streak is reached through `control_file`, which spells
  neither `state_dir` nor `.kstrl`.
- **Merging `origin/main` (`414d662`).** One conflict, in `CHANGELOG.md`, where
  both sides added an entry under `### Added`; both kept. The full suite was run
  on the merged tree, not on the branch alone.
- **`lanes/228/r2_verify_head.py`** re-runs the four round-2 properties against
  the head as a standalone probe with no pytest and no fixtures: five `--once`
  firings and five keepalive polls file one item each with identical
  `needs_human` sequences; a failed crossing write followed by a working inbox
  files exactly one item; an unreadable streak file is left byte-identical and
  one item is still filed; two instances filing on one failure make one item
  with `occurrences == 2`. **4 of 4 pass.**
- **Acceptance, re-run on this repository.** `uv run ks serve --dry-run --root .
  --no-color` reports `gate open-PR bound: ok` with 6 open pull requests, none of
  whose bodies ends with the footer. The dry run writes no streak file, as it
  does not enter `serve`. One real `uv run ks serve --once --root . --no-color`
  (empty queue, 0 items run) then wrote
  `~/.local/state/kstrl/kstrl-9c46970ababf1b90/pr_count_streak.json` containing
  `{"schema_version": 1, "consecutive": 0, "filed": false}` at mode **0644**,
  which is the umask default `atomicio` promises rather than `mkstemp`'s 0600.
  The file was removed afterwards so the control directory is as it was found.
  Round 1's scratch PR **#356** remains CLOSED with its branch deleted, as does
  round 0's **#346**.
- **Live marker counts, re-measured.** 6 open pull requests: the substring match
  counts **1**, the shipped anchored match counts **0**, and the one substring
  hit is PR #354's own Summary quoting the constant, which the anchor correctly
  rejects. **0** of the 6 bodies is a non-string, so the NIT 2 type check costs
  nothing today and is for the case it is for.

Assumed, not measured: the three declined skill findings and NIT 10's shared
decoder were read, not run. `docs/lessons/verify/pr-221/map_paths.py`, in the
same directory as the widget this PR corrects, imports `logical_model` from
`scripts/atlas`, which does not exist on `origin/main`; `check_widgets.mjs`
therefore dies after its widget checks. That is broken at the base commit too
and belongs to the atlas tooling, so it is reported rather than fixed. Nothing
else is outstanding.

## Handoff, measured rather than filed

**The idle daemon's GitHub cost.** The gate is evaluated before anything asks
whether there is work to admit, so an idle daemon makes one `gh pr list` per
poll forever. Re-measured on this repository with 6 open PRs, five calls:
**0.51, 0.41, 0.49, 0.47 and 0.81 seconds, returning 119,166 bytes each**
(machine load average 8.5, other lanes running). At the 60 second default poll
that is 1,440 spawns and about **172 MB** of GitHub traffic a day on an idle
queue. Round 1 reported 123,081 bytes at 7 open PRs and round 0 reported 91,644;
the conclusion does not move but the constants track the repository's PR count,
which is why they are re-measured each round rather than quoted.

The containment is one line, peeking `queue.next_ready` before the gates, but it
changes which reason an idle cycle reports, and the three eager gates
(`check_poison_breaker`, `check_cost_coverage`, `check_budget`) have exactly the
same property, so it is a decision about all four rather than about this one.
Not taken here.

**Two handoffs the second review added, both declined here with a reason.**

`gh pr list --search '"Generated by [kstrl]" in:body'` would make the gate cheap
by construction and would also dissolve the saturation case, since the server
would filter rather than the client. It is the better long-term shape and it is
a different gate: it moves the marker semantics into a GitHub search query whose
matching rules are not the anchored `endswith` this PR just spent a round
getting right, and it would need its own measurement of what that query actually
matches. Not a same-round change.

`decode_gh_json_list(payload, command)` beside `run_gh` would give
`count_open_kstrl_prs` and `intake_github.parse_issue_list` one decoder instead
of two that already differ: the new one catches `ValueError` per #318, the old
one catches only `json.JSONDecodeError`, so a `RecursionError` refuses here and
escapes there. The two also have deliberately different per-row policies -
`parse_issue_list` skips a non-dict row with `continue` and its docstring argues
for that tolerance - so sharing the decoder means deciding whether that
tolerance survives, which is a change to intake rather than to flow control.

## Quality pass

A four-angle cleanup review over the round-0 diff applied four changes, none of
which moves behaviour. `count_open_kstrl_prs` had re-implemented
`intake_github.run_gh` line for line, the same `capture_output`/`text`/`timeout`
call with the same `TimeoutExpired` and `OSError` pair and the same
stderr-or-stdout fallback; it now calls `run_gh` and decides only the shapes of a
bad payload, which is the half that is its own. The cost of that duplication was
visible in the pins: reuse takes the spawn sweep back to its base value, so two
census pins move in this PR instead of three. The `no_open_prs` fixture, pasted
as an autouse copy into four test modules, moved to `tests/conftest.py` with the
modules opting in through `pytestmark`; opt-in rather than autouse because
`tests/test_flow_control.py` asserts the bound's real behaviour and an autouse
stub it had to opt OUT of is a guard switched off in the one file where it
matters. The five failure tests for the counter, differing only in a patch
keyword and a match string, are one parametrized case. Two comments narrating
rejected drafts are cut to the fact.

Three findings were not applied. Making the `gates` tuple lazy so that
`_wait_gate_refusal` disappears is a real design point, and the loop already
branches the right way, but it moves the `ServeStateError` handling inside the
loop and rewrites every gate's shape. Dropping the `counter=` seam and the
`limit=` parameter: both are in the issue's stated signature, and a typed
parameter is a safer seam than patching a module attribute. The third is the
handoff above, with numbers.

Round 2 removed a mechanism rather than adding one, twice. The four
`if streak is not None` guards in `check_open_pr_bound` became a null-object
default, which took it from cognitive 13 to 7; round 1 had found this and
declined it because a mid-campaign production edit would have made its 36
mutation outcomes evidence for code that no longer existed, and this round
re-ran all 40 anyway. And `_record_count_failure` stopped taking the caller's
refusal string, which it could not tell apart from another gate's, and reads the
streak's own reason instead.

## Checklist

## Provenance

None. No `*_PROMPT` constant and no text reaching an LLM role's prompt is
touched, so H2 and H3 do not apply.


kstrl/serve.py
```diff
@@ -2891,6 +3197,10 @@ def serve(
     cfg = config or ServeConfig.load(root_dir)
     obs: ServeObserver = observer or _NullObserver()
     sleep = sleeper or time.sleep
+    # One per call, so a streak is a streak of THIS loop's polls. `once`
+    # gets one too: it is a list of one cycle, so nothing can reach the
+    # threshold, and passing it keeps the two paths one code path.
+    pr_count_streak = OpenPrCountStreak()
```
0xfauzi: **The streak can never fire in launchd `interval` mode, which is the deployment its own docstring names.**

`launchd_plist_dict` builds `interval` mode as `ks serve --once` on a `StartCalendarInterval`, so every firing is a **fresh process** and `pr_count_streak` starts at 0 each time. `threshold = 3` is unreachable, and the S5 inbox item is never filed - for exactly the two failures `OpenPrCountStreak`'s docstring names ("an expired `gh` token and a `gh` missing from launchd's PATH").

Measured on this branch:

```
serve(tmp, once=True, config=ServeConfig(max_open_prs=1))
-> cycles: 1  needs_human: [False]
   skipped: 'cannot count open kstrl PRs: gh pr failed (1): failed to run'
   inbox: empty, exit 0
```

That is the same silent-healthy surface the round-1 finding set out to remove, still fully intact for the scheduled deployment. `serve`'s comment ("`once` gets one too: it is a list of one cycle, so nothing can reach the threshold") states the fact but reads it as a harmless consequence rather than as the gap.

The streak has to survive the process to bound a per-process-restart failure: persist the counter next to the spend ledger in the state dir (it is already the daemon's own state, already fail-closed on an unreadable read), or file on the first failure when `once=True`.

kstrl/serve.py
```diff
@@ -2064,6 +2079,201 @@ def check_inbox_cap(root_dir: Path) -> Admission:
...
+
+    def record_conclusive(self) -> None:
+        self.consecutive = 0
+        self.filed = False
+
+    def should_file(self) -> bool:
+        """Whether THIS poll is the one that files, and mark it filed."""
+        if self.filed or self.consecutive < self.threshold:
+            return False
+        self.filed = True
```
0xfauzi: **`filed` is set before the write is attempted, and `_file_inbox_item` swallows every failure - so one failed or disabled inbox write permanently suppresses the alarm.**

`should_file()` mutates `self.filed = True` and returns True; `_record_count_failure` then calls `_file_inbox_item`, which catches `(OSError, TypeError, ValueError, KeyError, ControlStateError)` and returns `""`, and also returns `""` when `[inbox] enabled = false`. Nothing feeds that back, so the streak is burned and the only alarm for a permanently-stuck daemon is gone for the life of the process.

Measured, with `[inbox] enabled = false`:

```
_record_count_failure(root, streak, "cannot count", result)
-> result.inbox_items == ('',)   result.needs_human == True   streak.filed == True
second call -> inbox_items == ()  needs_human == False   (nothing ever files again)
```

Two defects in one: the empty string is recorded in `result.inbox_items` as if it were an item id (a caller counting `len(result.inbox_items)` sees one item that does not exist), and the suppression is permanent. A disk-full or permissions failure on the inbox is precisely the moment you want the item, and it is the moment it is thrown away.

Set `filed` from the outcome instead of before it:

```suggestion
    def should_file(self) -> bool:
        """Whether THIS poll is the one that files. The caller marks it."""
        return not self.filed and self.consecutive >= self.threshold
```

and in `_record_count_failure`, only `streak.filed = True` and append to `result.inbox_items` when `_file_inbox_item` returned a non-empty id.

kstrl/serve.py
```diff
@@ -2064,6 +2079,201 @@ def check_inbox_cap(root_dir: Path) -> Admission:
...
+
+    count: int
+    saturated: bool
+
+
+def count_open_kstrl_prs(cwd: Path, *, limit: int = 100) -> OpenPrCount:
+    """Open PRs whose body ENDS with :data:`PR_FOOTER_MARKER` (R10.7).
+
+    Runs ``gh pr list --state open --limit <limit> --json number,body``
+    under ``GH_TIMEOUT``. Raises ``RuntimeError`` on any gh failure,
```
0xfauzi: **The documented error contract is false, and it is the contract the next reader will code against.**

"Raises ``RuntimeError`` on any gh failure, timeout, missing binary, unparseable output or unrecognised row" is not what this function does. Two escapes are not `RuntimeError`:

- `run_gh` calls `subprocess.run(text=True)` with no `encoding=`, so a locale decode failure raises `UnicodeDecodeError` (a `ValueError`) straight through - this PR's own `test_a_decode_failure_in_the_real_counter_refuses` proves it.
- `json.loads` raises `RecursionError` on deeply nested input, which `except ValueError` does not catch.

`check_open_pr_bound` survives only because it catches `Exception`. But `PR_FOOTER_MARKER`'s docstring already advertises two more readers ("the dampener, the polled steering channel (#231)"), and the first thing either will do is read this docstring and write `except RuntimeError` - which is #318 reproduced, with a daemon that has no per-cycle handler underneath it.

Either state the real contract (the type is not part of it; catch `Exception`), or make it true by wrapping the body so everything leaves as `RuntimeError`.

kstrl/serve.py
```diff
@@ -2064,6 +2079,201 @@ def check_inbox_cap(root_dir: Path) -> Admission:
...
+    not tell "zero open" from "zero within the newest hundred", and the
+    second one is exactly the condition the bound exists for. Hence a
+    pair, not an ``int``.
+    """
+
+    count: int
+    saturated: bool
+
+
+def count_open_kstrl_prs(cwd: Path, *, limit: int = 100) -> OpenPrCount:
```
0xfauzi, line 2102: **`limit` is unreachable from anywhere, so a repository with 100 or more open PRs blocks the daemon permanently with no knob but `max_open_prs = 0`.**

`check_open_pr_bound` calls `count_fn(root_dir)` with no `limit`, and nothing in `kstrl/` passes one. On a repo with >= 100 open PRs the page is always saturated, so whenever the kstrl count is under the bound the gate takes the `counted.saturated` branch and refuses **every poll, forever**. 100 open PRs is ordinary for a busy repository, and the only escape is switching the bound off entirely.

`limit` being a parameter no production caller can set is the tell: it reads as configurable and is not. Either thread it from `[serve]` (`max_open_prs_scan_limit`, say), or - better - stop paging at all and let GitHub filter, see the note on `count_fn`.

kstrl/serve.py
```diff
@@ -2064,6 +2079,201 @@ def check_inbox_cap(root_dir: Path) -> Admission:
...
+        rows = json.loads(result.stdout)
+    except ValueError as exc:
+        raise RuntimeError(f"gh pr list returned unparseable JSON: {exc}") from exc
+    if not isinstance(rows, list):
+        raise RuntimeError(f"gh pr list returned {type(rows).__name__}, expected a list")
+    count = 0
+    for index, row in enumerate(rows):
+        if not isinstance(row, dict) or "body" not in row:
+            raise RuntimeError(f"gh pr list row {index} is not a PR record: {row!r:.120}")
+        if str(row["body"] or "").rstrip().endswith(PR_FOOTER_MARKER):
```
0xfauzi: **The one field the count actually reads is never type-checked, so a non-string body is silently counted as unmarked instead of refused.**

The row validator above checks `isinstance(row, dict)` and `"body" in row`, then this line does `str(row["body"] or "")`. A row of `{"number": 1, "body": 123}` or `{"number": 1, "body": {}}` passes validation and is coerced to `"123"` / `"{}"`, neither of which ends with the marker - so it counts as a non-kstrl PR rather than raising.

That is the S1 fail-open one field over, and the same shape as the round-2 finding recorded in CLAUDE.md ("the new raw validator took `severity` verbatim while `_parse_spec_issues` checked it against `_VALID_SEVERITIES`"): the validator and the reader disagree about what a valid record is, and the weaker definition is the one that decides. `None` is the only non-`str` body gh actually emits and it deserves the `or ""`; every other type is a payload shape change and should raise with its index like the rest.

```python
body = row["body"]
if body is not None and not isinstance(body, str):
    raise RuntimeError(f"gh pr list row {index} has a non-string body: {body!r:.120}")
```

kstrl/serve.py
```diff
@@ -2228,6 +2438,51 @@ def _file_inbox_item(
...
+    result.needs_human = True
+    result.inbox_items += (
+        _file_inbox_item(
+            root_dir,
+            kind_name="halted_run",
+            title="Continuous intake cannot count open pull requests",
+            detail=(
+                f"{streak.consecutive} polls in a row: {reason}. The daemon is "
+                "waiting and will admit nothing until the count works. Check "
+                "that `gh` is on the daemon's PATH and authenticated, or set "
```
0xfauzi: **This remediation is wrong for the saturated-page case, which is one of the two ways the count becomes unusable.**

`check_open_pr_bound` refuses with `"cannot count open kstrl PRs: gh returned a full page..."` when the page is full and the count is under the bound. That refusal increments the same streak, so after three polls the operator gets an item titled "Continuous intake cannot count open pull requests" telling them to "Check that `gh` is on the daemon's PATH and authenticated" - when `gh` is working perfectly, is on PATH, and is authenticated.

Combined with the hard-coded `limit=100`, a repository with >= 100 open PRs files an inbox item instructing the operator to fix a healthy `gh`, and the only advice that would actually work (`max_open_prs = 0`) is the last clause. Branch the detail on which refusal produced the streak, or drop the PATH/auth sentence in favour of the reason string that is already interpolated above it.

kstrl/serve.py
```diff
@@ -2330,6 +2585,61 @@ def _pause_queue(
...
+    if factory_lock_held(root_dir):
+        # Not a failure and not the item's fault: something else owns the
+        # repo. Wait rather than charging an attempt. This is a courtesy
+        # check only - it cannot make exit 2 unambiguous, which is why
+        # classify_run reads the child's output instead (#186 F6).
+        reason = "a factory run already holds this root"
+        obs.info(reason)
+        return reason
+
+    pr_gate = check_open_pr_bound(config, root_dir, streak=streak)
```
0xfauzi, line 2877: **The cost ordering puts a network call of up to `GH_TIMEOUT` (60s) between the factory-lock check and the claim.**

`factory_lock_held(root_dir)` used to be the last thing evaluated before `queue.lease(...)`. Now a `gh pr list` round trip sits between them, so the window in which a human's `ks factory` can take `.kstrl/factory.lock` after the daemon has already cleared that gate grows from microseconds to as much as a minute.

The comment above rightly calls the lock check "a courtesy check only", and the real protection is the child's own flock plus `classify_run`, so the cost is a wasted spawn and a charged attempt rather than two factories on one repo. But it is a real widening introduced by this ordering and it is not mentioned anywhere. Re-checking `factory_lock_held` immediately before the lease, or evaluating the bound before the lock check (it is the lock check that is cheap to repeat, not the gh call), closes it.

tests/test_open_pr_counter.py
```diff
@@ -0,0 +1,307 @@
...
+        Asserted through `_generate_pr_body`, the production writer, not
+        a test fixture imitating it: no network, no gh, no manifest on
+        disk. If a future edit appends anything after the footer, this
+        test goes red rather than the bound quietly counting zero.
+        """
+        manifest = _test_manifest()
+        body = _generate_pr_body(manifest.components[0], manifest)
+        assert body.rstrip().endswith(PR_FOOTER_MARKER)
+
+    def test_the_other_writer_joins_immediately_after_the_footer(self) -> None:
```
0xfauzi, line 257: **This is a CLEARING guard that cannot prove the site is compliant - the direction CLAUDE.md says a clearing guard must never be wrong in.**

> "A guard that CLEARS must be narrow, because over-matching converts a resolution into a clearing and deletes the mechanism... If a clearing guard cannot PROVE a site is compliant, it must flag."

It clears `create_single_pr` by checking that the first non-blank line after `lines.append(PR_FOOTER_MARKER)` contains `"\n".join(lines)`. That passes for a body that is *not* footer-terminated:

```python
lines.append(PR_FOOTER_MARKER)
body = "\n".join(lines)          # guard clears here
body += "\n\nCo-authored-by: ..."  # anchor broken, guard silent
```

and `create_single_pr` has no behavioural test at all to catch it, which is the stated reason for the source check ("the only way to reach that writer is a real `git push`"). That reason is removable rather than fundamental: extract the body construction into `_generate_single_pr_body(manifest, ...)` exactly as `_generate_pr_body` already is, and assert `body.rstrip().endswith(PR_FOOTER_MARKER)` on the real writer. That deletes the source-text guard instead of narrowing it, and makes both writers pinned the same way.

kstrl/serve.py
```diff
@@ -2064,6 +2079,201 @@ def check_inbox_cap(root_dir: Path) -> Admission:
...
+    count_fn = counter if counter is not None else count_open_kstrl_prs
+    try:
+        from kstrl.factory import FactoryConfig
+
+        if not FactoryConfig.load(root_dir).create_prs:
+            return Admission(
+                allowed=True,
+                reason="open-PR bound not applicable (create_prs = false)",
+            )
+        counted = count_fn(root_dir)
```
0xfauzi, line 2447: **An idle daemon downloads up to 100 full PR bodies every poll, and two cheaper forms exist that the handoff does not consider.**

The PR body measures it: 1,440 `gh` spawns and ~177 MB/day on an idle queue, the only network I/O in the cycle. The declined containment is peeking `queue.next_ready` before the gates. There is a second one that is strictly better and costs no reason-string change:

```
gh pr list --state open --search '"Generated by [kstrl]" in:body' --json number --limit N
```

GitHub does the filtering, the payload drops from ~123 KB to a few hundred bytes, `--json body` is no longer needed at all, and - the part that matters more than the bandwidth - the saturation problem disappears, because the page is now a page of *matching* PRs rather than a page of all open PRs. The `saturated` half of `OpenPrCount` and the permanent block on a >= 100-PR repository both go away with it.

Worth measuring before adopting (`in:body` search on markdown link text needs a check), but it is the version of this gate that is cheap by construction rather than cheap by ordering.

docs/control-loop-design.md
```diff
@@ -1024,7 +1024,7 @@ Cycle: R10, milestone [R10: Control Loop](https://github.com/0xfauzi/kstrl/miles
 | 4 | 5.9 name safe mode | [#225](https://github.com/0xfauzi/kstrl/issues/225) | `[x]` |
 | 5 | 5.9 adversarial budget: hard mode halts | [#226](https://github.com/0xfauzi/kstrl/issues/226) | `[x]` |
 | 6 | 5.5 dampener | [#227](https://github.com/0xfauzi/kstrl/issues/227) | `[ ]` |
-| 7 | 5.6 flow control | [#228](https://github.com/0xfauzi/kstrl/issues/228) | `[ ]` |
+| 7 | 5.6 flow control | [#228](https://github.com/0xfauzi/kstrl/issues/228) | `[x]` (this PR) |
```
0xfauzi, line 1027: **Ticking this box leaves two shipped teaching artifacts modelling code that no longer exists.**

`docs/lessons/verify/pr-221/admission_gates.py` is an *executable* model of the admission order, and it states the opposite of what this PR ships:

> "Issue #228 places the open-PR bound 'as the last gate, after check_budget' in the `gates` tuple at serve.py:1974-1978. That tuple is evaluated before the inbox cap and the factory lock, so the bound sits before those two, not after them. This script follows the issue's concrete instruction."

Its `ORDER` list and `admit()` both evaluate the bound **before** `inbox_at_cap` and `lock_held`. This PR deliberately evaluates it **after** both, and `TestGateOrderIsCost` is the control that keeps it there - so the script now sweeps 5.6M rows against the wrong rule and reports "holds" on all five claims.

The lesson HTML also still carries `"FlowControl": {"state": "planned", "state_label": "planned", "tracker": "R10.7"}` in its `DETAIL` map, so the system map renders the thing this PR builds as unbuilt. Both need to move with the checkbox.

kstrl/serve.py
```diff
@@ -2064,6 +2079,201 @@ def check_inbox_cap(root_dir: Path) -> Admission:
...
+    being merged or closed, which needs no operator action on the queue.
+    A count that stays unusable is the case a wait alone handles badly,
+    which is what ``streak`` is for; see :class:`OpenPrCountStreak`.
+
+    ``counter`` exists so tests inject a fake without putting a ``gh``
+    on PATH; it is resolved inside the body rather than as a default
+    argument so patching the module-level name reaches it.
+    """
+    if config.max_open_prs == 0:
+        return Admission(allowed=True, reason="open-PR bound disabled")
```
0xfauzi, line 2436: **Neither early-allow path resets the streak, so a streak survives the gate being switched off.**

`max_open_prs == 0` returns here and `create_prs = false` returns a few lines down; both are admissions, and neither calls `streak.record_conclusive()`. So `consecutive` accumulated before the switch persists: set `max_open_prs = 0` at `consecutive = 2`, run for a week, set it back to 1, and the first failure files immediately instead of on the third.

Small, but it is the same asymmetry `record_conclusive` exists to prevent - "the next failure after a recovery is new information" applies at least as strongly to "the next failure after the gate was off".

tests/test_encoding_readers.py
```diff
@@ -105,7 +105,12 @@
     "prd.py": 1,
     "proposals.py": 4,
     "security.py": 1,
-    "serve.py": 4,
+    # Four reads, plus the string literal "open" in the ``gh pr list
+    # --state open`` argv of ``count_open_kstrl_prs`` (#228). The net is
+    # over the TOKEN, not over reads, and a net that decided which
+    # spellings to leave out is a net that can be wrong about what it
+    # left out - so the row moves and says which one is not a read.
+    "serve.py": 5,
```
0xfauzi, line 123: **This pin now counts a token that is not a read, which weakens the census it belongs to.**

The comment is honest about it ("the net is over the TOKEN, not over reads"), but the consequence is that `"serve.py": 5` no longer means "five reads": it means "four reads plus one `--state open` argv literal". A future change that adds a genuine unguarded read in the same commit that drops `--state open` nets zero and the guard stays green - the exact skip-direction failure CLAUDE.md logs eleven instances of.

The cheap fix keeps the census meaning what it says: spell the argv as `"--state=open"` (gh accepts it), or hoist it to a module constant whose name the net can exclude, and put the row back to 4.

tests/helpers/fakegh.py
```diff
@@ -0,0 +1,103 @@
...
+
+import json
+import os
+from pathlib import Path
+from subprocess import CompletedProcess
+
+import pytest
+
+from kstrl.pr import PR_FOOTER_MARKER
+from tests.test_serve_seam import _write_executable
```
0xfauzi: **A shared helper importing a private symbol out of a test module inverts the `tests/helpers/` convention.**

`tests/helpers/` is the established home for shared machinery (`tests/helpers/procs.py`, `tests/helpers/astwalk/`), and nothing else in it depends on a test module. This makes `tests/helpers/fakegh.py` - and therefore `tests/test_flow_control.py` and `tests/test_open_pr_counter.py` - carry `tests/test_serve_seam.py`'s module-level import cost and its `pytestmark`, and it means a rename of a private function in a test file breaks two other suites.

`_write_executable` is four lines and has three consumers now. Move it into `tests/helpers/` (its own module or beside `fakegh.put_gh_on_path`, which is its only remaining caller here) and have `test_serve_seam.py` import it from there.

kstrl/serve.py
```diff
@@ -2064,6 +2079,201 @@ def check_inbox_cap(root_dir: Path) -> Admission:
...
+    args = ["pr", "list", "--state", "open", "--limit", str(limit), "--json", "number,body"]
+    result = run_gh(args, timeout=GH_TIMEOUT, cwd=cwd)
+    if not result.ok:
+        raise RuntimeError(result.error)
+    try:
+        rows = json.loads(result.stdout)
+    except ValueError as exc:
+        raise RuntimeError(f"gh pr list returned unparseable JSON: {exc}") from exc
+    if not isinstance(rows, list):
+        raise RuntimeError(f"gh pr list returned {type(rows).__name__}, expected a list")
```
0xfauzi, line 2158: **This top-level payload check is a copy of `intake_github.parse_issue_list`'s, with the command name changed.**

```python
# here
raise RuntimeError(f"gh pr list returned {type(rows).__name__}, expected a list")
# kstrl/intake_github.py:326
return [], (f"`gh issue list` returned {type(data).__name__}, expected a list")
```

Same decode-then-shape-check pair, same message, same reasoning recorded in both docstrings (#187 F7 there, S1 here). The PR already made the right call once by routing the spawn through `run_gh` rather than re-implementing it; the decode and the top-level shape check are the other half of the same seam. A `decode_gh_json_list(payload, command) -> list[Any]` in `intake_github` next to `run_gh` would leave each caller deciding only its own per-row policy, which is where the two genuinely differ (tolerant per entry there, strict per row here).

CHANGELOG.md
```diff
@@ -162,6 +162,24 @@ stage, runtime feedback, and an earned-autonomy ladder). See
...
 
+- **Breaking:** for daemon users, `ks serve` now stops admitting work
+  while a kstrl-authored pull request is open. A repository with one
+  open kstrl PR will admit nothing until it is merged or closed. The new
+  `[serve] max_open_prs` defaults to 1; set it to 0 to restore the old
+  unbounded behaviour, or raise it. The refusal is a wait rather than a
+  pause: nothing needs resuming, and the next cycle proceeds on its own
+  once the PR is gone. A PR counts as kstrl-authored when its body ENDS
+  with the footer line kstrl writes, so pull requests opened since the
+  footer took its current wording are counted too, and a pull request
```
0xfauzi, line 330: **This sentence states the false-positive direction and drops the false-negative one, which is the correction N1 asked for.**

> "so pull requests opened since the footer took its current wording are counted too, and a pull request that merely quotes the footer in prose is not"

"counted too" reads as an inclusion note. The behaviour that matters to a reader upgrading is the *exclusion*: a PR kstrl opened **before** the ralph -> kstrl rename (`bd19f2b`) is **not** counted, and neither is one whose body a human has edited. `docs/continuous-intake.md` states both directions correctly under "False negative" / "False positive"; the CHANGELOG entry is the one an upgrading operator reads first and it says the opposite of the interesting half.

PR #357, opened 2026-09-06
feat(sense): baseline and regression comparison for ks sense (#227)

### LINKED ISSUE #227: R10.6: Dampener: `ks sense` baseline in version control plus an advisory PR regression report
Part of #235 (R10: Control Loop tracking issue). Item R10.6.

Design section: 5.5. Sizing: M (one PR). Dependencies: **R10.1** (`ks sense --json`) must be merged first; this issue extends that command.

## Why

kstrl has no dampener. The mechanism: run a full deterministic scan on main, sort the results deterministically, track the count in version control, and on every new pull request check whether the branch added any violations. It is a disturbance dampener: it stops concurrent work from undoing the loop's progress while the loop runs. It ships advisory by default (it never fails a teammate's PR) with a documented path to blocking.

kstrl's disturbances are real and named: concurrent human commits, dependency bumps, a base branch that moves under a component. Nothing stops the measured property regressing while the factory improves it. The closest existing thing is fixture snapshot regression (`kstrl/fixtures.py:582-597`), which covers approved fixtures only and only inside a factory run.

The dampener is the same sensor as R10.1, pointed at a diff instead of a tree, plus a stored baseline.

### 1. A baseline file, in version control

Path: `scripts/kstrl/sense-baseline.json` (the `scripts/kstrl/` directory is the versioned per-project kstrl config home, alongside `prompt.md` and `prd.json`). Format:

```json
{
  "schema_version": 1,
  "generated_at": "2026-08-24T10:00:00Z",
  "base_ref": "<git rev-parse HEAD of the tree measured>",
  "passed": false,
  "signatures": {"linter:E501": 12, "typecheck:arg-type": 3}
}
```

`signatures` is a count per structured failure signature, produced by the existing signature scheme `kstrl.evolution.signatures_from_verification` (`kstrl/evolution.py:332`), so the dampener and the evolution journal speak the same vocabulary. Note `_MAX_SIGNATURES_PER_CHECK = 5` (`evolution.py:292`) caps how many signatures one check contributes to the journal; for the baseline, count **all** signatures (add a keyword argument `limit: int | None = None` to `signatures_from_verification` and pass `None`; the default keeps journal behaviour byte-identical).

Keys are sorted on write so the file diffs cleanly in git.

### 2. `ks sense --write-baseline [PATH]`

Runs the sensors (as R10.1 does) and writes the baseline to `PATH` (default `scripts/kstrl/sense-baseline.json` under `--root`). Prints one line: `baseline written: <path> (<n> signatures, <m> total findings)`. Exit code follows the sense result (0 pass, 1 fail, 2 error); a red baseline is allowed and expected on brownfield repos. Refuses to overwrite an existing file unless `--force`, printing `error: <path> exists; pass --force to replace it` (exit 2).

### 3. `ks sense --compare-baseline [PATH]`

Runs the sensors on the tree, computes the current signature counts, and compares to the baseline:

- **new**: signatures present now and absent in the baseline
- **increased**: signatures whose count now exceeds the baseline count
- **fixed**: signatures in the baseline absent now (reported so improvement is visible; never affects the verdict)

Output (human): a heading `sense regression report vs <baseline path> (<base_ref short sha>)`, then three lists. When new and increased are both empty: `no regression`.

`--format markdown` renders the same content as GitHub-flavoured markdown suitable for a PR comment, with a first line `<!-- kstrl-sense-dampener -->` so a workflow can find and update its own earlier comment.

`--json` adds `{"baseline": {...}, "current": {...}, "new": {...}, "increased": {...}, "fixed": {...}, "regressed": bool}` to the R10.1 document.

Exit code: **0 regardless of regression** by default (advisory). `--fail-on-regression` makes exit 1 when `regressed` is true. This is the graduation switch: advisory first, blocking by explicit choice.

If the baseline file is missing: print `error: no baseline at <path>; run ks sense --write-baseline first`, exit 2.

### 4. Dogfood: a workflow on this repository

Add `.github/workflows/sense-dampener.yml`:

- Trigger: `pull_request` (opened, synchronize, reopened).
- Steps: checkout, `uv sync`, `uv run ks sense --compare-baseline --format markdown > /tmp/sense.md || true`, then post or update a PR comment containing `/tmp/sense.md` via `gh` (find an existing comment by the `<!-- kstrl-sense-dampener -->` marker with `gh api repos/{repo}/issues/{n}/comments`; update it if found, create it otherwise).
- The job **never fails** on regression (no `--fail-on-regression`); it fails only if the command itself errors (exit 2), because that means the sensor could not run.
- Permissions: `pull-requests: write`, `contents: read`.

Generate this repository's baseline with `uv run ks sense --write-baseline` on `main` and commit it in the same PR.

### 5. Docs for target repositories

New section in `docs/runbook.md` or a new `docs/dampener.md` (your choice; link it from README's hand-written area): how to write a baseline, how to add the workflow to a target repo (copy the YAML), what advisory means, and how to graduate: (a) start advisory, (b) once the comments have been accurate on several PRs, add `--fail-on-regression`, (c) refresh the baseline deliberately with `--write-baseline --force` after an intentional change, in its own commit so the diff shows what moved.

CHANGELOG `### Added`. Regenerate README (new options).

## Out of scope

- Using the baseline inside the factory pipeline (a "fail only on new findings" verify mode). That is the graduation follow-up and needs its own issue once the standalone dampener has produced comments on real PRs.
- Any LLM-based comparison.

## Tests to write

`tests/test_sense_dampener.py`:

1. `test_write_baseline_round_trip`: write, read, keys sorted, `schema_version == 1`, `base_ref` is the repo HEAD sha.
2. `test_write_baseline_refuses_overwrite_without_force`: exit 2, message names the path; with `--force` succeeds.
3. `test_compare_no_regression`: baseline equals current; `regressed` false; human output contains `no regression`; exit 0.
4. `test_compare_detects_new_signature`: introduce a lint violation not in the baseline; `new` contains it; `regressed` true; exit 0 without the flag; exit 1 with `--fail-on-regression`.
5. `test_compare_detects_increased_count`: baseline `linter:E501: 1`, current 2; `increased` has the key with both counts.
6. `test_compare_reports_fixed_without_regression`: baseline has a signature the current tree lacks; `fixed` lists it; `regressed` false.
7. `test_compare_missing_baseline_exit_2`.
8. `test_markdown_format_has_marker`: first line is `<!-- kstrl-sense-dampener -->`.
9. `test_signatures_from_verification_limit_none_counts_all`: a check with 7 distinct codes yields 7 signatures with `limit=None` and 5 with the default (proving journal behaviour is unchanged).
10. Workflow YAML parses: `python -c "import yaml,sys; yaml.safe_load(open('.github/workflows/sense-dampener.yml'))"` in a test, or as a CI step; and `gh workflow view sense-dampener.yml` succeeds after merge (paste output in the PR).

## Acceptance (all must hold)

- [ ] `uv run ks sense --write-baseline` on this repository writes `scripts/kstrl/sense-baseline.json`; commit it. Paste the signature count in the PR body.
- [ ] `uv run ks sense --compare-baseline` on the same tree prints `no regression` and exits 0.
- [ ] Add a deliberate `E501` violation in a scratch file, run compare: the signature appears under `new`; exit 0; with `--fail-on-regression` exit 1. Revert the scratch file.
- [ ] The workflow runs on this PR itself (it triggers on `pull_request`) and posts one comment carrying the marker. Link the comment in the PR body. If the workflow cannot run on its own introducing PR for a permissions reason, say so and show it running on the next PR.
- [ ] Tests 1 through 10 pass; baseline suite passes; gen_docs check passes.
- [ ] The journal signature limit is unchanged for existing callers (test 9).

## Graduation (recorded, not built here)

Switch this repository's workflow to `--fail-on-regression` after the operator has seen the comment be right on real PRs. Then file the in-factory "new findings only" verify mode.

0xfauzi:
## Summary

R10.6, the dampener: a `ks sense` measurement recorded in version control, and a report of what a branch added to it.

`ks sense --write-baseline [PATH]` runs the sensors and records the structured failure signatures to `scripts/kstrl/sense-baseline.json`. `ks sense --compare-baseline [PATH]` runs them again and buckets the result into **new**, **increased**, **stopped measuring**, **fixed** and **unmeasured**. It is advisory: it exits 0 whether or not it found a regression, until `--fail-on-regression` is passed. `--format markdown` renders the report for a pull-request comment, and `.github/workflows/sense-dampener.yml` posts one on every pull request to this repository.

The signature vocabulary is the evolution journal's, from `kstrl.evolution.signature_counts_from_verification`, so the dampener and the journal cannot disagree about what a failure is called and the `"<check>:<code>"` spelling stays in one module.

The design question the issue does not ask is what makes `fixed` a claim rather than a guess. `new`, `increased` and `stopped measuring` FLAG, so they may over-match and the cost is a comment somebody reads. `fixed` CLEARS, and an over-matching clear deletes the mechanism with nothing failing. So a baseline signature that is absent now lands in `fixed` only when the check that produced it measured something in the current run, and in `unmeasured` otherwise.

## What round 1 of review changed

Four blockers, nine should-fix and seven nits. Every one is addressed or declined with a reason below; the four largest change the shape of the feature.

**A sensor that stopped measuring was invisible, and on this repository's own baseline that was exit 0.** `compare` bucketed only signatures, so a check that measured on the baseline side and nothing now contributed to no bucket at all. Reproduced against the committed baseline, which records `"signatures": {}`: on a branch where the test suite no longer finishes there is no signature anywhere to bucket, so the report read `no regression`, `regressed` was False and `exit_code_for(fail_on_regression=True)` returned 0. This branch had strictly REMOVED a signal `main` had, in the flagging direction, for the most important thing a pull-request check can catch. `compare` now carries a fifth bucket keyed on the CHECK rather than on a signature, with the reason the check gave, folded into `regressed` and rendered in both reports and the JSON.

**All six `measured=False` sites could be deleted with the suite staying green.** The review removed every one in a single edit and measured 5811 passed, rc 0 (its number, relayed). The field is now swept across every site that reports having measured nothing, and pinned by five censuses closed by construction plus a behavioural test per site. See "The measurement sweep".

**`read_baseline` caught `ValueError` around `json.loads`.** A baseline of 200000 nested arrays raises `RecursionError`, a `RuntimeError`, so a document this function is documented to refuse with exit 2 escaped as a traceback and exit 1. Now: all the I/O outside the guarded block, `except Exception` last, and only the two causes that can be named reported individually.

**The documented graduation to blocking did not block.** `docs/dampener.md` step 2 said adding `--fail-on-regression` was the whole change. The sense step runs under `set +e` and restores `set -e`, so its own status is always 0 and the exit code reaches the job only through `GITHUB_OUTPUT`; the only failing step keyed on `rc == '2'`. `bash -e -c 'set +e; false; echo "rc=$?"; set -e'` prints `rc=1` and exits **0**. An adopter following the instructions got a green job on every regression. The failing step now keys on `rc != '0'`, and the doc says why that detail is load-bearing.

The nine should-fix and the seven nits, one line each, so nothing is left to be inferred from the diff:

| item | what happened |
|---|---|
| S1 `timeout-minutes: 20` under a 1800s sensor budget | 40 minutes, pinned as a relationship rather than a number. Round 2 found the relationship was one gate's budget and the comment said three: it is 160 now, and DERIVED. See S5 below |
| S2 the comment step posts an empty body on any nonzero exit | gated on `[ -s sense.md ]`, and the file rather than the exit code is the gate |
| S3 `gh`'s status swallowed by `head`, so a failed lookup posts a DUPLICATE | `set -o pipefail`, the lookup captured to a file and checked on its own status, refusing rather than posting |
| S4 a vacuous pass is recorded as measured, so it CLEARS | `diff_scope` with no allowed paths and `bad_patterns` over an empty diff both measure nothing |
| S5 the doc's linter sentence was the design and not the code | the "command could not run" case measures nothing. Round 2 found that true only for a bare binary and false for the `uv run` shape this repository ships. See B1 below |
| S6 eight sites named in the handoff, plus two one level removed | all swept, and the `FixtureResult` half folded into its row with `all`. Round 2 found the CENSUS closed over `CheckResult` only; it covers both types now. See B2 below |
| S7 the baseline carries no identity of what it is OF | `verify_digest` (a REFUSAL on mismatch) and `project` (a NOTE), with `unmeasured_checks` gaining a reason per hole |
| S8 an explicit relative path resolves against the process directory | both forms resolve under `--root`, stated in `--help` |
| S9 "the one place the spelling is written" | corrected to one module, naming the three functions |
| N1 "seven refused flag combinations" | six; the body says six |
| N2 `_signatures_field` accepts a count of 0 | refused, `< 1` |
| N3 `frozen=True` with `Mapping` fields makes `__hash__` raise | declined: latent, nothing hashes them |
| N4 `--ui` and `--no-color` reach nothing | declined: `--json` already ignores them on `main` |
| N5 two module-public names with no caller outside their module | declined: style |
| N6 the live comment prints an absolute runner path | declined, handoff 3 |
| N7 `head -n 1` would SIGPIPE under an explicit `pipefail` shell | fixed: the lookup no longer pipes, and the step sets `pipefail` for any pipeline a later edit adds |

## What round 2 of review changed

Two blockers, eight should-fix, five nits. Both blockers are the same defect class as round 1's, arriving one level out from where the sweep had stopped, and both were reproduced end to end before anything was changed.

**`measured` on a failing gate was decided from the exit code, and the exit code belongs to the launcher.** Round 1 read it as `returncode not in {126, 127}`, which is the POSIX shell's vocabulary for a command word it could not run. This repository's gate commands are `uv run pytest`, `uv run mypy .` and `uv run ruff check .`, and uv spawns the child itself and reports its OWN status, which is 2. So the rule fired for the shape nobody runs and never for the shape everybody runs: `check_linter(root, "uv run <missing> check .")` returned `measured=True` and the comparison reported `fixed={'linter:E501': 12, 'linter:F401': 3}`, byte for byte what `docs/dampener.md` said was impossible. Adding 2 to the set is not the fix either, since pytest, mypy and ruff all use 2 for their own errors.

`measured` is now POSITIVE evidence, and the PARSER gives it. `ParsedOutput.recognised` is set by each of the six parsers at the point its own pattern matched: a diagnostic in the tool's format, or the tool's own FAILURE footer. `_failed_gate_result` reads it and is the only place a gate's measurement is decided, so the three call sites pass no opinion at all and `COMMAND_NOT_RUN_EXIT_CODES` is deleted. Measured on the head, all three gates:

| command | exit | measured |
|---|---|---|
| `uv run <missing> check .` | 2 | False |
| `<missing> check .` | 127 | False |
| exits 1 printing nothing | 1 | False |
| exits 1 printing a traceback | 1 | False |
| exits 1 printing the tool's own finding | 1 | **True** |
| exits 0 | 0 | **True** |

Two decisions inside that, stated rather than buried. **A PASSING gate still measures on its status, not on recognition.** Recognition as implemented is a failure report by construction: on a chained `uv run pytest && npm test` a passing pytest footer says nothing about the half that failed, so counting it would clear a baseline on the strength of the part that worked. That makes recognition always False on a clean run, so requiring it would mark every clean gate unmeasured and put every sensor permanently in the dark, tsc and eslint included, since neither prints anything on success. Exit 0 is evidence of its own: a command that never started returns uv's 2 or the shell's 127, never 0. **Recognition is ORed across every parser the gate tried**, including on the path that returns the primary's result, or an eslint footer the secondary understood would be dropped with the rest of its result.

**The census was closed by construction over `CheckResult` and open over `FixtureResult`, the sibling type this pull request gave the same field.** A new environment-failure row keeping the default was planted and the suite stayed green. The four nets now key on a SET of constructor names and read arguments by POSITION as well as by keyword, because `kstrl/fixtures.py` writes `FixtureResult(fixture, False, ...)` eighteen times and a keyword-only walk sees no `passed` at those sites. The pinned field order is checked against the real dataclasses, so a reorder fails loudly instead of making the walk read the wrong argument. `_fixture_file_text`'s `OSError` and `UnicodeDecodeError` rows are now `measured=False`: the caller reached them having SEEN the file exist, and its `contains` expectations are substring tests it could not run. That is the same pair, for the same reason, as `verify._self_critique_text`.

The eight should-fix, and where two of them part company:

| item | what happened |
|---|---|
| S1 `check_diff_scope` measured over an empty diff while `check_bad_patterns` did not | both report `measured=False` with the SAME message over an empty diff, from one constant: the dampener turns a row's message into the reason a check is unmeasured, and two spellings of one fact reach the operator as two holes |
| S2 `measured=bool(py_files)` counted the files the diff NAMED | it counts the files it opened, and the message says that number too. On a deletion-only commit that was "Scanned 3 Python files, no issues" with `measured=True` from a check that opened none |
| S3 `get_origin_slug` let a missing git escape as a traceback | `except (subprocess.TimeoutExpired, OSError)` on the three functions `ks sense` calls, which is the breadth `resolve_base_ref` already used. 21 handlers in `kstrl/git.py` still catch `TimeoutExpired` alone; they are not on this path, and sweeping them needs its own guard, so the count is recorded rather than half swept |
| S4 the workflow compared against the pull request's OWN baseline | the step reads it out of `origin/<base>` with `git show` into a temporary file and passes that path, so a branch cannot supply the yardstick it is judged by. A base ref that does not resolve is a red step, not a fallback; a base ref with no baseline is the bootstrap case, rc 0 and a report saying so |
| S5 `timeout-minutes: 40` sat below the worst case its own comment stated | 160. Five subprocesses in one sense run are each handed the full budget, the three gates and the two halves of the dead-code phase, and the multiplier is COUNTED out of `kstrl/verify.py` by the test with a control that fails if the counter stops matching |
| S6 nothing in the suite read this repository's own committed baseline | `tests/test_sense_committed_baseline.py`: the digest recomputed from this checkout at the workflow's own `KSTRL_TIMEOUT_VERIFY`, the schema version, a control that the digest varies with its inputs, and two the digest cannot see (every check the baseline names is still in the enrolled vocabulary, and its unmeasured reasons still match its unmeasured checks) |
| S7 a reason was interpolated raw into a markdown table cell | newlines collapse and pipes are escaped. `render_human` deliberately unchanged, because it prints one reason per line with no cell walls and the same string has to arrive there intact; a test says so |
| S8 the advisory footer was unconditional | the renderer takes the mode. The comment on a pull request the report had just failed used to deny that it could |

**S1 and S2 agree over an empty diff and part company on a deletion-only commit, and that is deliberate.** `bad_patterns` measured nothing there: it has to open a file to say anything and it opened none of the three. `check_diff_scope` DID measure: it applied the allowlist to three real paths, which is the whole of what scope means, and a deleted file outside the allowlist is a violation it must still catch. The review's probe list asked for `measured=False` from both. Making `diff_scope` unmeasured there would suppress a real measurement on the CLEARING side, so it is the one item in this round that is answered rather than carried out, and the verification probe asserts the divergence with the reason in the code.

**Three nits declined, with the reason rather than a silence.**

- **N1, the report never names the current run's own holes.** A feature gap, not a false statement: `current.unmeasured_checks` reaches `--json` and, through the set difference, `stopped_measuring`. A check that measured nothing now and was not in the baseline's `measured_checks` appears in neither renderer. Recorded for the owner.
- **N2, `check_bad_patterns` reads every changed `.py` with an unguarded `read_text`.** Pre-existing and byte-unchanged here. Guarding it means inventing a finding string whose later disappearance would read as a fix, which is the clearing direction this whole pull request is about, so it is a decision for its own change rather than a line added here.
- **N3, `_str_field(raw, "project")` accepts `""`.** Latent: `path.name` is `""` only for a root path, which `ks sense` refuses earlier.

**N4** (`README.md` omitted the `stopped measuring` bucket) and **N5** (a test name that overstated what it checked) are fixed.

**One defect CI found that no local run could.** `test_the_identity_reads_still_answer_when_git_is_present` built a temp repository and committed in it without setting `user.name` or `user.email`, so the commit took git's IMPLICIT identity from the account and the host. That exists on every machine this branch was written on and does not on a GitHub runner, where the gecos name is empty and git refuses with `empty ident name`. One test, red on CI and green everywhere else. Reproduced locally before fixing it rather than read off the log: a global config carrying `user.useConfigOnly = true` makes git refuse to guess exactly as the runner does, and under it the file is 1 failed and 37 passed; with the two config lines it is 38 passed; removing them again puts the failure back. The two sibling repositories this pull request builds already set both fields, so this was the single site that inherited the environment instead of stating it, and the whole suite was then re-run under that same identity-free environment to show no other test depends on one.

## Behaviour changes

**New on `ks sense`.** `--write-baseline [PATH]`, `--compare-baseline [PATH]`, `--force`, `--fail-on-regression`, `--format [human|markdown]`. Plain `ks sense` and `ks sense --json` are unchanged: with no dampener flag the command takes the path it took before. A RELATIVE `PATH` on either optional-value flag resolves under `--root`, exactly as the bare flag's default does; before, passing the exact path `--help` advertises as the default together with `--root` read a different file and reported "no baseline at ..." for a file that exists.

**Exit codes, complete.**

| invocation | condition | exit |
|---|---|---|
| `ks sense` | unchanged | 0 pass / 1 fail / 2 cannot measure |
| `--write-baseline` | written, sense passed | 0 |
| `--write-baseline` | written, sense failed | 1 |
| `--write-baseline` | file exists, no `--force` | 2 |
| `--compare-baseline` | no regression | 0 |
| `--compare-baseline` | regression, advisory | 0 |
| `--compare-baseline --fail-on-regression` | regression, including a sensor that stopped measuring | 1 |
| `--compare-baseline` | baseline missing, unreadable, malformed or at the wrong `schema_version` | 2 |
| `--compare-baseline` | baseline measured with different verify commands or a different timeout | 2 |
| either | a refused flag combination | 2 |

`--compare-baseline`'s exit follows the COMPARISON and never `result.passed`. A red tree is the normal state for the brownfield repository the dampener exists for.

**Six refused flag combinations, exit 2, each naming both flags.** Round 1 found the earlier claim of seven wrong; there are six `raise DampenerUsage` sites and six were exercised end to end. `--write-baseline` with `--compare-baseline`; `--force` without `--write-baseline`; `--fail-on-regression` without `--compare-baseline`; `--format` without `--compare-baseline`; `--json` with an explicit `--format`; `--json` with `--write-baseline`.

**The baseline carries the identity of what it is OF, and the reader checks it.** Two fields, and the difference between them is deliberate. `verify_digest` is a SHA-256 of the three resolved verify commands and the subprocess timeout, and a mismatch is a REFUSAL naming both digests: `docs/dampener.md` already said a baseline and a comparison measured at different timeouts are not a comparison, and the only mechanism behind that sentence was a literal `1800` typed into a workflow file. `project` is `owner/repo` from `origin`, and a mismatch is a NOTE, because a baseline copied between two checkouts of one project is legitimate and only a person can tell that from one copied between two projects. `owner/repo` and not the directory name: every kstrl lane measures inside a git worktree named after an issue number, so the directory name would record `227` and make every ordinary checkout report a mismatch that means nothing. `unmeasured_checks` gained the reason for each hole, and the reader refuses a document whose two key sets disagree.

**`CheckResult.measured: bool = True`**, a new field on an existing dataclass. `passed` still decides the verdict, the report table is untouched and the `ks sense --json` check objects are byte-unchanged, pinned by a test. What it buys is that a signature's disappearance can be told apart from the sensor's.

**No `SENSE_SCHEMA_VERSION` bump.** #335 took it to 3 while this branch was open; the `dampener` key is absent exactly when `--compare-baseline` was not given, it restates nothing, and no existing key changes meaning.

## The measurement sweep

One defect class, one pull request. On the merged tree `e81f608`: **39 `CheckResult` constructions, 18 of them stating what they measured, 9 failing rows keeping the default, and 3 places that READ a `CheckResult.measured` or a `FixtureResult.measured`.** The review's own census of the head it read was 41 constructions with 6 marked, which this census reproduces exactly at that commit. The total moved because #335 landed on `main` mid-review: `kstrl/verify.py` went from six `dead_code` rows to three, the rest becoming `NotMeasured` gaps, which is 41 to 38 and nothing to do with this branch.

Every site that reports a failure or a pass having measured nothing now says so: `self_critique`'s two read handlers, `prd_stories`' load failure, `policy_envelope`'s two infrastructure refusals, `test_adequacy`'s, the fixtures check's two PRD refusals, `_fixture_file_text`'s `OSError` and `UnicodeDecodeError`, the two vacuous passes, and fixture timeouts and launch failures through a new `FixtureResult.measured` folded with `all`. The three gates are the exception and they are decided differently after round 2: a gate has to READ its tool's output to know whether it looked at anything, so its row carries `ParsedOutput.recognised` and nothing about the exit status.

Two of them were measured CLEARING real findings on this repository's own committed baseline. `check_linter` against a binary that is not on PATH returned `passed=False measured=True "Linter failed (exit code 127)"` and the comparison reported `fixed={'linter:E501': 12, 'linter:F401': 3}`, so uninstalling a linter read as fixing every one of its findings, which is exactly what `docs/dampener.md` claimed the rule prevented. It now reports `fixed={}`, `unmeasured={'linter:E501': 12, 'linter:F401': 3}` and `stopped_measuring={'linter': ...}`. `check_diff_scope(cwd, base, allowed_paths=None)` returned `passed=True measured=True "No scope constraints"`, which `ks sense` takes on every run with no `--allowed-path`, and `diff_scope` was in the committed baseline's `measured_checks`.

Two sites are decided rather than inherited, and either could be argued the other way. `scope_unreadable` was going to be exempted; it is marked instead, because the new stopped-measuring bucket would otherwise report a REPAIRED harness as a sensor that went dark. A malformed fixture DEFINITION stays `measured=True`: it is a stable property of the PRD and its disappearance is a real fix. The line the sweep draws is the ENVIRONMENT failing, not the artifact.

**The guard is closed by construction, not a ledger.** `tests/test_check_result_measurement.py` runs five censuses over `kstrl/`, keyed by module and enclosing function rather than by line number so unrelated edits do not churn them.

1. Every `CheckResult` construction. A new check, a new failure branch or a new vacuous pass has to appear here first, whatever it then says about `measured`.
2. Every `measured` argument with its expression VERBATIM, so weakening `measured=bool(py_files)` to `measured=True` is a moved row rather than an unchanged count.
3. Every failing row that keeps the default, with one line each saying what it measured.
4. Every call to the three gates' shared `_failed_gate_result`, because without it a gate could stop passing the argument and no other dict would move.
5. Every READ of a `.measured`, which is the direction the other four cannot see: `measured` decides what a COMPARISON may call fixed, and the moment a verdict consults it, `all(c.passed for c in checks if c.measured)` turns a timed-out test suite into a passing run.

After round 2 the five censuses key on a SET of constructor names, so `FixtureResult` is closed by construction too, and they read arguments by POSITION as well as by keyword because `kstrl/fixtures.py` writes `FixtureResult(fixture, False, ...)` eighteen times. The pinned field order is checked against the real dataclasses, so a reorder fails loudly rather than making the walk read the wrong argument.

`tests/test_check_result_measurement_behaviour.py` is the other half: every row is driven from the real function, with PATH stubs where a tool has to be absent, plus controls that DID measure so that marking everything unmeasured fails the file rather than passing it. `tests/test_gate_measurement_behaviour.py` is the three gates, split out because their answer comes from a parser rather than from their own control flow, and `tests/helpers/measurement.py` holds the two assertions both files need, once, because the second copy is the one nobody would notice weakening.

## Tested

`origin/main` moved TWICE while this round was finishing, to `037f0e1` (#360) and then to `aad32e8` (#359), and both were merged rather than left for the merge queue. Every number below is measured on the final merged head `768537b` unless it names another commit, and where a measurement was taken at an earlier head in this round the commit is given so the two can be told apart. The whole set was re-run after each merge: the guard census on both sides, the full suite, the gates, the ratchets against the base ref, the 114 mutation plants and the probe script. Bytecode deleted first on every run, `PYTHONDONTWRITEBYTECODE=1`, `-p no:cacheprovider`, each long run in its own process group bounded by a timeout and killed by `os.killpg`, so a hang is recorded as a hang rather than waited on. The full suite and the gates ran in detached worktrees at that tree so the mutation sweep could hold the branch worktree; the sweep, the census and the probe script ran in the branch worktree itself.

**Gates on the MERGED tree.** `origin/main` moved to `037f0e1` (#360) while this round was in progress, so this is a real merge and not the fast-forward the earlier rounds reported. Every number below is measured on the merged head, the guard census was run on both sides of the merge, and the four-test delta it found is attributed.

| gate | result |
|---|---|
| `pytest tests/ -q -p no:cacheprovider` | 6374 passed, 37 skipped, 45 xfailed in 408.21s, rc 0 (6330 / 37 / 42 at `817446c`, before #359 merged) |
| `mypy kstrl/ --strict` | Success: no issues found in 137 source files |
| `ruff check kstrl/ tests/` | All checks passed! |
| `ruff format --check` on the changed `.py` files | 29 files already formatted |
| `gen_docs.py --check` | README.md generated sections are current |
| `cyclomatic_ratchet.py`, `file_length_ratchet.py`, `check_agent_docs_sync.py`, `check_no_mocks.py` | all four rc 0, the whole change staged on `aad32e8`, and rc 0 again on `037f0e1` before that |
| actionlint, zizmor and shellcheck on the workflow | pre-commit on the workflow passes actionlint, zizmor, the workflow schema check and the timeout-minutes hook; shellcheck 0.10.0 on all six `run:` blocks, 0 findings |
| CI on the pushed head | `dampener`, `lint`, `spine` and GitGuardian pass; the `test` job is what found the git-identity defect below, and it is green on the head after the fix |
| `pytest tests/` with NO ambient git identity | 6330 passed, 37 skipped, 42 xfailed in 389.71s, rc 0 at `817446c`, under a git config that refuses to guess an identity |
| em dashes and emoji in the diff | 6228 added lines against `aad32e8`, 0 em dashes, 0 non-ASCII characters, 0 codepoints above U+2500 |
| the reviewer's own probes re-run against this head | 26 probes, 0 failures, rc 0, run again at every head this round produced |

A repository-wide `ruff format --check .` reports 444 files already formatted and 4 it would reformat: `docs/spec-harness-engineering.md` and three `scripts/precommit/*.py`. Measured again on `origin/main` `037f0e1` in a tree of its own: the same 4 out of 431, so none is caused here, and none is touched here. The pre-commit hook excludes `scripts/precommit/` by design (those files stay byte-identical across repositories) and runs on Python only, so it trips on none of them.

**Complexity.** `sense` at cyclomatic 7, `_sense_dampener_report` 6, `_sense_report` 5, `_sense_verify_digest` 3, `_sense_document` 2. `kstrl/dampener.py` and `kstrl/dampener_report.py` are both clean under `ruff --select C901 --config lint.mccabe.max-complexity=10`, the highest being `compare` at 9. The cognitive ratchet, complexipy 7.0.1 with `--staged --max-complexity-allowed 15`, the way pre-commit runs it, is rc 0 with the whole change staged on `origin/main`: 207 new functions, 2 improved, and exactly one regression, `_sense_document` at cognitive 4 to 5.

**Three files were split because the ratchet is a gate rather than advice.** `kstrl/dampener.py` reached 927 lines; the rendering moved to `kstrl/dampener_report.py`, leaving 730 and 218. `tests/test_sense_dampener.py` reached 855; the on-disk artifact moved to `tests/test_sense_dampener_baseline.py` and the workflow tests to `tests/test_sense_dampener_workflow.py`, leaving 521, 360 and 327.

The third was found by running the ratchet the way it really compares, against `origin/main` with the whole change staged, which this branch had not done. `kstrl/fixtures.py` is 792 lines on `1775301` and on `037f0e1`, and was 838 here: CROSSED, exit 1. Per commit it never fired, because each commit compares against the branch's own HEAD and the file crossed 800 in round 1, after which every later commit saw it as already over. That is the class CLAUDE.md records under merging into a moving main, a failure attached to no single hunk and present in neither the branch's last commit nor `main`. `save_snapshot` and `check_snapshot_regression` moved to `kstrl/fixtures_snapshot.py` byte for byte, leaving 752 and 111, and the new module imports the two dataclasses back under `TYPE_CHECKING` only because a run-time import would be a cycle and neither function constructs either type. All four `scripts/precommit/` checks are rc 0 against `origin/main` `037f0e1` with the change staged after it.

That split moved one read between modules and turned three guards red, which is the by-file pin the same class of failure lives in: `EXPECTED_READ_SPELLINGS` and the cleared-reads inventory in `tests/test_encoding_readers.py` are keyed by module, and `tests/test_encoding_walk.py` reads its population from the first of them. The pins are re-derived by RUNNING the guards rather than edited to agree with the tree: `fixtures.py` drops from 3 read spellings to 2 and `fixtures_snapshot.py` carries the one that left. It is the reason a file split is not a free refactor here, and it was caught by the guard census being run rather than read.

**Mutations: 114 planted on the MERGED head, 114 caught, 0 still green, 0 hung, 0 could not plant, at every one of the three heads.** The sweep was run three times on merged trees, at `e81f608`, at `817446c` after the git-identity fix and at `768537b` after the second merge, rather than relayed from the pre-merge run, and re-run on this round's tree rather than the previous round's, because this round changed the same code and nine of the earlier anchors had MOVED. Each planted alone, its anchor asserted to appear an exact number of times before anything is written, `__pycache__` purged before every run with `PYTHONDONTWRITEBYTECODE=1`, bounded at 420s in its own process group and killed by `os.killpg`, tree restored from git and `git status --porcelain` asserted empty afterwards.

| set | plants | caught | still green | hung | could not plant |
|---|---|---|---|---|---|
| round 2's own fixes, re-planted | 36 | 36 | 0 | 0 | 0 |
| round 1's fixes, re-planted and retargeted | 43 | 43 | 0 | 0 | 0 |
| round 1's review's own mutations, its numbering | 14 | 14 | 0 | 0 | 0 |
| one `measured=False` line deleted at a time | 21 | 21 | 0 | 0 | 0 |
| **total** | **114** | **114** | **0** | **0** | **0** |

- **36** are round 2's own fixes put back as the defect each one fixed: the gate helper hard-coding `measured=True` or going back to the exit-code rule, each of the six parsers recognising anything, the union and the primary path dropping recognition, the census keying on one constructor again or reading keywords only, `check_diff_scope` measuring over an empty diff, `bad_patterns` counting named files, `get_origin_slug` narrowed back to `TimeoutExpired`, the workflow reading the branch's own baseline or falling back on an unresolvable base ref, `timeout-minutes` back to 40, the reason cell unescaped, the advisory footer unconditional again.
- **43** re-plant round 1's fixes, retargeted where this round moved their anchors.
- **14** are the review's own mutations that the first fix pass did not re-plant, carrying its numbering.
- **21** delete ONE `measured=False` line at a time, which is the question an all-at-once deletion cannot answer. Twelve in `kstrl/verify.py`, nine in `kstrl/fixtures.py` and none in `kstrl/fixtures_snapshot.py`, which the inventory walks anyway so that the split cannot hide a site from it; the count rose from eighteen because round 2 marked three more rows.

**Nine anchors had moved and each was RETARGETED rather than folded into a pass** (F2, F4, F5, F7, F14, F25, F28, X1 and S5b). A mutation that silently fails to plant is the same false all-clear as one aimed at the wrong file, so COULD-NOT-PLANT is its own outcome and it has fired: `S5b` reported it once against the assertion this round replaced, and was retargeted onto the replacement. On the final sweep, on the merged tree, it fired zero times and every anchor was found at the count it asserts.

**Two plants were STILL GREEN and both are closed rather than reported.** Both were guards that did not catch a weakening of THEMSELVES, which is this repository's most-repeated defect class.

`B1d` widened the pytest parser's recognition from a nonzero failure COUNT to `bool(raw_summary)`, so a PASSING footer counted as recognition, and all 184 tests in the target set stayed green. Every other case in the gate file either has a diagnostic to parse or prints nothing a summary pattern matches, so none of them separates the two rules. `tests/test_gate_measurement_behaviour.py` now drives the one shape that does: a test gate that exits 1 having printed nothing but a passing footer, which is `uv run pytest && npm test` with pytest green and the half after it red. One case per test parser, and a plant per parser to go with them (`B1d`, `B1j`), because `parse_vitest_output` carries the same claim. The second assertion in each is the control: the footer has to have reached `raw_summary`, or the case would be measuring output no parser matched and would hold with the recognition rule deleted.

`S5b` pinned the workflow cap against ONE gate's budget again, which demands 40 minutes, and the shipped 160 satisfied that too because the assertion was one-sided. The cap is DERIVED now, `math.ceil((spenders * verify_timeout + INSTALL_HEADROOM_SECONDS) / 60)`, so the arithmetic is the only thing that sets the number: 160 today, 40 under the mutation, and either half moving fails.

Red first for both, in that order: the new assertion fails with the plant applied and passes with it removed, measured as 186 collected and green on the head, 1 failed and 185 passed under each plant.

One limit stated rather than left to be noticed. Deleting an assertion outright is caught by no assertion, so no plant is offered for it; what the two new guards close is the WEAKENING direction, where the value or the arithmetic moves and the guard used to agree anyway.

Where round 1's review found six sites, checked rather than assumed. Three were the gate timeouts, `verify.py:1215, 1255, 1295` at `7f87b00`; they moved again this round and each is individually red at its current line, which the per-site set locates by reading the file rather than by carrying a number. The other three were `dead_code` rows at `2279, 2286, 2297`, which #335 replaced on `main` with `NotMeasured` gaps before this branch merged it, so they are no longer `CheckResult` sites at all; the gap path they became has its own behavioural test, that a gap lands where an unmeasured row does.

**Round 1's own four still-green plants, kept here because the fixes they forced are in this diff.**

`M8`, moving the digest refusal from above `run_mechanical_verification` to below it, survived 152 tests. The placement was stated in the function's own docstring ("refused in a tenth of a second rather than after five minutes of measurement") and pinned by nothing. A CLI test now drives it through the sensor spy the two neighbouring cases already used.

Three of the per-site plants were green on their own, all in `kstrl/fixtures.py`: the command fixture's launch failure, and both of the function fixture's sites. The reason is the same each time. The neighbouring branch was driven, so the function looked covered: the command fixture's TIMEOUT had a test and its OSError did not, and the function fixture, which runs its own subprocess and has its own pair, had neither. Three tests now drive them from the real function with no patching, using a cwd that is not there (`run_scrubbed` hands `cwd` to `Popen`, which raises `FileNotFoundError`), plus a fourth that is the control on the function path. 15 of 18 caught before, 18 of 18 after; the set is 21 plants now.

**Static-guard census, re-derived by RUNNING the guards at each revision, never by diffing a pin.** The set is derived two ways at every point and the union taken: `grep -rl -E 'ast\.parse|astwalk' tests/*.py`, which finds 39 modules, and an AST predicate that also sees `from tests.helpers import astwalk` and a module that walks without ever naming `ast.parse`, which finds 42. Neither is a superset of the other, so the union is what is run.

| revision | modules | passed | skipped | xfailed | rc |
|---|---|---|---|---|---|
| the head round 2 reviewed, `10c5df7` | 41 | 1225 | 9 | 41 | 0 |
| this branch before the merge, `da55f73` | 42 | 1246 | 9 | 41 | 0 |
| after merging `origin/main` `037f0e1`, `e81f608` | 42 | 1242 | 9 | 41 | 0 |
| after the git-identity fix, `817446c` | 42 | 1242 | 9 | 41 | 0 |
| after merging `origin/main` `aad32e8`, `768537b` | 43 | 1259 | 9 | 44 | 0 |

**The before-merge census was RED on its first run**, and that is the reason it is run rather than read: 1243 passed and 3 failed, all three the by-file pins the `kstrl/fixtures.py` split moved a read across. They are fixed by re-deriving the pins from the walk, and the row above is the re-run.

The module set does not move across the FIRST merge, and the four tests that do are attributed rather than waved at. Per-module collection, parsed rather than tailed, at both revisions: 1296 collected before and 1292 after, and the whole delta is `tests/test_procgroup.py` going from 63 tests to 59. #360 rewrote that file on `main`; this branch does not touch it. Across the second merge the set gains exactly one module, `tests/test_config_guard.py`, which #359 adds. The single module this branch itself adds to the set is `tests/test_sense_dampener_workflow.py`.

**The `CheckResult` census, run at four revisions by a second implementation rather than by the guard restating itself.**

| tree | `CheckResult(...)` | declaring `measured=` | `measured=False` | failing rows on the default |
|---|---|---|---|---|
| `origin/main` `aad32e8` | 38 | 0, no field | 0 | 22 |
| this branch as round 1 was reviewed, `7f87b00` | 41 | 6 | 6 | 19 |
| after round 1's sweep, `10c5df7` | 38 | 17 | 14 | 9 |
| this head, `768537b` | 39 | 18 | 15 | 9 |

The `7f87b00` row's 41 / 6 / 19 reproduces round 1's review census exactly, which is what makes the rest of the column comparable. 41 to 38 is #335 landing on `main` mid-review, not this branch. Round 2 moved the gates off `measured=False` and onto the parser's answer, and marked `_fixture_file_text`'s two read handlers, which is where this head's numbers come from. The one extra construction is `check_diff_scope`'s new empty-diff return, which is S1. The census now covers `FixtureResult` too, and the same second implementation counts it at 27 constructions, 6 declaring `measured=`, 6 of them `measured=False` and 17 failing rows on the default, against 4 / 4 / 19 at the reviewed head and 0 / 0 / 23 on `origin/main`. The two that moved are `_fixture_file_text`'s `OSError` and `UnicodeDecodeError`.

**The repository's own baseline, regenerated by running the command**, `KSTRL_TIMEOUT_VERIFY=1800 uv run ks sense --write-baseline --force --base HEAD`, on the merged tree. The timeout is not decoration: the digest is a function of the three resolved verify commands AND the subprocess timeout, so generating this file at the 300s default writes `4b633754fe6ab448` instead of `c446b9151d1cd264` and every comparison on a runner is then refused with exit 2. `--base HEAD` gives the run an empty diff, which is the state the workflow compares against on `main`, and it is why `bad_patterns` and `diff_scope` are recorded as unmeasured rather than as vacuous passes.

It was regenerated again on round 2's tree, by running the command rather than editing the file. The sweep moved one measured field, so `bad_patterns` records the reason `no files in the diff` instead of `Scanned 0 Python files, no issues`; everything else that moved is provenance, `generated_at` and `base_ref`. `verify_digest` is unchanged at `c446b9151d1cd264`, and `tests/test_sense_committed_baseline.py` now recomputes it from this checkout rather than trusting it.

**The workflow ran on a runner with the base-ref fix live, and took the bootstrap path.** Run 34167331975, job `dampener`, 15 seconds, conclusion success, and again on the next push. `git show 037f0e1:scripts/kstrl/sense-baseline.json` fails on the runner because the baseline file does not exist on `main` yet: this pull request is what adds it. So the step wrote the bootstrap report, recorded `rc=0`, and UPDATED the existing comment rather than posting a second one (id 5561736414, still one comment carrying the marker), with the text `No baseline on main at 037f0e1..., so there is nothing to compare this branch against`. That is S4 end to end on a runner: before the fix the same job read the pull request's OWN copy of the baseline and reported no regression against a yardstick the branch supplied, which is how it spent 5m28s on run 34130546828 and reported `no regression`. The bootstrap path spends no sensor time at all, so the first comparison with real measurement in it is the first pull request opened after this one puts the baseline on `main`. The digest check is therefore exercised by `tests/test_sense_committed_baseline.py` and by the earlier run rather than by this one.

**Three shell defects reproduced by RUNNING the step's own script**, read out of the parsed YAML and executed under `bash -e` against stub executables on PATH. That the sense step records `rc=1` and still exits 0, which is why a separate failing step is needed at all; that the comment step posts nothing when `sense.md` is empty, with a stub that would print `CALLED` if it were reached; and that a failed lookup refuses instead of posting, with a stub that fails on `--paginate` and would succeed on a POST. A fourth test is the control: with the lookup working, the script reaches PATCH with the id it found and does not POST.

## Assumed

Stated as assumptions, in those words.

- **The exit-2 branch and the fork guard were exercised locally, never on a runner.** They are tested against the parsed YAML and against locally executed shell. The comment-UPDATE path is no longer assumed: the run above updated the existing comment rather than posting a second one.
- **`timeout-minutes: 160` is arithmetic, not a distribution.** It is five subprocesses that can each spend `KSTRL_TIMEOUT_VERIFY` (1800) plus ten minutes for checkout, uv install and `uv sync`, and the test DERIVES it rather than bounding it. The multiplier is counted out of `kstrl/verify.py`; the ten minutes is headroom over a measured 9 seconds of install on run 34130546828 with a warm uv cache, and it is a choice rather than a measurement. Four cold-runner measurements on this branch: 4m23s, 5m2s, 5m24s and 5m28s, none of which came near either cap.
- **The digest's stability across machines is by construction, not by measurement.** It hashes the three resolved commands and the timeout; `resolve_typecheck_command` reads the project's `pyproject.toml`, so a project that changes that file changes the digest, which is the intended behaviour but has not been exercised across two machines.
- **One merge commit on this branch carries no `Co-Authored-By` trailer.** `7223962`, an earlier merge of `origin/main`, already in the pushed history; adding the trailer would mean rewriting published history for a commit with no content of its own. The latest merge was unpushed when the omission was noticed, so it was amended in place and does carry both trailers.

## Handoffs

**1. `require_self_critique` with no `progress_file_path` and no PRD emits neither a row nor a gap.** Named in `SENSE_SCHEMA_VERSION`'s own comment block as a pre-existing follow-up on #306, and it is the one remaining way a check the operator asked for can be silent. Out of scope here because it is about the sidecar rather than about `measured`, and the census in this pull request cannot see it: there is no `CheckResult` to count.

**2. The workflow duplicates this repository's whole test suite on every pull request.** `ci.yml` bills about 10 minutes per pull request; the dampener job measured 5m28s across two runs of it, so 6 billed minutes. What it buys over `ci.yml` is two sensors, because the committed baseline is `"passed": true, "signatures": {}` and `ci.yml` already gates on tests, mypy and ruff. Three ways to spend less, in ascending order of how much they change: `paths-ignore` for markdown (2 of the last 30 merged pull requests, about 7 percent); scoping the dampener's test sensor, since the baseline records zero test signatures; or triggering on `workflow_run` after `ci.yml` succeeds.

**3. The live comment prints an absolute runner path.** ``Baseline: `/home/runner/work/kstrl/kstrl/scripts/kstrl/sense-baseline.json` at `9715ec5`.`` Confirmed again on comment 5561736414 after this round, on run 34130546828. Cosmetic, and declined rather than fixed, because it is neither a false statement nor a wrong number: the path IS that path on a runner. Rendering it relative to `--root` when it is under the root would read better.

**4. Ticking #235's checkbox** is a `gh issue edit` after merge.

**Nits declined, one line each.** `Baseline` and `Comparison` are `frozen=True` with `Mapping` fields, so the synthesised `__hash__` raises `TypeError`: latent, nothing hashes them, and adding `__hash__ = None` would assert a decision nobody has needed to make. `--ui` and `--no-color` reach nothing on either dampener path: true, and `--json` already ignores them on `main`, so it is pre-existing behaviour extended rather than a new silent no-op. `refuse_existing_baseline` and `short_ref` are module-public with no caller outside their own module: style, and `short_ref` is now in a different module from its sibling so privatising it would move it again. The `head -n 1` and `pipefail` nit is fixed rather than declined: the lookup no longer pipes, and the step sets `pipefail` for any pipeline a later edit adds.

Closes #227


kstrl/dampener.py
```diff
@@ -0,0 +1,689 @@
...
+        f"**{title}**",
+        "",
+        header,
+        "|" + "|".join([" --- "] * columns) + "|",
+        *rows,
+        "",
+    ]
+
+
+def render_markdown(comparison: Comparison, baseline: Baseline, path: Path) -> str:
```
0xfauzi: **A sensor that goes dark on the branch is invisible in the report and is never a regression.**

`render_human` and `render_markdown` take `comparison`, `baseline` and `path` — they never receive `current`, so `current.unmeasured_checks` reaches only the `--json` block. The `unmeasured` bucket holds *baseline signatures absent now*, so with this repository's own baseline (`"signatures": {}`) it is always empty no matter how many sensors stopped measuring.

Reproduced against this branch:

```
BASELINE: {} ('linter', 'test_suite') ()          # green tree, both measured
branch:   test_suite -> CheckResult(passed=False, "Test suite timed out after 1800.0s", measured=False)

REGRESSED: False   exit (--fail-on-regression): 0
<!-- kstrl-sense-dampener -->
## sense dampener
Baseline: `scripts/kstrl/sense-baseline.json` at `06cd16d`.
**no regression**
```

Before `measured=False` existed, that timeout produced `test_suite:test-suite-timed-out-after-s` as a **new** signature, i.e. a flagged regression. The field narrows `fixed` correctly and, in the same move, deletes the only signal that the suite stopped running. A branch that hangs the tests gets a green "no regression" comment.

`--write-baseline` prints `unmeasured: ...` always, `none` included, precisely because a hole has to be visible at the moment it is recorded; the compare path drops that discipline. Suggest passing `current` into both renderers and naming its unmeasured checks unconditionally, and treating "a check the baseline measured that measures nothing now" as a flagging bucket rather than silence.

.github/workflows/sense-dampener.yml
```diff
@@ -0,0 +1,134 @@
...
+            gh api --method PATCH "repos/$REPO/issues/comments/$id" -F body=@sense.md
+          else
+            gh api --method POST "repos/$REPO/issues/$NUMBER/comments" -F body=@sense.md
+          fi
+
+      - name: Fail only when the sensor could not run
+        # Exit 2 means the measurement could not happen: a missing baseline, a
+        # malformed one, a bad kstrl.toml, or git could not produce the diff.
+        # A regression is NOT a failure here, on purpose.
+        if: steps.sense.outputs.rc == '2'
```
0xfauzi: **The documented graduation to blocking does not block.**

`docs/dampener.md` says: *"add `--fail-on-regression` to the `ks sense` invocation in the workflow. That is the whole change."* It is not. The sense step runs under `set +e` and records the code into an output; the step itself always exits 0 (its last command is `set -e`). The only failing step is this one, gated on `rc == '2'`. With `--fail-on-regression` added, `ks sense` exits **1**, no step keys on 1, and the job stays green.

So an adopter follows the documented step 2, believes the dampener is now blocking, and gets a control that looks like a control — the exact failure the same doc closes on. `test_the_workflow_is_advisory` pins that the flag is absent, so nothing covers the state the doc tells people to move to.

Either key the failure on `rc != '0'`, or make the doc's step 2 name both edits.

kstrl/verify.py
```diff
@@ -215,6 +215,20 @@ class CheckResult:
...
+    # one of these into a gap would make `result.passed` true on a timeout.
+    #
+    # Read by :mod:`kstrl.dampener` and nothing else today. It changes no
+    # existing behaviour and no published surface: `passed` still decides the
+    # verdict, the report table and the `ks sense --json` check objects are
+    # untouched. What it buys is that a signature's disappearance can be told
+    # apart from the sensor's, which a fallback signature cannot say for
+    # itself: `signature_slug` strips digits, so "timed out after 300.0s" and
+    # "timed out after 1800.0s" are the same string.
+    measured: bool = True
```
0xfauzi, line 233: **The clearing side defaults to "compliant", and 8+ known sites still clear.**

`fixed` CLEARS, and it clears whenever `check.measured` is True. `measured` defaults to `True`, so every `CheckResult(...)` construction that nobody audited votes "I measured" by omission. CLAUDE.md's rule (3) for this repo's dominant defect class: *"A guard that CLEARS must be narrow... If a clearing guard cannot PROVE a site is compliant, it must flag."* A default of True is the opposite: it assumes compliance for all 38 construction sites and proves it for 6.

The PR body already lists 8 unfixed sites and reports 5 of them reproduced landing in `fixed` (`policy_envelope` at verify.py:1686/1710, `test_adequacy` at :1879, `self_critique` at :610/:617, `prd_stories` at :818, `fixtures` at fixtures.py:655/:664). Two more the body does not list:

- `verify.py:2211` — ruff's own timeout inside `check_dead_code` is swallowed with `pass  # Non-fatal`, then the row is emitted with the default `measured=True` and a `ruff reports 0 auto-removable` note. Half the check measured nothing, in the very function this PR patched three times.
- `fixtures.py:164` / `:393` — `FixtureResult` command and function-fixture timeouts roll into a `fixtures` row with no `measured` signal.

`measured: bool = False` with the ~30 measuring sites stating `measured=True`, or the census guard the body describes as a follow-up, are the two shapes that fail red. Shipping the clearing half without either makes this instance twelve of the class.

kstrl/dampener.py
```diff
@@ -0,0 +1,689 @@
...
+    """
+    try:
+        text = path.read_text(encoding="utf-8")
+    except FileNotFoundError:
+        raise BaselineError(f"no baseline at {path}; run ks sense --write-baseline first") from None
+    except (OSError, ValueError) as exc:
+        raise BaselineError(f"cannot read the baseline at {path}: {exc}") from exc
+    try:
+        raw = json.loads(text)
+    except ValueError as exc:
```
0xfauzi: **`RecursionError` escapes this handler, so a malformed baseline exits 1 with a traceback instead of the documented exit 2.**

Reproduced on this branch:

```python
p.write_text("[" * 200000 + "]" * 200000, encoding="utf-8")
dampener.read_baseline(p)
# ESCAPED -> RecursionError -> traceback, exit 1, not exit 2
```

`json.loads` parses by recursive descent, so its taxonomy includes `RecursionError` (a `RuntimeError`) alongside `JSONDecodeError`. CLAUDE.md records this exact class from #318, where an enumeration of subclasses was wrong three times: *"`Exception` exactly: narrower is the defect itself"*, and *"ALL the I/O outside the guarded block"* — which this already does correctly, so only the exception class is wrong.

It matters beyond the traceback: `docs/dampener.md` promises exit 2 for "baseline missing, unreadable, malformed", and the workflow's failure step keys on `rc == '2'`. An exit 1 skips that step and falls through to the comment step (see the finding there). `kstrl/decisions.py:437` has the same narrow clause, so this is a class sweep rather than a one-line fix.

.github/workflows/sense-dampener.yml
```diff
@@ -0,0 +1,134 @@
...
+            echo '## sense dampener: the sensor could not run' >> "$GITHUB_STEP_SUMMARY"
+            echo '```' >> "$GITHUB_STEP_SUMMARY"
+            cat sense.err >> "$GITHUB_STEP_SUMMARY"
+            echo '```' >> "$GITHUB_STEP_SUMMARY"
+          fi
+
+      - name: Post or update the pull-request comment
+        if: >-
+          github.event.pull_request.head.repo.full_name == github.repository
+          && steps.sense.outputs.rc != '2'
```
0xfauzi: **Gating on `rc != '2'` posts an empty comment body on any other non-zero exit.**

`rc` is 0 or 2 only while the invocation is exactly the one on line 79. Any other code — a `RecursionError` traceback out of `read_baseline` (exit 1, see that finding), `uv run` failing, an OOM kill — leaves `sense.md` empty and `rc` at something that is not `'2'`, so this step runs and `gh api ... -F body=@sense.md` sends `{"body": ""}`. GitHub answers 422 "body cannot be blank", the step fails, and the job goes red with the useful diagnostic sitting unread in `sense.err` because line 130's step was skipped.

`steps.sense.outputs.rc == '0'`, or a `[ -s sense.md ]` test in the script, expresses "we have a report to post" — which is what this condition is trying to say.

.github/workflows/sense-dampener.yml
```diff
@@ -0,0 +1,134 @@
...
+        run: |
+          marker='<!-- kstrl-sense-dampener -->'
+          # `--jq` runs once PER PAGE, so a `first` inside it is the first
+          # match on THAT page, not the first overall. Measured against gh
+          # 2.73.0 on a 3-comment pull request forced to per_page=1: the
+          # `first` form printed three ids on three lines, which would hand
+          # the PATCH below a multi-line id once a pull request passes the
+          # 30-comment default page size. `--slurp` is refused alongside
+          # `--jq`, so the filter emits every match and `head` picks one.
+          id="$(gh api --paginate "repos/$REPO/issues/$NUMBER/comments" \
```
0xfauzi: **The lookup swallows `gh`'s exit status, so an API failure reads as "no earlier comment" and posts a duplicate.**

The assignment `id="$(pipeline)"` takes the status of the last stage (`head`), and the default Actions shell is `bash -e` with no `pipefail`, so a non-zero `gh api` is invisible. A rate limit, a 5xx, or a `null` body making the jq filter error (`null cannot be matched, as it is not a string`) all yield an empty `id` and fall into the POST branch — a second comment on every push, which is exactly the outcome the `--paginate`/`first` fix on this line was made to prevent, reached by a different route.

Capture the ids into a variable first and fail the step when `gh` does, rather than letting an empty read mean "not found". This is the repo's own rule from #260: *"make unreadable a REFUSAL, not an empty read, because an empty read is a silent removal of the mechanism."*

.github/workflows/sense-dampener.yml
```diff
@@ -0,0 +1,134 @@
...
+jobs:
+  dampener:
+    runs-on: ubuntu-latest
+    # A full `ks sense` on this repository was measured at 327 seconds, 326.7
+    # of them the test suite, on a warm developer machine with other work
+    # sharing it. A cold runner also pays mypy's and ruff's first-run cost,
+    # which is NOT measured and is not guessed at here; 20 minutes is the
+    # measured figure with headroom, against ci.yml's 15 for the same suite
+    # minus the spine tier.
+    timeout-minutes: 20
```
0xfauzi: **`timeout-minutes: 20` (1200s) is below the `KSTRL_TIMEOUT_VERIFY: "1800"` this same job sets, so the sensor timeout can never fire in CI.**

A hung suite is killed by GitHub at 20 minutes, minus the ~60s of checkout/uv/sync above it. The `subprocess` timeout at 1800s is unreachable, so the run produces no `sense.md`, no `sense.err`, no comment and no step summary — just a cancelled job. The whole `measured=False` mechanism this PR builds on the 1800s timeout is dead code on this runner.

The two numbers have to be ordered: either raise `timeout-minutes` above 1800s plus setup, or lower `KSTRL_TIMEOUT_VERIFY` under the job cap (and regenerate the baseline at whatever it becomes). `docs/dampener.md` step 3 tells adopters to set the env var without mentioning the job cap it has to sit under.

kstrl/dampener.py
```diff
@@ -0,0 +1,689 @@
...
+    #: timeout. They contribute no signatures at all.
+    unmeasured_checks: tuple[str, ...]
+    signatures: Mapping[str, int]
+
+    @property
+    def total_findings(self) -> int:
+        """Occurrences, not distinct signatures: 12 E501s are 12, not 1."""
+        return sum(self.signatures.values())
+
+    def to_document(self) -> dict[str, Any]:
```
0xfauzi, line 253: **The baseline records no provenance for the configuration it was measured under, so "measured the same way" has no mechanism.**

`docs/dampener.md` states the hazard in its own words — *"A baseline and a comparison measured at different timeouts are not a comparison"* — and then enforces it with a literal `1800` typed once in the YAML and once in a test. The document carries `generated_at`, `base_ref`, `passed`, `sense_schema_version`, `measured_checks`, `unmeasured_checks`, `signatures`, and nothing about the verify timeout or the test / typecheck / lint commands the signatures came from.

So an adopter who changes `[verify] test_command`, or runs `ks sense --compare-baseline` locally at the default 300s against a baseline written at 1800s, gets a comparison that is not one, with no note anywhere. Per CLAUDE.md: *"If there is no mechanism, there is no plan."*

Recording a fingerprint of the resolved `VerifyConfig` (timeout plus the command strings) and emitting a note — the same shape `sense_schema_changed` already has — closes it without a schema bump beyond `BASELINE_SCHEMA_VERSION`, which is still 1 and unreleased.

kstrl/dampener.py
```diff
@@ -0,0 +1,689 @@
...
+
+#: Two shapes, not one shape with a discriminant string. Roadmap doctrine 6
+#: forbids a new status vocabulary, and the earlier ``action="write"`` /
+#: ``action="compare"`` pair was one: it also forced three fields to be
+#: hard-coded dead on whichever side did not use them, and an
+#: ``assert baseline is not None`` in the CLI to say what the type could not.
+Mode = WriteMode | CompareMode
+
+
+def _baseline_path(value: str, root_dir: Path) -> Path:
```
0xfauzi, line 640: **An explicit relative path resolves against the process CWD while the bare flag resolves under `--root`.**

`ks sense --root /elsewhere --compare-baseline` reads `/elsewhere/scripts/kstrl/sense-baseline.json`; `ks sense --root /elsewhere --compare-baseline scripts/kstrl/sense-baseline.json` — the path `--help` literally prints as the default — reads `$PWD/scripts/kstrl/sense-baseline.json`, and reports `no baseline at ...` for a file that exists. The docstring says *"a value is taken as given"*, but nothing the operator reads (`--help`, `docs/dampener.md`) says it, and every test passes an absolute path, so the case is uncovered.

`root_dir / Path(value).expanduser()` (a no-op when the value is absolute) makes both spellings of the same location agree.

kstrl/dampener.py
```diff
@@ -0,0 +1,689 @@
...
+    """The bare flag means the default under ``--root``; a value is taken as given."""
+    if value == OPTIONAL_VALUE_SENTINEL:
+        return root_dir / DEFAULT_BASELINE_PATH
+    return Path(value).expanduser()
+
+
+def _refuse_dead_flags(
```
0xfauzi, line 655: **Two more flags that silently do nothing are not refused, against this function's own stated rule.**

The docstring is *"Refuse every combination in which a flag would silently do nothing"*, and the module argues a silent no-op is *"a flag an operator believes they passed"*. But `--ui` and `--no-color` are exactly that on both dampener paths: `_sense_dampener_report` never builds a console UI and prints through bare `click.echo`, so `ks sense --compare-baseline --ui gum` and `ks sense --write-baseline --no-color` accept a flag that reaches nothing.

`--ui` is the clearer of the two (`--no-color` at least produces uncoloured output by accident). Either refuse them alongside the six, or state in the docstring why these two are exempt — the enumeration reads as exhaustive today and is not.

kstrl/dampener.py
```diff
@@ -0,0 +1,689 @@
...
+    for name, count in value.items():
+        if not isinstance(name, str) or not name:
+            _fail(f"baseline {key!r} has a key that is not a non-empty string: {name!r}")
+        if isinstance(count, bool) or not isinstance(count, int) or count < 0:
+            _fail(f"baseline {key!r}[{name!r}] must be a non-negative integer, got {count!r}")
+        counts[name] = count
+    return counts
+
+
+@dataclass(frozen=True)
```
0xfauzi, line 213: **`frozen=True` generates `__hash__`, but `signatures: Mapping[str, int]` makes hashing raise.**

`@dataclass(frozen=True)` with the default `eq=True` synthesises `__hash__` from every field, so `hash(baseline)` raises `TypeError: unhashable type: 'dict'` at the first dict field. `Comparison` (line 338) has the same shape with four dict fields. Nothing hashes them today, so this is latent: `set(baselines)`, a `@lru_cache` on a function taking one, or a dict keyed by comparison all fail at runtime on a type that advertises immutability.

`field(hash=False)` on the mapping fields, or `eq=True, frozen=True` with an explicit `__hash__ = None`, makes the type say what it is. Storing the signatures as a sorted tuple of pairs would make it genuinely hashable, at the cost of the `.get`/`in` reads in `compare`.

kstrl/evolution.py
```diff
@@ -577,7 +578,60 @@ def category_for_check(check_name: str) -> str:
...
+    :func:`signatures_from_verification` cannot give, because it ends in a
+    dedupe - every count built from its return value would be 1 and a
+    dampener's "this got worse" bucket could never fire (#227).
+
+    ``limit`` caps the DISTINCT codes one check contributes, in first-seen
+    order, before anything is counted. ``None`` means no cap. The default is
+    the journal's own constant, so a defaulted call is byte-identical to what
+    the journal recorded before this parameter existed.
+
+    This is the one place the ``"<check>:<code>"`` spelling is written, and
```
0xfauzi: **"the one place the spelling is written" is not true, and it is the kind of claim a later author will act on.**

`signature_for_error` at line 523 also writes the format — `f"{check_name or 'unknown'}:{slug}"` — and `_check_signatures` reaches it on every check with no parsed codes, so a large share of real signatures are spelled there, not here. `split_signature` (line 526) re-implements the parse a third time.

The sentence is load-bearing for the dampener: `compare` recovers the check name with `split_signature` and clears `fixed` on the result, so the three spellings have to agree for the clearing side to be right. Narrow the claim to the codes path, or route the fallback through this function so it becomes true.

Issue #303, opened 2026-08-31
H3a sweep: seven more instruction-text builders reach a role's prompt unenrolled


0xfauzi:
> **Stabilisation queue: P1, rank 6 of 21** (set 2026-09-09; the full ordered list and the coordinator's rules are in #369). Seven instruction-text builders reach a role's prompt unenrolled, so H3 cannot see them change. Enrol under H3 (hash plus version, free); record the H2 obligation and spend estimate per role instead of running calibration (see #369 owner switches).

Found by an H3a sweep of `kstrl/` run while fixing #299. #299 closes the two instances it was filed for (`git.repo_change_source` and `pasted_change_source`) and deliberately stops there. This issue records everything else the sweep found so the list is not lost.

## Why the sweep was run

Policy H3a says harness-authored instruction text reaching a role's prompt is enrolled whatever assembles it, while operator-supplied values interpolated into it are not. #299 established that the AST enrollment walk in `tests/test_prompt_versions.py` sees module-level `*_PROMPT` constants only, so any instruction text a function assembles escapes enforcement regardless of how well it obeys the rule.

The sweep traced backwards from all five LLM call sites: `loop.py:721` (engineer), `decompose.py:390` and `:1300` (architect), `review.py:1086` (reviewer), `security.py:675` (security), `knowledge.py:1187` (distiller).

## Clear-cut: harness instruction text reaching a role, not enrolled

1. **`kstrl/init_cmd.py:1615-1709` `_generate_claude_md(ctx)`**, plus `_LANGUAGE_STANDARDS` (:1483), `_LANGUAGE_ANTIPATTERNS` (:1553), `_VERIFICATION_SECTION` (:1595). Role: engineer. This writes CLAUDE.md, which `loop.build_project_context` (`loop.py:447-458`) prepends verbatim to every iteration. The comment at `:1590` already says so. Same footing as `DEFAULT_PROMPT`, which is enrolled. The sweep called this the largest body of unenrolled steering in the repo.

   Note this one also interacts with policy H3b: `SCAFFOLDED_TEMPLATES` records the SHA-256 of every body each scaffolded template has ever shipped, and rows are appended, never edited. Enrolling the generated CLAUDE.md needs that interaction worked out first.

[...]

Issue #266, opened 2026-08-28
Stop pasting the diff into the review prompt: the reviewer already runs inside the worktree and can read the repo itself


0xfauzi:
Split out of #265 after a research pass the owner asked for. #265 asks how to stop the retry loop driving a component past the reviewer's chunking limit. This issue is the finding that the limit should not exist.

## The finding

**kstrl's reviewer is already running inside the worktree.** `kstrl/review.py:1032`:

```python
output_lines = collect_agent_output(
    agent, prompt, cwd=worktree_path, timeout=timeout, ...
)
```

The `Agent` protocol (`kstrl/agents/base.py:455`) is `run(prompt, cwd, timeout)`. `cwd` has always been passed.

So kstrl truncates the diff to 50,000 chars, splits it into chunks, fails closed when a single hunk will not fit, and burns retries on components it can no longer review, in order to hand a diff to an agent that is sitting in the repository, with git on its path, and is not permitted to look.

## Why hunk size is in the picture at all

It is not a property of review. It is a property of pasting.

`DEFAULT_PROMPT_DIFF_CHAR_LIMIT = 50_000` (`kstrl/git.py:774`) is a hand-picked prompt budget, not a model limit. Its own comment says "if the LLM context window changes, edit one place". A hunk is the splitting floor because it is the smallest unit of pasteable text that is still a valid diff.

Two things are being protected, and they are worth separating:

- The cap protects against context overflow. That is what goes away.
- Chunking rather than truncating protects against an engineer padding the first 50KB with benign churn so a malicious hunk lands after the cut. That is a real property and must be re-established, not dropped.

## The failure class is structural

A newly added file is always exactly one hunk. Measured: a 1200-test file, 65,775 characters, `grep -c '^@@'` returns 1. Through kstrl's real splitter:

```
DiffUnsplittableError: single hunk in test_big.py is 65661 chars, over the 49636 chars
left for hunk content after the 115-char file header (@@ -0,0 +1,2402 @@)
```

Ten seconds, no model call. **Any new file over roughly 50KB is permanently unreviewable**, and #265 shows the retry loop walks components into that state monotonically.

## Measured: a harness returns kstrl's exact contract while reading the repo itself

`codex exec --sandbox read-only --output-schema <schema> -o out.json --json`, given kstrl's PRD in the prompt and a schema mirroring `ReviewResult`, and **no diff pasted at all**:

- Top-level keys an exact match for the contract.
[...]

Issue #223, opened 2026-08-25
R10.2: Level-triggered retry context: render current failures, not the accumulated log


0xfauzi:
Part of #235 (R10: Control Loop tracking issue). Item R10.2.

Design section: 5.3 (and gap 3.3). Sizing: M (one PR). Dependencies: none (R10.1 makes the manual check cheaper but is not required).

## Why

The retry context is an integrator with no anti-windup.

`IterationContext` (`kstrl/context.py:20-46`) holds four lists: `records`, `review_findings`, `verification_failures`, `contract_failures`. The only mutators are `add_*` methods that append. There is no clear, no dedup, no expiry. The object is serialised into `component_contexts[comp.id]` on retry (`kstrl/pipeline.py:1170-1171`) and re-loaded with `IterationContext.from_json(...)` then appended to at nine call sites: `kstrl/pipeline.py` lines 1568, 1871, 2033, 2291, 2344, 2414, 2635, 2883 and `kstrl/factory.py:3138`.

`format_for_prompt` (`kstrl/context.py:48-88`) renders every accumulated entry and ends with the literal line `Fix ALL issues listed above before completing.` So on attempt 3 the agent is handed attempt 1's failures, which it may have fixed on attempt 2, unmarked as fixed, with an instruction to fix them all.

The codebase already knows the fix: `Component.review_findings` on the manifest *is* cleared at the start of each attempt (`kstrl/pipeline.py:1035`, `begin_attempt`). The prompt payload is the one that is not.

Vocabulary: an edge-triggered system acts on transitions ("this failure happened"); a level-triggered system acts on current state ("this failure is happening now"). Kubernetes controllers are level-triggered on purpose. The retry context must become level-triggered.

## The rule (this is the whole design; implement exactly this)

Phases run in a fixed order per attempt: engineer (guard/breaker/merge-conflict records) then verification, review, security, contract. Assign each phase an integer rank:

```python
PHASE_RANK: dict[str, int] = {
    "engineer": 0,      # guard violations, breaker trips, merge-conflict restarts, HITL retry requests
    "verification": 1,
    "review": 2,
    "security": 3,
    "contract": 4,
}
```

Each failure entry records the attempt number it came from and the phase that produced it. At render time, let `N` be the latest attempt with any entry and `Q` be the rank of the highest-ranked phase that has an entry from attempt `N` (an attempt stops at its first failing gate, so in practice `Q` is that gate; the max is the safe general form). Then:

[...]

PR #358, opened 2026-09-06
feat(context): a recorded pass retires a skippable-phase finding, and the contract gate merges it too (#247)

### LINKED ISSUE #247: R10.2 follow-up: record which phases ran, so an observed pass retires a review finding
Follow-up to #223 (R10.2), found by an independent review of PR #246. Sizing: M. Dependencies: reads better after #226.

## Why

The level-triggered retry context retires an older finding when a later attempt proves it is gone. It proves that two ways:

- **Observed**: the same phase produced a fresh reading in the latest attempt, which supersedes the old one.
- **Inferred**: a higher-ranked phase failed in the latest attempt, so every lower-ranked phase must have run and passed, or the attempt would have stopped earlier.

The inference is only sound for a phase that always runs once its predecessor passes. Review and security do not: `_phase_review` and `_phase_security` downgrade to `SKIP` when `max_adversarial_calls` is exhausted and let the component carry on (that is the defect #226 fixes). So a contract failure in attempt 2 does not prove the reviewer saw attempt 2.

`kstrl/context.py` therefore excludes review and security from the inference (`SKIPPABLE_PHASES`). That is the right side of the trade - dropping a live finding in silence is worse than showing a cleared one - but it has a cost:

**When the reviewer DID run in attempt 2 and passed, it records no entry.** A passing phase writes nothing to the context. So an attempt-1 review finding that the reviewer cleared still renders under `## Not re-measured`, and the agent is told to re-check something it has already fixed. Bounded and honest, but noise.

Pinned by `tests/test_context.py::test_cleared_skippable_finding_is_shown_not_dropped` so it cannot change without someone deciding to change it.

## What to change

Give `IterationContext` positive evidence of which phases ran in the attempt in flight, and retire a skippable-phase finding when its phase ran and produced no failure.

The structural obstacle, which is the real work: **the context is only deserialised on failure paths.** `IterationContext.from_json(comp_result.context_json)` is called at each failing gate; a phase that passes never touches the object, and a mutation made mid-attempt would not survive anyway, because the context only flows back into `component_contexts[comp.id]` through `PhaseFailure.context_json` on a failing path.

So this needs one of:

1. A per-attempt "phases that ran" set carried on `ComponentResult` (or alongside it) and merged into the context at whichever gate finally fails. Cheapest, keeps `IterationContext` a pure data object.
2. Constructing the context once per attempt and threading it through every phase. Larger, and it makes every phase a writer of the retry prompt.

Option 1 looks right. It also gives the renderer something to say: a phase that ran and passed could be named in the "Resolved or superseded" line rather than inferred silently.

## Acceptance

- [ ] Attempt 1 fails review; attempt 2 passes review and fails security. Attempt 3's prompt does NOT contain attempt 1's review finding, and the resolved line names review.
- [ ] Attempt 1 fails review; attempt 2 skips review (budget exhausted) and fails security. Attempt 3's prompt DOES contain attempt 1's review finding under `## Not re-measured`. This is the case `SKIPPABLE_PHASES` exists for and it must not regress.
- [ ] `test_cleared_skippable_finding_is_shown_not_dropped` is replaced by a test of the new behaviour, not deleted.
- [ ] No new pipeline phase (doctrine 5) and no new status enum (doctrine 6).

## Note on #226

#226 makes hard-mode review and security halt rather than skip when the budget is exhausted. That removes the cause in hard mode, but not in advisory mode, which keeps the downgrade by design. So this issue is still worth doing after #226; it is just less urgent.

0xfauzi:
## Summary

The level-triggered retry context retires an older finding two ways: observed, when the same phase produced a fresh reading in the latest attempt, or inferred, when a higher-ranked phase failed and every lower-ranked one must therefore have run and passed. The inference is unsound for review and security, because both downgrade to a skip when `max_adversarial_calls` is spent in advisory mode, so `SKIPPABLE_PHASES` excludes them from it. The cost was that a reviewer which DID run in attempt N and passed recorded nothing, so a finding it had cleared still rendered under `## Not re-measured` and the agent was told to re-check something it had already fixed.

That is measured, not argued. A repro driver runs the real `ComponentPipeline.process_result` twice per scenario with only `git.get_diff_content` and `agents.get_agent` stubbed, feeding `component_contexts` back in the way `_submit_args` does, and renders the block the worker actually builds. At the base commit, scenario B2 (the budget skipped the reviewer) and scenario C (the reviewer ran and passed) produced **byte-identical prompt blocks**. The prompt could not tell "the reviewer looked again and cleared it" from "the reviewer never looked". That driver and every raw measurement file named below live in the lane's scratchpad and are NOT in this repository: `git ls-files | grep '^lanes/'` is empty, so a reader of this PR cannot open them and the numbers are quoted here instead of pointed at.

`IterationContext` now carries positive evidence. `PhaseReading(attempt, phase)` records that a phase ran and returned a verdict; `_buckets` consults it and only the skippable phases with no reading in the latest attempt are held back, so a skippable entry is retired by an observed reading rather than by an inference. Nothing in `format_for_prompt` changed: the resolved line already derives its phase names from the resolved entries, so moving a review entry into `resolved` makes it name review.

"Ran" means **produced a reading**, and the narrowness is load-bearing. Three states are excluded, each for the same reason:

- a reviewer that CRASHED. In advisory mode it yields `passed = review_mode != HARD` = True with `infrastructure_error=True`, records no entry, and returns `ran=True`. Keying on `ran` would let it retire a live finding.
- a reviewer #266's coverage check says did not see the whole change. `apply_coverage_check` sets `diffstat_disagreement` in EVERY mode and converts it to an infrastructure error only in HARD mode; advisory returns early and `run_review` then forces `passed = True`, which also makes `coverage_refused` False, so the #266 wall does not fire either. Round 1 of review found this one live: the result arrived as ran, present and not an infrastructure error, and retired an earlier finding on a verdict reached without the whole change in hand.
- every skip route, all of which return `ran=False` and record nothing: `review_mode = "skip"`, the advisory budget downgrade, security not configured, and hard mode's #226 refusal.

The predicate is `ran and result is not None and not result.infrastructure_error and not result.diffstat_disagreement`, written once as `pipeline._produced_a_reading` and delegated to by both phase results.

**Where the readings are merged.** The context is only deserialised on failure paths, so a phase that passes has no failure path to write on. The record therefore lives on the pipeline as `_phase_readings`, keyed by component and attempt, cleared at the attempt boundary, and merged into the context at whichever gate finally fails. The attempt number is captured at record time (`comp.retries + 1`), not at merge time, because `retry_or_fail` increments `retries` before it stores. A record carried forward from an older attempt is inert by construction: `_buckets` only consults readings whose attempt equals the latest.

**The contract-gate instance.** `component_contexts` had exactly two writers, measured with an astwalk census: `retry_or_fail` in `pipeline.py` and the contract-breaker reset in `factory.py`. Only the first could merge the record, so a contract failure in attempt N would re-raise a review finding the reviewer cleared in attempt N. Those two lines move into `ComponentPipeline.record_contract_failure`, which the factory calls, leaving both writers behind the same merge.

**The retirement rule, stated once.** An entry from a skippable phase is retired only by an observation: a fresh entry at the same rank in attempt N, or a `PhaseReading` for attempt N. Never by inference from a higher-ranked failure. Branch order in `_buckets` is load-bearing: the `rank == q` crashed-sensor branch stays above the readings branch, so a reviewer that crashed still fails to retire its own phase, and mutation S6b (the two swapped) is red across four tests.

Closes #247

## Behaviour changes

- **What the retry prompt now says.** When review or security ran in the latest attempt and passed, an earlier finding from that phase is dropped from the next attempt's prompt and the `## Resolved or superseded` line names the phase: `1 earlier finding(s) from review passed or were re-measured in attempt 2 and are omitted.` Counted across the four repro scenarios: `## Resolved or superseded` sections 0 before / 2 after, `## Not re-measured` 4 before / 2 after.
- **An attempt that recorded no failure entry now retires what its readings measured.** The reachable case is the merge-conflict restart: `_retry_after_merge_conflict` fires at the `pr` phase, which is reached only after review and security have both run and passed, so both recorded a reading and `retry_or_fail` merged it into an attempt whose only evidence is an `IterationRecord`. Round 1 of review measured those records existing and being discarded, because every older entry outranked a `q` of -1. The plain engineer-loop failure is the same shape with the opposite answer: no sensor ran, so there is no reading and nothing moves.
- **What it still does not say.** A phase that did not run records nothing and retires nothing. A finding from a reviewer the operator turned off (`review_mode = "skip"`), or one an exhausted adversarial budget skipped in advisory mode, still renders under `## Not re-measured`. A reviewer that crashed is not a reading, and neither is one whose diffstat git disagrees with.
- The contract gate goes through the same merge, so a contract failure no longer re-raises a review finding an earlier attempt cleared.
- No `format_for_prompt` literal moved. What changed is which entry lands in which bucket.
- `IterationContext` JSON gains a `readings` key. A context written by an older parent process has no such key, reads back as no readings, and therefore degrades to showing the finding rather than to dropping it. A reading naming a phase outside `PHASE_RANK` is now REFUSED on read rather than accepted, so `_require_known_phase` is one definition of the vocabulary on both sides.

### Acceptance criterion 2 is unreachable as written, and how the intent is pinned

The issue's second criterion asks for an attempt that "skips review (budget exhausted) and fails security". That cannot happen, and it is measured rather than assumed: `_phase_review` and `_phase_security` consult the same `max_adversarial_calls` counter, `adversarial_budget_consume` only ever increments, and review runs first in `process_result`. A budget that has skipped review has already skipped security. Scenario B2 confirms it: with `max_adversarial_calls=1`, attempt 2 skips BOTH, and the failing gate has to be ranked above security.

The criterion's INTENT is that a skippable phase which did not run in attempt N retires nothing. Three tests pin it: the budget skip with the HITL checkpoint as the failing gate (`pr`, rank 6), the operator's explicit `review_mode = "skip"`, and the no-reading case at the unit layer. Round 1 of review measured the first two passing with the mechanism they pin deleted, because a reviewer that runs and CRASHES also records no reading and also retires nothing, so the consequence could not tell the two apart. Both now assert the cause - `outcome.review.ran is False` plus the skip reason the pipeline wrote - and the shared queue stub raises a `BaseException` subclass on an unexpected call, which `_phase_review`'s `except Exception` cannot turn into a crashed review. Mutations B2a (`adversarial_budget_ok` always True) and B2b (the SKIP dispatch disabled) are both red. Getting the literal scenario needs a per-phase adversarial budget, which is a different issue.

## Round 1 of review: what changed

Two blockers, six should-fixes and nine nits, each reproduced before it was accepted. What this diff now carries:

| finding | what changed | the mutation that proves it |
|---|---|---|
| B1 blind advisory reviewer counted as a reading | the fourth conjunct in `_produced_a_reading`, plus a behaviour test and its control | B1a caught, B1b caught |
| B2 both criterion-2 tests passed with their mechanism deleted | assert the cause; stub raises a `BaseException` subclass | B2a caught, B2b caught |
| S1 writer census had one control for two disjuncts | three controls, one per shape | S1a caught, S1b caught |
| S2 the per-component key of `_merge_phase_readings` was untested | a two-component test with comp-a as its own control | S2 caught |
| S3 `from_json` was a second, weaker definition of the phase vocabulary | it goes through `add_phase_reading` | S3 caught |
| S4 a disclosed miss with no test behind it | a `blind_spot` row under `xfail(strict=True, raises=AssertionError)` | S4 caught, by XPASS |
| S5 the recording census could not see a direct `add_phase_reading` writer | the net watches both names; the unkeyable sites key by location and are measured, not trusted | S5a, S5b, S5c all caught |
| S6 readings never reached an attempt with no entry | the routing fix, not the docstring: the readings branch moved above the rank comparison | S6a caught, S6b caught |

The assist's suggestion block for finding 5 was NOT applied as written: round 1 measured it turning the guard red on the unmutated tree, because `_merge_phase_readings`' own legitimate call is then counted and keys to `<unfolded>`. What shipped instead teaches the key the one-positional spelling and inventories the unkeyable sites by location.

Nits applied: the `to_json` sort is pinned (N1); the `begin_attempt` pop's comment says it BOUNDS the dict and that no test is owed (N2); `PhaseReading`'s docstring no longer claims the type enforces `SKIPPABLE_PHASES` when the census does (N3); the sweep's `assert moved > 0` becomes the measured `== 1944` (N5); the dead `run_review` assignment is gone (N6); `SKIPPABLE_PHASES` states the retirement rule once (N7); the test helper says which production paths an advisory `passed=False` verdict stands in for (N8).

Nits declined: N4 (one representation for `_phase_readings` and `PhaseReading`) relocates a shape and removes nothing, which is why the quality pass rejected it the first time. N9 (`docs/atlas/atlas.json` still names the replaced test) stays disclosed rather than edited: `scripts/atlas/` does not exist here, so the file cannot be regenerated, and nothing reads it.

## Round 2 of review: what changed

Zero blockers, one should-fix, five nits. Round 2 re-verified every round-1 finding by replanting round 1's own mutation, and all of them stayed caught.

**S1, the should-fix.** `IterationContext.readings` is a public attribute of a non-frozen dataclass, so `ctx.readings.add(PhaseReading(...))` was a THIRD recording route: it names neither enrolled writer, so both censuses were blind to it, and it skips the `_require_known_phase` check the two named routes go through. The dangerous variant - a reading written for the attempt in flight - was already caught behaviourally by five tests in `TestPhaseReadingsRetireSkippableFindings`, so this was defence in depth rather than a live drop, but `_READING_WRITERS`' comment claimed to be "the two names a phase reading can be written through", which was a claim about the tree that was not true, and the sibling net in the same file discloses ITS one miss with a strict-xfail row while this one was undisclosed.

Three changes, one per problem. `_records_a_reading` gains a SHAPE branch for `.add` on a `readings` receiver, because that route has no name to enrol. The file gains a third control string, one per route, so the shape branch cannot be deleted while the two existing controls still pass. And the sanctioned `self.readings.add` inside `IterationContext.add_phase_reading` is then counted, so it is inventoried by location as a third row rather than exempted, which is what makes a SECOND write to the set land as a fourth row instead of merging into a count nobody reads. `_FAN_OUT_SITES` is renamed `_UNKEYABLE_SITES` with it, since one of the three rows is a sanctioned writer rather than a fan-out point.

Measured green then red, both directions:

| mutation | outcome |
|---|---|
| A3a, the reviewer's own plant: `ctx.readings.add(PhaseReading(attempt=1, phase="security"))` in `ComponentPipeline._merge_phase_readings` | **caught**. The file was 3 passed / 1 xfailed with this planted at the reviewed head; it is now 2 failed / 1 passed / 1 xfailed, and both censuses name `pipeline.py::ComponentPipeline._merge_phase_readings: 2` |
| the shape branch itself deleted (`writes_the_set = False`), which is the guard mutated rather than the code | **caught**, by the third control: "the census predicate matched nothing in this control, so the inventory below is indistinguishable from what a switched-off net returns" |

**Nits.** N1 and N2 were two comments making false statements. The `read_in_n` comment said the pre-commit ratchet "fails a rise in a function that is over"; complexipy 7.0.1 measures `_buckets` at 14 against the gate of 15, so it is one point UNDER, and the hook runs `--staged` and fails a change that takes a function past the gate. `_blind_review_scenario`'s docstring said the mid-run `review_mode` switch is how an advisory attempt comes to hold a review-rank entry; every write of `factory_config.review_mode` in `kstrl/` is CLI, config or TUI parse time or the autonomy bundle before the first component is submitted, and `pipeline.py`'s two are locals of `_phase_review`, so the mode is fixed for the life of a run and that switch exists only in the test. It now says so and points at `_setpoint_failure` under `setpoint_agreement = "block"`, which the sibling `_pipeline` docstring already named. N3 and N4 are answered in the H4 section below.

N5 was a reuse cluster, verified rather than taken on sight. The location test now counts through `astwalk.census`, which the hand-rolled loop was that helper written out, and `_enclosing_site` goes through `scope_of`, hoisted out of `tests/test_append_opens_have_one_home.py` into `tests/helpers/astwalk`: three callers, one implementation, and attribution by OWNERSHIP rather than by line range, so a call inside a nested helper is credited to the helper rather than to whichever `def` encloses it on the page. The other citation in that finding was declined and the reason is measured: `enclosing_function` in `tests/test_settle_discipline.py` takes a parent map the top-down walk does not keep. On placement, this class stays in `tests/test_pipeline.py` because it uses five module-private helpers of that file, counted rather than estimated and now stated in the class docstring; the file-length ratchet does not decide it, because the rule is "newly crosses 800" and this file was already over.

Two disclosures the assist asked for and round 2 accepted as documentation: an attempt that ends WITHOUT writing a context - a terminal failure, a halt, a pass - drops what it observed, which is the safe direction (a reading that never arrives retires nothing, and there is no next attempt to show it to), and that note now sits on `_phase_readings` where a reader meets it.

Tested, by running it:

- **Gate numbers, re-measured for this round on the merged tree.** `Merged tree at `c7189f8`, with `origin/main` at `543e1aa`: **6101 passed, 37 skipped, 42 xfailed, 0 failed** in 352.05s, rc=0, bytecode deleted first and `PYTHONDONTWRITEBYTECODE=1`, bounded at 1800s in its own process group. Round 1's merged-tree run was 5907/37/42 against `origin/main` at `414d662`; the +194 passed are the two commits main has gained since. **One claim from round 1 is now false and is corrected here**: `git log f9e3802..origin/main -- kstrl/context.py kstrl/pipeline.py kstrl/factory.py` is no longer empty. `543e1aa` (#355) touches `kstrl/factory.py`, so main HAS moved under one of the three files this PR changes. Its hunks are at lines 102 to 3803 and this branch's single `factory.py` hunk is at 4254, they do not overlap, the branch's `record_contract_failure` delegation is present at `HEAD` verbatim, and the full suite on the merged tree is the control. `kstrl/context.py` and `kstrl/pipeline.py` are still untouched by main`. `uv run mypy kstrl/ --strict` clean (`132 source files`), `uv run ruff check kstrl/ tests/` clean, `uv run python scripts/gen_docs.py --check` current. `cyclomatic_ratchet.py`, `file_length_ratchet.py`, `check_no_mocks.py` and `check_agent_docs_sync.py` all exit 0 on the changed files, and no file newly crosses 800 lines.
- **Complexity, measured with complexipy 7.0.1 rather than reasoned about.** `_buckets` was frozen at cognitive 18 against a gate of 15. The routing fix is written as a branch chain with two conditional appends rather than nested `if`/`else` blocks, and that shape measures **14**, with cyclomatic 8 to 7. The alternative shapes were measured, not guessed: the same fix with nested blocks is 22, and with the conditional bound to a local first it is 20.
- **The round-2 mutations, planted here rather than relayed.** Two, one at a time, `__pycache__` deleted first and `PYTHONDONTWRITEBYTECODE=1` on every run, each anchor asserted to occur exactly once before the edit so a silent failure to plant is COULD-NOT-PLANT and not a pass. Both caught, in the table above. **0 hung, 0 could-not-plant.**
- **Round 1's 37 mutations: 34 caught, 3 still green, 0 hung, 0 could-not-plant**, on the same discipline. Round 2 of review planted 46 of its own and reports 38 caught, 7 still green, 1 green-as-expected control, 0 hung, 0 could-not-plant, replanting every finding-bearing member of round 1's set independently.
- **The still-green mutations, and they are not the same thing as a miss.** Round 1 named three. Round 2 of review found four more and asked for them to be recorded here rather than re-derived by the next reader. Round 1's: R3 (delete the `_phase_readings` pop in `begin_attempt`) is inert for correctness because the attempt number is captured at record time, and the comment says so; R8 (drop the `result is not None` clause) is unreachable today and the docstring says so; R13 (drop the keyword branch of the census key) is diagnostics rather than detection, which R16 proves by planting R13 together with a keyword-spelled site: caught. Round 2's four, on the reviewer's measurement rather than mine: R23 and R24 (either recording site moved BELOW its own failure return) are inert exactly as `_produced_a_reading`'s docstring predicts, because recording it changes no bucket; R25 (the census key stops reading the public spelling's first positional) is diagnostics like R13, which R27 proves by planting it with a real extra writer: caught; and BK3 (`< q` to `<= q`) is an EQUIVALENT mutant, because the `== q` branch above it makes the two predicates the same. None is a coverage hole.
- **Merged into a moving main and re-checked, twice.** `origin/main` moved again after round 2 of review, to `543e1aa`. The static-guard census was re-derived by EXECUTING the guards before and after the merge rather than by diffing a pin, over a derived file set (every test module that imports `astwalk` or parses source itself): `**38 files, 1183 passed / 9 skipped / 41 xfailed** before the merge at `4aa2b19`, **40 files, 1220 / 9 / 41** after at `c7189f8`. The delta is fully accounted for rather than asserted: the two guard files the merge added (`test_helper_import_direction.py`, `test_open_pr_counter.py`) run alone at exactly **34 passed**, and the three existing guard files the merge CHANGED (`test_encoding_readers.py`, `test_prompt_staleness.py`, `test_tui_config_guard.py`) go 56 to 59 passed. 1183 + 3 + 34 = 1220. The same arithmetic from the other end: the pre-merge 38-file set run at the post-merge tree gives 1186, and 1186 + 34 = 1220. No walk in this branch sees more or less after the merge, and this branch's own guard file is 3 passed / 1 xfailed on both sides`.
- **The two H4 claims round 1 refuted are corrected above.** "The criterion's intent is pinned three ways" was false as a claim about what was MEASURED: two of the three passed with their mechanism deleted. It is true now, and B2a/B2b are the evidence. "The predicate excludes the crashed sensor, and every skip route records nothing" was true of crashes and skips and false of the blind reviewer; the predicate now excludes that state too, and B1a is the evidence.

Assumed, not measured:

- **The contract-gate instance in production.** Nothing in `tests/` drives `run_factory`'s contract loop, so the instance itself is analysed from `factory.py` rather than driven end to end. The seam is measured: `record_contract_failure` has a behaviour test and two mutations (M7, M8).
- **The writer census proves location, not correctness.** `{"pipeline.py": 2}` says where the writers are, not that both merge. Its docstring says so and names the one behaviour test per writer that does prove it. Its one blind spot, a write through a local alias, is a strict-xfail `blind_spot` row rather than a sentence.
- **A JSON reading's `attempt` is not type-checked on read.** A payload carrying `"2"` rather than `2` deserialises and then matches no attempt, so it under-retires, which is the safe direction. The phase is checked; the integer is not.
- **Round 2 of review's four extra still-green shapes are relayed, not replanted here.** They are attributed to that review above, in those words. The two mutations this round turns on were planted and measured in this worktree.

## Quality pass

A four-angle cleanup review over the diff landed one `refactor:` commit. What it applied: `_require_known_phase` is now the single definition of the phase-name check, which matters because an entry is retired when a reading names the same phase and two definitions of a vocabulary means the weaker one decides (round 1 found the deserialiser was that weaker one, and this round closed it). `produced_a_reading` is written once as a module function over the union, since a shared frozen base is not available (narrowing `result` per subclass is an invariant-attribute override `mypy --strict` rejects, checked rather than guessed). `IterationContext.readings` became a `set`, so idempotency is structural, with `to_json` sorting where determinism is actually needed and a test pinning the order. `_with_phase_readings` became `_merge_phase_readings` and takes the object rather than JSON, so `record_contract_failure` stops parsing and serialising the same document twice. The bucket sweep's builder is one module-level `build_sweep_context`, so the class its own docstring calls the control is the same corpus by construction.

What it rejected, with reasons: keying `_phase_readings` on `PhaseReading` rather than a pair (relocates a shape, removes nothing); renaming it `_phases_measured` (reintroduces a word collision with `_buckets`' `measured_ranks`); `record_contract_failure` reading `comp.retries` instead of taking `attempt` (trades an explicit parameter for an implicit dependency on the caller's increment order); dropping the `begin_attempt` pop (it bounds the dict); and `lru_cache` on `astwalk.package_sources` (54 call sites across 22 files, outside this diff).

## Checklist

## Provenance

None. No `*_PROMPT` constant changed and no renderer text changed: no literal in `IterationContext.format_for_prompt` moved, and the `## Resolved or superseded` wording predates this diff. What changed is which entry lands in which bucket, which is the mechanism #223 shipped. H2 is not engaged and no calibration run is owed. `format_for_prompt` remains an unenrolled H3a instance, which is issue #303 item 2 and deliberately out of scope: enrolling it needs a version constant, a snapshot row and a paid run.

## Handoffs, not fixed here

- **`_buckets` still holds two definitions of "produced a reading", and the older one is pre-existing.** The `rank == q` crashed-sensor branch keys on `not entry.infrastructure`, so a non-infrastructure review entry at attempt 2 resolves an attempt-1 review finding with zero readings. Round 2 of review reproduced it and then measured that the branch is BYTE-IDENTICAL to `origin/main`, so it is #223 behaviour this diff does not touch. Closing it changes #223's retirement rule and carries that rule's own sweep obligation.
- **Collapsing the two context writers into one `_store_context` would make the merge closed by construction** rather than proved by a census plus one behaviour test per writer. It is a genuine improvement and it is a refactor arriving at round 2 of review with its own risk, so it is a handoff.
- A per-phase adversarial budget would make the issue's second acceptance criterion reachable as literally written. Different issue.
- `astwalk.package_sources()` re-parses the package per call, with no cache and 54 call sites across 22 test files. The cache belongs in `astwalk`, not in this diff.
- The `entries` loop in `from_json` is as lenient about phase names as the readings loop was. There an unknown phase would `KeyError` inside `_buckets` rather than under-retire, so it is the worse of the two, and it is pre-existing rather than this PR's.


kstrl/pipeline.py
```diff
@@ -186,6 +186,44 @@ class DiffPhaseResult:
...
+
+    Review and security answer this the same way, and one definition is
+    the point: two copies drift into two meanings of "the reviewer
+    looked", and the looser one is the one that retires a finding. A
+    shared frozen base is not available, because narrowing ``result``
+    per subclass is an invariant-attribute override ``mypy --strict``
+    rejects; both concrete results declare ``infrastructure_error``, so
+    a function over the union is.
+    """
+    return ran and result is not None and not result.infrastructure_error
```
0xfauzi: **An advisory review that admits it did not see the whole change counts as a reading, and retires a live finding.**

`apply_coverage_check` (`kstrl/review.py:1006-1020`, `kstrl/security.py:606-624`) sets `diffstat_disagreement` in *every* mode and only sets `infrastructure_error = True` in HARD mode; it returns early otherwise. `run_review` then forces `result.passed = True` for ADVISORY at `kstrl/review.py:1142`. So an advisory reviewer whose numstat disagreed with git's returns `ran=True, passed=True, infrastructure_error=False` -> `produced_a_reading == True`, a `PhaseReading` is recorded, and `_buckets` moves the earlier attempt's review finding into `resolved`, where `format_for_prompt` reports it as "passed or were re-measured" and drops it from the prompt entirely.

#266's own words for that state are "the verdict was reached without the whole change in hand". This is the silent-drop direction this function's docstring says it exists to prevent - it excludes the crashed sensor and admits the blind one.

Reachable without a mode change: advisory review + set-point blocking. Attempt 1's `_setpoint_failure` writes a review-rank entry with `infrastructure=False`; attempt 2's advisory reviewer disagrees on the diffstat, passes, records a reading, and attempt 1's unfixed set-point finding is deleted from attempt 3's prompt.

```suggestion
    return (
        ran
        and result is not None
        and not result.infrastructure_error
        and not result.diffstat_disagreement
    )
```

tests/test_pipeline.py
```diff
@@ -2239,3 +2241,282 @@ def test_an_unparseable_journal_config_does_not_break_a_retry(
...
+        assert self._attempt(pipeline, manifest).transition == Transition.RETRYING
+        assert self._attempt(pipeline, manifest).transition == Transition.COMPLETED
+        pipeline.record_contract_failure("comp-a", 2, "tier 0 broke")
+
+        block = self._prompt_block(pipeline)
+        assert "tier 0 broke" in block
+        assert "SQL-IN-USERS" not in block
+        assert "from security passed or were re-measured in attempt 2" in block
+
+    def test_a_review_the_budget_skipped_retires_nothing(
```
0xfauzi, line 2445: **This test passes with the mechanism it pins deleted. Measured, not argued.**

I replaced `ComponentPipeline.adversarial_budget_ok` with `return True` (so the budget never skips anything), deleted `__pycache__`, and ran with `PYTHONDONTWRITEBYTECODE=1`: **1 passed**.

The cause is the stub. `run_review` is `lambda *a, **k: next(review_queue)` (line 2317) with one queued result, so a second call raises `StopIteration` - which is an `Exception`, so `_phase_review`'s `except Exception` catches it and synthesises a crashed advisory review (`passed=True`, `infrastructure_error=True`). A crashed reviewer also records no reading and also retires nothing, so the assertions cannot tell "the budget skipped the reviewer" from "the reviewer ran and blew up".

Same for `test_a_review_turned_off_by_the_operator_retires_nothing`: replacing the `review_mode == ReviewMode.SKIP -> _review_did_not_run` dispatch with `if False:` leaves **all six** tests in this class green.

These are two of the three pinnings the PR body offers for acceptance criterion 2, and neither pins the cause. Assert the cause directly - `outcome.review.ran is False` and the expected `skip_reason` - or have the stub raise a `BaseException` subclass the phase's handler cannot swallow so an unexpected call fails loudly.

kstrl/context.py
```diff
@@ -312,7 +392,7 @@ def _buckets(self) -> _Buckets:
                     resolved.append(entry)
                 else:
                     not_remeasured.append(entry)
-            elif entry.phase in SKIPPABLE_PHASES:
+            elif entry.phase in unread_skippable:
```
0xfauzi: **A `PhaseReading` for attempt N retires nothing when attempt N recorded no failure entry, so the fix does not reach the merge-conflict retry.**

When `latest` is empty, `ranks` is empty and `q = -1`. Every phase rank is >= 0, so the `PHASE_RANK[entry.phase] > q` branch three lines up catches every older entry and sends it to `not_remeasured` before this branch is ever consulted.

`_retry_after_merge_conflict` (`kstrl/pipeline.py:1905-1925`) is exactly that shape: it adds an `IterationRecord` for the attempt and no `FailureEntry`. So a reviewer that ran and cleared a finding in attempt 2, followed by a PR merge conflict, still shows the cleared finding.

Reproduced against the branch head:

```python
ctx = IterationContext()
ctx.add_review_finding('criterion X', attempt=1, phase='review')
ctx.add_iteration(IterationRecord(iteration=3, success=False, attempt=2, error='merge conflict'))
ctx.add_phase_reading('review', attempt=2)
print(ctx.format_for_prompt())
# ## Not re-measured since attempt 1
# (attempt 1, review) criterion X
```

The direction is safe, but the rule this PR states - "an entry from a skippable phase is retired only by an observation: a fresh entry at the same rank in attempt N, **or a `PhaseReading` for attempt N**" - is not what the code does, and `_buckets`' docstring does not disclose the exception. Either consult the readings before the `> q` branch for skippable phases, or write the exception down.

tests/test_phase_reading_sites.py
```diff
@@ -0,0 +1,160 @@
...
+    ``record_contract_failure``. Saying so explicitly is the point: a
+    guard whose scope is misread is how a clearing guard gets written by
+    accident.
+    """
+    assert_census(
+        sources=package_sources(),
+        sees=_writes_the_retry_context,
+        key=lambda source_file, _node: label(source_file),
+        expected={"pipeline.py": 2},
+        control=_CONTEXT_WRITE,
```
0xfauzi: **The `update`/`setdefault` disjunct has no control, so deleting it leaves this guard green.**

`assert_census`'s own contract (`tests/helpers/astwalk/net.py`, "ONE CONTROL PER DISJUNCT"): "A single control is a SCALAR proof over the whole predicate, so `anchor(node) or namer(node)` stays green on it with either half deleted... Pass a sequence and each string is proved separately."

`control=_CONTEXT_WRITE` is one subscript assignment. Measured: deleting the entire `ast.Call` / `update` / `setdefault` branch of `_writes_the_retry_context` and running this file gives **2 passed**. M10 proves the disjunct catches an `update()` writer *while it is present*; nothing catches its removal, which is the same skip-direction blindness the widening was added to close.

Pass a sequence, e.g. `control=[_CONTEXT_WRITE, "self.component_contexts.update({comp.id: ctx.to_json()})"]`.

tests/test_phase_reading_sites.py
```diff
@@ -0,0 +1,160 @@
...
+self._note_phase_reading(comp, "review", review.produced_a_reading)
+"""
+
+#: A write to the retry context, in the shape both writers use.
+_CONTEXT_WRITE = """
+self.component_contexts[comp.id] = ctx.to_json()
+"""
+
+
+def _records_a_reading(node: ast.AST) -> bool:
```
0xfauzi, line 137: **The recording-site census keys on the private helper, so a direct `ctx.add_phase_reading(...)` writer is invisible to it.**

The dangerous mutation this inventory exists to catch is a reading recorded for a phase that did not measure anything - and that can be written straight onto the context (`IterationContext.add_phase_reading`, `kstrl/context.py:278`) without ever touching `_note_phase_reading`. The census stays `{"review": 1, "security": 1}` and the docstring's claim ("One `_note_phase_reading` call per skippable phase, and none for any other phase") is then true and irrelevant.

`add_phase_reading` has exactly one caller in `kstrl/` today (`_merge_phase_readings`), so widening the predicate to see both names costs nothing now and closes the hole:

```suggestion
def _records_a_reading(node: ast.AST) -> bool:
    """Every call that records a phase reading, whatever shape its
    arguments take, at BOTH layers: the pipeline's recorder and the
    context API it delegates to. Keying on the private helper alone
    would leave a direct ``ctx.add_phase_reading(...)`` writer - the
    exact mutation that records a reading for a phase that measured
    nothing - invisible to this inventory.

    Deliberately not conditioned on arity or on the phase being
    readable. A conjunct like ``len(node.args) >= 2`` would make a site
    spelled ``_note_phase_reading(comp, phase="review", ...)`` invisible
    and leave the census green, which is the second-site case the tests
    below exist to catch, failing in the skip direction.
    """
    return isinstance(node, ast.Call) and leaf_name(node.func) in {
        "_note_phase_reading",
        "add_phase_reading",
    }
```

(`_phase_argument` already keys on the `phase` keyword or positional index 1, which matches `add_phase_reading(phase, *, attempt)` only for the keyword spelling - so this change also needs `_phase_argument` taught the one-positional shape, or the expectation adjusted to 2 per phase.)

kstrl/context.py
```diff
@@ -440,6 +527,18 @@ def from_json(cls, data: str) -> IterationContext:
...
             )
+        # Placed ahead of the two shape branches so one loop serves both.
+        # An absent key means no readings, which is exactly the rule
+        # that held before #247: a context written by an older parent
+        # process degrades to showing the finding rather than to
+        # dropping it.
+        for reading_data in parsed.get("readings", []):
+            ctx.readings.add(
+                PhaseReading(
+                    attempt=reading_data["attempt"],
```
0xfauzi: **`from_json` builds a `PhaseReading` without `_require_known_phase`, so the deserialiser and the adder hold two definitions of a valid reading.**

`add_phase_reading` calls `_require_known_phase`, whose docstring says it is "One definition of a phase name this object will accept... Two copies of the check are two definitions of the vocabulary, and the weaker one is the one that decides." This path is the weaker one, and it is the path that actually crosses the process boundary.

CLAUDE.md states the rule twice: "Validate against the SAME vocabulary the parser uses, from the same constant", and "Two definitions of a valid record mean the weaker one is the one the gate consults."

The `attempt` field has the same problem one field over: `reading_data["attempt"]` is taken verbatim, so a JSON `"2"` deserialises to a string that can never equal the integer `n` in `_phases_read_in`, and the record silently stops working with nothing failing. Run both through the same validation the writer uses.

kstrl/context.py
```diff
@@ -419,6 +499,13 @@ def to_json(self) -> str:
                 }
                 for e in self.entries
             ],
+            # Sorted because the source is a set and this string crosses
+            # a process boundary: unsorted, the same context would
+            # serialise differently under a different PYTHONHASHSEED.
+            "readings": [
+                {"attempt": attempt, "phase": phase}
+                for attempt, phase in sorted((r.attempt, r.phase) for r in self.readings)
```
0xfauzi, line 527: **The `sorted(...)` here is claimed load-bearing and nothing tests it.**

Measured: replacing it with the raw generator (`for attempt, phase in ((r.attempt, r.phase) for r in self.readings)`) leaves all 58 tests in `test_context.py`, `test_context_readings.py`, `test_phase_reading_sites.py` and `TestPhaseReadingsRetireSkippableFindings` green - **STILL GREEN**.

So either the determinism the comment above argues for matters, in which case it needs an assertion (build the same context two ways, compare `to_json()` byte for byte), or it does not, in which case the comment's reasoning is wrong: crossing a process boundary needs round-trip fidelity, which a set already gives, not a stable byte ordering.

tests/test_context_readings.py
```diff
@@ -0,0 +1,237 @@
...
+        """Diff the two sweeps rather than restating either.
+
+        The entries that change bucket between ``TestBucketRuleSweep``'s
+        contexts and these are exactly the skippable-phase entries
+        ranked strictly below ``q``. Asserting the difference is what
+        makes this a statement about the record's effect rather than a
+        second copy of the rule.
+        """
+        moved = sum(self.check_one_difference(seq) for seq in sweep_sequences())
+        assert moved > 0
```
0xfauzi: **`assert moved > 0` does not assert what the test's name claims.**

Measured on the branch head: `moved` is **810**. The assertion holds at 1. The name says "moves exactly the entries it should" and the docstring says "Asserting the difference is what makes this a statement about the record's effect" - but the only quantitative claim is non-emptiness, so a change that retires 809 fewer entries passes here.

`check_one_difference` does prove each moved entry is one that MAY move, and the sibling partition test catches the whole-set regressions, so the gap is narrow. Still, this repo pins its censuses to exact numbers for exactly this reason. `assert moved == 810` (or a per-sequence expectation derived from the rule) is the version that fails when the effect shrinks.

kstrl/pipeline.py
```diff
@@ -1227,6 +1284,10 @@ def begin_attempt(self, comp: Component) -> None:
         comp.status = ComponentStatus.RUNNING.value
         comp.started_at = _iso_now()
         self.component_failure_signatures.pop(comp.id, None)
+        # #247: the readings describe the attempt in flight, so the
+        # attempt boundary is where they are cleared. Anything the
+        # previous attempt observed is already in the context's JSON.
+        self._phase_readings.pop(comp.id, None)
```
0xfauzi, line 1321: **This line is unguarded: replacing it with `pass` leaves every test green.**

Measured, `__pycache__` cleared, `PYTHONDONTWRITEBYTECODE=1`: **58 passed**. The quality pass rejected removing this pop "because it bounds the dict", but nothing fails if a later author removes it anyway - the readings are attempt-scoped, so stale entries are inert for bucketing and only cost memory and a growing `readings` list in every serialised context.

Per CLAUDE.md, "A guard you did not mutate is a guard you did not test." Either add the one assertion that the dict does not accumulate across attempts, or say in the comment that the pop is hygiene with no behavioural consequence, so the next reader does not have to re-derive it.

tests/test_pipeline.py
```diff
@@ -2239,3 +2241,282 @@ def test_an_unparseable_journal_config_does_not_break_a_retry(
...
+            config=_factory_config(
+                max_retries=5,
+                review_mode="advisory",
+                security_config=SecurityConfig(mode="advisory"),
+                max_adversarial_calls=1,
+                create_prs=True,
+                pause_before_pr_merge=True,
+            ),
+            reviews=[
+                ReviewResult(passed=False, mode="advisory", overall_notes="CRITERION-X-UNMET"),
```
0xfauzi, line 2471: **`ReviewResult(passed=False, mode="advisory")` is a combination `run_review` can never return.**

`kstrl/review.py:1135-1142` forces `result.passed = True` for every ADVISORY review, after `apply_coverage_check` and after downgrading every FAIL verdict. So attempt 1 of this test (and of `test_a_crashed_review_retires_nothing` at line 2467) drives `_review_failure` from a state production cannot reach - an advisory reviewer that fails the component on its own verdict.

It still exercises the retry path, so the test is not wrong about attempt 2. But the setup misdescribes how an advisory run actually accumulates a review-rank entry (in production that comes from `_setpoint_failure`, or from an earlier hard-mode attempt), which matters because it is the same asymmetry finding 1 turns on: in advisory mode the reviewer's own verdict never lands as an entry, but its coverage disagreement still records a reading.

kstrl/pipeline.py
```diff
@@ -186,6 +186,62 @@ class DiffPhaseResult:
...
+    shared frozen base is not available, because narrowing ``result``
+    per subclass is an invariant-attribute override ``mypy --strict``
+    rejects; both concrete results declare ``infrastructure_error``, so
+    a function over the union is.
+    """
+    return (
+        ran
+        and result is not None
+        and not result.infrastructure_error
+        and not result.diffstat_disagreement
```
0xfauzi, line 241: **An advisory reviewer that still reports the criterion unmet counts as a reading, and retires it.**

`run_review` forces `passed = True` for every ADVISORY run (`review.py:1134-1142` downgrades each FAIL verdict to ADVISORY and then sets `result.passed = True`), and `_passes_threshold` returns True unconditionally for advisory security (`security.py:587`). So in advisory mode `passed` carries no information about what the reviewer found. `_produced_a_reading` never looks at `passed` or at the criterion verdicts, so "the reviewer looked and cleared it" and "the reviewer looked, said criterion X is STILL unmet, and was forced to report pass" are the same reading.

An advisory review also writes no `FailureEntry`, so its own verdict never reaches the retry context either. The engineer's next prompt loses both.

Measured, on this branch, by adding one method to `TestPhaseReadingsRetireSkippableFindings` (attempt 1 fails review in hard mode; the operator drops to advisory exactly as `_blind_review_scenario` does; attempt 2's reviewer returns `criteria=[CriterionReview(criterion="criterion X", verdict=ADVISORY, explanation="criterion X is STILL unmet")]` with `passed=True`; security then fails):

```
=== PREVIOUS ATTEMPT CONTEXT (Attempt 3) ===

## Current failures (measured in attempt 2, security)
Security review infrastructure error: SQL-IN-USERS

## Resolved or superseded
1 earlier finding(s) from review passed or were re-measured in attempt 2 and are omitted.
```

`second.review.produced_a_reading is True`. Before this PR `CRITERION-X-UNMET` rendered under `## Not re-measured` with "re-check them yourself". Now it is gone and the still-unmet advisory verdict is nowhere in the prompt. That is the silent drop of a live finding the PR body says the halt-over-heroics doctrine rules out, and `test_a_review_git_agrees_with_still_retires_its_own_finding` currently enshrines it (its stub is `ReviewResult(passed=True, mode="advisory")`, which stands in for exactly this state).

The two ways an advisory run accumulates a review-rank entry are the two the helper's own docstring names, so this is reachable on both: an earlier hard-mode attempt, and `_setpoint_failure` under `setpoint_agreement = "block"`.

Suggested predicate: in advisory mode key the reading on the reviewer's own verdicts, not on `passed` - e.g. refuse the reading when `result.criterion_advisory_count` (or any advisory concern that was a downgraded fail) is non-zero, or record the reading only when `mode == HARD`. Whatever the shape, `passed` cannot be the input in a mode where it is a constant.

kstrl/context.py
```diff
@@ -305,17 +405,14 @@ def _buckets(self) -> _Buckets:
...
             elif entry.attempt == n:
                 current.append(entry)
-            elif PHASE_RANK[entry.phase] > q:
-                not_remeasured.append(entry)
             elif PHASE_RANK[entry.phase] == q:
-                if q in measured_ranks:
-                    resolved.append(entry)
-                else:
-                    not_remeasured.append(entry)
+                (resolved if q in measured_ranks else not_remeasured).append(entry)
```
0xfauzi, line 412: **Two definitions of "this phase produced a reading in attempt N" now live in `_buckets`, and the weaker one still decides.**

The readings route (line 411) is gated on `_produced_a_reading`, whose fourth conjunct exists to refuse a reviewer whose diffstat git disagrees with. This branch is gated on `q in measured_ranks`, i.e. `not entry.infrastructure`, which does not know about `diffstat_disagreement` at all.

The gap is reachable. `apply_coverage_check` sets `diffstat_disagreement` and returns early in advisory mode (`review.py:1019`), `run_review` then forces `passed=True`, so `infrastructure_error` stays False. With `setpoint_agreement` blocking, `_setpoint_failure` writes the review-rank entry for attempt N with `infrastructure=review_result.infrastructure_error` = **False** (`pipeline.py:3721`). Demonstrated:

```python
ctx = IterationContext()
ctx.add_review_finding("criterion X", attempt=1, phase="review")
ctx.add_review_finding("setpoint disagreement", attempt=2, phase="review", infrastructure=False)
# readings: set()
# resolved: [(1, 'review', 'criterion X')]
```

"criterion X" is retired with no reading at all, on a verdict #266 describes as reached without the whole change in hand - the exact state the new conjunct refuses one branch lower.

CLAUDE.md: "Two definitions of a valid record mean the weaker one is the one the gate consults." `_produced_a_reading` should be the one definition on both routes: derive `measured_ranks` from the same predicate (record a reading for a failing phase too - the PR already does - and test membership in `read_in_n` here instead of `measured_ranks`), or at minimum have `_review_failure`/`_setpoint_failure` set `infrastructure=True` when `diffstat_disagreement` is non-empty.

Note that `test_a_review_git_says_did_not_see_the_whole_change_retires_nothing` cannot catch this: its failing gate is security (rank 4), so the review entry goes through the skippable branch and never reaches `rank == q`.

tests/test_phase_reading_sites.py
```diff
@@ -0,0 +1,318 @@
...
+
+#: The two names a phase reading can be written through: the pipeline's
+#: private recorder and the context's own public method. Both, because
+#: a census that watched only the private one would clear the dangerous
+#: mutation: a reading for a phase that measured nothing can be written
+#: straight onto the context with ``ctx.add_phase_reading(...)`` and
+#: never touch the recorder. Measured in round 1 of review: with only
+#: the private name in this set, that planted writer left the file 2
+#: passed.
+_READING_WRITERS = frozenset({"_note_phase_reading", "add_phase_reading"})
```
0xfauzi, line 64: **The recording census is blind to a third writer, and this one is not disclosed.**

`_READING_WRITERS` is documented as "the two names a phase reading can be written through", but `IterationContext.readings` is a public mutable `set` (`context.py:223`), so `ctx.readings.add(PhaseReading(...))`, `|=` and `.update(...)` are all writers. They are invisible to `_records_a_reading` (which keys on `leaf_name(node.func)`) **and** they bypass `_require_known_phase`, which `from_json` was deliberately routed through so there would be exactly one definition of the vocabulary.

Measured on this branch: planting

```python
from kstrl.context import PhaseReading
ctx.readings.add(PhaseReading(attempt=1, phase="review"))
```

inside `_merge_phase_readings` leaves this file at **3 passed, 1 xfailed**. That is the same shape as the round-1 miss the `_PUBLIC_RECORDING_CALL` control was added for, one spelling over.

CLAUDE.md's guard rule: prefer **closed by construction** over a ledger of names, and express a genuine miss as an explicit `blind_spot` row plus a strict xfail (which this file does for the local-alias miss on the *other* net, and not for this one). Two options that close it rather than widen it: make the field private (`_readings`) with `add_phase_reading` the only writer and a read-only view for tests, or add a `spells("readings")`-style outer layer so any expression naming the attribute is counted.

kstrl/context.py
```diff
@@ -440,6 +544,23 @@ def from_json(cls, data: str) -> IterationContext:
...
+        # process degrades to showing the finding rather than to
+        # dropping it.
+        #
+        # Through ``add_phase_reading`` rather than straight into the
+        # set, so the read path checks the phase name against the same
+        # ``_require_known_phase`` the write path does. A deserialiser
+        # that accepted a name the writer refuses would be the second,
+        # weaker definition of the vocabulary that helper exists to
+        # prevent, and the weaker one is the one that decides.
+        for reading_data in parsed.get("readings", []):
```
0xfauzi, line 562: **`from_json` is now a raising parser, and its callers are on a path documented as un-wrapped.**

Three new ways this loop raises on a payload that previously round-tripped as an opaque string: `ValueError` from `_require_known_phase`, `KeyError` from `reading_data["phase"]` / `["attempt"]`, and `TypeError` if a `readings` element is not a mapping.

Before this PR `retry_or_fail` stored `context_json` verbatim and could not raise. It now parses (`pipeline.py:1487`), and so does `record_contract_failure` (`pipeline.py:1432`). `_setpoint_failure`'s own comment states the consequence: "process_result is not wrapped by its caller (factory.py, the scheduler loop), so an exception escaping here would abort the whole run and take every other component's work with it." The contract-loop caller is no better placed.

`test_a_reading_naming_an_unknown_phase_is_refused_on_read` pins the refusal but not where it lands. CLAUDE.md: "Route the refusal through the surface the other pre-spend refusals use ... not out of the top of the run as a traceback."

The safe shape is a per-component degrade: catch the malformed-readings case at the two pipeline writers and continue with no readings (which under-retires, the direction this PR already calls safe) plus a warning, rather than letting one component's bad context kill every sibling's work.

kstrl/pipeline.py
```diff
@@ -1304,6 +1388,52 @@ def _record_failure_signatures(
                 signature_for_error(phase or "unknown", error),
             ]
 
+    def _note_phase_reading(self, comp: Component, phase: str, produced: bool) -> None:
```
0xfauzi, line 1399: **The record only survives through a gate that stores a context, and nothing says which gates drop it.**

`_note_phase_reading` writes into `self._phase_readings`, and the only two things that ever read it are `retry_or_fail` (under `if context_json:`) and `record_contract_failure`. Every other terminal transition silently discards the attempt's readings: `_park_merge_pending`, `_fail_pr_flow`, `fail_for_budget` (the `breached_ceiling` return in `_phase_review`/`_phase_security` records a reading and then routes to a gate that stores nothing), `_coverage_failure` (FAIL, no context), and `retry_or_fail`'s own retries-exhausted branch. So does any resumption in a fresh process - `ks inbox retry` on a parked component starts with an empty dict.

Every one of those is the safe direction (under-retire), which is exactly why it will not show up as a failing test. The cost is that the mechanism is off for those paths with nothing recording it. The class comment says "In-run only", which covers the process boundary but not the terminal gates.

Worth stating in the docstring at minimum: which gates carry the record and which drop it, so a later reader does not assume the record is complete. `_park_merge_pending` is the one that actually matters, because an inbox retry resumes and the readings are gone.

tests/test_phase_reading_sites.py
```diff
@@ -0,0 +1,318 @@
...
+            "SKIPPABLE_PHASES plus the two fan-out points. A phase in "
+            "that set with no site is a finding that an observed pass "
+            "can never retire; a phase outside it with a site "
+            "duplicates the rank rule; a new location row is a writer "
+            "whose phase nobody can read from the source."
+        ),
+    )
+
+
+def test_the_retry_context_has_exactly_two_writers_and_both_are_in_the_pipeline() -> None:
```
0xfauzi, line 308: **Altitude: this census proves LOCATION, and the property that matters is MERGING - which its own docstring admits it cannot show.**

"It does not prove the two merge: a writer could be added inside `pipeline.py` that ignores the record and this stays green." That is the failure mode the PR exists to close (`factory.py` was the writer that did not merge), and the guard is one indirection away from being closed by construction.

The simpler, more general change: give `ComponentPipeline` one private `_store_context(self, comp_id: str, ctx: IterationContext) -> None` that merges and writes, have both `retry_or_fail` and `record_contract_failure` call it, and pin this census at `{"pipeline.py": 1}`. Then a third writer that forgets the merge is not merely located, it is impossible - the only subscript assignment to `component_contexts` is inside the function that merges - and the census stops needing a docstring explaining what it does not say.

That also removes the need for the two per-writer behaviour tests to carry the whole proof, and it shrinks the disclosed local-alias blind spot to one function.

tests/test_phase_reading_sites.py
```diff
@@ -0,0 +1,318 @@
...
+    attempt=...)`` takes it first, ``_note_phase_reading(comp, phase,
+    produced)`` second. Keying both off ``args[1:2]`` would file
+    ``ctx.add_phase_reading("security", attempt=1)`` under the location
+    row instead of under ``security``, which still FAILS the census but
+    reports the wrong thing.
+    """
+    return folded_str(_phase_argument_node(node)) or _enclosing_site(source_file, node)
+
+
+def _enclosing_site(source_file: Path, node: ast.AST) -> str:
```
0xfauzi, line 219: **Reuse: node -> innermost scope already has two implementations in `tests/`, both exact.**

`_enclosing_site` derives the innermost function by scanning every scope's `[lineno, end_lineno]` range and keeping the greatest `lineno`, and re-reads the source file (`parsed()` caches the AST by text but calls `source_file.read_text()` every time) once per unkeyable node.

`tests/test_append_opens_have_one_home.py:480` already builds the same map exactly, with no line arithmetic:

```python
owner: dict[int, str] = {}
for node, qualified in scopes(tree):
    for child in own_nodes(node):
        owner[id(child)] = qualified
```

`own_nodes` stops at a nested function, so a helper defined inside another one is credited to itself - the property the `scope.lineno > best` heuristic is reaching for. `tests/test_settle_discipline.py:487` has `enclosing_function(node, parents)` for the parent-map flavour.

If the map form is adopted, this belongs in `tests/helpers/astwalk/scope.py` next to `scopes`/`own_nodes` rather than as a third private copy - the third copy is how the sixteen-guard sprawl #324 records started.

tests/test_phase_reading_sites.py
```diff
@@ -0,0 +1,318 @@
...
+
+    The census below pins two rows whose phase the walk cannot fold. A
+    location pinned in a table is a claim about the tree, so it is
+    measured here rather than trusted: each named site really does hold
+    exactly the number of unkeyable writers the table gives it, and
+    ``_enclosing_site`` really does resolve to that qualified name. A
+    typo in either row would otherwise sit in the table forever, failing
+    nothing, while the row it was meant to pin went unwatched.
+    """
+    found: dict[str, int] = {}
```
0xfauzi: **Reuse: this loop is `astwalk.census` written out.**

`census(sources, sees, key)` in `tests/helpers/astwalk/net.py:80` is exactly `for source_file in sources: for node in all_nodes(parsed(source_file)): if sees(node): built[key(...)] += 1`, and it is already imported into this file's namespace via `assert_census`. The hand-rolled version can be:

```python
found = census(
    package_sources(),
    lambda node: _records_a_reading(node) and not folded_str(_phase_argument_node(node)),
    key=_enclosing_site,
)
```

Same answer, and it stops this file having two ways to count the same corpus - which is what lets the two drift when someone edits one.

tests/test_pipeline.py
```diff
@@ -2239,3 +2242,503 @@ def test_an_unparseable_journal_config_does_not_break_a_retry(
...
+    """
+
+    def __init__(self, hook: str) -> None:
+        super().__init__(
+            f"{hook} was called and the test queued no result for it: "
+            "a phase this test says did not run, ran."
+        )
+
+
+class TestPhaseReadingsRetireSkippableFindings:
```
0xfauzi, line 2270: **Test placement: shared helpers in a test module, and 503 lines into a file already at 2241.**

Two things pull in opposite directions in the same diff. `tests/test_context_readings.py` was created because "`tests/test_context.py` is at 687 lines against an 800-line ratchet" - for 29 lines of growth - while this class puts 503 lines into `tests/test_pipeline.py`, taking it from 2241 to 2744. The ratchet allows it (already-oversize files are reported, not failed), but the two decisions in one PR do not agree.

Separately, `section`, `CURRENT`, `NOT_REMEASURED`, `RESOLVED`, `build_sweep_context` and `sweep_sequences` are now imported out of `tests/test_context.py` by two other test modules. `tests/helpers/` is this repo's home for exactly that (`tests/helpers/procs.py`, `tests/helpers/astwalk/`), and the corpus docstring in `astwalk/corpus.py` records what a from-import out of a collected module costs when a name happens to start with `test` - which `sweep_sequences` narrowly does not. Moving the six names to `tests/helpers/context_prompt.py` removes the hazard and the cross-module coupling in one step.

Issue #257, opened 2026-08-28
The architect phase records no usage: `ks decompose` reports no cost at all, and the factory cost cap cannot see architect spend


0xfauzi:
Found while dogfooding kstrl as a first-time user, building a real web app from a spec (`writers-room`). Measured on a real paid run, not read out of the code alone.

## What happened

I ran the architect over a 957-word spec:

```
cd ~/Documents/code/writers-room
KSTRL_NO_TUI=1 ks decompose --spec docs/spec.md --project-name writers-room --ui plain --no-color --no-tui
```

It worked. It found 21 spec issues (7 blocker, 11 major, 3 minor), halted correctly, wrote `scripts/kstrl/spec-issues.json`, and exited 2. No complaint about any of that.

**It never told me what it cost, and there is nowhere to look it up.**

- 238 lines of output, zero lines matching `cost|usd|token|usage`.
- `.kstrl/progress.jsonl` was never created.
- `.kstrl/evolution.jsonl` holds one `spec_issues` record, no cost field.
- `.kstrl/runs/decompose-.../events.jsonl` holds 111 events. Exactly one mentions cost, and it is the plan stating the caps, not the spend:

```json
{"event": "run_plan", "component": "architect", "data": {"max_total_tokens": 0, "max_adversarial_calls": 0, "max_cost_usd": 0.0}}
```

There is **no `component_usage` event for the architect**. Compare a real factory run on an external project (deckgen, 2026-08-03), which emitted one per phase:

```json
{"event": "component_usage", "data": {"phase": "engineer", "cost_usd": 4.269656, ...}}
{"event": "component_usage", "data": {"phase": "review", ...}}
{"event": "component_usage", "data": {"phase": "security", ...}}
{"event": "component_usage", "data": {"phase": "distill", "cost_usd": 0.998211, ...}}
```

Four phases, four usage events. The architect is the fifth LLM role and emits none.

## Why it happens

`_record_usage` (`kstrl/pipeline.py:478`) has exactly four call sites, and none is the architect:

- `pipeline.py:1829` and `record_engineer_usage` at `pipeline.py:528` - engineer
- `pipeline.py:2676` - review
- `pipeline.py:3136` - security
- `pipeline.py:3364` - distill

`kstrl/decompose.py` contains no reference to `_record_usage`, `collect_usage`, `component_usage` or `UsageTotals`. I grepped for all four.

## The second consequence, which is the one that matters

[...]

Issue #193, opened 2026-08-03
Split the factory composition root (RunContext / RunServices / RunState) before R8.7


0xfauzi:
> **Stabilisation queue: P3, rank 16 of 21** (set 2026-09-09; the full ordered list and the coordinator's rules are in #369). Re-plumbing sequenced before R8.7; large and destabilising if rushed. On hold until the owner lifts the P3 hold in #369.

## Problem

`kstrl/factory.py` is 3,344 lines importing 31 kstrl-internal modules; it is simultaneously the workflow coordinator and the integration point for every feature. `ComponentPipeline.__init__` (`kstrl/pipeline.py:264-287`) takes 22 keyword-only parameters - roughly 9 config, 6 services/seams (including a 9-callable `PipelineHooks`), and 6 shared mutable structures its own docstring enumerates - plus 9 self-owned mutable attributes.

Every R8 feature so far has added parameters to both. R8.7 (release stage) touches manifest states, resume, the TUI, Linear sync, and evolution consumers; doing the split after R8.7 lands is strictly more expensive than before.

## Shape

```python
@dataclass(frozen=True)
class RunContext:   # identity + validated config
    root: Path; run_id: str; manifest_path: Path; config: AppConfig

@dataclass(frozen=True)
class RunServices:  # injected seams
    agents; git; evaluators; events: EventBus; interaction; ui; clock

@dataclass
class RunState:     # the shared mutable set, in one named place
    manifest; factory_result; worktree_paths; component_contexts
    fresh_base_retry_ids; component_failure_signatures; usage
```

The scheduler and pipeline then depend on three stable objects instead of accumulating one parameter per feature.

## Constraints

- No behavior change; this is a pure re-plumbing.
- `uv run mypy kstrl/ --strict` stays clean.
- Test churn limited to construction sites.
- Sequenced BEFORE R8.7 (#154).
- Pairs naturally with the config consolidation issue (load once, pass down); doing them together avoids re-plumbing twice.

Filed from the 2026-08-03 architecture-review session (user-approved item P11).

PR #359, opened 2026-09-07
fix(config): one kstrl.toml load per run, carried as a frozen envelope (#192)

### LINKED ISSUE #192: Config: load kstrl.toml once per run - mid-run re-loads diverge from the recorded policy hash
## Problem

Configuration is loaded independently, repeatedly, and at different times by 21 `*Config` dataclasses. Every `.load()` re-opens and re-parses `kstrl.toml` (`kstrl/config.py:366-384` parses on every call). `cli.py` loads `KstrlConfig` four separate times (:549, :807, :1179, :2217); `launch.py:76-94` is a parallel composition path re-loading seven configs.

## Motivating defect: audit-trail integrity

The pipeline re-loads `PolicyConfig`, `AdequacyConfig`, and `AutonomyConfig` from disk **per component, mid-run** (`kstrl/pipeline.py:1973-1976`), while `manifest.policy_hash` is computed once at run start (`kstrl/factory.py:2377`). An edit to `kstrl.toml` while a factory run is in flight makes the enforcement actually applied diverge from the envelope hash the manifest records. The hash exists precisely so "what policy was this run held to?" has one answer; today it can have two.

## Historical precedent

The progress writer/reader reconciliation shim (`kstrl/config.py:73-159`) exists because two independently loaded config domains drifted, twice - its own docstring records the second-order bug where the strings compared equal while the runtime paths did not.

## Fix

- Compose an `AppConfig` once at process entry (cli/launch) from the existing section dataclasses; pass validated objects down.
- Feature modules stop re-opening `kstrl.toml`; the per-section `from_env()`/`load()` methods remain for tests and section semantics, but production wiring calls them exactly once.
- Mid-run re-loads in `pipeline.py` are removed; the configs used are the ones the hash was computed from.

## Acceptance

- One parse of `kstrl.toml` per process.
- A mid-run edit to `kstrl.toml` has no effect on the running factory (test).
- The policy hash provably matches the enforced envelope for the whole run.

Filed from the 2026-08-03 architecture-review session (user-approved item P10).

0xfauzi:
## Summary

`manifest.policy_hash` exists so "what policy was this run held to?" has one answer. It had two. The hash was computed once at run start while Phase 1 re-read `[policy]`, `[adequacy]` and `[autonomy]` from `kstrl.toml` for every component, so an edit to `kstrl.toml` while a run was in flight changed what later components were held to without changing the record of it.

Measured at `414d662` on a two-component run with `[autonomy] enabled = false`, which is the default: component B was enforced at `max_files_changed=500`, `deps_allow_new=true` against a manifest recording the hash of `5` and `false`; the adequacy posture flipped from enabled to disabled with nothing recording either posture; and a malformed mid-run edit (`max_files_changed = "two"`) raised `ValueError` out of `_phase_verify` and aborted the whole run rather than failing one component.

A frozen `RunEnvelope` (`kstrl/runenvelope.py`) carries every section a run enforces: `[policy]`, `[adequacy]`, `[autonomy]`, the stored ladder state, `[sandbox]`, `[fixtures]`, `[inbox]` and `[divergence]`. `_run_factory_locked` resolves one before the run directory exists, the autonomy ladder clamps THAT object before the pipeline is constructed, `ComponentPipeline` takes it as a required keyword parameter with no default and no `or ...load()` fallback and resolves nothing of its own, and `manifest.policy_hash` is read back off the pipeline. This is #269 applied to config: `RunScope` already carries the run's plan-time path snapshot for the same reason and says so in `pipeline.py`.

Round 2 also makes a section the run cannot resolve a refusal rather than a traceback. That is the round-1 review's blocker 1 and it is the reason the envelope grew from three sections to seven; the detail is under "Round 2".

The issue's acceptance 1, "one parse of `kstrl.toml` per process", is NOT met and this PR does not claim it. `cli.py` alone holds 41 of the 48 command-entry sites; collapsing them is the composition root, which is #193. See Handoffs.

## Round 2: what the review found and what changed

Round 1 was reviewed at `9a32ef9`: 2 blockers, 5 should-fix, 5 nits. Every one is answered below, with the measurement that closes it.

### Blocker 1: a malformed `[inbox]` killed the run as a traceback, above the architect-spend record

Round 1 moved four config loads into `ComponentPipeline.__init__`, which runs in front of the whole run with no handler above it. Driven on both trees over an empty manifest, `__pycache__` purged:

| `kstrl.toml` | `origin/main` | round 1 (`23b5f40`) | round 2 |
|---|---|---|---|
| `[inbox] open_item_cap = "many"` | returned, exit 0 | **escaped `ValueError`** | **exit 2, names `[inbox]` and `open_item_cap`** |
| `[policy] max_files_changed = "two"` | escaped `ValueError` | escaped `ValueError` | **exit 2, names `[policy]` and `max_files_changed`** |

The second half is the ordering. `record_architect_usage` is the first thing done to the pipeline and the comment above it states the invariant in its own words: "nothing between the run directory's sinks and this line returns early... An early exit inserted above would break it silently", with the named consequence that `serve` charges $0 for a launch that spent real money (#257). Round 1's resolution sat between the sinks and that line.

The fix resolves the whole envelope at the TOP of `_run_factory_locked`, **above the run directory's sinks**, and refuses through `_report_preflight` with exit code 2, the surface every other pre-spend refusal uses (the rule CLAUDE.md already states from #260). Because the refusal happens before the run directory exists, there is no run to report $0 for and no early exit inside the protected window. `config_preflight.load_or_report` is split so `resolve_or_report` produces the same rejection line without a `toml_parse_scope` of its own: a nested scope replaces the outer cache rather than inheriting it, so the per-call variant would cost one parse per section.

The window is not theoretical. The entry preflight resolves all 22 sections before the command body, but on `ks factory --spec` the architect runs between that check and this one, at a measured 119 to 210 seconds against a frontier model.

`blame_env` is False in the envelope's resolution, and the reason is stated rather than defensive: naming an environment variable means clearing `os.environ` process-wide and the caller may be a TUI worker thread, and it cannot lose anything, because the entry check already resolved every one of these sections with `blame_env=True` and a running process's environment does not change underneath it. The only input that can have gone bad since is the file, which `_blamed_toml_value` still names.

### Blocker 2: `factory_config.policy_config` was an input and an output at once

Deleted. Unconditional, the field was read as the injection seam at the envelope resolution and written back with the clamped envelope at the ladder, so a second `run_factory` on the same `FactoryConfig` resolved nothing from `kstrl.toml`. The reviewer measured it: with `max_files_changed` edited from 5 to 777 between two runs, run 2 enforced 777 on `origin/main` and 5 on round 1, and recorded a hash that did not match disk.

`TestTheEnvelopeDoesNotOutliveItsRun` now pins the cross-run behaviour and asserts `policy_config` is still `None` after a run. Its one reader, `tests/test_autonomy_ladder.py`, reads `pipeline.run_envelope.policy`. The round-1 body's claim that M5 was an equivalent mutant is **withdrawn**: it was not, and restoring the write is caught (R2M4 below).

### Should-fix 3: the guard was blind to five call shapes

`tests/helpers/configwalk.py` now walks `ast.Lambda` as a scope of its own and resolves a surface reference through `astwalk.bindings` rather than matching text. It counts REFERENCES rather than calls, which is what makes `partial(PolicyConfig.load, root)` and `f = PolicyConfig.load` visible at all. Measured, planted into a `ComponentPipeline` method, offenders reported by layer B:

| shape | round 1 | round 2 |
|---|---|---|
| `PolicyConfig.load(...)` (control) | 1 | 1 |
| `mod.PolicyConfig.load(...)` | 0 | **1** |
| `PC.load(...)` after `import ... as PC` | 0 | **1** |
| `getattr(PolicyConfig, "load")(...)` | 0 | **1** |
| `partial(PolicyConfig.load, root)()` | 0 | **1** |
| `f = PolicyConfig.load; f(root)` | 0 | **1** |
| `(lambda: SecurityConfig.load(root))()` | 0, and layer A's census untouched | **1** |
| `(lambda: load_toml_section(...))()` | 0 by layer B, caught by layer A | **1 by layer B, and layer A attributes it** |
| list comprehension | 1 | 1 |

`test_the_walk_flags_a_per_phase_read_however_it_is_written` is one parametrized row per shape. Two spellings are offered for every node and the caller keeps whichever names a config class: trying the literal one and stopping there is what an earlier draft of this fix did, and it lost `PC.load`, because `PC` is not a config class and the walk never asked the resolver.

The DISCLOSED LIMIT moves rather than disappearing. Layer 1 still matches a primitive by bare `Name` only, so `c.load_toml_section` after `import kstrl.config as c` enrols nothing, and both the raw and the attributed census stay at zero together. `test_a_module_qualified_primitive_is_invisible` is the strict xfail behind that paragraph. Measured in `kstrl/`: 54 bare-name primitive calls, 0 in the attribute form, so it is latent.

One shape was fixed while writing this: dropping the literal spelling entirely took `serve.py` from 10 sites to 8, because `astwalk.bindings` deliberately does not bind a class DEFINED in the module being walked, so `ServeConfig.load` inside `kstrl/serve.py` resolves to nothing. That is the skip direction, and the literal spelling stays in front of the resolver for it.

### Should-fix 4: layer B covered `pipeline.py` only

It now walks `kstrl/pipeline.py` (0 sites, every read gone), `kstrl/factory.py` (7 sites, allowed in `_run_factory_locked` and `_open_health_breach_items`) and `kstrl/serve.py` (10). The reviewer's plant, `_p = PolicyConfig.load(root_dir)` as the first statement of `_submit_args`, which runs once per component, passed the whole round-1 file; it is now caught (R2M9).

`pipeline.py`'s count is zero, so it cannot be its own control, and that is said where the pin is: the control is the other two modules, walked by the same call in the same run, whose counts are not zero.

### Should-fix 5: the malformed-edit test never ran a run

It called `_phase_verify` in a `for` loop, so `process_result`, the scheduler and the `try` around `future.result()` were never entered. `TestAMalformedMidRunEditDoesNotAbortTheRun` drives the real scheduler over two components with the engineer stubbed and the edit made by the first component's verification hook. **Red first at `origin/main`**: `ValueError: invalid literal for int() with base 10: 'two'` escapes `run_factory` and neither component is verified. The phase-level test keeps its assertion under a name that says what it asserts and points at the run-level one.

### Should-fix 6: `AutonomyState.load` ran twice per ladder-on run

The envelope carries the state and the ladder reads it off there. This is done by REMOVING the second read rather than by adding a `state=` parameter, which is the same fix one mechanism smaller. `TestTheLadderStateIsReadOnce` counts the loads up to pipeline construction and pins them at 1; no parse count covers this, because the ladder state is JSON rather than `kstrl.toml`, which is how a 17.2 ms duplicate nine lines from the first survived a review. The epilogue's read is a different question and is asserted separately: `_record_autonomy_outcome` re-reads AFTER the run because `apply_demotion` may have written the file while components ran.

### Should-fix 7: the pipeline was built with the unclamped envelope and patched 155 lines later

The ladder now RESOLVES before the pipeline is constructed, so the pipeline holds the clamped envelope by construction, and `pipeline.run_envelope = run_envelope` is deleted. The reporting stays where it was, and the split is the point: the bus and its file sinks do not exist at the decision point, and `_adversarial_phase_gates` runs in between and must keep reading the CONFIGURED `review_mode` rather than the bundle's. `_LadderOutcome` carries the level, the bundle, the clamps and the overrides across that gap.

### Nits

- **8, the census.** The PR body's 99 came from a script not in the repo, and the shipped walk said 98. Both numbers are gone: every census figure below is `tests/helpers/configwalk.py`'s, re-derived by running it, and the command is in this body.
- **9, hedging.** The CHANGELOG stated as measured what the body called READ. It now names the mechanism and says a run was driven both before and after, which is what round 2 does.
- **10, the timing comment.** It claimed 0.617 ms for "the one document parse" against 0.538 ms for "the single `SandboxConfig` load it replaced"; the reviewer measured those labels transposed (0.524 ms and 0.609 ms). The claim is deleted rather than relabelled: round 2 moves those loads out of the constructor, so construction is pure attribute assignment again and there is nothing to justify.
- **11, the daemon control.** `>= 9` is now `== 10`.
- **12, `[sandbox]` twice in one run.** Fixed as one envelope field, so it did not need a handoff. `ComponentPipeline.__init__` and `_run_factory_locked` both read the envelope.

## Round 3: what the second review found

Round 2 was reviewed at `6012553`: 1 blocker, 4 should-fix, 8 nits, plus a false-positive nit answered rather than fixed. `origin/main` `037f0e1` (#360) is merged in as `111e232`. Every number below was re-measured on the merged tree.

### Blocker 1: the config surface did not contain `RunEnvelope`, so the pipeline guard could not fire on the one fallback the constructor forbids by name

Round 2's layer 1 enrolled a class only when one of its own methods called a parse primitive by bare name. `RunEnvelope.resolve` reaches `kstrl.toml` through `config_preflight.resolve_or_report` and seven section loaders instead, so the envelope was never surface. The review planted `run_envelope: RunEnvelope | None = None` plus `run_envelope = run_envelope or RunEnvelope.load(root_dir)` in `ComponentPipeline.__init__`, which is the fallback `pipeline.py` forbids in its own words, and the whole suite stayed green at 6141 passed with mypy strict clean.

`configwalk.surface` now iterates to a FIXED POINT. Round 0 is the direct readers; each further round adds every class with an annotated field typed by the surface, and every method of it, because obtaining a holder is obtaining the config it holds. Measured on `kstrl/`: **2 rounds, 22 classes to 25** (`RunEnvelope`, `EnvelopeResolution`, `FeatureParams`), **122 sites to 123**. The new site is `RunEnvelope.resolve` in `_run_factory_locked`, so the `factory.py` pin goes 7 to 8.

The closure runs over TYPES rather than over calls, and that is measured rather than argued: "any scope that references the surface" runs 5 rounds, 22 classes to 33 and 8 free readers to 92 on this tree, and by round 5 it enrols `ComponentPipeline` itself, which makes the assertion this guard exists to state unstatable. Both numbers are in the module docstring. A class-level annotation counts and a PARAMETER does not, because `ComponentPipeline.__init__` takes the resolved envelope and counting parameters would enrol the subject.

`test_the_pipeline_reads_no_config_at_all` gets a positive control in the same test: it walks an in-memory copy of `kstrl/pipeline.py` with the fallback planted and asserts the walk reports a site, before asserting the real file reports none. `pipeline.py` is the module whose count is zero, so it could not be its own control the way `factory.py` and `serve.py` are each other's. Both plant anchors are asserted to occur exactly once, so a rename that makes the plant apply to nothing fails instead of planting nothing quietly.

Red first, at `6012553` with the control added and the fixed point absent: `assert [] != []` at `tests/test_config_guard.py:316`. The review's exact plant is mutation R3M1 below and it is now **caught**.

### Should-fix 1: the `policy_config` field comment still described the write-back blocker 2 deleted

It read as an instruction to restore the cross-run leak. `kstrl/factory.py:341-349` now says INPUT ONLY, read once by `RunEnvelope.resolve` rather than by `.load`, never written back, what the write used to do wrong (a caller reusing one `FactoryConfig` had run 1's clamped envelope sitting in the field, so run 2 resolved nothing and recorded run 1's hash), and where the run's resolved policy actually lives.

### Should-fix 2: the base parse-count pair `(15, 10)` was not reproducible and it shipped inside a test's failure message

Re-measured with the shipped counters and fixture in detached worktrees at the branch base `414d662` and at `origin/main` `037f0e1`, both `[autonomy]` values, `__pycache__` purged:

| revision | `[autonomy] enabled` | pair | run |
|---|---|---|---|
| `414d662` | false | **(10, 10)** | exit 0, 1 run dir |
| `414d662` | true | **(10, 10)** | exit 0, 1 run dir |
| `037f0e1` | false | **(10, 10)** | exit 0, 1 run dir |
| `037f0e1` | true | **(10, 10)** | exit 0, 1 run dir |

`(15, 10)` is nowhere. The failure message of `test_a_run_resolves_each_section_once` now carries the measured pairs and the correction: the CALLS rise, 10 to 13, because seven sections are resolved at run start whatever the component count; only the PARSES fall, 10 to 7, because those seven share one document. `(14, 9)` is the intermediate state after the second `[autonomy]` read went.

### Should-fix 3: a ladder-OFF run read `.kstrl/autonomy.json` and migrated the control directory, which it never did before

Round 2 made `AutonomyState.load` unconditional so the ladder could read the object off the envelope. That removed a duplicate read on a ladder-ON run and added a FIRST read to every ladder-OFF run, which is the default. Measured on one empty-manifest run: `origin/main` 0 `AutonomyState.load` and 0 `ensure_control_state`; round-2 head 1 and 1, plus a `RuntimeWarning` about a ladder the run does not use when `.kstrl/autonomy.json` is corrupt, plus a control-directory migration a project that never opted in did not run before.

`RunEnvelope.autonomy_state` is `AutonomyState | None` and is loaded only when `[autonomy] enabled`. `_resolve_ladder` already returned early on that branch, so it never sees the `None`; the two conditions are one condition and the code says so rather than asserting. Red first with a call counter: ladder off 1, expected 0.

`test_the_stored_level_is_read_only_when_the_ladder_is_on` keeps its name, which is true again, and now asserts `autonomy_state is None` beside the level, so the name is what is checked.

### Should-fix 4: the refusal moves a launch from charged to charged-$0, and the record did not say so

`kstrl/factory.py:3344-3352` now states it: a refusal inside the `ks factory --spec` architect window lands on `RunSpend.unmetered_phases`' blocker-halt path, so `serve` charges the launch $0 and labels the day's total a floor with `architect` unmetered. That is the same treatment a spec blocker already gets. Measured: a launch whose architect spent $4.20 is charged $4.20 on `origin/main`, where a malformed `[policy]` happened to crash BELOW `record_architect_usage`, and $0.00 here. Charging it exactly needs a sink that exists before the run directory does; that is a measured handoff below, not a new issue.

### Nits

- **N4** the inbox guard's disclosure named the wrong file for `EXPECTED_ENVELOPE_SECTIONS`; it lives in `test_config_guard.py`.
- **N5** the two undisclosed blind shapes get a strict xfail each: a primitive named as a string through `getattr`, and a receiver built at run time through `importlib.import_module`. The disclosed-limit paragraph now also states the DEPTH, one level.
- **N6** splitting `load_or_report` narrowed the walk as a side effect; the one site that left (`ConfigProblemBanner.load`) is named where the free-reader set is pinned.
- **N7** a loader that returned `None` without raising gave exit code 2 with nothing printed. `_resolved` writes a line for that case at the one place a `None` section is produced, and a test drives it. Red first: `assert []` on an empty narration.
- **N8** `_section_for` raises `LookupError`, which is outside `SURFACE_REJECTIONS`, so a section in the envelope but not in `config_sections()` would reproduce blocker 1. The envelope test resolves all seven through `_section_for`.
- **N9** three of the four pipeline alias names were created by this branch's own round 1. Counted at `414d662`: `sandbox_config` 3 readers, the other three 0 each. The comment says that instead of citing readers that predate the change.
- **N10** the surface was walked nine times per run of the guard file. A module-scoped fixture walks it once: **9.82 s to 2.75 s**, median of three, bytecode purged before each.
- **N11** `_LadderOutcome` was `frozen=True` over two `list` fields, so `hash()` raised and `.clamps.append()` succeeded. Both fields are `tuple[str, ...]`.
- **N12 is answered rather than fixed.** Free readers are matched by unqualified name, so a module defining its own `run` collides, and the review verified it by planting one. That is the FLAGGING direction, which CLAUDE.md permits: the cost is one line somebody reads on an unrelated rename, and the plant is kept as mutation R3M9 so the false positive stays visible rather than becoming folklore. Qualifying `Surface.free` by module is a change to shared walking machinery, and it is listed under Handoffs.

### Two guards that passed the mutation they exist to catch

Running the mutation set to the end found two, both in tests this branch added. They are the repo's dominant defect class and both are now closed and re-planted.

- **R3M8 was still green.** Teaching layer 1's per-scope matcher the attribute form closes one of the two disclosed limits, and the strict xfail behind the disclosure kept xfailing: its probe returned `Surface.raw`, which comes from `_raw_primitive_calls`, a SECOND bare-name matcher the plant did not touch. The disclosure would have gone stale with nothing red. `_everything_layer_one_saw` sums raw, attributed and both enrolment sets, so either matcher moving flips the row, and `configwalk`'s docstring names the two matchers. Re-planted: **caught**.
- **R3M6 was still green.** `TestTheLadderOutcomeIsReallyFrozen` built its own `_LadderOutcome` from tuple literals, and an annotation is not enforced at run time, so reverting both annotations AND the `tuple(...)` at the construction site left it green. It now drives `_resolve_ladder` with a stored L4 over `max_level = 1` (the clamp) and `[policy] deps_allow_new` granted then withheld below L3 (the override), and asserts both fields are non-empty before asserting their type. Re-planted: **caught**.

## Behaviour changes

- **A mid-run edit to `kstrl.toml` takes effect at the next run, not this one.** There is deliberately no mid-run change detector, no warning and no mtime check: after the fix there is no per-component read left to detect a change against, and the rule that removes a mechanism beats the rule that adds a check.
- **A section the run cannot resolve refuses the run with exit code 2, naming the section and the key.** New in round 2, and the only path to it is an edit made while the architect was running, because the entry preflight already checked every section. It was a traceback in round 1 and, for `[policy]` and `[autonomy]`, a traceback on `origin/main` too.
- **Phase 1 and the set-point gate use the CLAMPED autonomy level.** Measured: a run clamped to L1 had Phase 1 judging at the stored L4. No verdict changes at either level today, because `adequacy.layer0_blocks` and `review.setpoint_blocks` both test only `>= 1`.
- **The crash path on a malformed mid-run edit is deleted rather than handled.** No handler was added, because there is no read left to raise from.
- **A `FactoryConfig` reused across two runs resolves `kstrl.toml` again.** Round 1 would have had run 2 inherit run 1's clamped envelope. Latent in production, where every caller assembles a fresh config, which is why it has a test rather than an argument.
- **Config PARSES per run fall; config CALLS rise.** A whole `run_factory` over an empty manifest costs 13 `load_toml_document` calls and 7 `tomllib.loads` parses against a re-measured (10, 10) at the branch base `414d662` and at `origin/main` `037f0e1`. The calls rise because seven sections are resolved at run start whatever the component count; the parses fall because those seven share one document. Round 2 of this body said the base was (15, 10), which is not reproducible at either revision. The count that matters is that it is CONSTANT in the component count, where the base was 5 calls and 5 parses per component.
- **A run with `[autonomy] enabled = false` reads no `.kstrl/autonomy.json` and runs no control-directory migration.** That is the default path, and round 2 had briefly made both happen on it.

## Tested

Every number below was measured on the tree with `origin/main` `037f0e1` (#360) merged in, which is the current tip.

### Census, re-derived by RUNNING the shipped walk

Reproduce with the walk that ships in this PR:

```
uv run python -c "
from pathlib import Path
from tests.helpers.astwalk import package_sources
from tests.helpers.configwalk import read_sites, surface
found = surface(package_sources())
print(found.rounds, len(found.classes), len(found.free), found.raw, found.attributed)
print(sum(len(read_sites(f, found)) for f in package_sources()))"
```

| | round 1 (`23b5f40`) | round 2 | round 3 |
|---|---|---|---|
| total call sites | 98 | 122 | **123** |
| `kstrl/pipeline.py` | 4, all in `__init__` | 0 | **0** |
| `kstrl/factory.py` | 8 | 7 | **8** |
| `kstrl/runenvelope.py` | 3 | 8 | **8** |
| `kstrl/serve.py` | 10 | 10 | **10** |
| `kstrl/config_preflight.py` | 3 | 26 | **26** |
| `kstrl/cli.py` | 41 | 41 | **41** |
| layer 1 | 22 classes, 8 free, 54 raw = 54 attributed, 0 rounds | identical | **2 rounds, 25 classes, 8 free, 54 raw = 54 attributed** |

Round 2's growth was the walk changing, not the code: counting REFERENCES rather than calls makes `config_sections()`'s 22 registry entries sites, which is the direction a FLAGGING guard may be wrong in and is what makes the `partial` and bound-name shapes visible. Round 3's one new site is `RunEnvelope.resolve` in `_run_factory_locked`, which the fixed point can now see, so the `factory.py` pin goes 7 to 8.

### Parses per run

Two numbers, because they are two claims: `load_toml_document` CALLS is how many times something asked for the document, and `tomllib.loads` is how many times the bytes were lexed.

| components | base `414d662` | round 1 | round 2 and 3 |
|---|---|---|---|
| 1 | 5 calls, 5 parses | 7, 2 | **7, 1** |
| 8 | 33, 33 | 7, 2 | **7, 1** |

Constant in the component count, pinned by `test_one_component_and_eight_cost_the_same`.

A whole `run_factory` over an EMPTY manifest is a different number and round 2 stated its base wrongly. Re-measured in detached worktrees with the shipped counters and fixture: `414d662` and `037f0e1` are both **(10, 10)** under either `[autonomy]` value, with the run completing and one run directory written. This PR is **(13, 7)**, pinned by `test_a_run_resolves_each_section_once`. So the CALLS rise, 10 to 13, because seven sections are resolved at run start whatever the component count, and only the PARSES fall, 10 to 7, because those seven share one document. The remaining 6 calls and 6 parses are `LinearConfig`, `NotifyConfig`, `KnowledgeConfig`, `TimeoutConfig`, `BreakerConfig` and `EvolutionConfig`, each outside the envelope and each once per run.

### Red first

| test | revision it is red at | how it fails there |
|---|---|---|
| `test_it_refuses_with_exit_2_and_names_the_key[[inbox]]` | round 1 `23b5f40` | `ValueError: invalid literal for int() with base 10: 'many'` out of `run_factory` |
| same, `[policy]` | round 1 and `origin/main` | `ValueError ... 'two'` out of `run_factory` |
| `test_it_refuses_before_the_pipeline_is_built` | round 1 | same traceback |
| `test_both_components_are_verified_and_the_run_returns` | `origin/main` | `ValueError ... 'two'` out of `run_factory`, 0 of 2 components verified |
| `test_the_pipeline_reads_no_config_at_all` (the new control) | round 2 `6012553` | `assert [] != []`: the walk reports nothing on a copy of `pipeline.py` with the forbidden fallback planted |
| `test_a_ladder_off_run_does_not_read_the_state_at_all` | round 2 `6012553` | `assert 1 == 0`, one `AutonomyState.load` on a ladder-OFF run |
| `test_a_loader_returning_none_without_raising_still_prints_a_reason` | round 2 `6012553` | `assert []`, exit code 2 with an empty narration |

### Mutations

**26 distinct plants in round 3**, one at a time, `__pycache__` purged and `PYTHONDONTWRITEBYTECODE=1` before each, bounded at 420 s with `start_new_session=True` and `os.killpg` on expiry, anchor occurrence count asserted before the edit so a plant that does not apply is COULD-NOT-PLANT and never a pass. Every plant was restored and `git status --porcelain` checked empty after the run.

**25 caught, 1 still green by construction, 0 hung, 0 could-not-plant.**

The eighteen the round-2 review left, replanted on this tree: R2M1 to R2M9, RM3, RM14, G1 to G5, R3M1 and R3M9. All eighteen caught. **R2M17 is the one that was still green at round 2**, the `RunEnvelope.load` fallback in `ComponentPipeline.__init__`; it is R3M1 here and it is now caught by the pipeline guard. R3M9 is the review's own false positive (a local `def run` in `factory.py` reading no config) and it is still caught, which is nit 12 above.

Eight are round 3's own:

| # | plant | outcome |
|---|---|---|
| R3M1 | `run_envelope or RunEnvelope.load(root_dir)` in `ComponentPipeline.__init__` | **caught** (was still green at round 2) |
| R3M2 | the surface fixed point stops at round 0 | caught (4 failed) |
| R3M3 | round 0 AND the pipeline test's control faked, that one test run alone | **still green, by construction** |
| R3M3b | the same pair, run against the whole guard file | caught (3 failed) |
| R3M4 | the ladder state is loaded unconditionally again | caught (6 failed) |
| R3M5 | a loader returning `None` reports nothing | caught |
| R3M6 | `_LadderOutcome` carries lists again | still green, then **caught** after the test drives `_resolve_ladder` |
| R3M7 | `[divergence]` leaves `config_sections()` while staying in the envelope | caught |
| R3M8 | layer 1 learns the attribute form, closing a disclosed limit | still green, then **caught** after the probe sums every count |

R3M3 is the only green one and it is the demonstration rather than a gap: with the closure blinded AND the control's own line replaced by a literal, the single test it targets passes, which is what makes the control the load-bearing line in that test. Run against the whole guard file the same pair is caught by three tests, because the round count and the class census pin the closure independently of the control. Both are recorded rather than one of them.

### Static guards, executed rather than read

The guard set is DERIVED at each revision by `grep -rl -E 'ast\.parse|astwalk' tests/*.py`, not read off a list, and each side is RUN. The pair around the `037f0e1` merge is not available because the merge is already committed; the pair that is available is `origin/main` `037f0e1` against this head, both executed today.

| | `origin/main` `037f0e1` | this head |
|---|---|---|
| guard files | 37 | **38** |
| passed | 1204 | **1221** |
| skipped | 9 | **9** |
| xfailed | 41 | **44** |

Per-file, from `--junitxml`, the delta reconciles to zero outside two files:

- `tests/test_config_guard.py`: absent, to 19 passed + 3 xfailed. The file is this branch's; the three xfails are the disclosed blind shapes.
- `tests/test_inbox_write_guards.py`: 29 passed to 27. The two `EXPECTED_CONFIG_LOADS` rows that left when `load_or_report` was split, replaced by the disclosure nit 4 corrected.
- Every other guard file: delta zero on all four counts. No guard went red and none lost a site.

### Gates

| gate | round 2 | round 3 |
|---|---|---|
| `pytest tests/ -q -p no:cacheprovider` | 6141 passed, 37 skipped, 43 xfailed, exit 0 | **6182 passed, 37 skipped, 45 xfailed, 413.0 s, exit 0** |
| `mypy kstrl/ --strict` | clean, 133 files | **clean, 134 files** |
| `ruff check kstrl/ tests/` | clean | **clean** |
| `gen_docs.py --check` | current | **current** |

`scripts/precommit/`, run against `origin/main` `037f0e1` with the lane's fourteen Python files staged in a detached worktree, so the comparison is against main rather than against the branch's own HEAD:

- file-length ratchet: exit 0. Five already-oversize files grew and are reported, not failed: `kstrl/factory.py` 4502 to 4632, `kstrl/pipeline.py` 4397 to 4425, `tests/test_autonomy_ladder.py` 1025 to 1045, `tests/test_pipeline.py` 2770 to 2772, `tests/test_progress_scope.py` 1675 to 1677. Nothing crossed the limit and nothing was added over it. This ratchet is why the tests are in three files: run with the lane staged it FAILED at `tests/test_config_read_once.py` 812 lines, which a file the lane adds is failed for rather than grandfathered.
- cyclomatic ratchet: exit 0. No NEW and no REGRESSED function, one IMPROVED (`config_preflight.load_or_report` 2 to 1) and one risen under the limit (`test_deps_allow_new_withheld_below_l3` 1 to 2).
- `check_no_mocks.py` on the four production files: clean.
- em dashes: 0 in the branch diff, 0 in the commit messages, 0 in this body. Emoji in the diff: 0.

The suite total is reconciled by collection rather than by argument. `pytest tests/ --collect-only -q` is **6217** at `origin/main` `037f0e1` and **6264** here, and the +47 is four files: `test_config_guard.py` +22 (19 passed and the 3 disclosed xfails), `test_run_envelope.py` +16, `test_config_read_once.py` +11, `test_inbox_write_guards.py` -2 for the two `EXPECTED_CONFIG_LOADS` rows that left the call walk. 6182 + 37 + 45 = 6264, so nothing collected went unaccounted for.

`kstrl/config.py` is untouched: `git diff origin/main...HEAD -- kstrl/config.py` is empty, and it stays at 799 lines.

H2/H3: no `*_PROMPT` constant is touched. `git diff origin/main...HEAD -- kstrl/ | grep -c _PROMPT` is 0.

## Assumed, and stated as such

- That `_phase_verify`'s `ValueError` aborted the whole RUN was READ at round 1 and is now DRIVEN: `test_both_components_are_verified_and_the_run_returns` fails at `origin/main` with the exception escaping `run_factory` and neither component verified. What remains read rather than driven is the claim about `414d662` specifically; the mechanism was re-measured at `037f0e1`, where the same code stands.
- That the seven envelope sections are safe to resolve ahead of the run rests on `config_preflight.config_sections()` covering all of them, which is checked in the source. Round 2 makes the consequence of being wrong a refusal with a message rather than a traceback, which is why that check is no longer the only thing standing between an operator and a stack trace.
- The 17.2 ms `AutonomyState.load` figure is the round-1 reviewer's measurement, not re-taken here. What is re-measured is the COUNT, which is what the fix changes.
- The base-revision suite count at `414d662` (5884 passed) is the implementer's, taken in a pristine detached worktree.

## Design decisions

1. **The refusal goes above the run directory's sinks, not below them.** Below them it would be an early exit inside #257's protected window, so a run would exist and report $0. Above them there is no run yet.
2. **Should-fix 6 is answered by removing the second read, not by adding a `state=` parameter.** The reviewer proposed the parameter; the envelope carrying the state is the same fix with one mechanism fewer, which is the tie-break CLAUDE.md states.
3. **The guard counts references, not calls.** It over-matches by 22 in `config_sections()`. A flagging guard may over-match; a walk that enumerates call shapes goes blind on the next shape somebody writes, which is what round 1 measured five times.
4. **`ComponentPipeline.run_envelope` stays required, with no default.** A default is a second source that can silently disagree, which is the defect.
5. **The tests are in three files.** Forced by the file-length ratchet, split at the seam they already had: static walk, phase, run.
6. **The surface closure runs over TYPES, not over calls.** Over calls it does not converge on this tree (5 rounds, 22 classes to 33, 8 free readers to 92) and by round 5 it enrols `ComponentPipeline` itself, which makes the assertion the guard exists to state unstatable. Both numbers are in the module docstring so the next author does not have to re-derive them to reject the idea.
7. **Nit 12 is answered rather than fixed.** Matching a free reader by unqualified name over-matches, and over-matching is the direction CLAUDE.md permits a FLAGGING guard to be wrong in. Qualifying `Surface.free` by module is a change to shared machinery and is a handoff, not a nit fix folded into this PR.

## Handoffs

- **`serve.resolve_merge_gate` (`kstrl/serve.py:1434`, `:1439`)** loads `AutonomyConfig` and `PolicyConfig` to decide a per-item merge gate milliseconds before the `ks factory` child re-resolves both in its own process. Left open: it is a per-item admission decision rather than a run envelope, the daemon enforces nothing, and closing it means passing resolved config across a process boundary, which is #193's composition root. The round-1 review adds a refinement worth carrying: a channel DOES already exist, a single `--pause-before-pr-merge` flag, which `run_factory` then overrules with the ladder's bundle, so the open window is "the channel carries one boolean the child overrules" rather than "no channel exists".
- **The 48 cli-entry sites** (`cli.py` 41, `launch.py` 7) are what "one parse of `kstrl.toml` per process" actually costs and what this lane cannot meet. Acceptance 1 of #192 should be restated on #193 with that number attached.
- **Layer 1 still matches a primitive by bare `Name` only.** Disclosed in `configwalk`'s docstring and carried by a strict xfail, so teaching it `astwalk.bindings` XPASSes and forces the disclosure to be edited in the same diff. Latent today: 54 bare-name primitive calls in `kstrl/`, 0 in the attribute form.
- **Charging a refused launch its exact architect spend needs a sink that exists before the run directory does.** Measured: a launch whose architect spent $4.20 is charged $4.20 on `origin/main`, where a malformed `[policy]` happened to crash BELOW `record_architect_usage`, and $0.00 here, because the refusal lands on `RunSpend.unmetered_phases`' blocker-halt path. The day's total is a floor with `architect` unmetered, which is the same treatment a spec blocker already gets, so the accounting is consistent rather than silent. Making it exact means a spend sink that outlives the decision to refuse, which is the composition root again (#193). Carried here with the numbers attached rather than filed as a sibling issue.
- **`Surface.free` holds unqualified names**, so a module defining and calling its own `run` is counted as a config read. Verified by the review with a plant, kept as mutation R3M9. The cost is one flagged line on an unrelated rename against an equality pin. Qualifying the set by module removes it and touches the walking machinery the other guards share.
- **`astwalk`'s `scopes()` and `own_nodes()` are not memoized** while `all_nodes` is. Measured at round 1 with a memo injected as a pytest plugin: `configwalk.surface()` warm 198.6 ms to 45.3 ms, the eight guard files sharing astwalk 14.52 s to 7.07 s, results byte-identical. Not done here: it is shared machinery sixteen guards depend on.

Closes #192


kstrl/pipeline.py
```diff
@@ -451,16 +454,34 @@ def __init__(
...
+            # one pair of roles that now runs shell commands inside the
+            # tree under review. read_only is a permission-layer posture
+            # on the claude adapters; the operator's OS-level
+            # enforcement is a separate payload and both are wanted.
+            self.sandbox_config = SandboxConfig.load(root_dir)
+            # R7.2: fixtures resolve from toml/env when the caller did
+            # not inject one; enabled=false (the default) makes
+            # run_mechanical_verification skip the check entirely.
+            self.fixtures_config = factory_config.fixtures_config or FixturesConfig.load(root_dir)
+            self.inbox_config = InboxConfig.load(root_dir)
```
0xfauzi: A malformed `[inbox]` section now aborts the whole run, and the two builders still document the opposite contract.

`InboxConfig.load` casts per key (`int(section["open_item_cap"])`, `float(section["snooze_hours"])`). Until this PR that cast only ever ran inside `_inbox_resolve` / `_inbox_add`, both of which catch `(OSError, TypeError, ValueError, ...)` and warn, because "bookkeeping must not be able to fail a run". Here it runs at pipeline construction with nothing in front of it and nothing above it: `run_factory` has no handler, so it leaves the process as a raw traceback.

Measured both ways, `__pycache__` purged, `PYTHONDONTWRITEBYTECODE=1`, with `kstrl.toml` containing `[inbox]\nopen_item_cap = "many"` and an empty manifest:

- at `543e1aa` (origin/main): `run_factory` returns, `exit_code 0`.
- on this branch: `ESCAPED: ValueError invalid literal for int() with base 10: 'many'`.

The entry preflight does cover `[inbox]`, but it is not the whole window: on `ks factory --spec` the architect runs for 119 to 210s between the preflight and this line, which is the exact window `config_preflight`'s own docstring measured for #272. Three comments in this file now assert something that is no longer true and should move with the code: line 501 ("lazily built so a disabled inbox costs nothing and a broken one can never fail a run"), `_inbox_add`'s "Never fatal" docstring, and "a broken inbox degrades to a warning". `[divergence]` and `[fixtures]` join `[inbox]` in the same block, widening the fatal set from 2 sections to 6.

If this is the intended trade (the disposition row in `tests/test_inbox_write_guards.py` says it is), the refusal should still route through the surface the other pre-spend refusals use rather than out of the top of the run as a traceback.

kstrl/factory.py
```diff
@@ -3323,11 +3327,20 @@ def _run_factory_locked(
...
-    autonomy_config = AutonomyConfig.load(root_dir)
+    # #192: the run's config envelope, resolved ONCE here and injected
+    # into the pipeline, which never resolves one of its own.
+    # kstrl/runenvelope.py records the divergence that costs. Resolved
+    # here rather than at the ladder resolution below because the #262
+    # probe gate needs [autonomy] first; the ladder still resolves in
+    # its original place for the reason its own comment gives (the
+    # policy hash must record the CLAMPED envelope), and it clamps this
+    # object before the hash is taken.
+    run_envelope = RunEnvelope.load(root_dir, policy_override=factory_config.policy_config)
```
0xfauzi: This moves two raise-capable config loads above `pipeline.record_architect_usage`, which is the #257 invariant stated 160 lines below.

`factory.py:3482-3497` says it explicitly: "The money is already gone by this point, so every early exit from here has to leave it recorded rather than report a run that cost nothing... The invariant this rests on, nothing between the run directory's sinks and this line returns early, holds today and is checked by hand, not by a mechanism. An early exit inserted above would break it silently."

Before this PR the only load in that window was `AutonomyConfig.load`; `PolicyConfig.load` sat at old-line 3533, below `record_architect_usage`. This PR puts `PolicyConfig.load` and `AdequacyConfig.load` here at 3338, and `FixturesConfig.load` / `InboxConfig.load` / `DivergenceConfig.load` inside `ComponentPipeline.__init__` at 3448, all of them above `pipeline.record_architect_usage(architect_usage)` at 3504. Any of the five raises on an operator edit made during the architect's run, and the exception is exactly the silent early exit the comment warns about: no architect row, so `serve.read_run_spend` charges $0 for a launch that spent real money.

Verified reachable: `[inbox] open_item_cap = "many"` raises `ValueError` out of `run_factory` on this branch and returns `exit_code 0` at `543e1aa`. The same class reaches this line through `[policy]` and `[adequacy]` too.

Two ways out, both cheap: resolve the envelope after `record_architect_usage`, or record the architect's spend before the envelope resolves. The second is the one that also closes the pre-existing `SandboxConfig` case.

tests/helpers/configwalk.py
```diff
@@ -0,0 +1,181 @@
...
+class ReadSite:
+    """Layer 2: one call of the config surface, and the scope that owns it."""
+
+    module: str
+    line: int
+    scope: str
+    target: str
+
+
+def read_sites(source_file: Path, found: Surface) -> list[ReadSite]:
```
0xfauzi, line 314: Layer B goes blind on a config read written inside a lambda, and that shape is not the disclosed one.

`read_sites` iterates `scopes(tree)` and `own_nodes(node)`. `astwalk.scope._walk_scopes` yields only `FunctionDef` / `AsyncFunctionDef` / `ClassDef`, never `ast.Lambda`, and `own_nodes` explicitly stops at `ast.Lambda`. So a call inside a lambda belongs to no scope: it is not an offender, and it does not appear in `read_sites` at all, so `EXPECTED_PIPELINE_SITES == 4` still holds. Both halves of the layer go quiet at once, which is the shape the class docstring says it exists to prevent.

Measured, one plant at a time, `__pycache__` purged and `PYTHONDONTWRITEBYTECODE=1`:

- plant `_sec = (lambda: SecurityConfig.load(self.root_dir))()` at the top of `ComponentPipeline._phase_review`: `tests/test_config_read_once.py` gives 23 passed, 1 xfailed, byte-identical to clean.
- the same call written plainly, `_sec = SecurityConfig.load(self.root_dir)`: 2 failed (`test_every_pipeline_config_read_is_in_init`, `test_the_pipeline_site_count_is_pinned`), 21 passed.

Layer A does not cover it either. `_raw_primitive_calls` counts only bare-`Name` calls to `PARSE_PRIMITIVES`, so the `attributed == raw` census catches `lambda: load_toml_section(...)` (confirmed: attributed 0, raw 2) but sees nothing at all for a surface-class `Cls.load(...)` in a lambda. The module docstring discloses only the `mod.Cls.load` shape, and there is no `blind_spot` row for this one.

Cheapest fix that keeps the shape: have `read_sites` walk `ast.Lambda` nodes as their own scopes (qualified as `<enclosing>.<lambda>`), which makes any lambda read an offender rather than an absence. If it stays open, it needs a strict-xfail `blind_spot` row and a line in the disclosure, the way the `mod.Cls.load` limit got one.

tests/test_config_read_once.py
```diff
@@ -0,0 +1,691 @@
...
+
+
+class TestPipelineReadsConfigOnlyAtConstruction:
+    """Layer B. Scoped to ``kstrl/pipeline.py``, which is why a per-poll
+    read in ``kstrl/serve.py`` is outside it by construction rather than
+    by an exemption somebody has to remember."""
+
+    def test_every_pipeline_config_read_is_in_init(self) -> None:
+        found = configwalk.surface(package_sources())
+        sites = configwalk.read_sites(KSTRL_PACKAGE / "pipeline.py", found)
```
0xfauzi: Layer B is scoped to `pipeline.py`, but the per-component loop it is guarding also lives in `factory.py`, and nothing in this file covers that half.

`read_sites` is called on `pipeline.py` and `serve.py` only. The behavioural half does not fill the gap either: `TestTheParseCountDoesNotGrowWithComponents` drives `phase_verify_envelopes`, which counts nothing `_run_factory_locked` does, and `test_a_run_resolves_each_section_once` pins `(14, 9)` over an EMPTY manifest, so nothing on the per-component path of `factory.py` is counted by any assertion here.

Measured, `__pycache__` purged and `PYTHONDONTWRITEBYTECODE=1`: planting `_p = PolicyConfig.load(root_dir)` as the first statement of `factory.py::_submit_args`, which runs once per component, gives 23 passed and 1 xfailed. Identical to clean. That is the same defect this PR fixes, in the file that owns the scheduler, passing every layer.

Layer A does not see it: adding a `Cls.load` call does not make `factory.py` part of the surface, since `surface()` only enrols scopes that call a PRIMITIVE.

Two options. Extend layer B to `factory.py` with its own allowed-scope set (the run-start scopes), which keeps the closed-by-construction shape; or give `_empty_run` one component so the `(14, 9)` census becomes sensitive to the per-component path. The second is smaller and also gives the malformed-edit test below a real run to assert about.

kstrl/runenvelope.py
```diff
@@ -0,0 +1,107 @@
...
+        reads JSON, which ``toml_parse_scope`` does not cache, and it
+        costs 18.3ms (14.6 of them ``ensure_control_state``) against the
+        sub-millisecond window that scope's docstring says makes a
+        stale document safe.
+        """
+        with toml_parse_scope():
+            policy = policy_override or PolicyConfig.load(root_dir)
+            adequacy = AdequacyConfig.load(root_dir)
+            autonomy = AutonomyConfig.load(root_dir)
+        level = AutonomyState.load(root_dir).level if autonomy.enabled else 0
```
0xfauzi: On the ladder-on path this 18.3ms read is always thrown away, and `AutonomyState` is then loaded a second time nine lines later in the factory.

The docstring four lines up prices this call itself: 18.3ms, 14.6 of them `ensure_control_state`. But `_run_factory_locked` does `autonomy_state = AutonomyState.load(root_dir)` for `resolve_runtime_level` and then immediately `run_envelope = replace(run_envelope, autonomy_level=int(autonomy_level))`, so whenever `autonomy.enabled` is true, which is the only case in which this line reads the file at all, the value it computed is dead on arrival and `ensure_control_state` has run twice.

This is the same shape the PR just deleted one field over: the `[autonomy]` section was read bare at run start and again in `RunEnvelope.load`, and the fix was to carry the resolved object rather than resolve it twice. `AutonomyState` is the remaining instance of it, and it costs about 30x what the duplicate `[autonomy]` parse cost (18.3ms against the measured 0.988ms).

`RunEnvelope.load` could take `state: AutonomyState | None = None` and the factory could resolve the state once and pass it in, which also removes the window in which a concurrent `ks autonomy promote` makes the two reads disagree.

kstrl/factory.py
```diff
@@ -3567,8 +3588,19 @@ def _run_factory_locked(
...
+    # autonomy_active:` was measured STILL GREEN: the envelope reaches
+    # the pipeline through the required constructor parameter, which is
+    # the fix, and this is only defence for a future clamp added outside
+    # the ladder block. Deleting the second line is the non-equivalent
+    # mutation of the same pair, and it is caught. `policy_config` has
+    # no production reader left besides RunEnvelope.load's override seam
+    # above; it stays because the ladder's own test reads the clamped
+    # envelope off it (#192).
+    factory_config.policy_config = run_envelope.policy
+    pipeline.run_envelope = run_envelope
```
0xfauzi: The constructor parameter is required "with no default and no fallback", and then this line writes over it 155 lines later. That reopens a narrowed version of the window the PR closes.

`ComponentPipeline` is built at 3448 with the UNCLAMPED envelope and only reaches the clamped one here at 3603. Between those two lines the pipeline holds a policy the ladder has already decided to withhold: today `deps_allow_new=true` at L1, which is exactly the value `resolve_runtime_level` exists to refuse. Nothing in that window reads it (only `record_architect_usage` at 3504 and the DAG check), so it is latent, but it is latent for the same reason the original defect was: a hand-checked ordering rather than a mechanism, in the same 160-line stretch that already carries one hand-checked ordering invariant (#257 at 3495).

The comment above says the envelope "reaches the pipeline through the required constructor parameter, which is the fix, and this is only defence". That is true only on the ladder-off path. On the ladder-on path the constructor parameter carries the wrong object and this assignment is the whole mechanism, which is why M9 is caught and M5 is not.

The shape that makes the constructor parameter true in both cases is to resolve the ladder BEFORE constructing the pipeline, then delete this line. The PR declined a different hoist (above `_adversarial_phase_gates`) for a real reason: probe spend under a skip-mode run. That reason does not apply to hoisting the ladder to just below `_adversarial_phase_gates` and above `ComponentPipeline(...)`, which is a move of about 100 lines within one function and leaves the gates where they are.

If it stays, the pin is worth strengthening: nothing today asserts that the pipeline is unused between 3448 and 3603, and that is the invariant this shape rests on.

kstrl/factory.py
```diff
@@ -3323,11 +3327,20 @@ def _run_factory_locked(
...
+    # here rather than at the ladder resolution below because the #262
+    # probe gate needs [autonomy] first; the ladder still resolves in
+    # its original place for the reason its own comment gives (the
+    # policy hash must record the CLAMPED envelope), and it clamps this
+    # object before the hash is taken.
+    run_envelope = RunEnvelope.load(root_dir, policy_override=factory_config.policy_config)
+    # Off the envelope rather than a second AutonomyConfig.load: the
+    # two reads were nine lines apart and cost two parses of the same
+    # file (measured 0.988ms, and a nested toml_parse_scope does not
+    # collapse them because the inner scope shadows the outer cache).
```
0xfauzi: The nested-scope limitation is documented in three places rather than fixed in the one place it lives.

`config.toml_parse_scope` is `token = _PARSE_SCOPE.set({})`: the inner scope starts empty, which is why nesting buys nothing. That is a property of six characters, not of the design. Seeding the new dict from the enclosing one makes nesting free:

```python
token = _PARSE_SCOPE.set(dict(_PARSE_SCOPE.get() or {}))
```

A copy rather than the same dict keeps the exit semantics exactly as they are (the inner scope's reads do not leak outward when it closes), and inheriting cannot widen the staleness window beyond what the OUTER scope already promised, since the outer scope is what defines it.

With that, the two scopes this PR adds collapse into one whenever a caller wraps them, `RunEnvelope.load` and `ComponentPipeline.__init__` become composable, and the measured `(7, 2)` in `test_one_component_and_eight_cost_the_same` becomes `(7, 1)`. It also deletes the special case from all three places it is currently written out: this comment, `runenvelope.py`, and `TestTheFactorySideParseCountIsPinned`'s docstring.

Per CLAUDE.md's altitude rule, the change that removes the mechanism beats the three that document it. `kstrl/config.py` is untouched by this lane, so if that is out of scope here, the honest form is one sentence saying the shadowing is fixable and where, rather than three sites stating it as a fact about the world.

tests/test_config_read_once.py
```diff
@@ -0,0 +1,691 @@
...
+        readings = phase_verify_envelopes(
+            tmp_path,
+            comps,
+            between=_edit(tmp_path, _AFTER.format(autonomy=autonomy)),
+            run_envelope=envelope,
+        )
+
+        assert {r.adequacy.enabled for r in readings} == {True}
+
+    def test_a_malformed_mid_run_edit_does_not_abort_the_run(self, tmp_path: Path) -> None:
```
0xfauzi: This test never runs a run, so it cannot pin what its name and its docstring claim.

The docstring names the mechanism precisely: `_phase_verify`'s caller `process_result` runs outside the `try` that wraps `future.result()`, so one bad edit aborted the whole run. `phase_verify_envelopes` calls `pipeline._phase_verify(...)` directly in a `for` loop. `process_result`, the scheduler, and the `try` in question are never entered, so what is asserted is "`_phase_verify` does not raise", which is the necessary condition and not the claim. The PR body states it as "one test asserting a mid-run edit to `max_files_changed = "two"` completes the run".

`_empty_run` in this same file already drives the real `run_factory` and returns the pipeline. Giving it one component (or a `components` parameter) turns this into the assertion its name makes, and would also make the `(14, 9)` census sensitive to the per-component path in `factory.py`, which nothing currently is.

Smaller point on the fixture: `_MALFORMED` drops `[adequacy]` as well as breaking `[policy]`, so if the read did come back it could fail on either. A fixture that changes only the field under test is the one that names the cause.

kstrl/factory.py
```diff
@@ -3567,8 +3588,19 @@ def _run_factory_locked(
...
+    # Unconditional, though a mutation restricting both lines to `if
+    # autonomy_active:` was measured STILL GREEN: the envelope reaches
+    # the pipeline through the required constructor parameter, which is
+    # the fix, and this is only defence for a future clamp added outside
+    # the ladder block. Deleting the second line is the non-equivalent
+    # mutation of the same pair, and it is caught. `policy_config` has
+    # no production reader left besides RunEnvelope.load's override seam
+    # above; it stays because the ladder's own test reads the clamped
+    # envelope off it (#192).
+    factory_config.policy_config = run_envelope.policy
```
0xfauzi: This write is now unconditional, which turns `FactoryConfig.policy_config` into both the input override and the output, and makes a second `run_factory` on the same instance inherit the first run's clamp.

Line 3338 reads the field as `policy_override`; this line writes the resolved and possibly ladder-clamped policy back into it. So for any caller that reuses a `FactoryConfig` across two `run_factory` calls in one process, run 2 resolves nothing from `kstrl.toml` at all: it gets run 1's clamped envelope, `max_files_changed` and `deps_allow_new` included, and records its hash as if it had read the file. Before this PR the write sat inside `if autonomy_active:`, so on the default path (`[autonomy] enabled = false`) the field stayed `None` and run 2 re-read the file.

No caller does this today, and I checked: the three `cli.py` sites and `tui/embed.py` run once per process, and `tui/session.start_run_session` calls `_prepare_factory` per session, so the config is rebuilt. So this is latent rather than live. It is worth naming anyway, because a field that is an input on line 3338 and an output on line 3602 is the two-answers ambiguity the envelope was introduced to remove, and the comment above already records that nothing in production reads it.

Two clean forms: keep the field write-once (`if factory_config.policy_config is None`), or drop the round trip and have `_empty_run`-style callers read the clamp off `pipeline.run_envelope`, which is what the guard test already does and what `tests/test_autonomy_ladder.py:776` could be pointed at.

Issue #204, opened 2026-08-03
poll_interval_seconds measures awake time; interval mode measures wall clock and coalesces


0xfauzi:
> **Stabilisation queue: P1, rank 8 of 21** (set 2026-09-09; the full ordered list and the coordinator's rules are in #369). The runbook presents two launchd modes as equivalent; measured across a suspend they are not. Correct the doc from what was measured; leave the one unmeasured number open and say so; do not fabricate it.

Found while running the R8.6 sleep/wake verification. The two launchd packagings in `docs/continuous-intake.md` §4 are presented as two ways to get the same cadence. Measured across a real suspend, they are not.

## Measured

`ks serve --root <scratch> --max-cycles 8` with `poll_interval_seconds = 60`, started 16:40:02, exited 17:07:45. Full sleep/wake accounting from `pmset -g log`:

| Period | Duration |
|---|---|
| awake 16:40:02 -> 16:41:25 (clamshell close) | 83s |
| **suspended** 16:41:25 -> 16:57:40 | 975s |
| awake 16:57:40 -> 16:57:42 (DarkWake, SleepService) | 2s |
| **suspended** 16:57:42 -> 17:01:57 | 255s |
| awake 17:01:57 -> 17:02:02 (DarkWake, Maintenance) | 5s |
| **suspended** 17:02:02 -> 17:02:44 | 42s |
| awake 17:02:44 -> 17:07:45 (lid open) | 301s |
| **total awake / suspended** | **391s / 1272s** |

391 + 1272 = 1663s = the 27m43s wall clock exactly, so nothing is unaccounted for.

The process survived three suspends totalling 21 minutes, held `serve.lock` throughout, and exited normally.

## The claim this supports

**`poll_interval_seconds` measures awake time, not wall clock.** A laptop asleep for eight hours does not accumulate 480 missed polls in keepalive mode; it resumes mid-poll and continues. Interval mode under launchd is the opposite: it fires against the wall clock and **coalesces** missed firings - three boundaries were missed across the 975s sleep and exactly one catch-up run followed (3 `service inactive` events matching 3 `ks serve on ...` header lines in `serve.err.log`, 1:1).

Neither behaviour is wrong. But an operator choosing between the two modes is currently told they differ only in process lifetime, and after a long sleep they behave measurably differently.

## What is NOT resolved

[...]

Issue #203, opened 2026-08-03
caffeinate -i may not hold the system awake for a run started inside a dark wake


0xfauzi:
> **Stabilisation queue: P1, rank 9 of 21** (set 2026-09-09; the full ordered list and the coordinator's rules are in #369). A real bug in unattended operation, but its remaining items need a real suspend only the owner can run. Prepare the experiment script and the doc; do not block the queue on the suspend itself.

Found while running the R8.6 sleep/wake verification (the test `docs/continuous-intake.md` §7 listed as "not verified"). **Measured, on a real suspend.**

## What was observed

Interval mode (`--plist-mode interval`) schedules `ks serve --once` on a `StartCalendarInterval`. That mechanism works - see the companion result below - but one of the catch-up firings landed inside a **2-second dark wake with the lid still shut**:

```
16:41:25  Sleep     Clamshell Sleep, Using Batt ....................  975 secs
16:57:40  DarkWake  from Deep Idle [CDNP] : rtc/SleepService .......    2 secs
16:57:40.783        launchd: service inactive: com.kstrl.serve.<hash>   <- our run completed
16:57:42  Sleep     Sleep Service Back to Sleep ....................  255 secs
...
17:02:44  Wake      lid SMC.OutboxNotEmpty / HID Activity              <- lid actually opened
```

The run fit because a cycle over an **empty** queue takes 0.08-0.12s (measured, 3 runs of `/usr/bin/time -p ks serve --once`). A cycle that actually claims an item does not: the live factory runs measured earlier in this work took 10-20 minutes.

## Why this may be a problem

`caffeinate -i` asserts `PreventUserIdleSystemSleep` (measured previously with `pmset -g assertions`). A dark wake ends on a **SleepService timer**, not an idle timer. Whether that assertion prevents the `Sleep Service Back to Sleep` transition is **not tested and I am not going to guess.**

If it does not, unattended interval mode can start a 15-minute factory run inside a 2-second window and be suspended a second or two later.

## What is NOT at risk (checked, so nobody re-derives it)

`serve_lock` wraps the entire run including the `once` branch (`kstrl/serve.py:2339`) and uses `LOCK_EX | LOCK_NB`, raising `ServeLockedError` on contention. So while a suspended run holds the lock:

- the next scheduled firing exits immediately instead of double-running the item;
- it never reaches `reap_leases`, so it cannot reap the live-but-suspended run.

[...]

PR #360, opened 2026-09-07
fix(serve): caffeinate forks, so the rationale changes and the lock order is pinned (#209, #203)

### LINKED ISSUE #209: caffeinate -i execs in place: the factory is the daemon's direct child, not a grandchild as the process-group rationale states
Found while writing the seam tests in #208. **Measured, not inferred.**

## The claim in the code

`subprocess_factory_runner`'s docstring (`kstrl/serve.py:1193-1202`) explains why the timeout signals a process group rather than the direct child:

> **Why a process group and not `subprocess.run(timeout=...)`.** That helper signals only its DIRECT child. On macOS the direct child is the `caffeinate` wrapper, so the factory itself is a grandchild and outlived the timeout while this code recorded an infrastructure failure and requeued the item - two factories on one repo, which is exactly what `factory.lock` exists to prevent (#186 F1, reproduced: descendants were still running after `TimeoutExpired`).

The stated causal chain is: caffeinate is a wrapper process, therefore the factory is a grandchild, therefore signalling the direct child is insufficient.

## What was measured

macOS 25.5 (Darwin), `/usr/bin/caffeinate`:

```python
p = subprocess.Popen([caff, "-i", sys.executable, "/tmp/whoami_probe.py"],
                     stdout=subprocess.PIPE, text=True, start_new_session=True)
```

```
caffeinate at: /usr/bin/caffeinate
Popen pid (what on_spawn sees): 17364
child reported: child pid=17364 ppid=17363
```

The target process reports **the same pid Popen returned**. `caffeinate -i <utility>` execs the utility in place; it is not a surviving wrapper process. Reproduced a second time through the real `subprocess_factory_runner` in `tests/test_serve_seam.py`, where `on_spawn`'s pid equals the pid the spawned interpreter reports for itself.

So on this machine the factory is the daemon's **direct child**, not a grandchild, and the first link in the docstring's chain does not hold.

## Why this is worth a look rather than a doc fix

The conclusion may well still be right for a different reason. The #186 F1 repro is real and is still covered by a passing test (`test_the_supervised_timeout_reaps_descendants` spawns `sh -c "sleep 30 & ..."` and asserts descendant reaping) - but in that test the descendants come from the **shell**, not from caffeinate. A real `ks factory` also spawns descendants of its own (agent subprocesses, git, the verify commands), and those are genuine grandchildren regardless of how caffeinate behaves.

So the likely position is: **the process-group design is correct, and the reason given for it is wrong.** That is worth confirming rather than assuming, because:

- If the grandchild-via-caffeinate premise is what justified `start_new_session=True`, someone reading this later could conclude the group logic is unnecessary when `caffeinate = false` and simplify it away. The factory's own descendants would still need it.
- `_safe_pgid`'s guards and `terminate_process_group`'s "confirm the group is gone" logic are sized for the wrapper case. Whether they behave identically when the direct child IS the factory has not been checked.
- It is unknown whether this exec-in-place behaviour is stable across macOS versions, or whether it was different when #186 F1 was reproduced. If caffeinate changed, the original repro may have been diagnosed against behaviour that no longer exists.

## Second-order question this raises

If caffeinate execs in place, the process that holds the `PreventUserIdleSystemSleep` assertion is the factory itself, having inherited it across the exec - not a separate supervising process. That is a different lifetime model from "a wrapper holds the assertion until the utility exits", and it may matter to #203, which asks whether the assertion survives a dark wake. The two questions should probably be measured together with `pmset -g assertions`.

## What is already done

#208 pins the measured topology as a canary:

```python
assert seen == [ran.pid], (
    "caffeinate no longer execs in place, so the daemon's direct "
    "child is not the factory - re-check the process-group "
    "termination path in subprocess_factory_runner"
)
```

so a future change in caffeinate's behaviour fails a test rather than silently invalidating the reasoning. Nothing in the shipping code was changed.

## Acceptance

- A recorded verdict on whether the process-group termination path is justified by the factory's own descendants independently of caffeinate.
- `subprocess_factory_runner`'s docstring states the actual reason, with the measurement, and the #186 F1 reference points at what was really reproduced.
- The macOS-version dependence of the exec-in-place behaviour is noted, or explicitly declared not worth tracking.

0xfauzi:
## Summary

`subprocess_factory_runner`'s docstring explained why a factory run is supervised as a
process GROUP by blaming `caffeinate`: "on macOS the direct child is the `caffeinate`
wrapper, so the factory itself is a grandchild and outlived the timeout". #209 measured
one pid and concluded caffeinate execs in place. Neither reading enumerated the run's
process group, and that is the step that decides it.

Measured 2026-09-07 on macOS 26.6.2 build 25G83, Darwin 25.6.0 arm64 (xnu-12377.161.14,
T8142), `/usr/bin/caffeinate`, spawning exactly as `run_supervised` does
(`caffeinate_prefix(True)`, `start_new_session=True`) and then reading the whole group
with `ps -o pid=,ppid=,pgid=,stat=,command= -g <pgid>`:

```
caffeinate = True    Popen returned 72146, pgid 72146, daemon pid 72140
  72146 72140 72146 Ss   python probe_child.py        <- the pid Popen returned IS the utility
  72147 72146 72146 S    /usr/bin/caffeinate -i ...   <- a surviving caffeinate, CHILD of the utility
  72150 72146 72146 S    sleep 45                     <- the utility's own descendant

caffeinate = False   Popen returned 72177, pgid 72177
  72177 72140 72177 Ss   python probe_child.py
  72178 72177 72177 S    sleep 45
```

`pmset -g assertions` prints TWO lines for that assertion, and the round-1 review found
the PR reporting one of them. Re-measured on this machine today:

```
   pid 43437(caffeinate): [0x...] 00:00:00 PreventUserIdleSystemSleep named: "caffeinate command-line tool"
	Details: caffeinate asserting on behalf of '/bin/sleep' (pid 43436)
```

The row HEADER is keyed on the helper (43437) and carries the `named:` field. The
`Details:` line under it names the UTILITY (43436), which is the pid the daemon holds.
Section 5 of the docs had said the `Details:` wording was "not what this machine
prints"; it is, on every run, and that sentence is gone.

So `caffeinate -i` FORKS. The utility keeps the pid `Popen` returned, a second
`caffeinate` runs as a child of it inside the same group and session, and that child is
the assertion holder. The daemon's direct child is the factory, which is the half #209
got right, and the assertion is not inherited across an exec, which is the half it did
not measure. Pid identity cannot separate exec-in-place from fork-the-helper, because
both leave the utility on the pid the daemon was given, so the earlier probe's data was
consistent with what is measured here and its conclusion did not follow from it.

The reason the group is needed is the factory's own descendants. A factory-shaped child
(`sh -c "sleep 40 & ... "`) SIGKILLed on the direct child alone, which is exactly what
`subprocess.run(timeout=...)` does on expiry, left its descendant running with
`caffeinate` on and with it off alike; `killpg` on the spawn-time pgid emptied the group
in both cases. The design is correct and the reason recorded for it was wrong.

Ten prose sites state the measured reason: two docstrings in `kstrl/serve.py`, four
in `tests/`, two sections of `docs/continuous-intake.md`, and the PR #186 review record.
Each names the sentence it replaces, so the correction is auditable rather than silent.

#203 item 3 ships alongside it, because it is the same mechanism seen from the queue's
side: `serve()` takes `serve_lock` before the cycle reaches `reap_leases`, and that
ordering is the only thing stopping a second scheduled firing requeueing a suspended run
whose wall-clock lease has lapsed. It was correct on the head and nothing tested it.

Every number in this body was re-measured on the current head, which already contains
`origin/main` `1775301`. Rounds 1 and 2 of the review are answered in full below; every
number either round found wrong is corrected in place rather than left standing.

## Round 1 review: what changed

Two blockers, six should-fix, eight nits. Every one was reproduced here before it was
fixed; nothing below is relayed unchecked. Five commits sit on top of `7fd4b70`.

**B1. `read_group_members` answered a group it could not see with a confident `()`.**
`_interpret` had a second refusal the count never consulted: when a complete listing
shows no row for the group, it asks the kernel, and a kernel that says the group is
occupied means the listing did not show every process. Reproduced on `7fd4b70`: the same
listing gave `GroupLiveness(live=None, ...)` and `GroupMembers(pids=(), reason='')`.

The three refusals a parsed listing can earn now live in one function,
`_listing_refusal`, which BOTH reads consult: uid-filtered, unreadable row,
kernel-disagrees. `_listing_for`'s two `ps`-failure heads got the same treatment. Each
read appends its own consequence and nothing else differs, so a refusal added for either
read is a refusal for both. That also fixes the second half of nit N2: the members read
was reporting a `ps` failure with the word "liveness" in it.

**B2. The docs denied a line `pmset` does print.** Corrected in Summary above.

**S1. A row the parse could not read flipped liveness from LIVE to GONE**, and the PR
body said the kill path did not change. Both are fixed. Reproduced on all three trees
with `"1 1 Ss\nbad 7 Ss\n50 7 Z\n"` for group 7: `live=True` on `543e1aa`, `live=False`
on `7fd4b70`, which `serve.terminate_process_group` reads as "the group is gone" for a
group whose only running member is the row it could not read. The handoff this PR
recorded is taken: such a row makes the whole listing unreadable and both reads refuse
it. A blank line does not, because it claims nothing.

The cost was measured before taking it, since a refusal a real `ps` could earn would
stop the daemon reaping anything at all: 20 reads of `PS_ARGV` gave **16320 rows, every
one exactly three columns with a numeric pid and pgid**. The reviewer measured 16331 on
another load, with **0 non-conforming rows both times**; the count is a property of the
machine's load and the shape is not. Round 2 found the "would stop the daemon reaping
anything" clause wrong in the other direction, which is B2 below.

**S2. "Exactly two members with caffeinate, one without" was the census of the test's
childless sleeper**, written as a claim about a factory run's group. The census now
parametrizes a second utility that spawns one child into the run's group: 3 members with
caffeinate and 2 without, beside the sleeper's 2 and 1. What is pinned is the DIFFERENCE
the prefix makes.

**S3. NOT resolved by round 1, and round 2 measured why.** The mark was attached to
every census case and could not fire, because the predicate behind it answered True on
exactly the `ps` it exists to skip. See B1 in the next section.

**S4. The class-level darwin skip cost CI the whole census.** Measured on both trees
with `_NEEDS_CAFFEINATE` forced to skip, which is the CI shape: on `7fd4b70`, 4 passed
and 4 skipped, and all four skips are the census class. Here, 7 passed and 3 skipped.
And the `start_new_session=True` mutation, run under that same simulation, is now caught
by the two `bare` cases where it was invisible before. The marker moved onto the
`caffeinate=True` parameters.

**S5. The forked helper holds the run's stdout and stderr.** Reproduced with `lsof`: fds
1 and 2 of the utility (46863) and of the helper (46864) are the same pipe object
`0xa503981d70e6f31e`. So `communicate()` cannot see EOF until the HELPER exits, and
`factory_timeout_seconds` defaults to `0.0`, which that call turns into `timeout=None`.
Latent: over 8 spawns the group is empty at the first sample after the utility exits.
Recorded in `caffeinate_prefix`.

**S6. Neither "load-bearing" claim was.** Deleting the census hold loop leaves the file
green, and so does deleting it TOGETHER with the mutation its own comment says it exists
to catch. Deleting the lease fixture's `time.sleep` leaves the file green; raising the
TTL does not. Both are now described as the insurance they are, with the measurements.

**Nits.** N1 renamed (the group is pytest's, not one the test created). N2 above, plus
"running" back in `_FILTERED_VIEW` with a test that pins the word. N3 `except
BaseException` narrowed to `except Exception`. N4 `_census` says which sample `>=` keeps
and why it does not delegate. N5 documented, not closed: the reaper class now records
that it pins the path and not the property, with the green mutant that shows it and the
call-site census that would close it, which is Handoff 4. N6 both drafting-history
docstrings deleted. N7 re-measured, below. N8 merged.

**N7, re-measured.** The PR reported the caffeinate helper visible 12/12 at a median of
2.1 ms and a max of 2.4 ms. Same instrument, same loop, on this machine: 12/12, **median
21.6 ms, max 22.4 ms**, while one `read_group_members` costs a median of **21.7 ms** over
40 samples. The helper is in the FIRST sample every time and the fork is faster than
anything this census can time, so no latency figure is claimed any more.
`_SETTLE_DEADLINE_SECONDS` is stated as 223x the first sample, which is what it is.

**One thing round 1 created and did not hide.** `kstrl/procgroup.py` went to exactly 800
lines, the ratchet limit, with prose compressed to fit. That prediction came true in one
round: round 2 asked for three more paragraphs of recorded reasoning, and the file was
split rather than trimmed. See "The split" below.
## Round 2 review: what changed

Two blockers, five should-fix, seven nits. Every one reproduced here before it was
fixed. Both blockers were in the direction this repo calls its dominant defect: a guard
that goes blind rather than red, and a behaviour change whose harmful half the record did
not state.

**B1. The skip predicate returned True on exactly the `ps` it exists to skip.**
`procs.ps_is_readable` was `read_group_liveness(os.getpgrp()).live is True`. `_interpret`
answers True off a visible runner BEFORE it consults the refusal table, correctly, since
a partial listing can only show fewer processes; the caller's own group always holds the
caller; and under a `ps` filtered to one uid our own processes are precisely the ones
still visible. So it was True by construction on every machine while
`read_group_members` refused the same listing and raised through `on_spawn`. It is now
`read_group_members(os.getpgrp()).pids is not None`, which consults the shared refusal
table the rest of this PR built.

The previous body called this guard the one thing it could not mutate. The mutation
exists and is the one every other test in the file already uses. Measured here, one real
`ps` read with the pid-1 row stripped and put back:

| listing | old spelling | this one |
|---|---|---|
| honest (pid 1 present) | True | True |
| uid-filtered (pid 1 stripped) | **True** | **False** |

`TestTheSuiteSkipPredicateCanActuallyFire` pins both rows off ONE real read, so the
difference between them is that row and nothing about load. It also asserts the MARK's
condition, not just the predicate: a skipif built the way `procs` builds it evaluates
True under the filtered listing, so the skip fires rather than merely being attached.

**B2. The unreadable-row refusal has a second direction, and it is the one that poisons
a run.** `readable` is a whole-listing flag, so one short row anywhere in `ps -A` refuses
every group read taken from that listing. For a group holding only a zombie that turns a
reapable answer into an unreadable one, and `serve._group_liveness_for_reap` then falls
back to `signal_probe_alive`, which counts an unreaped zombie as alive by design. The
reviewer measured it on three groups they created:

| group | honest read | under a refusal |
|---|---|---|
| one live member | live=True, reap alive=True | alive=True, degraded |
| **one zombie, nothing running** | **live=False, reap alive=False** | **alive=True, degraded** |
| genuinely empty (kernel ESRCH) | live=False, reap alive=False | alive=False, still reaps |

**The refusal is KEPT, and the decision is now recorded in three places** (this section,
the CHANGELOG entry, and `_read_listing`'s docstring). A mangled row is evidence about
the STREAM, not about the row: a listing carrying one row the parse could not read is a
listing whose other rows cannot be trusted either. The refusal's failure direction is
over-reporting alive through the signal-probe fallback, which is the direction
`kstrl/procgroup.py`'s module docstring already declares as that fallback's only error.
The concrete cost is that a zombie-only group under a refusal reads as alive, the item is
poisoned and set `needs_human`, and that is VISIBLE; the alternative error, a running
group called gone, is silent, and silence is #186 F1 itself. A genuinely empty group
still reaps, because the kernel answers ESRCH without consulting the listing.

The cost sentence at the old `kstrl/procgroup.py:654` said a refusal "would stop the
daemon reaping anything". The table shows it would not, and it now reads: "would leave
every zombie-only group unreapable, because the fallback signal probe counts an unreaped
zombie as alive; a genuinely empty group still reaps". Unreachable on real output by both
measurements, so this is a decision about the record rather than about a live case.

**S1. The pgid was proved to parse as an int and then compared as a string.**
`_reads_as_int(group)` converts the cell to establish readability and the next line
compared `group != str(pgid)`, so any spelling `int()` accepts and `str` does not produce
was attributed to no group at all, with `readable` still True and no refusal. The
reviewer measured `007`, `+7`, `0_7` and an Arabic-Indic seven each returning a confident
`(51,)` for a group holding `(50, 51)`. Now `int(group) != pgid`. `pid == "1"` one line
up got the same treatment, and that half needed its own pin: reverting it alone left the
four pgid rows green, so `test_pid_1_is_pid_1_however_it_is_spelled` was added.

**S2. An unreadable pid row was only unreadable when its neighbour column named the group
being asked about.** The pid check moved above the group filter. `readable` is a property
of the (listing, pgid) pair. Measured cost, by the reviewer: +0.082 ms per parse of a
real 813-row listing (0.199 to 0.281 ms, median of 200) against a `ps` read costing about
22 ms, and it is in the docstring where the order is explained.

**S3. The module docstring still called the census macOS-only**, which is what S4
changed. It now says only the `caffeinate=True` cases are, and that the two bare cases
run on CI, which is where the `start_new_session` mutation is caught.

**S4. The assertions whose message names the round-1 N2 defect could not detect it.**
They checked the read's OWN consequence was present, not that the foreign one was absent,
so a read reporting both passed. The reviewer's P5 and P6 measured exactly that: still
green, twice. Both count tests now assert `_UNMEASURABLE not in members.reason`, and the
liveness pair asserts `_UNCOUNTABLE not in liveness.reason`. Both constants are imported
from the module rather than spelled again. P5 and P6 replanted: **caught**, and so are
the two mirror plants that put the COUNT's consequence into the liveness heads.

**S5. The one test that reads the real `ps` had no environment guard.**
`test_it_reads_the_real_group_this_process_is_in` now carries the same fixed mark.

**N1. One home for both skip predicates.** `ps_is_readable` had three spellings and the
caffeinate pair had three more, in a PR whose own commit was titled "one home for each
thing the lane wrote twice". Both predicates and both marks now live in
`tests/helpers/procs.py`: `NEEDS_A_READABLE_PS` and `NEEDS_CAFFEINATE`. Every call site
goes through them.

**N2. `_listing_refusal` reads as a pure predicate and issues a syscall.** The name is
kept (the rename would touch more than the definition and its call sites) and the
docstring's first sentence now says the third branch issues `killpg(pgid, 0)`, that the
count read makes that syscall on a path that did not exist before, and that `_interpret`
lost its explicit `_kernel_says_group_is_empty` call at the point of decision.

**N3. The 800-line ratchet.** B2, S2 and N2 add prose, and no reasoning was deleted to
fit, so the file was split. See "The split" below.

**N4-N7.** The hook line now says `check_agent_docs_sync.py` is `pass_filenames: false`
and receives zero filenames. "62 tests to 60" is now "62 passed to 60". The row count is
stated as 16320 on one load, 16331 on the reviewer's, 0 non-conforming both times.
`TestTheRealRunnerExecsItsArgv` is now `TestTheRealRunnerRunsItsArgvAsAChild`, which is
what its own docstring had been saying for a round.

## The split, and the four pins it moved

`kstrl/procgroup.py` was at exactly 800, which is the ratchet limit, and round 2's N3
recorded that the next added line would fail the hook. B2, S2 and N2 add prose. Nothing
was deleted to fit, so the file was split.

`kstrl/procgroup_listing.py` (270 lines) holds the layer that turns bytes into a
`_Listing` and nothing else: `PS_ARGV`, `PS_TIMEOUT_SECONDS`, `PS_KILL_GRACE_SECONDS`,
`_read_ps`, `_Listing`, `_read_listing`, `_reads_as_int`. `kstrl/procgroup.py` imports
every one of them back and declares `__all__`, so no public name moved and no caller
changed. It is 675 lines now.

`_listing_refusal` and the three refusal-head constants deliberately did NOT move, and
that is a departure from the plan this round was given. Its third branch consults the
kernel, which is exactly what N2 is about, so moving it drags
`_kernel_says_group_is_empty`, then `_probe_says_gone`, then `_may_signal_group` with it,
and `_may_signal_group` is the guard `safe_pgid`, `_refuse_group` and `signal_group` all
rest on. The new module would then be "the listing parser plus the signal primitives",
which is not a cut. The cut taken is the one with no back-edge: `procgroup_listing`
imports nothing from `procgroup`.

**Every census pin was re-derived by RUNNING the guard, never by editing the literal.**
Four moved, and each one FAILED first, which is the guards working:

| Pin | Where | `origin/main` | Here |
|---|---|---|---|
| `EXPECTED_PS_COMMAND_SPELLINGS` | `tests/test_procgroup.py` | `{procgroup.py: 5, tests/test_process_scoping.py: 2, tests/test_serve.py: 6, tests/test_shutdown.py: 1}` | `{procgroup.py: 5, procgroup_listing.py: 1, tests/helpers/procs.py: 1, tests/test_procgroup_members.py: 2, tests/test_process_scoping.py: 2, tests/test_serve.py: 6}` |
| `EXPECTED_PROCESS_MODULES` | `tests/test_process_lifecycle.py` | `procgroup.py` holds `communicate` | `communicate` moves to a new `procgroup_listing.py` row |
| `EXPECTED_SPAWNERS` | `tests/test_process_lifecycle.py` | `procgroup.py: SpawnerRules(1,1,1,1)` | `procgroup_listing.py: SpawnerRules(1,1,1,1)`, `procgroup.py` absent (pinned to zero on all four) |
| `POPEN_ALLOWLIST` / `CHILD_WAIT_SCOPE` | `tests/test_timeout_enforcement.py` | holds `kstrl/procgroup.py` | holds `kstrl/procgroup_listing.py` |

`_PS_OWNER` moved with the call site, and the uniqueness claim moved with it: the
sentence "the only place in `kstrl/` or `tests/` that shells out to `ps`" is now in the
file that has the call, which is what `test_the_claim_names_the_roots_the_net_actually_walks`
reads. `procgroup.py` points at it rather than repeating it.

The `POPEN_ALLOWLIST` move is the same shape as the miss recorded on `CHILD_WAIT_SCOPE`
as instance eleven of the guard-goes-blind class: `kstrl/procdispose.py` was split out of
`procgroup` and landed on neither list, and the gate's wait count fell from 11 to 7 with
nothing failing. What caught it this time is that gate's own per-file anti-vacuity check.
Leaving `kstrl/procgroup.py` on the list as well as the new file fails with "these
deadline-managed modules show no wait sites at all: kstrl/procgroup.py", so a stale entry
cannot be carried quietly. That was measured, not assumed: the first edit did leave both
names on the list and the gate went red.

One thing a future test can get wrong, said in the new module's docstring rather than
discovered: patching `kstrl.procgroup.PS_ARGV` now reaches a re-exported binding that
`_read_ps` does not read. Measured when the file was cut, the four bounded-call tests
that did this went RED rather than quietly green, and they were retargeted at
`kstrl.procgroup_listing`. Red is not guaranteed for the next such test.

## Behaviour changes

**One behaviour change, latent rather than active, and BOTH of its directions.** Round 1
put it here and round 2 found half of it missing. A `ps` row this parse cannot read now
makes the whole listing a REFUSAL for both public reads, where the PR head dropped it and
reported `live=False`, which `serve` reads as "the group is gone".

| group | honest read | under a refusal |
|---|---|---|
| one live member | live=True, reap alive=True | alive=True, degraded |
| **one zombie, nothing running** | **live=False, reap alive=False** | **alive=True, degraded** |
| genuinely empty (kernel ESRCH) | live=False, reap alive=False | alive=False, still reaps |

So refusing costs the middle row: a zombie-only group used to read as gone and reapable,
and under a refusal `serve._group_liveness_for_reap` falls back to `signal_probe_alive`,
which counts an unreaped zombie as alive by design, so the run is not called reaped and
the item is poisoned and set `needs_human`.

**The refusal is KEPT.** A mangled row is evidence about the STREAM, not about the row: a
listing carrying one row the parse could not read is a listing whose other rows cannot be
trusted either. Its failure direction is over-reporting alive, which
`kstrl/procgroup.py`'s module docstring already declares as that fallback's only error,
and the cost is VISIBLE: a poisoned item shows up in the inbox, where a released one is
silent, and silence is the #186 F1 hazard this whole reading exists for. A genuinely
empty group still reaps, because the kernel answers ESRCH without consulting the listing.

Unreachable on real `ps` output, measured twice: 16320 rows on one load, 16331 on the
reviewer's, **0 non-conforming rows both times**. So this is a decision about the record
rather than about a live case, and it is now recorded in three places: here, the CHANGELOG
entry, and `_read_listing`'s docstring.

Two narrower readings changed with it, both unreachable on real output for the same
reason: the pgid column is compared as an integer rather than against `str(pgid)`, and
the pid column's readability is checked before the group filter. Round 2's S1 and S2.

Everything else in production is small:

- `kstrl/procgroup.py` gains `read_group_members(pgid)` and `GroupMembers`. It is
  READ-ONLY, it has no caller in `kstrl/` (the daemon still uses `read_group_liveness`),
  and it shares the one `ps` call, the one parse and now one refusal table with the
  liveness read. It exists because the census the test needs is a question about
  MEMBERSHIP that a bool cannot answer, and because the first draft put a second `ps` in
  `tests/helpers/procs.py`, where `tests/test_procgroup.py::TestOnlyOneModuleShellsOutToPs`
  caught it. Only the census layer caught it; the walking layer missed it, because the
  argv began with a starred name it cannot fold.
- `_Listing` carries the pids it parsed (`listed`, `running_pids`) and a `readable` flag,
  and derives `rows`/`running` from the tuples.
- Both reads refuse the same five listings: `ps` did not run, `ps` exited non-zero, the
  view is uid-filtered, a row could not be read, and the kernel contradicts an empty
  listing. They differ only in the consequence each appends.

Everything else in `kstrl/serve.py` is a docstring.

## Design decisions

- **The new tests live in a new file**, `tests/test_serve_process_tree.py`. The
  reaper-ordering class is not itself about a process tree, so that is a slightly loose
  home; the module docstring says plainly what the two have in common. Splitting is a
  one-line change if preferred.
- **The #186 record in `docs/dark-factory-roadmap.md` is ANNOTATED, not rewritten.** It
  is a record of what that review found, and rewriting it would erase that a wrong
  mechanism was believed for a year. The correction is a dated italic paragraph naming
  the wrong sentence and the measurement.
- **No `pmset` assertion in the suite.** The property that matters is that the power
  assertion is released when the group is killed, and that follows from the helper being
  inside the group, which the census pins, plus the measured release. Parsing
  `pmset -g assertions` would add a dependency on an output format for a derived fact.
- **No command lines in the census.** `procgroup`'s listing asks for `pid=,pgid=,stat=`
  deliberately (command lines measured 23.5ms per call against 11.6ms). So the argument
  that the extra member is caffeinate's is DIFFERENTIAL: the parametrized runs differ in
  nothing but `caffeinate_prefix`, and one has a member the other does not. A second,
  `ps`-free test closes the same fact from inside the utility, by asking whether it
  inherited a child it never created.
- **The census pins a difference, not a constant.** Four parametrizations, two utilities
  (childless, and one that spawns a child) crossed with the prefix on and off. The
  caffeinate cases are marked; the two bare cases run on CI, which is what S4 bought.
- **#203 items 1 and 2 stay open and are owner-run.** Whether `caffeinate -i` holds the
  machine up against a dark wake's return to sleep needs a real suspend on real hardware.
  The runnable procedure is in `docs/continuous-intake.md` section 7, written as five
  steps plus what each result decides. `Refs #203`, never `Closes`.
- **`tests/test_procgroup.py` shed parse tests** to `tests/test_procgroup_members.py`. It
  was at 796 of the 800-line ratchet. Nothing was deleted; it is 799 now, after
  round 3 moved two more (see "Round 3 gates").

## Tested, round 1

The gates as they stood at `bca5af7`, kept because round 2 re-measured every number in
this table and agreed with all of them. The CURRENT gates are in "Round 3 gates" below.

| Gate | Result |
|---|---|
| `uv run pytest tests/ -q -p no:cacheprovider` | **6126 passed, 37 skipped, 42 xfailed, 569.93s, 0 failures** |
| `uv run mypy kstrl/ --strict` | Success, 132 source files |
| `uv run python scripts/gen_docs.py --check` | README.md generated sections are current |

The four `scripts/precommit/` hooks, run in a detached worktree at `origin/main` with
this branch's tree staged over it, each with the argv its own hook config hands it:

| Hook | Result |
|---|---|
| `file_length_ratchet.py` (9 python files) | rc=0. Three informational GREW rows for files already over the limit. Nothing CROSSED. |
| `cyclomatic_ratchet.py` (9) | rc=0. `_interpret` 5 to 3 IMPROVED, `read_group_liveness` 3 to 2 IMPROVED, `_read_listing` 5 to 8, under the limit. |
| `check_no_mocks.py` (2, after the hook's own tests exclude) | rc=0 |
| `check_agent_docs_sync.py` (`pass_filenames: false`, so 0 filenames) | rc=0 |
| em dashes / emoji in the staged diff | 0 / 0 |

Two of those were WRONG on the first attempt and the failures are worth naming, because
both return a plausible number. Passing the file list to a zsh command as `$files` gives
ONE argument containing newlines rather than nine, so `file_length_ratchet` measured
nothing and returned rc=0, which is indistinguishable from a pass. And `check_no_mocks`
over the whole file list returns rc=1 with 120 hits, all of them tests doing their job,
because the hook config excludes `tests/` and a hand invocation does not. Both re-run
with a real argv array and the hook's own filter.

File lengths, `origin/main` to this branch: `kstrl/procgroup.py` 616 to **800**, which is
the ratchet limit exactly; `kstrl/serve.py` 3704 to 3752; `tests/helpers/procs.py` 429 to
457; `tests/test_procgroup.py` 796 to 798; `tests/test_serve_seam.py` 749 to 767;
`tests/test_serve.py` 3409 to 3419; `tests/test_timeout_enforcement.py` 1699 to 1703; and
the two new files at 275 and 585.

The suite is 6126 against 6117 on the reviewer's own merge of the same main into
`7fd4b70`. The +9 is this round's tests: `tests/test_procgroup_members.py` 10 to 17 and
`tests/test_serve_process_tree.py` 8 to 10.

### Mutations, round 1

32 planted, each alone except one deliberate composite, `__pycache__` deleted before
every run, `PYTHONDONTWRITEBYTECODE=1`, each in its own session with `os.killpg` on
expiry, bounded at 300 to 600 s, every anchor asserted to appear exactly once before
planting so COULD-NOT-PLANT is a distinct outcome. **Zero could-not-plant, zero hung.**
The tree was `git status` clean after every batch.

The review's 21 rows, re-run here: **13 caught, 8 still green**, and every row agrees
with the review's own verdict.

| # | Mutation | Outcome |
|---|---|---|
| R1 | `caffeinate_prefix` returns `[]` always | caught |
| R2 | `caffeinate_prefix` ignores its `enabled` flag | caught |
| R3 | `run_supervised` drops `start_new_session=True` | caught |
| R4 | census: drop the HOLD loop | still green, S6 |
| R5 | census: keep the FIRST sample, not the largest | still green |
| R6 | `read_group_members` returns `listed`, counting zombies | caught |
| R7 | the uid-filtered refusal is skipped | caught |
| R8 | the "undercount" tail is dropped | caught |
| R9 | a non-numeric pid becomes an anonymous member | caught |
| R10 | `serve()` runs the cycle BEFORE `serve_lock` | caught |
| R10b | the same mutant, the pre-existing lock tests only | still green, 18 passed 196 deselected, deliberate |
| R11 | the reap spy stubs an empty result instead of delegating | still green |
| R12 | the lease fixture never waits for the lapse | still green, S6 |
| R13 | census expects 1 member with caffeinate | caught |
| R14 | fork-probe expectations swapped | caught |
| R15 | `group_member_pids` degrades to `[]` | still green, unreachable: `ps` is readable here |
| R16 | COMPOSITE: R2 plus the hold loop removed | caught, which is what falsifies the hold's stated justification |
| R17 | the lease TTL never lapses | caught, so the TTL is the load-bearing half |
| R18 | census settle loop removed | still green |
| R19 | `_FILTERED_VIEW` drops "did not list pid 1" | caught |
| R20 | `serve()` runs the cycle AFTER releasing the lock | still green, N5 |

Two had to be RETARGETED, and reporting that matters because the first attempt at each
gave a DIFFERENT answer. R10/R10b: a first mutant that returned before the lock without
taking it at all was caught by the pre-existing lock tests, because it removes the lock
rather than reordering the reaper. The faithful mutant keeps the lock and runs the cycle
first, and reproduces the review and this PR's own M3b byte for byte. R11: a first mutant
returning `None` crashed `serve_cycle` and read as caught; the faithful one returns an
empty `ReapResult` and is still green.

Eleven new plants, one per thing this round changed. **All eleven caught.**

| # | Mutation | Outcome |
|---|---|---|
| F1 | B1 replanted: the kernel cross-check leaves the shared table | caught |
| F2 | B1 the other way: the count never consults the table | caught |
| F3 | S1 replanted: an unreadable pid row is dropped, as on `7fd4b70` | caught |
| F4 | S1 the other way: the readable check leaves the shared table | caught |
| F5 | the count reports a `ps` failure without its own consequence | caught |
| F6 | N2 replanted: `_FILTERED_VIEW` loses the word "running" | caught |
| F7 | S2: the factory-shaped case expects 2 members instead of 3 | caught |
| F8 | S2: the spawner's child escapes into its own group | caught |
| F10 | a blank line counts as unreadable, so every listing refuses | caught |
| F12 | `_reads_as_int` uses `str.isdigit` instead of `int()` | caught |
| F15 | `_interpret` checks the refusal before the runner | caught |

Round 1 said S3's `_NEEDS_A_READABLE_PS` "cannot be shown red here because `ps` IS
readable on this machine". **That was wrong**, and it is the sentence round 2's B1 came
out of: the mutation exists, it is the one every other test in the file already uses, and
it is planted in round 3 as RF1. What remains genuinely unmutable from round 1 is N3's
`except Exception` narrowing, which no test can see, and the prose edits.

### Static guards, round 1

Every census re-derived by RUNNING it, never by diffing a pin, at `origin/main` and
before and after the merge. Superseded by "Round 3 static guards" below, which re-ran
both sides after the module split.

- 51 files under `tests/` AST-walk, 40 of them executable guard test files, the same on
  `origin/main` and here. Executing all 40: **1220 passed / 9 skipped / 41 xfailed at
  `origin/main`, 1218 / 9 / 41 here.** The delta is **-2**, not the -3 this body reported
  before: `tests/test_procgroup.py` drops from 62 passed to 60 because two parse tests
  moved into `tests/test_procgroup_members.py`, which does not AST-walk. The third parse
  test in that file is new, not moved. Round 3 re-derived all of this; the current
  numbers are in "Round 3 static guards" below.
- `EXPECTED_PS_COMMAND_SPELLINGS` moves twice, and both moves are diagnostic sentences
  that open with the word rather than second `ps` call sites: `procgroup.py` 5 to 6 for
  `_UNREADABLE_ROW`, plus new rows `tests/test_procgroup_members.py: 1` and
  `tests/test_serve_process_tree.py: 1`. Layer 1 flagged the last one on its own, which
  is the census working. The walk still finds exactly one call site in `procgroup.py`.
- `astwalk.test_sources()` and the `process_scoping` corpus are 232 on the pre-merge
  branch and 234 after, and 232 at `origin/main`: the +2 is this PR's two new test files.
- `EXPECTED_APPEND_OPENS`, `EXPECTED_ROUTED_APPENDS`, `EXPECTED_JOURNAL_PATH_SITES`,
  `EXPECTED_EVENT_NAME_SPELLINGS` and `EXPECTED_MACHINE_WIDE_SPELLINGS` are byte-identical
  across the merge and against `origin/main`.

## Round 3 gates

On the committed head, from the worktree, `__pycache__` deleted first and
`PYTHONDONTWRITEBYTECODE=1`:

| Gate | Result |
|---|---|
| `uv run pytest tests/ -q -p no:cacheprovider` | **6138 passed, 37 skipped, 42 xfailed, 350.30s, exit 0** |
| `uv run mypy kstrl/ --strict` | Success, 133 source files |
| `uv run python scripts/gen_docs.py --check` | README.md generated sections are current |

The four `scripts/precommit/` hooks, run in a detached worktree at `origin/main` with
this branch's tree staged over it (in place compares the tree with itself), each with the
argv its own hook config hands it:

| Hook | Result |
|---|---|
| `cyclomatic_ratchet.py` (12 python files) | rc=0. Four IMPROVED rows, nothing worse. |
| `file_length_ratchet.py` (12) | rc=0. Three informational GREW rows for files already over the limit. Nothing CROSSED. |
| `check_no_mocks.py` (3, after the hook's own exclude) | rc=0 |
| `check_agent_docs_sync.py` (`pass_filenames: false`, 0 filenames) | rc=0 |
| em dashes in the diff / the four commit messages / this body | 0 / 0 / 0 |

**The length ratchet caught this round's own change and was then mutated.** On the first
run it failed: `CROSSED tests/test_procgroup.py 796 -> 825 lines`. Nothing was deleted to
fit; the two liveness ps-failure tests moved to `tests/test_procgroup_members.py` to sit
with their count twins, which is where round 2's S4 says they belong anyway. Re-run:
rc=0. Mutated by appending two blank lines and re-staging: rc=1,
`CROSSED tests/test_procgroup.py 796 -> 801 lines`. Restored: rc=0.

File lengths, `origin/main` to here: `kstrl/procgroup.py` 616 to **675** (was 800 before
the split); `kstrl/procgroup_listing.py` new at 270; `tests/helpers/procs.py` 429 to 520;
`tests/test_procgroup.py` 796 to 799; `tests/test_procgroup_members.py` new at 493;
`tests/test_serve_process_tree.py` new at 582; `tests/test_serve_seam.py` 749 to 779;
`tests/test_shutdown.py` 720 to 718; `tests/test_process_lifecycle.py` 588 to 601;
`tests/test_timeout_enforcement.py` 1699 to 1713.

A standalone re-check of the six things round 2 asked for is committed to the lane's
scratchpad as `r3_verify_head.py`; run from the worktree root it prints one line per
check with the measured value. All six PASS on this head.

### Round 3 red-first

Every fix below has a test that fails without it. "Red against `bca5af7`" here means the
one production hunk reverted to its `bca5af7` shape with the rest of the tree at the
head, which is the same edit and is re-runnable.

| Fix | Test | Red without the fix |
|---|---|---|
| B1 predicate | `TestTheSuiteSkipPredicateCanActuallyFire` (3 tests) | yes, 2 failed |
| S1 pgid as int | `test_a_pgid_int_accepts_is_the_group_it_names` (4 params) | yes, 4 failed |
| S1 pid 1 as int | `test_pid_1_is_pid_1_however_it_is_spelled` (4 params) | yes, 4 failed, and **still green until this test was added** |
| S2 pid before the filter | `test_a_row_the_parse_cannot_read_...`, new param `1 1 Ss / bad 999 Ss / 50 7 S` | yes, 1 failed |
| S4 foreign consequence | the four `TestAPsThatGaveNoAnswerIsRefusedByBOTHReads` tests | yes, 1 failed on each of four plants |

The B1 pair is measured off ONE real `ps` read with the pid-1 row stripped and put back,
so the difference between True and False is that row and nothing about the machine. The
test also asserts the old spelling, `read_group_liveness(os.getpgrp()).live is True`,
still answers True on the same filtered listing: without that control a False could mean
the wrapper had broken the read rather than filtered it.

### Round 3 mutations

Twelve plants. Each planted alone, `__pycache__` deleted before every run,
`PYTHONDONTWRITEBYTECODE=1`, each in its own session with `os.killpg` on expiry, bounded
at 300 s, and every anchor asserted to appear exactly once before planting, so
COULD-NOT-PLANT is a distinct outcome. **Zero could-not-plant, zero hung.** The tree was
restored and `git status` clean after each.

The first four are the round-3 fixes replanted: each reverts the one production hunk to
its `bca5af7` shape, which is red-first measured against the head with the rest of the
tree unchanged.

| # | Mutation | Outcome |
|---|---|---|
| RF1 | B1: `ps_is_readable` back to the liveness spelling | **caught** (2 failed) |
| RF2 | S1: the pgid compared against `str(pgid)` again | **caught** (4 failed) |
| RF2b | S1: the pid-1 completeness compared as text again | **caught** (4 failed), after its own pin was added |
| RF3 | S2: the pid check back below the group filter | **caught** (1 failed) |
| P5 | S4: `_listing_for` head 4 embeds the liveness consequence | **caught** (1 failed), still green in round 2 |
| P6 | S4: `_listing_for` head 5 embeds the liveness consequence | **caught** (1 failed), still green in round 2 |
| P5b | the mirror: head 4 embeds the COUNT consequence | **caught** (1 failed) |
| P6b | the mirror: head 5 embeds the COUNT consequence | **caught** (1 failed) |
| P1 | `_listing_refusal` drops the uid-filtered head | caught (4 failed) |
| P2 | `_listing_refusal` drops the unreadable-row head | caught (2 failed) |
| P3 | `_listing_refusal` drops the kernel-disagrees head | caught (2 failed) |

**RF2b is the one worth reading.** Reverting `int(pid) == 1` to `pid == "1"` was **still
green** on the first attempt: the four pgid spellings do not exercise the completeness
column, so nothing pinned it. The result is reported rather than folded into RF2, and
`test_pid_1_is_pid_1_however_it_is_spelled` was added to close it. Its failure direction
is the opposite of the pgid one and worth naming: an unrecognised pid-1 row leaves
`complete` False, so a full listing is REFUSED as filtered to this uid rather than
undercounted.

### Round 3 static guards

The guard set derived the review's way, `grep -rl -E 'ast\.parse|astwalk' tests/*.py`, and
EXECUTED at both revisions rather than read off a pin.

| | `origin/main` `1775301` | head `b714bf7` |
|---|---|---|
| AST-walking files under `tests/` | 37 | 37, the same set |
| executing all 37 | **1208 passed, 9 skipped, 41 xfailed (51.5s)** | **1204 passed, 9 skipped, 41 xfailed (52.9s)** |

The **-4** reconciles exactly: `tests/test_procgroup.py` goes from 62 passed to 58, and
all four are tests that moved into `tests/test_procgroup_members.py`, which does not
AST-walk. Two moved in round 1 (the parse tests) and two in round 3 (the liveness
ps-failure pair, moved to sit with their count twins). Measured on both trees rather than
inferred: 62 and 58.

Four census pins moved, every one re-derived by RUNNING the guard and every one RED
first. They are in "The split" above.

## Assumed

- **Nothing about macOS 25.5.** Whether `caffeinate` behaved differently then is
  untestable from here, because the only evidence is a probe that read pid identity, and
  pid identity is the one thing exec-in-place and fork-the-helper agree on. No claim
  either way is written in the code or the docs.
- **Nothing about a `ps` that a container or `hidepid` would produce.** Round 3 narrows
  this rather than closing it: the skip predicate IS now measured against a filtered
  listing, but the listing is a real one with its pid-1 row stripped, which is the only
  thing a `hidepid` mount changes about the answer, not a `ps` configured to filter. No
  container or `hidepid` mount was set up.
- **The docs' 2-second dark-wake observation and the 0.08-0.12s empty-cycle timing** are
  quoted from #203 and #204 and were not re-measured.
- **Whether `caffeinate -i` holds the machine up against a dark wake's return to sleep**
  is not measured, and is marked as such in `docs/continuous-intake.md` section 7.

## Handoffs

1. **`run_supervised` calls `on_spawn` OUTSIDE its `try`.** An exception from the
   lease-adoption callback escapes with the factory child neither waited on nor signalled.
   In production `_adopt` catches `(QueueError, OSError)`, so the reachable failures are
   handled; an unexpected exception type, or a `MemoryError`, would leak a running
   factory. Measured earlier in this lane: the child is still a running member of its
   group 3.0s after a timeout that never ran. A different defect class, so it is reported
   rather than fixed. The census test captures and re-raises after the supervisor is done
   with the child, and says so.
2. **Taken in round 3.** `kstrl/procgroup.py` was at exactly 800 lines and this round adds
   prose, so the `ps` read and its parse moved to `kstrl/procgroup_listing.py`. 675 and 270
   now. See "The split" above for the four census pins that moved with it.
3. **#203 items 1 and 2 are owner-run and stay open.** The suspend experiment is a
   runnable procedure in `docs/continuous-intake.md` section 7: start
   `caffeinate -i /bin/sleep 300`, confirm the assertion row against the forked helper's
   pid, close the lid on battery until a `DarkWake ... SleepService` entry appears, then
   read `pmset -g log` for whether a `Sleep Service Back to Sleep` line follows while the
   assertion's pid is alive, and repeat with `caffeinate -s`. #204 wants the same hardware.
4. **`TestTheReaperRunsOnlyUnderTheDaemonLock` pins the path, not the property.**
   Measured: a `serve()` that takes `serve_lock`, releases it, then runs the cycle leaves
   all four tests green, because the contended case still raises at acquisition (R20). A
   call-site census over the one call site each of `reap_leases` and `serve_cycle` has,
   shaped like `TestOnlyOneModuleShellsOutToPs` next door, would close it. Not taken here:
   it is new static-guard machinery, and this repo's own rule is that a guard needs its
   own two layers and its own mutation. The number is recorded in the class docstring so
   it does not have to be re-derived.

Closes #209

#203 stays OPEN: only item 3 ships here, and items 1 and 2 need a real suspend on real
hardware. `Refs #203`, never `Closes`.

AI code review will be run on this PR as an assist at the owner's request; the owner
remains the gating reviewer (H1).


kstrl/procgroup.py
```diff
@@ -407,16 +451,51 @@ def read_group_liveness(pgid: int) -> GroupLiveness:
...
+    """
+    listing, failure = _listing_for(pgid)
+    if listing is None:
+        return GroupMembers(None, failure)
+    if not listing.complete:
+        return GroupMembers(
+            None,
+            f"{_FILTERED_VIEW.format(pgid=pgid)} A count taken from it would be an undercount.",
+        )
+    return GroupMembers(listing.running_pids)
```
0xfauzi, line 511: **`read_group_members` drops the kernel cross-check `_interpret` applies, so it reports a confidently EMPTY group where the liveness read refuses to answer.**

`_interpret` (L574-L583) has two trust checks on a complete listing: `listing.rows` and then `_kernel_says_group_is_empty(pgid)`. `read_group_members` forwards only the `complete` check and then returns `listing.running_pids` unconditionally.

Measured on this branch, with a `ps` that lists pid 1 but hides every member of a real, definitely non-empty group (the hidepid / truncated-listing case the module exists for):

```
liveness: GroupLiveness(live=None, reason="ps listed no process in group 79011, but the kernel reports that group is not empty, so the listing did not show every process. ...")
members : GroupMembers(pids=(), reason='')
```

That is exactly the failure `GroupMembers`'s own docstring says this read refuses: "a caller counting members is usually asserting that there is no OTHER member, and a listing filtered to one uid would answer that with a confident undercount." The uid filter is not the only way to get one, and the module already owns the second detector. `tests/test_procgroup_members.py:129` (`test_an_empty_group_in_a_complete_listing_is_an_empty_answer`) pins the fail-open as intended, so nothing will catch it.

The docstring line "That is the one undercount this read is able to recognise" is false as written.

```suggestion
    if not listing.listed and not _kernel_says_group_is_empty(pgid):
        return GroupMembers(
            None,
            f"ps listed no process in group {pgid}, but the kernel reports "
            f"that group is not empty, so the listing did not show every "
            f"process. A count taken from it would be an undercount.",
        )
    return GroupMembers(listing.running_pids)
```

kstrl/procgroup.py
```diff
@@ -520,12 +605,28 @@ def _read_listing(stdout: str, pgid: int) -> _Listing:
...
+        # line with fewer than three columns the same way - so this is
+        # unreachable on real output. It is reachable in principle, and
+        # then it is an UNDERCOUNT, which is the direction this module
+        # otherwise refuses; the stronger reading, where a row the parse
+        # could not read marks the whole listing untrustworthy and both
+        # public reads refuse it, is #209's handoff rather than its
+        # change, because it moves what the daemon's kill path does.
+        try:
+            member = int(pid)
+        except ValueError:
```
0xfauzi: **This skip is not only an undercount: it flips `read_group_liveness` from LIVE to GONE, which is the direction the module opens by calling dangerous.**

The row is dropped from `listed` as well as from `running`, so a RUNNING member with an unparseable pid becomes invisible to `_interpret`. When the group's other visible rows are zombies, `rows` is now truthy and `running` is empty, and `_interpret` takes the `#298` branch and returns `live=False`.

Measured, same input against both trees, `__pycache__` cleared:

```
input:  "1 1 Ss\nbad 7 Ss\n50 7 Z\n"   (complete listing, one running member with an unreadable pid, one zombie)

origin/main : _Listing(complete=True, rows=2, running=1)                -> GroupLiveness(live=True)
this branch : _Listing(complete=True, listed=(50,), running_pids=())    -> GroupLiveness(live=False)
```

`live=False` reaches `serve.terminate_process_group` as `GroupTermination(reaped=True)`, which is `serve` releasing the item while a factory may still be writing to the repo - #186 F1, the thing this module was built for. So the PR body's "Nothing in the daemon's kill path changed" and "behaviour on real output is unchanged" hold for real `ps` output only; for this input class the behaviour change is in the fail-open direction, and the comment above the `continue` describes it as an UNDERCOUNT, which understates it.

Minimal fix that keeps the base's fail direction without taking the handoff: carry the count of rows the parse could not read on `_Listing`, have `_interpret` add it to both `rows` and `running` (so an unreadable running member still reads as live), and have `read_group_members` refuse a listing with any of them. Then no reading is silently narrowed and the handoff stays a handoff.

kstrl/serve.py
```diff
@@ -1480,10 +1480,30 @@ def resolve_merge_gate(item: QueueItem, root_dir: Path) -> MergeGate:
...
+    utility in place: it FORKS. The process the daemon spawned keeps the
+    pid ``Popen`` returned and becomes the utility, and a second
+    ``caffeinate`` process appears as a CHILD of it, in the same process
+    group and session. That child is the assertion holder - ``pmset -g
+    assertions`` names it ``PreventUserIdleSystemSleep ...
+    "caffeinate command-line tool"`` against the child's pid, not the
+    utility's - so the assertion is not inherited across an exec by the
+    factory. Measured lifetime: the helper is gone at the first sample
+    after the utility exits, and gone after a ``killpg`` on the
+    spawn-time group, with the ``pmset`` row gone in both cases. It
```
0xfauzi: **The fork has a second consequence this section does not record: the helper inherits the run's stdout/stderr pipe, so `communicate()` cannot see EOF until the helper exits - and on the default config that wait has no deadline.**

Measured on this machine with `lsof` against a live helper spawned exactly as `run_supervised` does:

```
COMMAND     PID   FD   TYPE   DEVICE              NAME
caffeinat 97255    1   PIPE   0x31794f771f78259e  ->0x8440722e4cabc810
caffeinat 97255    2   PIPE   0x31794f771f78259e  ->0x8440722e4cabc810
```

The fork happens before the exec, so the helper holds fds 1 and 2 - the same pipe `run_supervised` reads. `Popen.communicate` returns on EOF, and EOF needs every writer to close. `ServeConfig.factory_timeout_seconds` defaults to `0.0`, and `run_supervised` turns that into `communicate(timeout=None)`, so on the default configuration the daemon's wait on a finished factory is now bounded by the caffeinate helper's exit rather than by the factory's.

Measured today it is fine - 5 spawns each, `caffeinate` 0.015-0.016s against 0.012-0.013s bare, a ~3ms difference - so this is latent, not active. But "the helper is gone at the first sample after the utility exits", two lines below, is now a liveness dependency of the no-timeout path and not just a tidiness observation. Given this PR exists because an unmeasured process-tree claim went unchallenged for a year, it is worth stating here (and it is an argument for `factory_timeout_seconds > 0`).

tests/test_serve_process_tree.py
```diff
@@ -0,0 +1,502 @@
...
+    while time.monotonic() < hold_until:
+        time.sleep(_POLL_SECONDS)
+        sample = procs.group_member_pids(pgid)
+        if len(sample) >= len(pids):
+            pids = sample
+    return pids
+
+
+@_NEEDS_CAFFEINATE
+class TestTheRunGroupMembership:
```
0xfauzi, line 210: **The census has no `procs.ps_is_readable()` guard, so on a machine with a uid-filtered `ps` it fails instead of skipping - and it fails with a message pointing at kstrl.**

`_NEEDS_CAFFEINATE` gates on darwin plus the binary, but not on whether `kstrl.procgroup` can actually measure here. On a machine where `ps` does not list pid 1, `read_group_members` refuses (correctly), `procs.group_member_pids` raises `AssertionError("ps did not list pid 1, ... A count taken from it would be an undercount.")`, `on_spawn` captures it and `_run_and_census` re-raises it - so the test goes red over the environment.

`tests/helpers/procs.py:377` exists for exactly this, and `tests/test_shutdown.py:624` is the precedent:

```python
if not ps_is_readable():
    pytest.skip(...)
```

Its docstring names the failure mode verbatim: "a #298 assertion would fail with a message pointing at kstrl rather than at the missing binary". Folding it into the class marker keeps the one-marker simplification this PR made:

```suggestion
class TestTheRunGroupMembership:
```

(with `_NEEDS_CAFFEINATE` extended to `or not procs.ps_is_readable()`, or a second class-level marker beside it).

kstrl/serve.py
```diff
@@ -1480,10 +1480,30 @@ def resolve_merge_gate(item: QueueItem, root_dir: Path) -> MergeGate:
...
+    leaks neither a process nor a power assertion.
+
+    Because the helper is INSIDE the spawned group, the group kill on
+    the timeout path releases the power assertion as well as reaping the
+    factory. ``tests/test_serve_process_tree.py`` pins that membership:
+    exactly two members with caffeinate, one without.
```
0xfauzi: **"exactly two members with caffeinate, one without" is true of the TEST's utility, not of a factory run - and this docstring is the one a reader consults about a factory run.**

`tests/test_serve_process_tree.py` spawns `python -c "import time; time.sleep(30)"`, a utility with no children, so its group holds the utility plus the helper. A real factory run's group holds the factory plus the helper plus every agent subprocess, `git` and verify command the factory spawns - which is the point `subprocess_factory_runner`'s corrected docstring makes forty lines down ("what outlives a signal to the direct child is the factory's OWN descendants").

Same sentence in `docs/continuous-intake.md:364` ("the group has two members rather than one") and :376.

This is the shape the PR is correcting: a measurement taken on one process tree, written as a general claim about the run's process tree, in a docstring nobody re-derives. Suggest scoping it, e.g. "pins that membership on a childless utility: exactly two members with caffeinate, one without".

The substantive claim - the helper is INSIDE the run's group, so the group kill releases the assertion - is unaffected and is what the test actually establishes.

tests/test_serve_process_tree.py
```diff
@@ -0,0 +1,502 @@
...
+
+    def spy(queue: object, **kwargs: object) -> object:
+        calls.append(queue)
+        return real_reap(queue, **kwargs)  # type: ignore[arg-type]
+
+    monkeypatch.setattr(serve_module, "reap_leases", spy)
+    return calls
+
+
+class TestTheReaperRunsOnlyUnderTheDaemonLock:
```
0xfauzi, line 482: **Altitude: this class pins the ordering on the one path `serve()` takes today; nothing closes the property by construction.**

The hazard as stated is a reachability claim - "the reaper is only ever reachable under the daemon lock" - and it holds today because `reap_leases` has exactly one call site (`kstrl/serve.py:2977`, inside `serve_cycle`) and `serve_cycle` has exactly one (`kstrl/serve.py:3481`, inside `serve()`'s `with serve_lock`). I checked: no other caller in `kstrl/`.

Three behavioural tests pin the current path. A new caller - a CLI subcommand, a TUI action, a future `ks queue reap` - would reintroduce the defect with all three still green, because they only exercise `serve()`. That is the shape CLAUDE.md's "prefer **closed by construction** over a **ledger of give-ups**" is about, and the file next door (`tests/test_procgroup.py::TestOnlyOneModuleShellsOutToPs`) is the worked example: a census of the call sites, so a new one shows up as an unexplained delta rather than as a silently uncovered path.

A short AST census of `reap_leases` / `serve_cycle` call sites in `kstrl/`, pinned at 1 each, would close it and costs a dozen lines given `tests/helpers/astwalk/`. Worth it here because the tests below are the only mechanism, and their own docstring says a mutated `serve()` left all 18 pre-existing lock tests green.

tests/test_serve_process_tree.py
```diff
@@ -0,0 +1,502 @@
...
+        spawned: int | None = None
+        members: list[int] | None = None
+        failure: BaseException | None = None
+
+        def on_spawn(pid: int) -> None:
+            nonlocal spawned, members, failure
+            try:
+                spawned = pid
+                members = _census(os.getpgid(pid), expected)
+            except BaseException as exc:  # noqa: BLE001 - re-raised below
```
0xfauzi: **`except BaseException` here also captures `KeyboardInterrupt` and `SystemExit` and defers them for up to ~26 seconds.**

The capture-and-re-raise is right for `Exception` - the reason given (a failing callback would leak a 30-second sleeper) is sound. But `BaseException` means a Ctrl-C landing inside the census is stashed in `failure` and not re-raised until `run_supervised` has finished with the child: `_RUN_TIMEOUT_SECONDS` 1.0s, then `GROUP_TERM_GRACE_SECONDS` 15.0s, then `GROUP_KILL_GRACE_SECONDS` 10.0s, then `ABANDON_GRACE_SECONDS`. So an interrupt during a `ps` poll reads as a hung suite.

`Exception` gets the stated protection (`procs.group_member_pids` raises `AssertionError`, `os.getpgid` raises `ProcessLookupError`, both are `Exception`) and leaves the interrupt behaving as an interrupt:

```suggestion
            except Exception as exc:  # noqa: BLE001 - re-raised below
```

(`kstrl/procgroup.py::_read_ps` catches `BaseException` for the opposite reason - it re-raises immediately after disposal, so nothing is deferred.)

kstrl/procgroup.py
```diff
@@ -172,6 +172,16 @@
...
 
+#: The other sentence two branches report, said once for the same reason.
+#: Both reads refuse a uid-filtered listing and each appends its own
+#: consequence; #209 wrote the head out a second time, which moved this
+#: module's ``ps``-spelling census for a copied sentence rather than for
+#: a new mechanism. ``{pgid}`` is filled in by the caller.
+_FILTERED_VIEW = (
+    "ps did not list pid 1, so the view is filtered to this uid and a "
+    "member of group {pgid} owned by another uid would be invisible."
+)
```
0xfauzi, line 220: **Deduplicating the sentence dropped a word the liveness message needed, and the two callers now append inconsistent tails.**

The message `_interpret` used to emit was "... and a **running** member of group N owned by another uid would be invisible". The shared text says "a member". `read_group_liveness` is answering "is anything still EXECUTING in this group", and the degraded reason is what `serve` puts in front of an operator deciding whether a factory is still writing to the repo; "a member" no longer says which question went unanswered. No test pins the word - the three assertions all match on `"did not list pid 1"` - so this went unnoticed.

Second, smaller: every other refusal in this module ends with `_UNMEASURABLE` ("Process-group liveness cannot be measured here, and reporting 'no live member' would be a false negative"). `read_group_members` appends its own sentence instead, so the one reading whose fail direction the docstring argues is *stricter* is the one whose message does not carry the module's standard "this was not measured" clause. Worth appending both.

Suggest restoring the word and letting each caller add its noun, e.g. `"...and a {kind} of group {pgid} owned by another uid would be invisible."` with `kind="running member"` / `kind="member"`.

tests/test_serve_process_tree.py
```diff
@@ -0,0 +1,502 @@
...
+    "    sys.stdout.write('NO_CHILDREN')\n"
+)
+
+#: The deadline the fork probe gets. It exits on its own in milliseconds;
+#: this only has to be long enough that a loaded machine does not make
+#: the test measure its own timeout path instead.
+_PROBE_TIMEOUT_SECONDS = 60.0
+
+
+def _census(pgid: int, expected: int) -> list[int]:
```
0xfauzi, line 177: **Reuse: this is the third bounded poll over a process group, and the other two live in `tests/helpers/procs.py`.**

`tests/helpers/procs.py` already holds `wait_for_pid_to_die` (:83) and `wait_for_group_to_die` (:421), and this PR just added `group_member_pids` (:101) beside them. `_census` is a settle-then-hold poll over the same subject, sitting in a test module instead - and the argument the new helper's own docstring makes for delegating ("a copied fixture is one that stops matching the helper it feeds", `dead_group`'s wording) applies to the loop as much as to the read.

Moving it next to `group_member_pids` costs nothing here (it takes `pgid` and `expected` and returns pids) and gives the second caller somewhere to land. Not blocking; the loop itself is correct, and the settle/hold split with the reason for each is the good part.

One nit inside it: the docstring says "taking the largest count it sees", but `if len(sample) >= len(pids)` keeps the LAST sample of maximal size, not the first. No effect on the length assertion; `>` would match the prose.

tests/test_serve_process_tree.py
```diff
@@ -0,0 +1,585 @@
...
+#: does not list pid 1 - a ``hidepid`` mount, a container - the read
+#: under ``procs.group_member_pids`` refuses, correctly, and raises
+#: through ``on_spawn``; the red test then reads as a defect in
+#: ``kstrl/procgroup.py`` rather than as an environment that cannot be
+#: measured. ``tests/test_shutdown.py`` skips on the same helper for the
+#: same reason. Evaluated once at import, which costs one ``ps``.
+_NEEDS_A_READABLE_PS = pytest.mark.skipif(
+    not procs.ps_is_readable(),
+    reason="ps here is absent or filtered to one uid, so a group census would be an undercount",
+)
```
0xfauzi: **`_NEEDS_A_READABLE_PS` cannot see the environment it names, so S3's guard never fires.**

`procs.ps_is_readable()` is `read_group_liveness(os.getpgrp()).live is True`, and `_interpret` answers `True` off a visible runner *before* it consults `_listing_refusal`. Under a uid-filtered `ps` our own group is ours, so a runner is always visible and the completeness refusal is never reached.

Measured on this branch by replacing `procgroup._read_ps` with one that strips the pid-1 row out of a real listing:

```
=== uid-filtered ps (no pid 1), the environment the skip names ===
ps_is_readable(): True  -> skipif fires? False
group_member_pids RAISED AssertionError: ps did not list pid 1, so the view is
filtered to this uid and a running member of group 7139 owned by another uid ...
```

So on a `hidepid` mount or a container every census case runs, `_census` raises out of `on_spawn`, and the red test points at `kstrl/procgroup.py` rather than at the environment - the exact outcome this comment says the mark prevents.

The predicate is blind at its pre-existing site too (`tests/test_shutdown.py:624`, where a filtered listing makes the zombie-only group refuse and `process_group_alive` degrade). One fix closes both: consult the refusal table, e.g. `read_group_members(os.getpgrp()).pids is not None`.

kstrl/procgroup.py
```diff
@@ -471,61 +569,147 @@ class _Listing:
...
+    running: list[int] = []
     for line in stdout.splitlines():
         parts = line.split()
-        # A row missing a column would IndexError below. Real ps does not
-        # emit one; a filtered or truncated listing might.
+        if not parts:
+            continue
         if len(parts) < 3:
+            readable = False
             continue
```
0xfauzi: **The unreadable-row refusal has a second direction, and it is the one the CHANGELOG does not state: a zombie-only group flips from reaped to POISONED.**

`readable` is a whole-listing flag, so one short row anywhere in `ps -A` output refuses every group read taken from that listing. Measured on both trees with `"1 1 Ss\nbad\n50 7 Z\n"` for group 7:

```
origin/main 1775301 : read_group_liveness(7).live = False   (gone, #298's case, reapable)
this branch bca5af7 : read_group_liveness(7).live = None    (ps emitted a row this parse could not read...)
```

`None` routes through `_group_liveness_for_reap` to `signal_probe_alive`, which counts a zombie as alive by design, so `terminate_process_group` never confirms the reap, `serve_cycle` sets `needs_human` and poisons the item. That is the "every timed-out run is poisoned" hazard `_group_liveness_for_reap`'s own docstring warns about, reached through a single mangled row rather than through a missing `ps`.

The PR body and the CHANGELOG describe only the safe direction ("dropping it turned a live group into a gone one"). Both directions belong in the behaviour-change note, and it is worth deciding whether an unreadable row should refuse a group whose rows the parse *did* read.

kstrl/procgroup.py
```diff
@@ -471,61 +569,147 @@ class _Listing:
...
         pid, group, state = parts[0], parts[1], parts[2]
         complete = complete or pid == "1"
+        if not _reads_as_int(group):
+            readable = False
+            continue
         if group != want:
```
0xfauzi: **The pgid column is proved to parse as an int and then compared as a string, so a spelling difference silently drops members while `readable` stays True.**

`_reads_as_int(group)` converts the cell to establish readability, discards the conversion, and line 684 compares `group != want` where `want = str(pgid)`. Any spelling `int()` accepts but `str(pgid)` does not produce leaves the row attributed to no group at all, with no refusal. Measured:

```
pgid zero-padded   "1 1 Ss\n50 007 Ss\n51 7 Ss\n"  pgid=7 -> readable=True, running_pids=(51,)
pgid plus-signed   "1 1 Ss\n50 +7 Ss\n51 7 Ss\n"   pgid=7 -> readable=True, running_pids=(51,)
arabic-indic pgid  "1 1 Ss\n50 \u0667 Ss\n51 7 Ss\n"    pgid=7 -> readable=True, running_pids=(51,)
```

`read_group_members(7)` returns a confident `(51,)` where the group holds `(50, 51)` - the undercount the refusal table exists to prevent, reached through the one column the parse converts and throws away. (`read_group_liveness` degrades safely here because the kernel cross-check catches the resulting empty listing, so this is the count read's hazard.)

Fix: compare the converted value, `int(group) == pgid`. `complete = complete or pid == "1"` on line 680 is the same string-vs-int split one line up.

kstrl/procgroup.py
```diff
@@ -471,61 +569,147 @@ class _Listing:
...
             continue
-        rows += 1
+        if not _reads_as_int(pid):
+            readable = False
+            continue
```
0xfauzi: **The pid readability check sits *after* the group filter, so the rule `_read_listing`'s own docstring states is only enforced for rows already attributed to the requested group.**

The docstring three lines up says flatly: "Three shapes are unreadable: a non-blank row with fewer than three columns, a pid column that is not a number, and a PGID column that is not a number." The code makes the second conditional on `group == want`. Measured:

```
"1 1 Ss\nbad 9 Ss\n50 7 Ss\n"  pgid=7 -> readable=True
"1 1 Ss\nbad 7 Ss\n50 7 Ss\n"  pgid=7 -> readable=False
```

So `readable` is not a property of the listing, it is a property of the (listing, pgid) pair, and the same stream is trustworthy or not depending on which group you ask about. `test_a_row_the_parse_cannot_read_makes_the_listing_unreadable` only plants the bad pid *inside* the requested group, so nothing catches the divergence.

Either hoist the pid check above line 684 (making `readable` mean what the docstring says, closed by construction) or narrow the docstring to what the code does. The first is the direction this module's own rule points: a guard that CLEARS must refuse what it cannot prove.

tests/test_serve_process_tree.py
```diff
@@ -0,0 +1,585 @@
...
+   the WALL CLOCK, which advances across a suspend, so a run suspended
+   overnight blows its 3600s lease while its pid is alive. What stops a
+   second scheduled firing requeueing it is that the second firing takes
+   ``serve_lock`` before it reaches the reaper and exits. Measured while
+   writing this: a ``serve()`` mutated to run its cycle BEFORE acquiring
+   the lock left every existing lock test in the suite green.
+
+The census is macOS-only because ``caffeinate`` is, and is skipped on a
+mac that does not have it installed. The lock tests are neither: they
+need ``fcntl``, exactly as ``TestServeLock`` next door does.
```
0xfauzi: **The module docstring contradicts `_NEEDS_CAFFEINATE`'s own comment 40 lines below about which cases are macOS-only.**

Here: "The census is macOS-only because `caffeinate` is, and is skipped on a mac that does not have it installed."

Lines 74-80: "It is a per-parameter mark rather than a class marker, which was round 1 of #209's S4. CI is ubuntu, and under a class marker the whole census contributed nothing there - **including the two cases that need no `caffeinate` and no macOS**."

Two of the four parametrizations (`bare`, `bare-with-child`) carry no mark and run on ubuntu, and they are the ones S4 bought and the only ones that catch the `start_new_session=True` mutation on CI. A reader taking this paragraph at its word concludes the census contributes nothing on CI and could re-add the class marker, which is the S4 defect returning.

This is the defect class the PR exists to fix - a prose claim about behaviour the code does not have - landing in the new file.

tests/test_procgroup_members.py
```diff
@@ -0,0 +1,275 @@
...
+        assert "undercount" in members.reason
+
+    def test_it_reads_the_real_group_this_process_is_in(self) -> None:
+        """The control for all the fakes above: the fake could be
+        modelling a `ps` format that does not exist. This one asks about
+        pytest's own group, which certainly holds pytest. The group is
+        not one this test created, which is why it asserts only its own
+        membership and nothing about the size."""
+        members = read_group_members(os.getpgrp())
+        assert members.pids is not None, members.reason
```
0xfauzi, line 397: **The one test in this file that reads the real `ps` has no environment guard at all.**

Every other real-`ps` reader in the PR sits behind `_NEEDS_A_READABLE_PS` (itself blind - see the comment on `tests/test_serve_process_tree.py`). This one asserts `members.pids is not None` against the live listing, so under a `hidepid` mount or a container `ps` that does not show pid 1 it fails with `"ps did not list pid 1, so the view is filtered to this uid..."` - a red test pointing at `kstrl/procgroup.py` for an environment that cannot be measured, which is precisely what S3 argued must be a skip.

It is the control for the fakes above it, so it should stay; it needs the same (fixed) readable-`ps` guard the census cases get.

tests/test_serve_process_tree.py
```diff
@@ -0,0 +1,585 @@
...
+#: round 1 of #209's S4. CI is ubuntu, and under a class marker the
+#: whole census contributed nothing there - including the two cases
+#: that need no ``caffeinate`` and no macOS. Measured: the mutation that
+#: drops ``start_new_session=True`` from ``run_supervised`` is caught by
+#: this census and by nothing else in this file, and
+#: ``start_new_session`` has nothing to do with macOS.
+_NEEDS_CAFFEINATE = pytest.mark.skipif(
+    sys.platform != "darwin" or shutil.which("caffeinate") is None,
+    reason="caffeinate is macOS-only and must be installed",
+)
```
0xfauzi: **A third local spelling of a skip predicate, in a PR whose own commit is titled "one home for each thing the lane wrote twice".**

This constant's comment argues that copies drift ("the fourth copy of that pair next door in `tests/test_serve_seam.py` has already drifted in wording") and then adds another copy rather than hoisting it. Current census:

- `tests/test_serve_seam.py:433` - `@pytest.mark.skipif(sys.platform != "darwin", reason="caffeinate is macOS-only")`
- `tests/test_serve_seam.py:462` - `if shutil.which("caffeinate") is None: pytest.skip(...)`
- `tests/test_serve_process_tree.py:81` - this one, both halves in a mark

and the same for the readable-`ps` skip: `tests/test_shutdown.py:624` in-body, `tests/test_serve_process_tree.py:93` as a mark.

`tests/helpers/procs.py` already owns `ps_is_readable` and every other process helper; both predicates belong there as one exported mark each. That is also where the `ps_is_readable` fix has to land, so the two changes are the same change.

kstrl/procgroup.py
```diff
@@ -471,61 +569,147 @@ class _Listing:
...
+    The kernel is consulted LAST and only for a listing showing the
+    group holding nothing: it costs a syscall, and a listing that
+    already shows a member has answered the question.
+    """
+    if not listing.complete:
+        return _FILTERED_VIEW.format(pgid=pgid)
+    if not listing.readable:
+        return _UNREADABLE_ROW.format(pgid=pgid)
+    if not listing.rows and not _kernel_says_group_is_empty(pgid):
+        return _KERNEL_DISAGREES.format(pgid=pgid)
```
0xfauzi, line 567: **`_listing_refusal` reads as a pure predicate over a listing, and one of its three answers is a fresh syscall.**

The name and the docstring ("Why this listing cannot be believed about `pgid`") say it inspects `listing`. Two of the three branches do; the third calls `_kernel_says_group_is_empty(pgid)`, which issues `killpg(pgid, 0)`. Two consequences worth naming:

1. `read_group_members` now issues that syscall on a path that did not exist before, for every read whose listing shows no member of the group.
2. `_interpret` lost the explicit kernel consultation at the point of decision. The old body read `if _kernel_says_group_is_empty(pgid): return GroupLiveness(False)` where a reader could see the measurement; now the same decision is a `return ""` from a helper named for the listing.

The sharing is right - the B1 fix depended on it - but the third row is a *finding about the world*, not a property of the listing. Either the name should say so (`_refuses_pgid`, or split the kernel row into an explicit call in each read) or the docstring should open with the fact that consulting it can cost a syscall.

kstrl/procgroup.py
```diff
@@ -471,61 +569,147 @@ class _Listing:
...
     #: pid 1 was present, so the view is not filtered to our own uid.
     complete: bool
-    #: Rows carrying this pgid, and how many of them are not zombies.
-    rows: int
-    running: int
+    #: Every non-blank row parsed. False means at least one row could
+    #: not be attributed to a group, so this is not a full view of any
+    #: group. No default: a default is how a later constructor comes to
+    #: claim readability it never established.
+    readable: bool
```
0xfauzi: **`kstrl/procgroup.py` lands at exactly 800 lines and the PR body records that prose was compressed to reach it.**

Verified against the hook rather than the body: `scripts/precommit/file_length_ratchet.py` has `MAX_LINES = 800` and `classify` returns clean on `now <= MAX_LINES`, so this passes at 800 and fails at 801. Handoff 2 is accurate.

The cost is the part worth reconsidering before merge. The PR body says "The refusal table plus the readability field is about 35 lines of code and the rest was prose; the prose was compressed to fit" - documentation traded for a line budget, in the one module whose stated value is the reasoning it records, and the file this PR exists because its prose was wrong.

The split is a mechanical one (the module docstring's `ps` trust argument reads as `docs/`, or the listing parse splits from the signal helpers) and landing it here costs less than handing the next author a module that fails the hook on its first added line.

PR #362, opened 2026-09-07
feat(context): the memory file is loaded last, through the golden-patterns loader (#230)

### LINKED ISSUE #230: R10.9: Memory file: standing feedback loaded last, after the retry context
Part of #235 (R10: Control Loop tracking issue). Item R10.9.

Design section: 5.7. Sizing: S (one PR). Dependencies: **R10.8** (`kstrl/operator_context.py`, the shared loader).

## Why

A memory file: a version-controlled markdown file loaded deterministically into the actuator's context **after the controller**, on every run. What belongs in it: permanent scope exclusions, known false-positive areas, and reviewer feedback that should change future work. What does not: one-off instructions and single-run logs. It is the human-on-the-loop steering channel: edits change future behaviour, not one PR.

kstrl has three weak versions already: the `## Agent Learnings` section of `CLAUDE.md` (appended by `ks evolve --apply`), per-directory `AGENTS.md` files (the engineer prompt tells the agent to update them), and `codebase_map.md`. What is missing is the **loading rule**. None of them is guaranteed to load after the controller's output. `CLAUDE.md` is prepended **before** the templated prompt but **after** `context_prefix` (`kstrl/loop.py:529-541`), so standing feedback sits between the retry context and the instructions rather than framing the retry context. The ordering is the mechanism.

This file is distinct from golden patterns (R10.8): golden patterns say what good code looks like and are static; memory carries standing corrections and grows as reviewers steer (R10.10 writes to it).

### 1. The file

Path `scripts/kstrl/memory.md`, configurable as `[paths] memory` in `KstrlConfig` (env `KSTRL_MEMORY_FILE`), loaded like the other `[paths]` keys (`kstrl/config.py:424-436`).

`ks init` scaffolds it when absent with this template:

```markdown
# Memory

Standing feedback for future kstrl runs in this repository. Loaded into the
engineer prompt on every run, after the retry context, so an entry here
changes future behaviour rather than one PR. Keep durable guidance only:
permanent scope exclusions, known false-positive areas, reviewer feedback
that should change future work. Not one-off instructions, not run logs.
Budget: about 1000 tokens; the loader truncates past that and warns.

## Guidance

- <one durable rule per line>
```

### 2. Loading position

`MEMORY_MAX_CHARS = 4000`. Use `load_operator_file(OperatorFile(path, header="MEMORY (standing feedback)", max_chars=MEMORY_MAX_CHARS))` from R10.8.

Insert the block as the **last** element of the `parts` list in `kstrl/factory.py:1439-1450`, after the retry context:

```
knowledge_prefix, golden_patterns, feedforward_prefix, retry_context, memory
```

Rationale, to be written as a comment at the insertion point: the retry context is the controller's output for this attempt; memory is the operator's standing correction to how that output should be acted on, so it is read after it.

When the loader truncates, also emit a UI warning at run start naming the file, so the operator prunes it: `memory.md exceeds <max> characters; <n> shown; prune it`.

Read from the worktree first, root second, exactly as R10.8 does for golden patterns.

### 3. Docs

- README hand-written area: one paragraph "Steering the loop" describing memory.md, what belongs in it, and that R10.10 appends reviewer comments to it.
- `docs/runbook.md`: a short "Standing feedback" subsection: when a reviewer keeps correcting the same thing, put it in memory.md, not in a PR comment.
- Regenerate README. CHANGELOG `### Added`.

## Out of scope

- Writing to the file automatically (R10.10).
- Merging or migrating the `## Agent Learnings` section of `CLAUDE.md` into it. Leave `ks evolve --apply` alone; the R9 design owns that path.

## Tests to write

`tests/test_operator_context.py` (extend):

1. `test_memory_block_loads_with_memory_header`.
2. `test_memory_truncation_warns`: a 9000-character file produces a warning naming the path (capture via the UI stub or `caplog`).

`tests/test_pipeline.py` or the spine test that captures the engineer prompt:

3. `test_memory_is_last_in_prefix`: with knowledge, golden, feedforward, a retry context, and memory all present, `prompt.rindex("=== END PREVIOUS CONTEXT ===") < prompt.index("=== MEMORY (standing feedback) ===")`, and the memory end delimiter precedes `# Project Context (from CLAUDE.md)` (the CLAUDE.md prepend header at `kstrl/loop.py:533`).
4. `test_memory_absent_leaves_prefix_identical`.
5. `test_ks_init_scaffolds_memory`: created on first init; an edited file is not overwritten on a second init.
6. `test_paths_memory_config_env_beats_toml`.

## Acceptance (all must hold)

- [ ] With the file present, the memory block is the last part of `context_prefix` and precedes the CLAUDE.md prepend (test 3).
- [ ] With the file absent, prompts are byte-identical to before (test 4).
- [ ] Truncation is announced in the prompt and warned to the operator (tests 2 and the loader tests from R10.8).
- [ ] Capture one real engineer prompt (stub agent echoing stdin) with all five parts present and paste the last 15 lines of the prefix in the PR body, showing the memory block after the retry context.
- [ ] Tests 1 through 6 pass; baseline suite passes; gen_docs check passes.

0xfauzi:
## Summary

`scripts/kstrl/memory.md` is standing operator feedback read into the engineer's
prompt on every factory iteration and every `ks run`, AFTER the retry context.
The position is the feature: the retry context is what the loop measured about
this attempt, and memory is the operator's standing correction to how that gets
acted on, so it is read after it rather than framed by it.

It is not a second loader. `kstrl/operator_context.py` gains one
`OperatorFileKind` row and every function in it now loops the table, so
everything the two #229 review rounds forced on golden patterns holds for this
file because it is the same code path.

The order was measured, not read off the code: a real worktree, the real
`run_loop`, and a real agent subprocess whose whole job is to write its stdin to
a file, with every block present at once. No LLM was called.

Before (`543e1aa`, the merge-base with #229 in it):

```
=== KNOWLEDGE (test) ===
=== GOLDEN PATTERNS (operator-authored) KSTRL-DATA-<32 hex> ===
## Architect Decisions
=== CODEBASE CONTEXT (auto-generated) ===
=== PREVIOUS ATTEMPT CONTEXT (Attempt 2) ===
=== END PREVIOUS CONTEXT ===
# Project Context (from CLAUDE.md)
Read the PRD at ...                      <- the templated engineer prompt
```

After, the same fixture:

```
=== PREVIOUS ATTEMPT CONTEXT (Attempt 2) ===
...
=== END PREVIOUS CONTEXT ===

=== MEMORY (standing feedback) KSTRL-DATA-975b46362b55c9c4ecd1970c8df8ef80 ===
# Memory

## Guidance

- never touch migrations
=== END MEMORY (standing feedback) KSTRL-DATA-975b46362b55c9c4ecd1970c8df8ef80 ===

# Project Context (from CLAUDE.md)
```

Full captures: the acceptance criterion's stub-agent prompt is what
`tests/test_spine_engineer_loop.py::TestMemoryIsReadAfterTheRetryContext::test_memory_is_last_and_precedes_the_claude_md_prepend`
asserts, and the second half is a real `ks init` plus a real `ks run` through the
shipped CLI, asserted by
`tests/test_spine_golden_patterns_e2e.py::TestTheMemoryScaffoldGoesNowhereUntilItIsEdited`.

## Round 1 fixes

Review round 1 returned 0 blocker, 3 should-fix and 10 nits. All three
should-fixes are fixed, all ten nits are closed or recorded with the reason,
and six new commits carry them (`2df2941`, `fbaa333`, `895dd30`, `7fdedcc`,
`0046f8c`, `a0c9e04`), with `origin/main` 037f0e1 merged in on top.

| Finding | What changed | Where |
|---|---|---|
| **SHOULD-FIX 1** the scaffold preamble is prompt text and one sentence of it is false where it is read | `DEFAULT_MEMORY` 354 to 148 characters and `DEFAULT_GOLDEN_PATTERNS` 715 to 313: a title and one declarative sentence each. "Nothing is injected while this file is unchanged", "Keep `## Guidance` last", "Keep it short", "Replace the placeholders" and "Read into the engineer's prompt on every factory component iteration" are gone. Both bodies APPEND a `SCAFFOLDED_TEMPLATES` row (H3b); no existing row moved. The lifecycle text was already in `docs/runbook.md` and the README | `kstrl/init_cmd.py` |
| **SHOULD-FIX 2** the cut keeps the head of the one file that grows at the tail | `keep: Literal["head", "tail"]` on `OperatorFileKind`, copied onto `OperatorFile` by `operator_file_spec`. `_cut` implements both ends under the same `CUT_FLOOR` rule. One `_KEPT` lookup names the direction in both the prompt's `fact` and the terminal's `message` | `kstrl/operator_context.py` |
| **SHOULD-FIX 3** two docstrings say a new row reaches the worker's block and it does not | Both corrected to name only the parent notice and `KstrlConfig.validate`, and to say the worker's block is hand-ordered at SIX edits per row, the sixth being the entry in the `parts` tuple that IS the prompt order. The round-1 wording listed every edit but that one, which is round 2's blocker; see the Round 2 fixes section. Plus the guard that refuses an unpaid row: a four-layer closed-by-construction count, with layer 2b reading the prompt-order literal itself | `kstrl/operator_context.py`, `tests/test_operator_files_reach_the_prompt.py` |
| NIT 2 the runbook counts two commands and there are three | "`ks factory`, `ks retry` and `ks run` read it" | `docs/runbook.md` |
| NIT 3 one appended newline still injects the empty skeleton | `shipped_label` tries the raw digest FIRST and the newline-collapsed text second, so the widening can only ADD matches. Both callers move together | `kstrl/init_cmd.py` |
| NIT 4 `KSTRL_MEMORY_FILE=""` resolves to the repo root | Behaviour unchanged and inherited: measured, all six path rows of `STRING_KEYS` do the same. Characterised in a test parametrized over every path row so the next person to change it sees all six | `tests/test_config_toml.py` |
| NIT 5 "Not changed: `_report_operator_files`" is not quite true | Corrected below: its BODY did not change; its docstring gained four lines | this body |
| NIT 6 the `ruff format` count is one low | Corrected below. It is 379 on the merged head (375 at `cbc0f04`, plus this PR's new guard file and #360's two); the three exceptions are the three `scripts/precommit/` files the review named | this body |
| NIT 7 the distinguishability guard hand-lists five of six fields | Derived from `dataclasses.fields(OperatorFileKind)`, asserts checked plus exempt equals the field set, exempts only `keep` with the reason | `tests/test_operator_file_kinds.py` |
| NIT 8 `spec_for`'s docstring says "the way production builds one" and its default is not | The docstring was the defect. It now says the scaffold is opt-in and why the default stays | `tests/helpers/operatorfiles.py` |
| NIT 9 "`## Guidance` stays last" is pinned against the constant, never the operator's file | A real `append_records` against a real `ks init` scaffold, plus the failure mode recorded as what happens: with a section after `## Guidance` the append lands there and no gate goes red. The runbook says so too | `tests/test_init_cmd.py`, `docs/runbook.md` |
| NIT 10 five documents say `ks feature` and `ks understand` read neither file, and nothing holds it | Layer 3 of the new guard: a census of the `context_prefix` name across `kstrl/`, pinned at `{factory.py: 4, loop.py: 3}` | `tests/test_operator_files_reach_the_prompt.py` |
| NIT 11 the prompt is built once per ATTEMPT | Agreed, and not a doc correction. The number is now a test: two iterations, an agent that edits the root memory file on its first, both captured prompts byte-identical. Seven "iteration" wordings moved to "attempt" | `tests/test_spine_engineer_loop.py` |

One further fix nobody asked for, found by running the new notice rather than
reading it: with the direction written between the count and the filename the
prompt's fact read `... dropping the start from memory.md`, where the closing
phrase attaches to the direction. The direction now sits last in both notices
(`a0c9e04`).

### Round 1: red first

Each new case was run at the base commit `cbc0f04` with the head's test files
copied in and the head's production code absent. Thirteen are red there.

| Finding | Case | At `cbc0f04` |
|---|---|---|
| SHOULD-FIX 1 | `TestTheScaffoldPreambleTheEngineerReads` (2 cases) | **RED**, on the six sentences taken verbatim from the old bodies |
| SHOULD-FIX 2 | `test_the_end_the_row_declares_is_the_end_that_survives` (2), `test_both_audiences_are_told_which_end_went` (2), `TestTheNewestStandingCorrectionSurvives` (2), `test_every_row_declares_a_direction_the_cut_implements`, `test_the_memory_row_is_the_one_the_issue_asked_for` | **RED**, 8 cases |
| NIT 3 | `test_appended_newlines_are_not_an_edit[one]` and `[three]` | **RED**. The case is the round-2 assertion INVERTED, which is the red-first evidence in reverse |
| NIT 7 | `test_every_row_is_distinguishable_from_every_other` | **RED**, because the derived field set has no `keep` at the base |
| SHOULD-FIX 3 | `tests/test_operator_files_reach_the_prompt.py` (7 cases) | **COULD NOT MAKE RED.** Both declared rows are already wired in `factory.py` at the base, so the guard is green there. What it closes is a FUTURE third row, and its evidence is mutations G1, G2 and G3 below |
| NIT 4 | `test_an_empty_path_env_var_resolves_to_the_repo_root` (6) | **COULD NOT MAKE RED.** The behaviour is inherited and unchanged; the case characterises it |
| NIT 9 | `TestTheAppendLandsUnderGuidance` (2) | **COULD NOT MAKE RED.** `## Guidance` was already last at the base. The case is a missing pin, not a fix; mutation N9 is its evidence |
| NIT 11 | `TestThePromptIsBuiltOncePerAttempt` | **COULD NOT MAKE RED.** The seam is unchanged. The case is a missing pin; mutation N11 is its evidence |

Base run: 13 failed, 199 passed. Head: the same selection is green.

### Round 1: mutations

30 plants at `a0c9e04`, the head before `origin/main` 037f0e1 was merged in, in a
detached worktree, each deleting `__pycache__` first, `PYTHONDONTWRITEBYTECODE=1`, `-p no:cacheprovider`, its own
process group killed by `os.killpg` on a 420-second bound, anchor asserted
present exactly once before the plant so a silent no-op is COULD-NOT-PLANT and
never a pass.

**30 caught, 0 still green, 0 hung, 0 could-not-plant.**

| id | what it plants | outcome |
|---|---|---|
| G1 | a third `OperatorFileKind` row declared and given no `load_operator_file` site in `factory.py` | **CAUGHT** |
| G2 | the memory injection site deleted from `factory.py` | **CAUGHT** |
| G3 | the walk switched off, so it sees no calls at all | **CAUGHT** |
| D2a | the memory row's direction flipped back to `head` | **CAUGHT** |
| D2b | `_cut`'s tail branch takes the head window anyway | **CAUGHT** |
| D2c | the tail direction string names the wrong end to both audiences | **CAUGHT** |
| D2d | the prompt is told one end and the terminal the other | **CAUGHT** |
| D1a | the memory body's new ledger row removed | **CAUGHT** |
| D1b | the golden body's new ledger row removed | **CAUGHT** |
| D1c | the OLD memory row dropped instead of kept (H3b's actual prohibition) | **CAUGHT** |
| N3a | the digest narrows back to the raw text | **CAUGHT** |
| N3b | normalise-only, so the widening stops being additive | **CAUGHT**, by the case that supplies the row the table cannot hold |
| N9 | a section added after `## Guidance` in the shipped scaffold | **CAUGHT** |
| N11 | the prompt rebuilt per iteration rather than per attempt | **CAUGHT** |
| C3b | the `STRING_KEYS` re-export becomes a REAL copy | **CAUGHT** |
| R1-R9 | the reviewer's nine row-table plants, re-planted against the changed module | **CAUGHT**, 9 of 9 |
| F1-F3 | the reviewer's three loader plants (cut floor, fact/message swap, root resolution) | **CAUGHT**, 3 of 3 |
| S1-S3 | the reviewer's three ledger plants (digest mismatch, `ks init` stops scaffolding, suppression removed) | **CAUGHT**, 3 of 3 |

The merge changes no file any plant targets and none any plant's test selection
covers (`git diff --name-only a0c9e04 HEAD` against the six targeted files: no
overlap), so the verdicts are reported at the revision they were measured at
rather than re-measured on the merged tree.

G3 is the answer to CLAUDE.md's rule that `assert hits == []` is not a control:
switching the walk off makes the guard fail rather than agree. N3b is the one
that was STILL GREEN on the first pass, and it was green for a reason worth
recording rather than papering over: every body in `SCAFFOLDED_TEMPLATES` ends
in exactly one newline, so normalise-only is an equivalent mutant against the
real table. `0046f8c` adds the row the table cannot hold, a synthetic template
whose history carries the digest of a two-newline body, and the plant is caught.

### Round 1: truncation, before and after

Measured through the real loader on the head, the shipped `DEFAULT_MEMORY` plus
400 appended rules at the 4000-character budget, the two directions run over the
same fixture so the comparison is the direction and nothing else.

```
keep=head (what shipped for review):
    - rule 0000 .. - rule 0320    321 of 400 rules, 3999 characters
    fact: truncated: 3999 of 4948 characters shown from memory.md,
          keeping the start of the file and dropping the end

keep=tail (this fix):
    - rule 0067 .. - rule 0399    333 of 400 rules, 3995 characters
    fact: truncated: 3995 of 4948 characters shown from memory.md,
          keeping the end of the file and dropping the start
```

79 standing corrections were dropped and they were the 79 newest. They are now
the 67 oldest instead. The review measured the same direction on its own fixture
and lost rules 0303 to 0399; the fixtures differ because the scaffold shrank in
the same PR, and the direction and the conclusion are identical.

Golden patterns is the control and is unchanged: 600 appended rules at its own
6000-character budget deliver `- rule 0000` through `- rule 0472` and drop the
rest, `keeping the start of the file and dropping the end`.

## Round 2 fixes

Review round 2 returned 1 blocker, 2 should-fix and 8 nits, and confirmed the
mechanism: 30 re-planted mutations all caught, H3b clean by hand, the row
table, the truncation direction, the resolution, the ledger and the config move
all holding. Everything it found is fixed, in six commits on top of a merge of
`origin/main` `aad32e8` (#359 and #360).

| Finding | What changed | Where |
|---|---|---|
| **BLOCKER 1** the round-1 guard clears a row that is loaded and reaches no prompt, and the remedy it prints is short by one edit | Layer 2b reads the literal `_run_component` builds `parts` from, maps each loader call in that scope back to the local it binds, and asserts one element per declared row. The reviewer's plant is now caught: `memory,` removed from the tuple with the loader call left and `memory_chars = len(memory)` keeping ruff quiet. The loader-site count stays beside it, because a row can skip either edit. The remedy is a list, `WORKER_EDITS`, with the count derived from it everywhere it is printed, and a case fails if either docstring stops naming an edit the guard counts | `tests/test_operator_files_reach_the_prompt.py`, `kstrl/operator_context.py` |
| **SHOULD-FIX 1** the tail arm's `CUT_FLOOR` fallback is dead to the whole suite | A floor case parametrized over `OPERATOR_FILES` that builds its unwrapped paragraph at whichever end the row declares, so neither arm can lose its fallback quietly. The reviewer's mutant, which left the full suite green at 6213 passed, is caught | `tests/test_operator_file_kinds.py` |
| **SHOULD-FIX 2** the widened digest reaches `ks init --upgrade-prompts`, whose stated argument is byte identity | One table, two rules, each written at its call site. `shipped_label`'s default is the RAW digest and `_classify` takes it, so a scaffold that gained a trailing newline is unrecognised, reported and left alone rather than overwritten; the loader passes `ignore_trailing_newlines=True` because its decision is whether to inject. Both the end-to-end `--upgrade-prompts` case and a case pinning both answers about ONE file | `kstrl/init_cmd.py`, `kstrl/operator_context.py`, `tests/test_prompt_upgrade.py`, `tests/test_operator_context.py` |
| NIT 1 `read_operator_file` says the digest is taken on the rendered text | It is taken on the raw text and again on the text with its trailing newlines collapsed; the CRLF case is `read_text`'s universal newlines, before any digest. The docstring says that, and says which rule this reader asks for | `kstrl/operator_context.py` |
| NIT 2 `spec_for`'s docstring counts are stale | The rule stays, the arithmetic goes. Round 1 wrote eleven and three, round 2 measured fifteen and four on a set this PR grew | `tests/helpers/operatorfiles.py` |
| NIT 3 "all four behave identically" over six parametrized rows | The parametrization is the set and the numeral is gone, which is the treatment this branch already applied to `init_wizard`'s "the three" | `tests/test_config_toml.py` |
| NIT 4 the D5 paragraph carries pre-fix counts | Re-measured on this head and corrected below: 14 cases over `OPERATOR_FILES` (28 collected, 14 per kind), 5 tie tests, and 6 of the 7 `dataclasses.fields` attributes checked with `keep` exempted | this body |
| NIT 5 the tail cut discards a complete line that fitted | Both arms fixed, not one. The tail window that already starts at a line boundary is kept whole, and so is the head window that already ends on one: `rfind` had the same defect mirrored, giving up the last complete line rather than one newline. Each has its own case | `kstrl/operator_context.py`, `tests/test_operator_file_kinds.py` |
| NIT 6 `_KEPT[spec.keep]` raises where `_cut` treats the same value as head | One vocabulary, checked where the row is built. `_KEPT` moves above the rows and `__post_init__` on both `OperatorFileKind` and `OperatorFile` refuses a `keep` outside it, so `dataclasses.replace(spec, keep="middle")` fails at construction instead of truncating one way and dying about the other | `kstrl/operator_context.py`, `tests/test_operator_file_kinds.py` |
| NIT 7 `rendered[-0:]` with a zero budget is the whole file | The same `__post_init__` refuses a non-positive `max_chars`, rather than `max(1, ...)` inside the cut making a zero silently mean one. A case pins that 1 is still accepted and still truncates | `kstrl/operator_context.py`, `tests/test_operator_file_kinds.py` |
| NIT 8 the runbook's "steers every run" is corrected four sentences later | The sentence an operator acts on now carries the qualification in place and points at the paragraph that lists the commands | `docs/runbook.md` |

Two things round 2 did not ask for, both found by running rather than reading:

- **The head arm has nit 5's defect too.** The review measured the head arm
  giving up one character on an aligned window, which is the newline and is not
  content. The mirror case is a window that ends just before a newline, where
  `rfind` moves back past a complete line: measured at 94 of 100 characters
  with that line gone. It is the same defect class in the same function, so it
  is fixed in the same commit and has its own case and its own mutation (B7).
- **Making the raw digest the default made an existing case stop asking about
  the branch it is written for.** Round 3's own mutation N3b2 (normalise-only
  in the widened branch) came back STILL GREEN, because
  `test_a_historical_body_that_ended_in_two_newlines_is_still_found` asked the
  default, which the plant does not touch. It asks both now, and the plant is
  caught. That is the only still-green result in this round and the fix for it
  is `902b71e`.

### Round 2: red first

Run at the merge commit `8ea99f6` (this branch's head plus `origin/main`
`aad32e8`, before any round-2 fix) with the round-2 test files copied in and no
round-2 production code. **11 red, 183 passed.** Output:
`lanes/230/r3/red_first_base.txt`.

| Finding | Cases | At `8ea99f6` |
|---|---|---|
| BLOCKER 1, the record half | the two docstring cases | RED (2) |
| SHOULD-FIX 2 | the `--upgrade-prompts` case and the one-file-two-answers case | RED (2) |
| NIT 5 | the tail case and the head mirror | RED (2) |
| NITS 6 and 7 | the four refusal cases and the spec's | RED (5) |
| BLOCKER 1, the guard half | `test_every_declared_row_reaches_the_prompt_order` | COULD NOT MAKE RED. Both rows are delivered at the base; the evidence is mutations B1, B2, B5 and B10 |
| SHOULD-FIX 1 | the parametrized floor case | COULD NOT MAKE RED. The fallback is in the base code; the case is what was missing, and mutation B3 is its evidence |

### Round 2: mutations

46 plants at the head, in a detached worktree, harness
`lanes/230/r3_mutate.py`, which imports round 2's plant list rather than
copying it. One plant at a time, anchor asserted present exactly once or the
plant is COULD-NOT-PLANT and never a pass, `__pycache__` deleted before every
run, `PYTHONDONTWRITEBYTECODE=1`, `-p no:cacheprovider`, `start_new_session=True`
with `os.killpg` on a 420-second bound so an expiry is HUNG rather than green,
original restored after each and `git status --porcelain` clean afterwards.

**43 CAUGHT, 0 still green, 0 hung, 3 COULD-NOT-PLANT retargeted and then
caught.** Results `lanes/230/r3/mutations.json`, log `mutations.log`.

| Plant | What | Outcome |
|---|---|---|
| B1 | the reviewer's second plant: `memory,` out of the prompt-order tuple, loader call kept, one ordinary read of the local | CAUGHT, by the new layer 2b and by seven spine cases |
| B2 | the reviewer's first plant: the same omission with the local unused | CAUGHT |
| B3 | the tail arm's `CUT_FLOOR` fallback deleted | CAUGHT |
| B4 | `_classify` back on the widened rule, so `--upgrade-prompts` overwrites | CAUGHT |
| B5 | the delivery census switched off so it sees no elements at all | CAUGHT, red rather than an empty pass |
| B6, B7 | each line-boundary probe removed, tail and head | CAUGHT |
| B8, B9 | each construction-time refusal switched off | CAUGHT |
| B10 | the delivery walk's flag direction turned into a clear | CAUGHT |
| B11, B12 | the remedy stops naming the `parts` entry; the count goes stale by one | CAUGHT |
| N3a2, N3b2, N3c | round 2's N3a and N3b retargeted at the rewritten lookup, plus the loader's keyword removed | CAUGHT (N3b2 after `902b71e`; see above) |
| S3b | round 2's S3 retargeted at the rewritten suppression branch | CAUGHT |
| the round-2 set | G1-G3, D2a-D2d, D1a-D1c, N9, N11, C3b, R1-R9, F1-F3, S1, S2 | all 27 CAUGHT, re-planted at this head |

N3a, N3b and S3 reported COULD-NOT-PLANT because this round rewrote the exact
lines their anchors named. Each was retargeted at the same behaviour in its new
shape and then run; none was scored as a pass.

### Round 2: censuses across two moving mains

`origin/main` moved twice during this round, so the census was executed four
times, never read off a pin. Predicate: `grep -rl -E 'ast\.parse|astwalk'
tests/*.py`, the same one on every tree.

**Merge 1, `aad32e8` (#359, #360).** 38 to 39 guard files
(`tests/test_config_guard.py`), 1211 to 1228 passed, 34 inventories on both
sides, and NO inventory's contents move. Every delta is a corpus size:
`_package_sources` 134 to 135 for `kstrl/runenvelope.py`, and the two test
corpora by the five and six files #359 adds.

**`aad32e8` against this branch before merge 2.** 38 to 39 guard files (this
PR's `tests/test_operator_files_reach_the_prompt.py`), 1221 to 1234 passed, 32
to 34 inventories. `_package_sources` 134 to 135 for `kstrl/config_keys.py`; the
two test corpora by this PR's three and four new test files; the two new
inventories are this guard's layer 1 `{factory.py: 3, operator_context.py: 1}`
and layer 3 `{factory.py: 4, loop.py: 3}`; and
`test_prompt_staleness::test_every_filename_a_template_is_written_under_is_pinned`
moves by exactly the five counts its own comment predicts (`config.py` 2 to 3,
`factory.py` 1 to 2, `init_cmd.py` 8 to 10, `init_wizard.py` 4 to 5,
`operator_context.py` 1 to 2). Nothing else moves, and no CLEARING guard's
inventory moves. Layer 2b adds no inventory, because it is a resolving walk
rather than an `assert_census`.

**Merge 2, `0770d49` (#357).** 39 to 41 guard files
(`tests/test_check_result_measurement.py` and
`tests/test_sense_dampener_workflow.py`), 1234 to 1260 passed, 34 to 39
inventories. Every delta belongs to #357: `_package_sources` 135 to 138 for
`kstrl/dampener.py`, `kstrl/dampener_report.py` and
`kstrl/fixtures_snapshot.py`, with the matching corpus move on 27 walks and six
new walks at 138; the test corpora 241 to 250 and 376 to 388; five new
inventories, all of them `test_check_result_measurement.py`'s;
`test_encoding_walk`'s walk renamed from "the five measured" to "the six
measured" with `dampener.py` in it; and `test_encoding_readers` moving one
`fixtures.py` spelling into `fixtures_snapshot.py`. No inventory this branch
owns moves.

Mutations were measured on the pre-merge-2 tree and the four that carry this
round's decisions were re-planted on the merged tree and are still CAUGHT (B1,
B3, B4, B5, plus B7 and N3b2). #357 touches no file any plant targets and none
in any plant's test selection.

## Behaviour changes

Every one of them, including the two that deviate from the issue text.

1. **A new operator file, `scripts/kstrl/memory.md`.** Read from the repo ROOT,
   never from a component worktree. Budget 4000 characters, about 1000 tokens on
   the four-characters-per-token convention `operator_context` already
   documents. Its own budget, not a share of the golden-patterns 6000: golden
   patterns is pruned by hand while #231 makes the daemon a writer of this one,
   and a shared budget would let the file that grows by machine starve the file
   that does not, silently.
2. **The block is last in `context_prefix`,** and `run_loop` prepends the whole
   prefix to `CLAUDE.md` plus the templated prompt, so the block also lands
   before `# Project Context (from CLAUDE.md)`. The order is now ONE literal
   tuple in `_run_component`: the retry context moved from a trailing `if` into
   `_retry_block(previous_context_json)`, which returns `""` for nothing, so
   adding memory REMOVED a branch instead of adding one (`_run_component`'s ruff
   C901 number went 35 to 33, measured with pinned `ruff@0.16.4`).
3. **`ks init` scaffolds it,** and an unedited scaffold injects nothing:
   `DEFAULT_MEMORY` is enrolled in `SCAFFOLDED_TEMPLATES` with an APPENDED row,
   so `load_operator_file` recognises kstrl's own body and treats it as empty.
   A second `ks init` never overwrites an edited file. `## Guidance` is the LAST
   section of the scaffold on purpose, and the comment says so: #231 appends a
   `/memory` comment to the end of the file.
4. **New config key `[paths] memory`, env `KSTRL_MEMORY_FILE`,** field
   `KstrlConfig.memory_file`. It follows the existing precedence (env beats
   TOML beats default) with no new code, because `_apply_toml_overrides`,
   `_apply_env_overrides`, `anchored` and `validate` are all driven by the
   table. `ks config`, the generated README block, `docs/env-vars.md` and
   `kstrl.toml.example` all carry it.
5. **`STRING_KEYS` moved to a new `kstrl/config_keys.py`** and is re-exported
   from `kstrl.config` as the SAME object. This was forced, not chosen:
   `kstrl/config.py` was at 799 lines against the 800-line pre-commit ratchet,
   and the field plus the row took it to 801, `CROSSED`, rc=1. It is now 781
   with the feature in, 19 lines of headroom. `config_report` and
   `scripts/gen_docs.py` import from `config_keys`; the three loaders read
   `kstrl.config.STRING_KEYS`.
   `tests/test_string_keys_reach_every_surface.py::test_the_re_export_is_the_same_object`
   pins the two names to one object by IDENTITY, so they cannot become two
   tables that happen to agree.
6. **`ks feature` and `ks understand` read neither operator file,** and every
   shipped surface says so in those words. Verified rather than relayed: the
   three `run_loop` calls in `kstrl/feature_cmd.py` (301, 464, 589) and the one
   in `kstrl/cli.py:1551` pass no `context_prefix`, and `ks run` reaches the
   worker's prefix assembly because it is a `run_factory` invocation
   (`kstrl/cli.py:1182`).
7. **DEVIATION from the issue: root only, not "worktree first, root second".**
   The issue says to read the worktree first. #229's round-1 finding S3 is why
   this does not: the worktree is the tree the engineer has just been writing
   to, so a worktree read lets one component choose what the next component is
   told, unfiltered, under a header saying the operator wrote it. #231 raises
   the stake further, because the root copy becomes a file the daemon writes on
   an authorised reviewer's say-so. The deviation is deliberate and is the safer
   direction.
   `test_a_worktree_memory_copy_is_not_read` and
   `test_the_worktree_memory_copy_loses_to_the_root_copy` pin it.
8. **DEVIATION from the issue: the shared truncation wording.** The issue asks
   for `memory.md exceeds <max> characters; <n> shown; prune it`. The loader
   already has ONE wording built from one `shown` string for two audiences (the
   `fact` inside the prompt, the `message` on the terminal), which is what
   #229's round 2 forced so the two cannot disagree about the numbers. A
   memory-specific sentence would be a second wording of one measurement, so
   the shipped string differs from the issue's. The acceptance criterion
   ("truncation is announced in the prompt and warned to the operator") is met
   by the shared path: the prompt carries
   `[truncated: <n> of <total> characters shown ...]` and the terminal carries
   `shorten <absolute path>` once per run.

Its BODY did not change: `_report_operator_files` in `kstrl/factory.py`
gained four docstring lines in this diff and no code (round 1, nit 5). That is
the point. Its round-2 docstring claimed it would not have to change when
R10.9's row landed, and the row has landed and no caller learned the new
file's name.
`tests/test_operator_file_warnings.py::TestTwoFilesAreTwoSubjects` runs both
files misconfigured at once and checks the two lines carry different subjects,
different paths and different `[paths]` keys, which nothing exercised before.

## Design decisions

The five questions the plan left open, decided with the reason each way.

**D1. One row table, not a second loader.** The five loose `GOLDEN_PATTERNS_*`
constants collapse into a frozen `OperatorFileKind` row carrying key, field,
header, subject, max_chars and scaffold; `golden_patterns_spec` becomes
`operator_file_spec(kind, root, configured)`; `_rows` loops `OPERATOR_FILES` and
reads each path off the row's own `field` with `getattr`. The alternative was a
parallel path for the second file, which is how the parent and the worker came
to read different files in #229's round 2.

**D2. The order is one literal tuple, not a second `if`.** A second `if` after
the retry-context `if` would make "memory is last" a property of statement
sequence, true by accident. Hoisting `_retry_block` makes the whole order a
value one test can pin and removes a branch from an already-large function.

**D3. A separate 4000-character budget** (D-item 1 above).

**D4. `ks feature` and `ks understand` keep reading neither file**, following
#229, with the four `run_loop` call sites carried below as a measured handoff
rather than as a new issue.

**D5. Parametrized subset plus tie tests, not full parametrization.**
`tests/test_operator_file_kinds.py` runs 14 cases over `OPERATOR_FILES` reading
the header, subject, budget and scaffold off the row under test, 28 collected
and 14 per kind, plus 5 tie tests: every kind matches a `[paths]` row whose
field is a real `Path` field of `KstrlConfig`, every kind is enrolled in
`SCAFFOLDED_TEMPLATES`, every row is distinguishable from every other on the 6
of 7 `dataclasses.fields` attributes that are not `keep` (so the parametrized
cases are not vacuous), every row declares a direction that is a key of `_KEPT`,
and the memory row carries the numbers the issue asked for. All three counts
re-measured on this head after the round-2 fixes; they were 11, four and five
when the paragraph was written (round 2, nit 4). The #229 mutation re-plant below is the proof the migration did not blind
the guards it rewrote the fixture of.

**Q1 answered by measurement, not preference.** `config.py` could not take the
row. Compressing comment prose for two lines is what #229's round 2 had to do
and leaves the next `[paths]` row in the same place; `STRING_KEYS` is the part
of the file that grows, imports nothing so no cycle is possible, and carries no
template filename and no `read_text`, so neither static census moves because of
the refactor and every census delta in this PR is attributable to the feature.

### Every #229 review finding, closed for this file by construction

The claim is that the memory file inherits these rather than re-earning them.
That is a claim, so `tests/test_operator_file_kinds.py` checks it per row.

| #229 finding | How it is closed here |
|---|---|
| R1 BLOCKER 1: an unedited scaffold injected into every prompt | `MEMORY.scaffold = "memory.md"` plus an appended `SCAFFOLDED_TEMPLATES` row; `read_operator_file` already suppresses a body `shipped_label` recognises. Tested at three levels: the loader, `_run_component`, and a real `ks init` plus `ks run`. |
| R1 BLOCKER 2: the 34-slot positional `_submit_args` tuple | The 35th slot lands together with its entry in the whole-dict equality expectation, and `test_the_three_string_path_slots_are_distinguishable` became four. |
| R1 S3: a worktree read an agent can poison | `operator_file_spec` resolves against `root` only, for every kind. |
| R1 S4: a forgeable fixed delimiter | The delimiter carries a per-build token from `generate_data_delimiter()`. A body line spelling the memory END delimiter cannot close the block. |
| R1 nit 14: "byte identical" asserted as delimiter-absence | `test_memory_absent_leaves_the_prefix_byte_identical` compares the two prompts as strings with the temp path normalised out. |
| R2 BLOCKER 1: an unguarded `Path.exists()` stat | There is no stat on this path and none was added. The mode-000-parent and ENAMETOOLONG cases are parametrized over both kinds. |
| R2 S2: the parent and the worker resolving different files | One resolver, called by `_rows` in the parent and by `_run_component` in the worker, for both kinds. |
| R2 S3: a line cut delivering 17 characters of 6000 | `CUT_FLOOR` is on the shared cut and the floor cases are sized off `kind.max_chars`, never off the literal 6000. |
| R2 S4: an imperative and an absolute path inside the prompt | `fact` goes in the block, `message` goes to the terminal; asserted per kind. |
| R2 S5: `_report_operator_files` hardcoding a subject | This PR is the case that claim was made about. The moved-subject test now moves `OPERATOR_FILES`, and a new test runs both rows misconfigured at once. |
| #260 round 2 / #229 P5, one field over | `_rows` reads `getattr(config, kind.field)`; there is one expression and it names no field. Two plants (`Z2`, `Z3`) pointing the memory row at the golden field and the golden key are both caught. |

## Tested

Every number below was produced by running the thing on this tree. Bytecode is
deleted before every measurement and `PYTHONDONTWRITEBYTECODE=1` is set, because
two same-size edits inside one second reuse stale bytecode and report a false
pass.

| Gate | Result |
|---|---|
| `uv run pytest tests/ -q -p no:cacheprovider` | **6467 passed, 37 skipped, 45 xfailed** in 375.0s, rc=0 |
| `uv run mypy kstrl/ --strict` | Success: no issues found in 138 source files |
| `uv run ruff check kstrl/ tests/ scripts/` | All checks passed |
| `uv run ruff format --check kstrl/ tests/ scripts/` | 397 formatted; the 3 exceptions are `scripts/precommit/{check_agent_docs_sync,cyclomatic_ratchet,file_length_ratchet}.py`, which `.pre-commit-config.yaml` excludes on purpose so those files stay byte-identical across repos (round 1, nit 6: the review measured 375 at `cbc0f04`; this branch has added one test file and `origin/main` 037f0e1 has added two) |
| `uv run python scripts/gen_docs.py --check` | README.md generated sections are current |

Suite history on this branch, same command every time: 6147 / 37 / 42 before and
after the simplify commit (555.3s then 583.8s); 6176 / 37 / 42 after the round-1
fixes at `a0c9e04`; 6213 / 37 / 42 with `origin/main` 037f0e1 merged in, where
the +37 are #360's; and the number above with `origin/main` `aad32e8` merged in
and the round-2 fixes on top, where the +62 are #359's tests plus this round's
cases; and the number above with `origin/main` `0770d49` merged in as well,
where the +192 are #357's. The machine is shared with other lanes, so
wall time is not a measurement of this change. For comparison: base `543e1aa`
6078 passed; `origin/main` 1775301's own baseline as measured by the
orchestrator on 568bca4 was 5564 passed, 37 skipped, 38 xfailed.

Also run the way pre-commit runs it, so the numbers are the hook's and not a
reading of the file: `complexipy --staged --max-complexity-allowed 15` rc=0,
with `_run_component` cognitive complexity 51 to 48 IMPROVED and the only two
regressions 0 to 1 and 1 to 2. All 29 pre-commit hooks ran on the simplify
commit and passed.

### Mutations

33 plants on the branch head, run in a DETACHED worktree, every one deleting
`__pycache__` first, `PYTHONDONTWRITEBYTECODE=1`, `-p no:cacheprovider`, in its
own process group killed by `os.killpg` on a 420-second bound. Each plant
asserts its anchor is unique AND that the bytes changed before scoring, so a
silent no-op is COULD-NOT-PLANT and never a pass.

**30 caught, 3 still green, 0 hung, 0 could-not-plant.** The three still-green
are the three deliberate equivalence controls, which is what proves the harness
can report green at all:

- M11: `<= max_chars` written as `< max_chars + 1`.
- X15b: `tuple(t)` on a tuple IS `t`, the control for X15 (which plants a real
  COPY of the table and is caught).
- Z1: `_retry_block` dropping its whitespace guard. Measured rather than
  assumed: `IterationContext.from_json` on an empty context, on `{}` and on
  `{"iterations": [], "review_findings": []}` all return 177 characters
  beginning `=== PREVIOUS ATTEMPT CONTEXT (Attempt 1) ===`, so no JSON parsed
  through `IterationContext` reaches it. It is an equivalent mutant. The guard
  is kept rather than deleted because deleting it would be a behaviour change
  smuggled into a refactor, and
  `test_an_empty_iteration_context_still_renders_its_header` pins the
  measurement so the claim is checked rather than asserted.

The full set was run TWICE, before and after the simplify commit, and every one
of the 33 verdicts is the same in both runs. The one anchor the simplification
moved (`M7`, which pairs the memory row with the golden anchored default) was
retargeted onto the inlined `getattr` and is still caught; the whole set was
dry-run for anchor uniqueness first, so a retarget was reported rather than
scored as a pass.

**The twelve #229 loader-facing plants, re-planted against the migrated
guards, are all still caught**: P5 (root-only resolution), P8 (remedy in the
prompt), P9 (message instead of fact), P10 (absolute path in `display`), P11
(printer hardcodes the subject), P12 (row loses its subject), P13 (validate
reports a present file as missing), P14 (budget against the raw text), P15
(digest input normalised), X1 (scaffold suppression removed, round-1 BLOCKER
1), X6 (worker reads the worktree), X7 (fixed delimiter restored). Renaming a
guard set's fixture is how a guard goes blind, so this is the evidence the
migration did not blind them.

R10.9's own, all caught: M1 memory before the retry context, M2 memory dropped
from the tuple, M3 the golden and memory slots rotated in the 35-slot
positional tuple, M4 golden's header on the memory row, M5 golden's budget on
it, M6 the row dropped from `OPERATOR_FILES`, M7 above, M8 the ledger row
dropped, M9 the scaffold digest no longer consulted, M10 the worker reading the
worktree copy, M12 the `KEY_DESCRIPTIONS` row deleted, M13 the `[paths] memory`
row dropped from `STRING_KEYS`, M14 `ks init` no longer scaffolding
`memory.md`, Z2 the memory row pointing its `field` at the golden path, Z3 the
memory row claiming the golden `[paths]` key, X15 the re-export handing out a
copy, N8/N9 `config_report` and `gen_docs` hand-copying the paths rows.

### Static guard censuses, re-derived by running the walks

No pin was read. A pytest plugin wraps `assert_census` before any test module
imports it and recomputes each inventory from the same corpus and predicate the
guard uses, over the 33 test files that carry one (1074 passed, 9 skipped, 41
xfailed).

Base `543e1aa` to this branch, one guard moves, and it moves by exactly the
five counts the plan predicted:
`test_prompt_staleness.py::test_every_filename_a_template_is_written_under_is_pinned`
goes `config.py` 2 to 3 (the field default), `factory.py` 1 to 2 (the worker
kwarg default), `init_cmd.py` 8 to 10 (the ledger row and the
`_create_if_missing` call), `init_wizard.py` 4 to 5 (the scaffold preview),
`operator_context.py` 1 to 2 (the row's `scaffold` field). `cli.py` and
`launch.py` do not move: neither points an operator at this file. The package
corpus goes 132 to 133 (`config_keys.py`), which moves `_corpus_sizes` on every
guard and nothing else. **The encoding census does not move**, because the
refactor carries no `read_text` site out of `config.py`, which is why
`STRING_KEYS` was the block chosen to move: every census delta this PR shows is
attributable to the feature.

The review's H4 table was right that the earlier wording described a merge
that was not in this branch's history: at `cbc0f04`, `origin/main` 1775301 was
an ANCESTOR of the head. `origin/main` has since moved to 037f0e1 (#360) and
IS merged in, so there is now a real before-and-after to show, and it is below.

Across the simplify commit, every inventory is **byte-identical**, which is the
check that a code move in a test file did not change what a walk sees.

**Re-derived after the round-1 fixes and again after merging `origin/main`
037f0e1**, by EXECUTING every walk at each revision, never by diffing a pin.
The guard file set is 40 at `origin/main` and 41 here, and
`tests/test_operator_files_reach_the_prompt.py` is the difference. Guard suite:
`origin/main` 1216 passed, 9 skipped, 41 xfailed; head 1223 passed, 9 skipped,
41 xfailed, and the +7 are that file's seven cases.

34 inventories at the base, 36 at the head, and five move:

| Inventory | Delta | Why |
|---|---|---|
| `_package_sources` | 133 to 134 | `kstrl/config_keys.py` |
| `_corpus_sizes` | 133 to 134 on every `kstrl/` walk; `test_process_scoping` 233 to 236; `test_procgroup` 366 to 370 | the new package module, and the three new test files |
| `test_prompt_staleness.py::test_every_filename_a_template_is_written_under_is_pinned` | `config.py` 2 to 3, `factory.py` 1 to 2, `init_cmd.py` 8 to 10, `init_wizard.py` 4 to 5, `operator_context.py` 1 to 2 | the five counts above |
| `test_operator_files_reach_the_prompt.py::test_no_new_code_gets_hold_of_the_loader` | new: `{factory.py: 3, operator_context.py: 1}` | layer 1 of the new guard |
| `test_operator_files_reach_the_prompt.py::test_no_other_command_hands_run_loop_a_prefix` | new: `{factory.py: 4, loop.py: 3}` | layer 3 of the new guard |

Nothing else moves. No encoding, atomicio, TOML-reader, event-name, journal,
process-scoping, state-dir, TUI-config or prompt-enrolment inventory changes,
and #360's own `ps`-command inventory is identical at both ends of this diff.

Across the merge itself, head before to head after, every delta belongs to #360:
`_package_sources` 133 to 134 (`kstrl/procgroup_listing.py`), the corresponding
`_corpus_sizes` moves, and
`test_procgroup.py::test_nobody_names_the_ps_command_without_appearing_here`
gaining `procgroup_listing.py`, `tests/helpers/procs.py` and
`tests/test_procgroup_members.py` and losing `tests/test_shutdown.py`. This PR's
own two inventories are unchanged across the merge.
The checkpoint of this lane expected the `test_process_scoping` and
`test_procgroup` corpus counts to be RED and they are not: those numbers are
derived by the recording plugin, and the pins those two files actually carry
(`EXPECTED_MACHINE_WIDE_SPELLINGS`, `EXPECTED_PS_COMMAND_SPELLINGS`) key on
files that HAVE hits, which the new test files do not. Established by running
both guards, not by reading them.

### Ratchets

Run the way pre-commit runs them: a DETACHED worktree checked out at
`origin/main` with this branch's tree staged on top, so both ratchets (which
diff the INDEX against HEAD) see exactly this diff.

- `file_length_ratchet.py` rc=0. No `CROSSED` and no `ADDED` over the limit.
  `kstrl/config.py` is not reported at all, which is the whole reason
  `STRING_KEYS` moved: 799 to 781, 19 lines of headroom. Three files that were
  already over 800 grew and are reported rather than failed, which is the
  hook's documented behaviour. Against `origin/main` `0770d49` after the
  round-2 fixes: `kstrl/factory.py` 4632 to 4666, `kstrl/init_cmd.py` 1853 to
  1984, `scripts/gen_docs.py` 916 to 920.
- `cyclomatic_ratchet.py` rc=0, with pinned `ruff@0.16.4` C901:
  `IMPROVED kstrl/factory.py::_run_component cyclomatic 35 -> 33`. Two rises
  under the limit: `init_cmd.py::shipped_label` 2 to 4 (round 1, nit 3's second
  digest) and one test helper 3 to 4.
- `check_no_mocks.py` rc=0 and `check_agent_docs_sync.py` rc=0.
- Em dashes: zero U+2014 in all 28 changed files, whole files and not only
  added lines, because the `no-em-dash` hook is `types: [text]`. Zero in the
  ten commit messages. Zero emoji in any added line.
- `tests/test_operator_context.py` 741 to 758 and `kstrl/operator_context.py`
  389 to 559, both well under the limit.

## Assumed

Stated as assumptions rather than results.

- That an unattended multi-iteration `ks factory` run against a real LLM behaves
  as the seam tests say. No such run was made and no LLM was called in this
  lane. What WAS run is the seam itself: `_run_component` to `run_loop` to a
  real agent subprocess reading its stdin, and the real `ks init` plus the real
  `ks run` command group.
- `ks feature` and `ks understand` are covered by reading their four `run_loop`
  call sites, by a captured prompt for neither, and now by layer 3 of
  `tests/test_operator_files_reach_the_prompt.py`, which fails if either module
  ever names `context_prefix`. The claim is held by something that runs; it is
  still not held by a prompt capture.
- Each commit is not independently proven green; only the head is measured.
- The daemon-writes-memory case belongs to #231 and does not exist yet.

## Simplify

A `/simplify` pass ran four review agents over this diff. Six findings were
applied in `cbc0f04`; nine were declined. Every gate, all 33 mutations and the
guard census were re-run afterwards, and the census is byte-identical.

**Accepted.** (1) The spine fixture (`_repo_with_source`, `_run`,
`_prompt_after_run`) moves to module level in
`tests/test_spine_engineer_loop.py`: the R10.9 class had been reaching it by
INSTANTIATING the R10.8 class, a second path to one fixture. (2) That
indirection also carried a `_repo_with_source` wrapper nothing called. (3)
`assert_delimited(block, header)` moves into `tests/helpers/operatorfiles.py`
and is shared; the inline copy in `tests/test_operator_file_kinds.py` checked
each delimiter line on its own, so it passed a block whose two delimiters
carried different tokens. (4) `operator_context._configured` is inlined into
`_rows`: twelve lines for one expression, and its stated mypy justification was
re-measured and is false (`--strict` is green inlined). The rationale moves onto
`_rows`, where the loop that actually prevents the one-field-over defect is. (5)
`init_wizard.py`'s comment saying only "the three under scripts/kstrl/" can
carry a label was already wrong at four when R10.8 landed; the hand count is
removed rather than corrected. (6)
`tests/test_string_keys_reach_every_surface.py` claimed "patching either name is
patching the table the loaders use". Measured false: `from kstrl.config_keys
import STRING_KEYS` binds a second name, so after the monkeypatch that module
has 11 rows and `kstrl.config` still has 10. The docstring now says which
surfaces the fixture reaches and which file covers the other three.

**Declined.** Deleting about 180 lines of golden-only cases from
`tests/test_operator_context.py` because the parametrized file re-covers them:
that removes guards the twelve #229 re-plants were verified against, and
contradicts the recorded decision to keep the R10.8 cases with their audit-trail
docstrings. Parametrizing the six R10.9 spine cases over the kinds: larger than
what it removes, and it restructures a class the mutation set targets. Merging
`_golden_warnings` and `_memory_warnings`: lengthens 13 call sites to save two.
Sharing one `MEMORY_REL` across three test files by deriving it from
`KstrlConfig()`: that reads the expectation off the code under test. `spec_for`'s
no-op `max_chars` default: every alternative is the same length and less
explicit. Collapsing the "memory is last" rationale out of `ARCHITECTURE.md`:
that file is read by people who never open `factory.py`. Deleting three spine
cases to save 1.5 seconds of a 555-second suite. Two further findings were real
but too large for this PR and are carried as handoffs below.

## Handoffs

- **#231 (R10.10, polled steering) writes to this file.** Two corrections this
  PR owns and #231 should take: the append goes through
  `kstrl/appendio.py::append_records(path, payload, repair="", lock=True)`, NOT
  the `tempfile.mkstemp` plus `os.replace` pattern #231's text names, because
  CLAUDE.md forbids a hand-rolled `mkstemp` and `tests/test_atomicio.py`
  AST-walks `kstrl/` and fails a new one; and `## Guidance` must stay the last
  heading of the file, because that is what makes a tail append land in the
  right section. **Nothing in kstrl enforces that and this PR does not add it**
  (round 1, nit 9): shipping the constant with `## Guidance` last is not the
  same as reading the operator's headings, and
  `TestTheAppendLandsUnderGuidance` records the failure mode as a measurement,
  a `## Notes` section after `## Guidance` taking every append with no gate
  going red. #231 is what writes the file, so #231 is where the check belongs.
  One more thing #231 will hit: `append_terminated` raises
  `ValueError("append_terminated payload must end in a newline")`, so the
  payload carries its own newline.
- **`ks feature` and `ks understand` read neither operator file.** The four
  call sites are `kstrl/feature_cmd.py:301`, `:464`, `:589` and
  `kstrl/cli.py:1551`; none passes `context_prefix`. Wiring them is a decision
  about what those commands are for, not an oversight, and it is left where
  #229 left it. No sibling issue filed.
- Row 9 of the tracker in `docs/control-loop-design.md` is ticked in this PR.
  The box in the #235 tracking issue is ticked on merge.

Closes #230

AI code review will be run on this PR as an assist at the owner's request; the
owner remains the gating reviewer (H1).

Reviewed by four `/simplify` agents for quality only, and then by two AI
correctness reviews (round 1: 0 blocker, 3 should-fix, 10 nits; round 2: 1
blocker, 2 should-fix, 8 nits), whose findings are answered in the Round 1 and
Round 2 fixes sections above. No AI review was run on the round-1 or round-2 fix
commits themselves, per H1; the owner is the gating reviewer.

Co-Authored-By: Claude Fable 5.1 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01WtNQsD9BMWmd5razgnKbre


kstrl/operator_context.py
```diff
@@ -76,38 +88,86 @@
...
+#: the two files have different jobs and different growth rates. Golden
+#: patterns is written once and pruned by hand, while #231 makes the
+#: daemon a writer of this one, and under a shared budget a file that
+#: grows by machine would starve a file that does not, silently.
+MEMORY = OperatorFileKind(
+    key="memory",
+    field="memory_file",
+    header="MEMORY (standing feedback)",
+    subject="Memory",
+    max_chars=4000,
```
0xfauzi, line 255: **The cut keeps the HEAD, and memory.md is the one operator file documented to grow at the TAIL.**

`read_operator_file` truncates with `window = rendered[: spec.max_chars]` (line 329), so everything past the budget is dropped from the END. That is the right direction for golden patterns, which is hand-pruned. It is the wrong direction for this row: `DEFAULT_MEMORY` says "new entries are appended to the end of the file", the runbook says "appends land at the end of the file", the README says R10.10 "will append your `/memory` pull-request comments to the end of the file", and this PR's own handoff tells #231 to append there.

Measured on this branch, a memory.md of 400 appended rules (4823 characters):

```
fact: truncated: 3994 of 4823 characters shown from scripts/kstrl/memory.md
first line of body: # Memory
last line of body:  - rule 0330
contains newest rule 0399? False
contains oldest rule 0000? True
```

So the 69 newest standing corrections are silently absent from every engineer prompt while the oldest survive. Once #231 lands, a `/memory` comment appended to an over-budget file reaches no run at all and the daemon has no way to know. The announced `fact` says how much arrived but never which end was lost, so neither the engineer nor the operator can tell.

Suggested shape: put the cut direction on `OperatorFileKind` (golden keeps the head, memory keeps the tail) and say which end was dropped in the shared `shown` string, so the two audiences still get one wording.

kstrl/factory.py
```diff
@@ -2180,41 +2203,50 @@ def _run_component(
...
+    # component choose what the next component is told, unfiltered and
+    # under a header saying the operator wrote it (review round 1, S3).
+    # The same function resolves the parent's once-per-run notice, so the
+    # two cannot read different files (review round 2, should-fix 2). ""
     # when absent, empty, or an unedited `ks init` scaffold.
-    golden_patterns = load_operator_file(golden_patterns_spec(root_dir, golden_patterns_file_str))
+    golden_patterns = load_operator_file(
+        operator_file_spec(GOLDEN_PATTERNS, root_dir, golden_patterns_file_str)
+    )
+    memory = load_operator_file(operator_file_spec(MEMORY, root_dir, memory_file_str))
```
0xfauzi, line 2231: **The worker's prompt block does not follow `OPERATOR_FILES`, so a third row would be validated and warned about but never injected.**

`_rows` loops the table, so `_report_operator_files` and `KstrlConfig.validate` pick up any new row. This site does not: it names `GOLDEN_PATTERNS` and `MEMORY` by hand, and the order is hand-written into the `parts` tuple below.

Concretely, adding a third row (an `OperatorFileKind` plus its `STRING_KEYS` row plus its `SCAFFOLDED_TEMPLATES` row) gives you: all 11 parametrized cases in `tests/test_operator_file_kinds.py` green (they call `load_operator_file` directly), `test_every_kind_matches_a_paths_row` green, `test_every_kind_is_enrolled_in_the_ledger` green, the parent warning about a misconfigured `[paths] <key>`, and `validate` erroring on it - and no engineer prompt ever containing the block. Nothing in the suite fails. The operator is told their file is misconfigured for a file the run never reads.

The whole feature reaches the worker through five hand edits (`memory_file_str` at :2004, the load at :2218, the tuple slot at :2246, `memory_file_rel` at :3743, the positional slot at :3844) plus a 35th slot in a positional tuple. A third file needs the same five plus a 36th.

The general change: send ONE slot carrying `{kind.key: rel_path}` for every row of `OPERATOR_FILES`, and put the prompt position on the row (the thing that actually varies is "before or after the retry context"), so this site becomes two comprehensions over the table and a third file is a row and nothing else.

kstrl/operator_context.py
```diff
@@ -76,38 +88,86 @@
...
+class OperatorFileKind:
+    """One operator-authored file, declared once and read everywhere.
+
+    Everything that differs between two such files sits on this row, so
+    nothing below it spells any of them: the parent's once-per-run
+    notice, ``KstrlConfig.validate`` and every worker's prompt block all
+    follow the same declaration rather than a parallel one. R10.9 is the
+    case that claim was made about (review round 2, should-fix 5, and
+    ``_rows``' own docstring): the memory file is a second row, and no
+    function in this module or in ``kstrl/factory.py`` learned its name.
```
0xfauzi: **This claim is measurably false, and it is the claim the whole "second row, not second loader" argument rests on.**

> the parent's once-per-run notice, `KstrlConfig.validate` and every worker's prompt block all follow the same declaration rather than a parallel one [...] no function in this module or in `kstrl/factory.py` learned its name.

`kstrl/factory.py` learned its name in five places: the import at :107 (`MEMORY`), the load at :2218, the parameter `memory_file_str` at :2004, `memory_file_rel` at :3743, and the positional slot at :3844. The worker's prompt block is the one consumer that does NOT follow the declaration.

`_rows`' docstring (line ~349) carries the same claim one step further: "a row added there reaches the parent's notice, `KstrlConfig.validate` and **the worker's block** with no edit below this line". The first two are true; the third is not.

This repo's rule is that a docstring claim is checked rather than asserted - `_report_operator_files`' round-2 docstring made a claim about R10.9 and this PR is the case that proved it. Same treatment is owed here: either narrow the claim to the two surfaces that do follow the table, or make the third one follow it too.

kstrl/init_cmd.py
```diff
@@ -253,6 +253,30 @@
...
+# docs/runbook.md and in the README, where it costs no tokens.
+#
+# `## Guidance` is LAST and must stay last. R10.10 (#231) appends a
+# `/memory` comment to the END of the file and expects it to land in
+# that section; a section added after it would take the appends.
+DEFAULT_MEMORY = """# Memory
+
+Standing feedback for kstrl runs in this repository: durable rules that
+should change future runs, not one-off instructions. Read into the
+engineer's prompt after the retry context. Nothing is injected while
```
0xfauzi: **This preamble becomes engineer-facing prompt text the moment the operator adds one line, and it is imperative and self-contradicting.**

The comment above already states the mechanism: "the moment anyone edits this file, the whole file reaches the engineer on every iteration, preamble included". Three consequences of the specific wording:

1. `Nothing is injected while this file is unchanged since ks init wrote it.` is FALSE in every context where the engineer can read it. The only way it reaches a prompt is by having been changed.
2. `Keep ## Guidance last; new entries are appended to the end of the file.` is an instruction addressed to the reader, inside a block labelled `MEMORY (standing feedback)` and asserted to be operator-authored. The engineer is a plausible actor on it: under `ks run` (`use_worktrees=False`, `allowed_paths` empty, so the in-loop guard is inert) an agent that appends to `scripts/kstrl/memory.md` writes into the operator's tree and steers every later run.
3. This module's own H3a argument says exactly this: "Adding a sentence that tells the engineer what to DO with the block would make it a prompt body and would put it under H3." `DEFAULT_MEMORY` is not enrolled in `tests/test_prompt_versions.py`.

`DEFAULT_GOLDEN_PATTERNS` has the same shape but no imperative and no false statement, so this is not simply inherited precedent.

Suggested fix: keep the scaffold body to a heading, one non-imperative sentence and `## Guidance`, and leave the mechanism to `docs/runbook.md` and the README, which is where the comment already says it belongs.

docs/runbook.md
```diff
@@ -388,6 +388,28 @@ accounting files and no events), or its `events.jsonl` could not be
 opened. When that happens the last finished run's verdict is still
 reported beside it.
 
+## Standing feedback
+
+When you find yourself correcting the same thing on a second pull request,
+the correction belongs in `scripts/kstrl/memory.md`, not in another PR
+comment. A comment steers one change; a line in that file is read into every
+subsequent engineer prompt, after the retry context, so it steers every run
```
0xfauzi: **"read into every subsequent engineer prompt" is stronger than the code, and the gap is the one #231 will care about.**

`_run_component` assembles `context_prefix` once, before it calls `run_loop`, and `run_loop` builds `prompt` once at `kstrl/loop.py:628`, outside its iteration loop. So the memory block is frozen for a whole component attempt: an edit made while a component is running reaches no iteration of that attempt, only the next attempt or the next component.

That is fine for a hand-edited file. It is the whole latency question for R10.10, whose stated behaviour is that a `/memory` pull-request comment appends a line "the next run reads" (this module's own docstring). "Next run" is accurate; "every subsequent engineer prompt" is not, and an operator reading this section will expect the correction to land on the iteration after they write it.

Same overstatement in `README.md` line 227 ("It reaches every factory engineer prompt") and in the CHANGELOG entry. Suggest saying "every engineer prompt from the next component attempt onward".

tests/helpers/operatorfiles.py
```diff
@@ -0,0 +1,100 @@
...
+    Through ``operator_file_spec`` and not through a literal, so a field
+    added to :class:`~kstrl.operator_context.OperatorFile` reaches these
+    cases the same way it reaches the factory, and so a case parametrized
+    over ``OPERATOR_FILES`` exercises the row rather than a copy of it.
+    """
+    built = operator_file_spec(kind, path.parent, path)
+    return replace(
+        built,
+        max_chars=kind.max_chars if max_chars is None else max_chars,
+        scaffold=scaffold,
```
0xfauzi, line 70: **`spec_for` silently discards `kind.scaffold`, while its docstring says it builds a spec "the way production builds one".**

`operator_file_spec` sets `scaffold=kind.scaffold` on the line above; this `replace` then overwrites it with the parameter, whose default is `None`. Production never builds a spec with `scaffold=None` for an enrolled kind, so the default is the one shape that cannot occur in the factory.

The direction it fails in is the unsafe one. A case that writes a scaffold body and asserts the block is SUPPRESSED fails red if the argument is forgotten, which is fine. A case that writes a scaffold body plus an edit and asserts the block is ON passes either way, so `test_one_edited_line_turns_the_block_on` is currently proving less than it reads as: with `scaffold=None` it would pass even if `shipped_label` suppression were broken in the "edited scaffold" direction.

Fix: default `scaffold` to the built spec's own value (`built.scaffold`) and make the no-scaffold case an explicit argument, which is the rarer one. `max_chars` on the line above has the same shape but is a genuine no-op, since `operator_file_spec` already sets it to `kind.max_chars`.

tests/test_operator_file_kinds.py
```diff
@@ -0,0 +1,310 @@
...
+            assert isinstance(getattr(KstrlConfig(), kind.field), Path)
+
+    def test_every_kind_is_enrolled_in_the_ledger(self) -> None:
+        assert {k.scaffold for k in OPERATOR_FILES} <= {t.filename for t in SCAFFOLDED_TEMPLATES}
+
+    def test_every_row_is_distinguishable_from_every_other(self) -> None:
+        """The parametrized cases above compare a block against the OTHER
+        rows' headers, and a run of tests over identical rows would pass
+        every one of them. Said out loud rather than left to luck."""
+        for attribute in ("key", "field", "header", "subject", "scaffold"):
```
0xfauzi: **Five of `OperatorFileKind`'s six fields, hand-listed. `max_chars` is the one left out.**

This is the test whose job is to stop the parametrized cases going vacuous, and `test_over_the_kinds_own_budget_it_truncates_and_announces` is precisely the case that goes vacuous when two rows share a budget: it sizes its fixture off `kind.max_chars`, so with `MEMORY.max_chars = 6000` it passes for both rows and proves nothing about either. Only `test_the_memory_row_is_the_one_the_issue_asked_for` catches that today, and it is a literal pin that a third row would not have.

The tuple is also a ledger of what someone remembered, not a census: a seventh field added to the row is unchecked here with nothing failing, which is the "ledger of give-ups vs closed by construction" split CLAUDE.md names.

```suggestion
        for attribute in (f.name for f in dataclasses.fields(OperatorFileKind)):
```

(with `import dataclasses` added). That is closed by construction: a new field on the row is distinguishable-checked the day it lands.

kstrl/init_cmd.py
```diff
@@ -253,6 +253,30 @@
...
+this file is unchanged since `ks init` wrote it. Keep `## Guidance`
+last; new entries are appended to the end of the file.
+
+## Guidance
```
0xfauzi, line 287: **"`## Guidance` is LAST and must stay last" is load-bearing for #231 and nothing detects the one actor who can break it.**

The invariant is stated in four places (this comment, the scaffold body, `README.md`, `docs/runbook.md`) and pinned in one: `test_ks_init_scaffolds_memory` asserts `DEFAULT_MEMORY.rstrip("\n").endswith("## Guidance")`. That checks the constant, which only this repo can change and which the ledger row already guards.

The actor that actually breaks it is the operator, adding `## Notes` or `## Archive` after `## Guidance` in their own file, exactly as this feature invites them to. Nothing reads the file's headings. When #231 appends to the tail it will land under whatever heading is last, silently, and the entries will still be injected (so no gate notices) while sitting in a section that says something else.

Given this PR carries the constraint to #231 as a handoff, the check belongs with the appender: refuse or repair an append when `## Guidance` is not the last heading of the file, rather than relying on four documents.

README.md
```diff
@@ -207,7 +207,29 @@ Three fixture types: `cli` (run a command, check output), `function` (import and
...
+findings are known false positives, and the feedback you have given more than
+once. What does not belong in it is a one-off instruction or a run log; an
+entry changes future runs rather than one pull request, which is the whole
+point of the file.
+
+The position is the mechanism. It is read into the engineer prompt AFTER the
+retry context and before `CLAUDE.md`, so your correction frames how this
+attempt's failures get acted on rather than the other way round. It reaches
+every factory engineer prompt and every `ks run` prompt (`ks feature` and
+`ks understand` do not read it), from the repo root and never from a
```
0xfauzi, line 228: **Four shipped surfaces assert this and nothing tests it.**

"`ks feature` and `ks understand` do not read it" is also asserted in `docs/env-vars.md`, `docs/runbook.md` ("Two commands read it"), `CHANGELOG.md` and `ARCHITECTURE.md`. The PR body lists the evidence under **Assumed**: four `run_loop` call sites read by eye (`kstrl/feature_cmd.py:301`, `:464`, `:589`, `kstrl/cli.py:1551`), none passing `context_prefix`.

That is a negative claim about a data path, made on five user-facing surfaces, held up by nobody adding a keyword argument. It is one cheap test: assert that no `run_loop` call in `feature_cmd.py` or the `understand` command passes `context_prefix` (an AST walk over those modules, the shape `tests/helpers/astwalk/` already exists for), or capture a real `ks feature` prompt the way `tests/test_spine_golden_patterns_e2e.py` captures a `ks run` one and assert neither header appears.

Worth closing here rather than in #231, because #231 makes memory.md a file the daemon writes: "which commands read it" stops being a documentation detail once the content arrives from a pull-request comment.

